PR #17, opened 2022-05-17
TEST-7: add exchange rate

riley-stride:
DO NOT MERGE, LOGIC NOT FINISHED YET

**Summary**:

Implemented logic to calculate the exchange rate between stToken/Token at deposit and unbonding. The main complication along the way is not being able to figure out how to get return values out of asynchronous ICQ calls, so having to write the result to state rather than handle it in-memory. This required the creation of interchainstaking store variables. 

We need rigorous testing in this code (it’s exploitable); this is just a v0 of the logic. 

Pointers for review:
- Most logic lives in `stride/x/stakeibc/keeper/exchange_rate.go`, where two functions live:
      - `getStTokenExchRate` calcalates the exchange rate between stAsset/Asset using as a helper the `ICQQueryHostBalance` function. We use `BankKeeper`'s `GetSupply` function to calculate supply of stAsset (same approach as QS to calculating stAsset supply). For deposits, `inclOutstandingRewards` param should be true, for withdrawals, false. 
      - `ICQQueryHostBalance` submits an ICQ call to measure the amount of tokens on host chain either across all delegation accounts, delegated, or in outstanding accrued rewards. Results are written to the host_zone state rather than returned (due to technical limitations described in summary). 
- In `proto/stakeibc/host_zone.proto`, added `connectionID`, denominations and state variables used to calc exchange rate (they're updated to measure host zone delegations and balances). Removed `rewardsAccount`. 
- In `stride/proto/stakeibc/params.proto`, remove unused `ica_account.proto` import
- In `stride/x/stakeibc/keeper/exchange_rate.go`, replace BinaryCodec with Codec (the former does not allow for `cdc.UnmarshalJSON` calls required in exchange_rate calc, the latter does. Also, add `bankKeeper` for exchange_rate stAsset supply calcs. 
- Create `stride/x/stakeibc/keeper/zones.go`, containing helpers for zones. We could move these functions to `zones.go` to simplify. <==`**TODO` move to zones.go **
- In `stride/x/stakeibc/types/keys.go`, added prefixes. We may not need these. <==`**TODO` remove these **

For more details see Jira TEST-7

_Note I had to add third_party folder to get this working_


riley-stride:
Closing this as it's subsumed by TEST-50

PR #33, opened 2022-06-06
[TEST-6] Add `RedeemStake` message

asalzmann:
**Summary**
Add redemptions for liquid staking. The approaches considered for redemptions
- Tokenize stake using LSM and transfer it to the user
    - Not all chains support LSM, so this is nice-to-have but not feature complete
- Batching redemptions
    - This is likely more gas efficient / mirrors our staking logic, but is a poor UX because the user needs to wait until the next epoch to see their balance reappear 
- **Redeeming tokens directly**
    - This is the approach implemented
    - Bundle the following messages into a single TX
        - MsgSetWithdrawAddress(from=ica_delegatation_account_address, to=delegator_address_on_host)
        - MsgUndelegate(from=ica_delegatation_account_address, from=delegator_address_on_host, amount=coins)
        - MsgSetWithdrawAddress(from=delegator_address_on_host, to=ica_delegatation_account_address)

In particular, the following changes were made
- Add `RedeemStake` message
    - Add `RedeemStake` message to `tx.proto`
    - Add `RedeemStake` to CLI, handler, codec
    - Update `HostZone` schema to include a redemption rate and a last redemption rate
    - Implement `RedeemStake` functionality in `msg_server_redeem_stake.go`. This does some sanity checks on the input, escrows the user's coins and constructs the interchain tx. Note, the tx containing the 3 messages is atomic
- Add `OnAcknowledgementPacket` logic to `module_ibc.go`
- Clean up `LiquidStake` logic
- Add `ica_redeem.sh` script
- Update `ALLOW_MESSAGES` to include `MsgUndelegate`
- Remove `build/` and `scripts/state` which were still being tracked by git

Next steps
- This approach does not work for addresses that have a different bech32 address derivation path https://docs.cosmos.network/master/spec/addresses/bech32.html (e.g. Secret Network)
- The validator address is hardcoded, we need to expand this logic to handle (un)delegation (from)to multiple validators
- This needs more robust testing (using BAT) once TEST-11 is merged in
- Implement `OnAcknowledgementPacket` logic (TEST-28, TEST-33)

**Test plan**
`make init` then register the host zone
```
$STR1_EXEC tx stakeibc register-host-zone \
    connection-0 uatom statom --chain-id \
    STRIDE_1 --home /stride/.strided --keyring-backend test \
    --from val1 --gas 500000 -y
```
transfer tokens to the delegation account
```
$GAIA1_EXEC tx bank send $VAL_KEY $DELEGATION_ADDR 100uatom --chain-id GAIA_1 --home /gaia/.gaiad --keyring-backend test -y
```
verify tokens are being staked with `gaiad q staking delegations cosmos10ltqave0ml70h9ynfsp6py2pv925xuzys7ypmffr8ud92sj09dzs6xtq8e`
```
delegation_responses:
- balance:
    amount: "7"
    denom: uatom
  delegation:
    delegator_address: cosmos10ltqave0ml70h9ynfsp6py2pv925xuzys7ypmffr8ud92sj09dzs6xtq8e
    shares: "7.000000000000000000"
    validator_address: cosmosvaloper19e7sugzt8zaamk2wyydzgmg9n3ysylg6na6k6e
pagination:
  next_key: null
  total: "0"
```

undelegate some tokens
```
/ $ strided tx stakeibc redeem-stake 2 statom \
>     --chain-id STRIDE_1 --home /stride/.strided --keyring-backend test \
>     --from val1 --gas 500000 -y
code: 0
```
verify tokens were undelegated
```
/ $ gaiad q staking delegations cosmos10ltqave0ml70h9ynfsp6py2pv925xuzys7ypmffr8ud92sj09dzs6xtq8e
delegation_responses:
- balance:
    amount: "5"
    denom: uatom
  delegation:
    delegator_address: cosmos10ltqave0ml70h9ynfsp6py2pv925xuzys7ypmffr8ud92sj09dzs6xtq8e
    shares: "5.000000000000000000"
    validator_address: cosmosvaloper19e7sugzt8zaamk2wyydzgmg9n3ysylg6na6k6e
pagination:
  next_key: null
  total: "0"
```


x/stakeibc/keeper/ibc_handlers.go
```diff
@@ -0,0 +1,131 @@
...
+// 	sdk "github.com/cosmos/cosmos-sdk/types"
+// 	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
+// 	stakingtypes "github.com/cosmos/cosmos-sdk/x/staking/types"
+// 	channeltypes "github.com/cosmos/ibc-go/v3/modules/core/04-channel/types"
+// 	"google.golang.org/protobuf/proto"
+
+// 	icatypes "github.com/cosmos/ibc-go/v3/modules/apps/27-interchain-accounts/types"
+// )
+
+// Implements core logic for OnAcknowledgementPacket
```
asalzmann, line 23: Implement OnAcknowledgementPacket logic (TEST-28, TEST-33)

scripts/ica_redeem.sh
```diff
@@ -0,0 +1,54 @@
+#!/bin/bash
```
vish-stride, line 1: don't think we need to do this in this PR, but we should probably incorporate this into the bats test file

x/stakeibc/keeper/ibc_handlers.go
vish-stride: does this work right now if we return nil on ack? do we not need to process any info / emit any events?

vish-stride (approved):
Overall looks great, this is very exciting! 

Some minor comments, but overall definitely good to merge imo. 

2 overall things:
1. We should probably migrate the test cases to bats (not urgent though, and i'm happy to take this ticket), just to make sure this functionality doesn't break with subsequent commits
2. Changing redemption to allow arbitrary withdrawal addresses should be straightforward I think, given your implementation. Should be a quick ticket, also happy to take it

x/stakeibc/keeper/ibc_handlers.go
asalzmann: it can just return nil (any processing can be added before returning nil)

x/stakeibc/keeper/msg_server_liquid_stake.go
```diff
@@ -23,17 +23,22 @@ func (k msgServer) LiquidStake(goCtx context.Context, msg *types.MsgLiquidStake)
...
 	if err != nil {
-		return nil, sdkerrors.Wrapf(sdkerrors.ErrInvalidType, "failed to parse %s coins", coins)
+		return nil, sdkerrors.Wrapf(sdkerrors.ErrInvalidType, "failed to parse %s inCoin", inCoin)
 	}
 
 	// Safety checks
 	// ensure Amount is non-negative, liquid staking 0 or less tokens is invalid
-	if msg.Amount < 1 {
+	if !inCoin.IsPositive() {
 		k.Logger(ctx).Info("amount must be non-negative")
```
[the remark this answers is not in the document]
asalzmann: done

x/stakeibc/keeper/msg_server_redeem_stake.go
```diff
@@ -0,0 +1,101 @@
...
+		k.Logger(ctx).Info("Failed to send sdk.NewCoins(inCoins) from account to module")
+		panic(err)
+	}
+
+	// calculate the redemption rate
+	// when redeeming tokens, multiply stAssets by the exchange rate (allStakedAssets / allStAssets)
+	// TODO(TEST-7): Update redemption_rate via ICQ
+	var rate sdk.Dec
+	rate = hostZone.LastRedemptionRate
+	if hostZone.RedemptionRate.LT(rate) {
```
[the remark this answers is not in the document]
asalzmann: This is pulled from QS, I imagine they added it as a simple way to mitigate redemption rate manipulation. Needs more thought but pushing to TEST-7

scripts/ica_redeem.sh
asalzmann: good call, added test-74 - working on these now. going to push them to a separate PR

x/stakeibc/keeper/msg_server_redeem_stake.go
```diff
@@ -0,0 +1,101 @@
...
+	connectionId := hostZone.GetConnectionId()
+	
+	// Safety checks
+	// Redemption amount must be positive
+	if !inCoin.IsPositive() {
+		return nil, sdkerrors.Wrapf(sdkerrors.ErrInvalidCoins, "amount must be greater than 0. found: %s", msg.Amount)
+	}
+	// Denom is valid
+	// Should we register stAssets somewhere and add an additional check here?
+	if types.IsStAsset(msg.Denom) != true {
```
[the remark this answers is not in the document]
asalzmann: possibly not, although I think it's a cheap sanity check and would err on the side of checking more rather than less in this function

x/stakeibc/keeper/msg_server_redeem_stake.go
```diff
@@ -0,0 +1,101 @@
...
+		return nil, sdkerrors.Wrapf(sdkerrors.ErrInvalidCoins, "denom is not a valid stAsset. found: %s", msg.Denom)
+	}
+	// TODO: REMOVE THIS, TESTING PURPOSES ONLY
```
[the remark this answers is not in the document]
asalzmann: removed

asalzmann:
Added a ticket for arbitrary address schemas

proto/stakeibc/tx.proto
```diff
@@ -29,6 +30,15 @@ message MsgLiquidStake {
 message MsgLiquidStakeResponse {
 }
 
+message MsgRedeemStake {
+  string creator = 1;
+  int32 amount = 2;
```
riley-stride, line 35: cosmos.Dec might be better? Allow non-integer redemption amounts?

x/stakeibc/keeper/msg_server_redeem_stake.go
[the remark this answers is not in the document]
riley-stride: Resurfacing this since I don't see any reason we would want to use LastRedemptionRate and I rolled TEST-7 into TEST-50 before seeing this PR comment. 

What's the redemption rate manipulation attack vector we're preventing here? Let's discuss

x/stakeibc/keeper/msg_server_redeem_stake.go
```diff
@@ -0,0 +1,89 @@
...
+
+	// calculate the redemption rate
+	// when redeeming tokens, multiply stAssets by the exchange rate (allStakedAssets / allStAssets)
+	// TODO(TEST-7): Update redemption_rate via ICQ
+	var rate sdk.Dec
+	rate = hostZone.LastRedemptionRate
+	if hostZone.RedemptionRate.LT(rate) {
+		rate = hostZone.RedemptionRate
+	}
+	native_tokens := inCoin.Amount.ToDec().Mul(rate).TruncateInt()
```
riley-stride, line 67: Ideally we should not round here (imagine a world in which ATOM trades at $1000. Rounding in that world is a serious issue for the protocol!

x/stakeibc/types/message_redeem_stake.go
```diff
@@ -0,0 +1,58 @@
...
+	return &MsgRedeemStake{
+		Creator: creator,
+		Amount:  amount,
+		Denom:   denom,
+	}
+}
+
+// isStAsset checks if the denom of the asset matches our stAsset prefix
+// Note, this is not definitive (other f'st{something}' tokens could be minted)
+func IsStAsset(denom string) bool {
```
riley-stride, line 24: Probably safer to check against all HostZone stDenoms?

riley-stride (commented):
Overall LGTM! Solid logic in `x/stakeibc/keeper/msg_server_redeem_stake.go`, esp with the atomic txs.

Added some nit / questions throughout. 

Main substantive comments:
- In `x/stakeibc/keeper/msg_server_redeem_stake.go`, let's discuss using RedemptionRate vs LastRedemptionRate and possible attack vectors
Some int/dec and rounding points:
- We probably want to allow non-integer redemption rates? (See https://github.com/Stride-Labs/stride/pull/33/files#r894582608)
- Redemptions should probably be fractional (See https://github.com/Stride-Labs/stride/pull/33/files#r894596931)

-

PR #36, opened 2022-06-09
[Test 59] Automate Host Zone Transfer Args

vish-stride:
**Summary**

High level, this PR fixes testing framework (was broken after folder reorg) and fixes interchain-staking issues around module account (removing hard coded values)

Important things to look at:
1. RegisterHostZone now takes an extra argument for channel id (in files `tx_register_host_zone.go` and `msg_server_register_host_zone.go`
2. In `keeper/host_zone.go` we use DenomTrace to get the Host Zone from the ibc denom

There was also a slight amount of funniness on the bats repos (I previously just `mv`'d the submodules, but it looks like that is not correct). Everything is fixed now there, but you might see small changes to git files 

**Test plan**

`make init` now runs tests that tokens make it to host chain, and that those tokens are staked properly
To run just tests, please run `sh scripts/tests/run_all_tests.sh` 
All test cases should pass! 

Right now, the additional tests that are being run are 
1. Host Zone is registered, uatom is the BaseDenom, and the DelegateAddress is the expected one
2. Assets are being staked on Host Zone (e.g. if we wait 30 seconds, more assets than before are staked)

Follow Ups:
1. After the ICQ PR, we should make it so all Host Zone assets are staked 
2. After (1), we should uncomment the additional test that should make sure all assets are staked on Host Zone (still need to add a few lines to that test though)


proto/stakeibc/tx.proto
```diff
@@ -65,6 +65,7 @@ message MsgRegisterHostZone{
   string base_denom = 4 [ (gogoproto.moretags) = "yaml:\"base_denom\"" ];
   string local_denom = 5 [ (gogoproto.moretags) = "yaml:\"local_denom\"" ];
   string creator = 6;
+  string channel_id = 10 [ (gogoproto.moretags) = "yaml:\"channel_id\"" ];
```
riley-stride, line 68: why should we pass the channel into registerHostZone? Does this replace having the channel assigned automatically?

scripts/tests/basic_tests.bats
```diff
@@ -83,18 +96,23 @@ setup() {
...
+  assert_line '    address: cosmos19l6d3d7k2pel8epgcpxc9np6fsvjpaaa06nm65vagwxap0e4jezq05mmvu'
+}
+
+# add test to see if assets are properly being staked on host zone
+@test "tokens staking on host zone" {
+  # run below test once ICQ is deployed
+  # VAL_ADDR='cosmosvaloper19e7sugzt8zaamk2wyydzgmg9n3ysylg6na6k6e'
+  # $GAIA1_EXEC q staking delegation cosmos19l6d3d7k2pel8epgcpxc9np6fsvjpaaa06nm65vagwxap0e4jezq05mmvu cosmosvaloper19e7sugzt8zaamk2wyydzgmg9n3ysylg6na6k6e
+  #   amount: "240"
+  del_balance_atom=$($STR1_EXEC q bank balances $DELEGATE_ADDR --denom uatom | GETBAL)
```
riley-stride, line 113: Loving this testing framework

proto/stakeibc/tx.proto
```diff
@@ -65,6 +65,7 @@ message MsgRegisterHostZone{
   string base_denom = 4 [ (gogoproto.moretags) = "yaml:\"base_denom\"" ];
```
riley-stride, line 65: Nit: 

This PR feels like good PR to change the denom names from "base" and "local" to something more intuitive. I still every time I see them have trouble remembering which is which with the current names. 

Maybe baseDenom (e.g. uatom) and stDenom (stuatom)?

x/stakeibc/keeper/host_zone.go
```diff
@@ -50,22 +52,33 @@ func (k Keeper) GetHostZone(ctx sdk.Context, chain_id string) (val types.HostZon
...
+// GetHostZoneFromDenom returns a hostZone from the relevant coin denom (denom is ibc/...)
 func (k Keeper) GetHostZoneFromDenom(ctx sdk.Context, denom string) (val types.HostZone, found bool) {
 	store := prefix.NewStore(ctx.KVStore(k.storeKey), types.KeyPrefix(types.HostZoneKey))
 	iterator := sdk.KVStorePrefixIterator(store, []byte{})
+	var bval types.HostZone
 
 	defer iterator.Close()
 
+	// hash is the part of denom after ibc/
+	hash := strings.Join(strings.SplitN(denom, "ibc/", -1), "")
```
riley-stride, line 64: could we add a simple error check that complains if the denom is not prefixed by "ibc"?

scripts/tests/c.sh
```diff
@@ -1,13 +1,42 @@
 STR1_EXEC="docker-compose --ansi never exec -T stride1 strided --home /stride/.strided --chain-id STRIDE"
+GAIA1_EXEC="docker-compose --ansi never exec -T gaia1 gaiad --home /gaia/.gaiad --chain-id GAIA"
 GETBAL() {
     head -n 1 | grep -o -E '[0-9]+'
   }
-IBCATOM='ibc/9117A26BA81E29FA4F78F57DC2BD90CD3D26848101BA880445F119B22A1E254E'
+IBCATOM='ibc/27394FB092D2ECCD56123C74F36E4C1F926001CEADA9CA97EA622B25F41E5EB2'
```
asalzmann, line 6: does the setup logic affect this denom? I think this is a hash of source port + source channel, so it's possible initializing channels in a different order could affect this

x/stakeibc/keeper/host_zone.go
asalzmann, line 64: should this be
```suggestion
	hash := strings.Join(strings.SplitN(denom, "ibc/", 1), "")
```
?

asalzmann: realize this is somewhat pedantic given the lookup below _should_ fail if it's not valid, but I would err on the side of explicitly checking assumptions in the core liquid staking logic

x/stakeibc/keeper/host_zone.go
```diff
@@ -50,22 +52,33 @@ func (k Keeper) GetHostZone(ctx sdk.Context, chain_id string) (val types.HostZon
...
+	req := &ibctypes.QueryDenomTraceRequest{Hash: hash}
+	goCtx := sdk.WrapSDKContext(ctx)
+	denomTrace, err := k.transferKeeper.DenomTrace(goCtx, req)
+	if err != nil {
+		k.Logger(ctx).Error("unable to obtain chain from denom %s: %w", hash, err)
+		return bval, false
+	}
+	baseDenom := strings.ToUpper(denomTrace.DenomTrace.BaseDenom)
```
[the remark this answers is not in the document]
asalzmann: similar to above comment, I think it's worth adding an explicit check that the token has only traveled one hop, rather than implicitly verifying this with the loop below

asalzmann (commented):
Looks good! Left a few comments, mostly around checks in the denom trace code

scripts/init.sh
```diff
@@ -102,41 +102,38 @@ sh ${SCRIPT_DIR}/init_gaia.sh
...
 # Spin up docker containers
 #############################################################################################################################
-
-sleep 5
 docker-compose down
 docker-compose up -d stride1 stride2 stride3 gaia1 gaia2 gaia3
-echo "Chains created"
-sleep 10
+echo "Chains creating..."
+CSLEEP 10
```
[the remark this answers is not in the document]
vish-stride: Ah, I defined this in `vars.sh`, it's just sleep but it prints out how many seconds are left. It makes the long 60s sleeps a bit more manageable

proto/stakeibc/tx.proto
vish-stride: transfer channel so we know what channel to move the tokens over on. I don't think we strictly need this (we could back out the channel based on the denom trace when we create the DepositRecord), but I think it's slightly more standard to only support one transfer channel (I believe it's similar to Osmosis)

scripts/tests/basic_tests.bats
vish-stride: Thank you! Should hopefully be easy enough to customize / add more detailed tests

scripts/tests/c.sh
```diff
@@ -1,13 +1,42 @@
...
+
+docker-compose --ansi never exec -T stride1 strided --home /stride/.strided --chain-id STRIDE q ibc-transfer denom-trace C4CFF46FD6DE35CA4CF4CE031E643C8FDC9BA4B99AE598E9B0ED98FE3A2319F9
+docker-compose --ansi never exec -T stride1 strided tx stakeibc register-host-zone C4CFF46FD6DE35CA4CF4CE031E643C8FDC9BA4B99AE598E9B0ED98FE3A2319F9 ATOM stATOM --chain-id STRIDE --home /stride/.strided --keyring-backend test --from val1 --gas 500000 -y
+
 
 ibcaddr=$($STR1_EXEC q stakeibc module-address stakeibc | awk '{print $NF}') 
 module_atom=$($STR1_EXEC q bank balances $ibcaddr --denom $IBCATOM | GETBAL)
 echo $module_atom
 echo $ibcaddr
 echo $($STR1_EXEC q bank balances $ibcaddr --denom $IBCATOM)
```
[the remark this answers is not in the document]
vish-stride: Ah, sorry I've actually just been using this file as a "hot file" to run quick queries to debug. Not meant to be used outside of debugging. I could add it to the .gitignore?

proto/stakeibc/tx.proto
vish-stride: Also agreed these aren't intuitive but probably better to make this an "urgent fix" after the sprint compiles? I'll add a ticket for it right now

x/stakeibc/keeper/hooks.go
```diff
@@ -35,11 +35,11 @@ func (k Keeper) BeforeEpochStart(ctx sdk.Context, epochIdentifier string, epochN
 					continue
 				}
 				delegateAddress := delegateAccount.Address
-				// TODO TEST-59 grab proper port name and channel name
 				timeoutHeight := clienttypes.NewHeight(0, 500)
 				transferCoin := sdk.NewCoin(depositRecord.Denom, sdk.NewInt(int64(depositRecord.Amount)))
 				goCtx := sdk.WrapSDKContext(ctx)
-				msg := ibctypes.NewMsgTransfer("transfer", "channel-1", transferCoin, addr, delegateAddress, timeoutHeight, 0)
+
+				msg := ibctypes.NewMsgTransfer("transfer", hostZone.ChannelId, transferCoin, addr, delegateAddress, timeoutHeight, 0)
```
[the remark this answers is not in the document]
vish-stride: I think (??) the transfer channels are just assigned chronologically, so it's very possible it's not channel 1 (when I was testing, I had it as channel 0,1, and 2). I think as long as we know the transfer channel when we create HostZone (and we're able to update it if need be), this should be okay!

x/stakeibc/keeper/host_zone.go
vish-stride: This is a great catch, I think this might expose an attack vector! Per huddle, it sounds like this is taken care of with TEST-50

scripts/tests/c.sh
vish-stride: Agreed this is dependent on the ordering in the init file. If we change it, we'll have to change this

x/stakeibc/keeper/host_zone.go
vish-stride: they should result in the same output here, I just wanted to explicitly call out "go to the last section after ibc/" instead of "go to the first section after ibc/"

PR #39, opened 2022-06-11
[TEST-5] Reinvest rewards

asalzmann:
**Summary**

**Test plan**

PR #40, opened 2022-06-12
Setup code for withdraw, reinvest rewards

asalzmann:
**Summary**
This PR preps adding `MsgRedeemStake`, adds `ReinvestRewards`, does some cleanup around integer types and adds `ProcessDelegationStaking` to go through each HostZone and stake the delegation.

- Add (unimplemented) MsgRedeemStake
    - Note, this was originally going to be implemented per [TEST-5](https://github.com/Stride-Labs/stride/pull/39), but we realized that this approach would not work given we cannot withdraw to arbitrary addresses https://stride-labs.slack.com/archives/C03DP2J3R45/p1655076988479129 
- Change all int32 -> int64 (matters for small token denoms)
- Add a `DefaultDelegate` and `DefaultReinvest` interval to params
- Remove `stakeibc` and `bank` keepers from `interchainquery`. First, `stakeibc` uses `icq` so this introduced a circular dependency. Second, `icq` shouldn't need access to `stakeibc` or `bank` keepers
- Remove MsgQueryBalance, MsgQueryExchangerate, MsgQueryDelegatedbalance
- Restructure interchain queries to be run from the Keeper by adding `QueryBalances`, which takes in an arbitrary address to query and a callback
- Add `ReinvestRewards` function (not yet called in epochs)

**Next steps**
We need to keep records for
- Deposits
- Rewards
- Pending unbondings

Once we have these records, we can (atomically) during each epoch
1. Stake deposits
2. Distribute rewards from the previous epoch & stake rewards from the epoch before that
3. Distribute unbonded tokens

**Test plan**
BATS test for `ProcessDelegationStaking`: see "tokens staking on host zone"


x/interchainquery/client/cli/tx.go
```diff
@@ -25,55 +24,15 @@ func GetTxCmd() *cobra.Command {
 	}
 
 	cmd.AddCommand(
-		QueryBalanceCmd(),
+		//QueryBalanceCmd(),
 		// TODO(TEST-53) remove cli access to ICQ queries pre-launch
 		SubmitQueryResponse(),
```
riley-stride, line 29: We can remove this now

x/interchainquery/client/cli/tx.go
```diff
@@ -25,55 +24,15 @@ func GetTxCmd() *cobra.Command {
 	}
 
 	cmd.AddCommand(
-		QueryBalanceCmd(),
+		//QueryBalanceCmd(),
```
riley-stride, line 27: Nit: rm

x/interchainquery/types/msgs.go
```diff
@@ -12,7 +12,7 @@ const (
 
 var (
 	_ sdk.Msg = &MsgSubmitQueryResponse{}
-	_ sdk.Msg = &MsgQueryBalance{}
+	//_ sdk.Msg = &MsgQueryBalance{}
```
riley-stride, line 15: nit: rm

x/stakeibc/keeper/delegation.go
```diff
@@ -31,3 +37,61 @@ func (k Keeper) RemoveDelegation(ctx sdk.Context) {
...
+		delegationIca := zoneInfo.GetDelegationAccount()
+		if delegationIca == nil || delegationIca.Address == "" {
+			k.Logger(ctx).Error("Zone %s is missing a delegation address!", zoneInfo.ChainId)
+			return false
+		}
+		cdc := k.cdc
+		DelegateOnHost := k.DelegateOnHost
+
+		var queryBalanceCB icqkeeper.Callback = func(k icqkeeper.Keeper, ctx sdk.Context, args []byte, query icqtypes.Query) error {
+			k.Logger(ctx).Info(fmt.Sprintf("\tdelegation staking callback on %s", zoneInfo.HostDenom))
```
riley-stride, line 55: I believe you want to use the parent function's keeper. Easy fix it to rename `k icqkeeper.Keeper` on line 54 to `icqk`

x/stakeibc/keeper/hooks.go
```diff
@@ -35,7 +35,9 @@ func (k Keeper) BeforeEpochStart(ctx sdk.Context, epochIdentifier string, epochN
 					continue
 				}
 				delegateAddress := delegateAccount.Address
-				timeoutHeight := clienttypes.NewHeight(0, 500)
+				// TODO(TEST-89): Set NewHeight relative to the most recent known gaia height (based on the LC)
+				// TODO(TEST-90): why do we have two gaia LCs?
+				timeoutHeight := clienttypes.NewHeight(0, 10000)
```
riley-stride, line 40: higher is better here to be sure to avoid issues while testing

x/stakeibc/keeper/hooks.go
```diff
@@ -53,34 +55,38 @@ func (k Keeper) BeforeEpochStart(ctx sdk.Context, epochIdentifier string, epochN
...
-					k.Logger(ctx).Info("Successfully staked %s on %s", processAmount, zoneInfo.ChainId)
-				}
-				return false
-			}
-
-			// Iterate the zones and apply icaStake
-			k.IterateHostZones(ctx, icaStake)
+			k.ProcessDelegationStaking(ctx)
 		}
+		// process withdrawals
```
riley-stride, line 60: Let's rm the comments before merging into main

x/stakeibc/keeper/ibc_handlers.go
```diff
@@ -102,6 +102,20 @@ func (k Keeper) HandleAcknowledgement(ctx sdk.Context, modulePacket channeltypes
 	// 	case "/ibc.applications.transfer.v1.MsgTransfer":
 	// 		// TODO: Implement!
 	// 		continue
+	// 	case "cosmos.bank.v1beta1.MsgSend":
```
riley-stride, line 105: What's going on with the comments here? Probably OK but I don't entirely follow

x/stakeibc/keeper/msg_server_liquid_stake.go
```diff
@@ -19,10 +19,10 @@ func (k msgServer) LiquidStake(goCtx context.Context, msg *types.MsgLiquidStake)
 	// check that hostZone is registered
 	// strided tx stakeibc liquid-stake 100 uatom
 	hostDenom := msg.Denom
-	hostZone, hostZoneFound := k.GetHostZoneFromDenom(ctx, msg.Denom)
-	if !hostZoneFound {
+	hostZone, err := k.GetHostZoneFromHostDenom(ctx, msg.Denom)
```
riley-stride, line 22: Can we rename this Denom (msg.Denom)? I know we have it in the backlog

x/stakeibc/keeper/msg_server_redeem_stake.go
```diff
@@ -25,12 +23,15 @@ func (k Keeper) RedeemStake(goCtx context.Context, msg *types.MsgRedeemStake) (*
 	if err != nil {
 		return nil, sdkerrors.Wrapf(sdkerrors.ErrInvalidCoins, "could not parse inCoin: %s", coinString)
 	}
-	hostZone, err := k.GetHostZoneFromIBCDenom(ctx, msg.Denom)
+	// remove st prefix to get the base denom
+	baseDenom := msg.Denom[2:]
```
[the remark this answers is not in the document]
riley-stride: Also on this line, is there a more robust way to map between the two denoms? If we change the prefix down the line (ideally we don't have to but possible) we'd need to come back to change this

x/stakeibc/keeper/msg_server_redeem_stake.go
```diff
@@ -66,22 +67,19 @@ func (k Keeper) RedeemStake(goCtx context.Context, msg *types.MsgRedeemStake) (*
 	}
 	native_tokens := inCoin.Amount.ToDec().Mul(rate).TruncateInt()
 	outCoin := sdk.NewCoin(hostZone.HostDenom, native_tokens)
+	_ = outCoin
```
riley-stride, line 70: does this line do anything? same comment for ln 75

x/stakeibc/keeper/msg_server_submit_tx.go
```diff
@@ -91,6 +94,79 @@ func (k Keeper) DelegateOnHost(ctx sdk.Context, hostZone types.HostZone, amt sdk
...
+
+	// Fetch the relevant ICA
+	delegationAccount := hostZone.GetDelegationAccount()
+	withdrawAccount := hostZone.GetWithdrawalAccount()
+
+	SubmitTxs := k.SubmitTxs
+	GetParam := k.GetParam
+	cdc := k.cdc
+
+	var cb icqkeeper.Callback = func(k icqkeeper.Keeper, ctx sdk.Context, args []byte, query icqtypes.Query) error {
```
riley-stride, line 118: k => icqk so that we don't override the parent keeper definition and can distinguish

x/stakeibc/keeper/msg_server_submit_tx.go
```diff
@@ -91,6 +94,79 @@ func (k Keeper) DelegateOnHost(ctx sdk.Context, hostZone types.HostZone, amt sdk
...
+		queryRes := bankTypes.QueryAllBalancesResponse{}
+		err := cdc.Unmarshal(args, &queryRes)
+		if err != nil {
+			k.Logger(ctx).Error("Unable to unmarshal balances info for zone", "err", err)
+			return err
+		}
+		// Get denom dynamically
+		balance := queryRes.Balances.AmountOf(hostZone.HostDenom)
+		balanceDec := sdk.NewDec(balance.Int64())
+		commission := sdk.NewDec(int64(GetParam(ctx, types.KeyStrideCommission))).Quo(sdk.NewDec(100))
```
riley-stride, line 129: Dividing by 100 is a bit of a hack here, we might be better off defining the commission as a Dec and passing it in directly. I just want to avoid any kind of rounding issues (I don't see obvious issues but maybe better to not manipulate/transform key parameters too much)

x/stakeibc/keeper/msg_server_submit_tx.go
```diff
@@ -91,6 +94,79 @@ func (k Keeper) DelegateOnHost(ctx sdk.Context, hostZone types.HostZone, amt sdk
...
+		// Dec type has 18 decimals and the same precision as Coin types
+		strideAmount := balanceDec.Mul(commission)
+		reinvestAmount := balanceDec.Sub(strideAmount)
+		strideCoin := sdk.NewCoin(hostZone.HostDenom, strideAmount.TruncateInt())
+		reinvestCoin := sdk.NewCoin(hostZone.HostDenom, reinvestAmount.TruncateInt())
+
+		// transfer balances from the withdraw address to the delegation account
+		sendBalanceToDelegationAccount := &bankTypes.MsgSend{FromAddress: withdrawAccount.GetAddress(), ToAddress: delegationAccount.GetAddress(), Amount: sdk.NewCoins(reinvestCoin)}
+		msgs = append(msgs, sendBalanceToDelegationAccount)
+		// TODO: get the stride commission addresses (potentially split this up into multiple messages)
```
riley-stride, line 139: add a ticket

x/stakeibc/types/params.go
```diff
@@ -11,7 +11,8 @@ import (
 var (
 	// these are default intervals _in epochs_ NOT in blocks
 	DefaultDepositInterval      uint64 = 5
-	DefaultDelegateInterval     uint64 = 2
+	DefaultDelegateInterval     uint64 = 5
```
riley-stride, line 14: we need to raise these for robust testing; will include in my PR TEST-96

riley-stride:
Overall on a good track, nice! 

Lots of different things going on in this PR, a lot of which are necessary scaffolding for our next set of features (including TEST-96), however I don't think we should merge this into main until it's either one notch cleaner or functional & tested. Small comments throughout, nothing major

x/interchainquery/client/cli/tx.go
asalzmann: Good call, removed. Also, closed out TEST-80 and TEST-84 (tracking updates to this code)

x/interchainquery/client/cli/tx.go
asalzmann: Done.

x/interchainquery/types/msgs.go
asalzmann: Done.

x/stakeibc/keeper/delegation.go
asalzmann: Great catch, done.

x/stakeibc/keeper/hooks.go
asalzmann: Pushed to 1000000

x/stakeibc/keeper/ibc_handlers.go
asalzmann: This hasn't been implemented yet, but I've have been adding bits and pieces of pseudocode to track the updates we're going to need to make.

Tracked in TEST-28 and TEST-33

Slightly cleaner approach might be to open a draft PR and add to that, although it makes the process of adding code slightly more complex because we'd have to switch between the working PR and the draft to add comments. Going to leave this in for now because the implementation is up next

x/stakeibc/keeper/msg_server_liquid_stake.go
asalzmann: Done. Updated `tx.proto`, the type files and keeper functions. Also updated denom on `MsgRedeemStake` to `StAssetDenom`.

x/stakeibc/keeper/msg_server_redeem_stake.go
asalzmann: The unimplemented part of this function requires these coins, but the linter removes these lines unless you do this

x/stakeibc/keeper/msg_server_submit_tx.go
asalzmann: Done.

x/stakeibc/types/params.go
asalzmann: sgtm

x/stakeibc/keeper/msg_server_submit_tx.go
asalzmann: Done.

x/stakeibc/keeper/hooks.go
asalzmann: Updated.

x/stakeibc/keeper/msg_server_submit_tx.go
asalzmann: Per @vish-stride comment in params.go
"you apparantly cannot safely encode floats, so we make commission * 100"
Given that, we need to divide by 100 somewhere. I think `Dec` is the best place to do so; the `Dec` type has 18 decimals so rounding errors shouldn't be an issue. More importantly, our commission is `10`, so 10/100 will produce a `Dec` type like `0.100000...`

asalzmann:
Fixed everything you mentioned in your comments, thanks for the thorough review.

Merging this PR is a tradeoff between having everything functional vs. having smaller, reviewable PRs. For velocity I'd bias towards the latter, so going to merge this.

PR #41, opened 2022-06-13
[DEPRACATED -- NEWER VERSION IS TEST-119] Test 96: Epochly Redemption Rate + Used in MintStAsset

riley-stride:
## Summary
Add logic to automatically update the redemptionRate, and use that exchange rate as the rate for minting new stAssets. Note this approach is _unsafe_ and _lagged_. 

Note that the current implementation of the redemptionRate is flawed in two key ways (the first is most important). It is:
- **_unsafe_**: updates to components of the redemptionRate occur asynchronously and read from differents blocks on the hostZone, so it is possible and likely that the redemptionRate will be use stale balances (some next steps on how to address this are below in _outstanding_). Empirically in my testing, the redemptionRate has mostly behaved as expected, but occasionally jumped below 1 (should never happen) or otherwise jumped down then back up again drastically due to mis-timed queries. 
- _lagged_: the redemptionRate, even if it were _safe_, is lagged because it takes time to submit and receive the response to an ICQ query, therefore the hostZone balances used to calculate the query may be lower than their true updated balances, as rewards could have accrued and been claimed in the time between query delivery and minting/redemption. This is probably not a big deal, much lower priority than the _safe_ty of the redemptionRate calculation.

Repeat the following:
- On epoch `0`, we query `undelegatedBalances` on the hostZone ICA delegate account and store them in `hostZone.delegationAccount.balances`. This logic lives inside the `ProcessDelegationStaking` function, which already queries hostZone delegationAccount balances, sparing us the need to query balances again redundantly. 
- On epoch `3`, we query `delegatedBalances` (staked balances) on the hostZone ICA delegate account (we add these up across all validators the ICA has staked to) and store them in `hostZone.delegationAccount.DelegatedBalances`
- On epoch `6`, we query the outstanding supply of `stAssets` for the hostZone, then calculate the redemptionRate as `(undelegatedBalances+delegatedBalances) / stAssetSupply`. 
- Repeat, modding epoch numbers...

---

## Main logic:
* In `msg_server_submit_tx.go`, add functions that iterate over hostZones (using `IterateHostZones`), define callbacks, and invoke ICQ through helper methods in `interchainquery`, passing in the relevant callback (which typically saves the result to store). Added fns `ProcessUpdateUndelegatedBalance` and`ProcessUpdateDelegatedBalance`.
     * Also add `ProcessUpdateExchangeRate`, which reads the balances from the stakeibc store, queries stAsset outstanding supply and saved the redemptionRate to state. 
* Add `QueryDelegatorDelegations` to the `interchainquery` module, which accepts a callback, queries delegations from a specified address and executes the callback. We invoke this fn when querying delegated balances from `stakeibc`. 
* In `delegation.go`, save `undelegatedBalances` (which we've already collected from our staking logic) to the store. Also add some logging.
* In `msg_server_liquid_stake.go`, add logic to calculate the redemption rate for minting new stAssets.

Misc:
* Logging script augmented, _very_ helpful for debugging. Run `while true; do make init build=logs ; sleep 3 ; done` in the background. I find it easiest to open a new stride VSCode window with folder access on a separate monitor with CMD+SHIFT+P, select Workspaces: Duplicate as workspace in new Window , then open all the logs in different tabs there to easily click between them. 
* Increase the timeout block height in `hooks.go`: `timeoutHeight := clienttypes.NewHeight(0, 1000000000000)`. This was giving stochastic errors.
* Tweak `IterateHostZone`'s signature so allow ingesting `ctx sdk.Context`
* Raise the trusting_period for more robust testing
* Point Dockerfile.gaia back at regular gaia and `go mod tidy` and `go mod download` on the image.
* Added rough test to the bottom of `basic_tests.bats`. 

---

## Test plan
- I added some testing commands to the bats scripts but need some help parsing the outputs to do a proper check. Can you lend a hand here @vish-stride ? See the bottom of `basic_tests.bats`. Execute with `sh ./scripts/tests/run_all_tests.sh`

## Outstanding
- specify a height when querying the hostZone, to ensure no balances are stale
       - query stAsset balances at the same block height as the balances above!
- use the redemptionRate in redemptions

---

Good example of the danger of stale balances (_unsafe_):

    time 1
    staked: 0
    unstaked: 10

    time 2
    staked: 10
    unstaked: 0

if query 1 returns that unstaked balances = 10 (from time 1) and query 2 returns that staked balances are 10 (from time 2), the ex rate could be 0.5 (10 stXXX / 20 XXX) when it should be 1


x/stakeibc/keeper/msg_server_submit_tx.go
```diff
@@ -165,6 +167,143 @@ func (k Keeper) ReinvestRewards(ctx sdk.Context, hostZone types.HostZone) error
...
+				sdk.EventTypeMessage,
+				sdk.NewAttribute("totalDelegatedBalance", delegatorSum.Amount.String()),
+			),
+		})
+
+		return nil
+	}
+	// 1. query delegation account delegated balances using icq
+	// 2. sum up the resulting delegations (across validators)
+	// 2. write the result to hostZone.delegationAccount.delegatedBalance
```
asalzmann, line 212: nit: 3

x/stakeibc/keeper/msg_server_submit_tx.go
```diff
@@ -165,6 +167,143 @@ func (k Keeper) ReinvestRewards(ctx sdk.Context, hostZone types.HostZone) error
...
+// icq to read host delegation account undeleted balance => update hostZone.delegationAccount.Balance
+// TODO(TEST-97) add safety logic to query at specific block height (same as query height for delegated balances)
+func (k Keeper) UpdateUndelegatedBalance(ctx sdk.Context, index int64, hostZone types.HostZone) error {
+	_ = ctx
+	// Fetch the relevant ICA
+	delegationAccount := hostZone.GetDelegationAccount()
+
+	var cb icqkeeper.Callback = func(icqk icqkeeper.Keeper, ctx sdk.Context, args []byte, query icqtypes.Query) error {
+
+		queryRes := bankTypes.QueryAllBalancesResponse{}
```
asalzmann, line 226: Pretty sure this does not include staked balances, but just wanted to double check that's also your understanding

x/stakeibc/keeper/msg_server_submit_tx.go
```diff
@@ -165,6 +167,143 @@ func (k Keeper) ReinvestRewards(ctx sdk.Context, hostZone types.HostZone) error
...
+}
+
+// Update the redemption rate using values of delegatedBalances, balances and stAsset supply
+// TODO(TEST-97) add safety logic that checks balance, delegatedBalance and stAsset supply's block_height_updated are all equal
+func (k Keeper) UpdateExchangeRate(ctx sdk.Context, index int64, hostZone types.HostZone) error {
+	_ = ctx
+
+	// Assets: native asset balances on delegation account + staked
+	delegatedBalance := hostZone.GetDelegationAccount().DelegatedBalance
+	unDelegatedBalance := hostZone.GetDelegationAccount().Balance
```
asalzmann, line 270: nit: undelegatedBalance

x/stakeibc/keeper/msg_server_submit_tx.go
```diff
@@ -165,6 +167,143 @@ func (k Keeper) ReinvestRewards(ctx sdk.Context, hostZone types.HostZone) error
...
+		balance := queryRes.Balances.AmountOf(hz.HostDenom)
+
+		// Set delegation account balance to ICQ result
+		hz, found := k.GetHostZone(ctx, hostZone.ChainId)
+		if found {
+			k.Logger(ctx).Error("invalid chain id, zone for \"%s\" already registered", hostZone.ChainId)
+		}
+
+		da := hz.DelegationAccount
+		da.Balance = balance.Int64()
```
asalzmann, line 244: Actually, scratch the above comment. I think this should be `DelegatedBalance`.

Can we make this naming less confusing? E.g.
- DelegatedBalance
- UndelegatedBalance

x/stakeibc/keeper/msg_server_submit_tx.go
```diff
@@ -165,6 +167,143 @@ func (k Keeper) ReinvestRewards(ctx sdk.Context, hostZone types.HostZone) error
...
+			sdk.NewEvent(
+				sdk.EventTypeMessage,
+				sdk.NewAttribute("delegatedBalance", strconv.FormatInt(delegatedBalance, 10)),
+				sdk.NewAttribute("unDelegatedBalance", strconv.FormatInt(unDelegatedBalance, 10)),
+				sdk.NewAttribute("stAssetBalance", stAssetSupply.String()),
+			),
+		})
+		return errors.New("Exchange rate calculation error: ")
+	} else {
+		// ExchRate = Assets / Claims
```
asalzmann, line 288: Should we refer to "redemption rate" from now on?

Exchange rate seems slightly confusing because you don't know which asset is the base. E.g. with 10 assets and 5 stAssets, the exchange rate is 2 and 0.5. If we refer to a redemption rate, it's clear that it's 2 (redeem 1 stAsset for 2 assets)

asalzmann (commented):
Overall LGTM!  Just a few minor comments

x/stakeibc/keeper/msg_server_submit_tx.go
riley-stride: Agreed

x/stakeibc/keeper/msg_server_submit_tx.go
```diff
@@ -165,6 +167,143 @@ func (k Keeper) ReinvestRewards(ctx sdk.Context, hostZone types.HostZone) error
...
+		for _, delegation := range response.DelegationResponses {
+			delegatorSum = delegatorSum.Add(delegation.Balance)
+			if err != nil {
+				return err
+			}
+		}
+
+		// Set delegation account balance to ICQ result
+		da := hz.DelegationAccount
+		da.Balance = delegatorSum.Amount.Int64()
```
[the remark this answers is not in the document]
riley-stride: good call, done

asalzmann:
Let me know when you want me to review. I merged in #40, so you should be able to merge main into this branch.

asalzmann:
Screen Shot 2022-06-16 at 1 22 52 PM
lol

riley-stride:
Haha perfect timing

Sent via Superhuman iOS ***@***.***>

asalzmann:
Should we close?

PR #42, opened 2022-06-13
rm unsupported handler calls

asalzmann:
**Summary**
Our testnet should only support liquid staking.

**Test plan**


x/stakeibc/handler.go
```diff
@@ -20,18 +20,18 @@ func NewHandler(k keeper.Keeper) sdk.Handler {
...
-		case *types.MsgRegisterAccount:
-			res, err := msgServer.RegisterAccount(sdk.WrapSDKContext(ctx), msg)
-			return sdk.WrapServiceResult(ctx, res, err)
-		case *types.MsgSubmitTx:
-			res, err := msgServer.SubmitTx(sdk.WrapSDKContext(ctx), msg)
-			return sdk.WrapServiceResult(ctx, res, err)
-		case *types.MsgRegisterHostZone:
-			res, err := msgServer.RegisterHostZone(sdk.WrapSDKContext(ctx), msg)
-			return sdk.WrapServiceResult(ctx, res, err)
-		case *types.MsgRedeemStake:
```
riley-stride, line 32: I realize we are not supporting redemption, but why remove MsgRegisterAccount, MsgSubmitTx and MsgRegisterHostZone? @asalzmann

asalzmann: I don't think we'll want them to be callable for mainnet, so was thinking we should also remove them for testnet

riley-stride: Makes testing slightly harder though for now, maybe let's wait to merge this in? We need to register a host zone in the init script in order to do any cross chain logic. Maybe we could register it elsewhere?

riley-stride:
closing in favor of WL approach

PR #46, opened 2022-06-16
Add record keeping

asalzmann:
**Summary**
Update the following to be records-based
1) Token transfers (controller -> host zone)
2) ICA staking
3) Reinvesting rewards
4) Withdrawals [this is coming in the next PR]

**Test plan**


notes.txt
```diff
@@ -0,0 +1,241 @@
+# Record keeping
+
+Unfortunately I think this complicates unbonding logic a bit.
+Given this reality,
+1) I think the naive approach is to track user's unbondings in stakeibc module state, then transfer 
+the unbonded tokens from the delegation account to the user's account as soon as the unbonding period 
+ends. Staking module's unbonding-delegations object has completion_time: "2022-06-12T23:10:51.070726210Z"
```
riley-stride, line 7: Note: I am not sure, but seems like unbondings are triggered based on clock time rather than block height. This means we need to be able to check clock time reliably in epochs

notes.txt
```diff
@@ -0,0 +1,241 @@
...
+(by summing up deposit receipts), rather than staking the entire delegation account balance. This would 
+alleviate the need to transfer the funds out before delegation account token are re-staked (because the 
+unbonded tokens would remain in the account until transfered to the user).
+
+3) Another approach could be to track unbonded amounts each in a FIFO queue in stakeibc module state, 
+and, at each epoch, while the balance in the delegation account is  >= the amount of the top item in 
+the queue, transfer funds to that user's account and pop the queue. If the queue is emptied, stake the 
+remaining funds. Haven't thought this one through fully
+
+I vote for (2) for its simplicity, but (3) could be interesting down the line bc it allows
```
riley-stride, line 21: I agree #2 is the best approach given our time constraints and need for safety

notes.txt
```diff
@@ -0,0 +1,241 @@
...
+******************************DEPOSITS*****************************
+*******************************************************************
+
+# SECTION 1
+# 0 Update DepositRecord type
+  UPDATE DEPOSIT RECORD
+  message DepositRecord {
+    uint64 id = 1; // epoch_num:chain_id
+    int64 amount = 2; // total amount
+    string denom = 3; // tokens
```
riley-stride, line 39: nit: specify which denom e.g. hostDenom

notes.txt
```diff
@@ -0,0 +1,241 @@
...
+  hz = getHz(record)
+  if record.status == TRANSFER:
+    def ibcTransferAck(ack):
+      if ack == success:
+        record = getRecord(ack)
+        delete(record)
+    # add record deletion logic to ack
+    ibcTransfer.ackFunction = ibcTransferAck
+    # ibc transfer the tokens over
+    ibcTransfer(to=hz.depositAccount(), amount=record.amount)
```
riley-stride, line 90: Logic makes sense, good thinking to use the ack to delete records. I thought we were planning to batch transfers from the module account to the delegation account?

notes.txt
```diff
@@ -0,0 +1,241 @@
...
+  ICABankSend.Ack = bankSendAck
+  ICABankSend(from=withdrawal, to=delegat, amount=record.amount)
+
+# add new records for rewards that accrued in the previous epoch
+for hz in getAllHostZones():
+  def addRewardsRecord(hz):
+    # define a query callback function
+    def queryCB(hz):
+      recordBalance = sumAllRewardsRecords()
+      newBalance = balance - recordBalance
```
riley-stride, line 137: why are you subtracting these here?

riley-stride: Don't we need to exclude the stride commission from the rewardRecord amount? Seems like the pseudocode queries for the full balance of the withdrawal account and adds a record that will sweep the entire balance back to the delegation account (that'd mean stride gets no commission!)

notes.txt
```diff
@@ -0,0 +1,241 @@
...
+      createNewRewardsRecord(newBalance)
+    wA = hz.withdrawalAccount()
+    query(address=wA, queryType=balance, cb=queryCB)
+
+
+# NOTE: this is executed after processing records because although ICA calls
+# _are_ ordered (meaning we could first send balances from withdrawal -> delegate account),
+# there's no way to enforce that the records are updated on stride without chaining these.
+# unclear whether chaining is a robust approach. What happens if one step fails?
+# what do we have to roll back?
```
riley-stride, line 147: good question, need to think this through

notes.txt
```diff
@@ -0,0 +1,241 @@
...
+# easier to think about these in isolation IMO
+1. ICQ wA balance
+2. ICQ CB: create records, BankSend ICA call them from withdrawal -> delegate account
```
riley-stride, line 150: edge case logic: when you BankSend ICA Call, mark records 'in progress' in case the ICA ack takes longer to return than an epoch (in which case you risk triggering the BankSend ICA Call another time in the next epoch before the first BankSend ICA Call acks).

notes.txt
```diff
@@ -0,0 +1,241 @@
...
+3. Bank send ICA Ack: update deposit records, iterate through them and MsgDelegate ICA call them
```
riley-stride, line 151: edge case logic: same approach as ln 138, mark records 'in progress' when you begin an ICA call that processes them

notes.txt
```diff
@@ -0,0 +1,241 @@
...
+
+
+>>> EPOCH START
+----------
+
+# users call redeem stake
+def redeemStake(tokens):
+  unbondingRecord = newUnbondingRecord()
+  # getUnbondingBlockHeight queries the unbonding record
+  # in the creation ack and adds a dequeue unbonding height
```
riley-stride, line 192: not sure we can do this by block height, might need to use clock time. ideally we can do it by height

notes.txt
```diff
@@ -0,0 +1,241 @@
...
+  # in ack
+    # fetch record
+    # delete from unprocessed records
+    # add unbonding height
+    # add to processed records
+
+# CACHE MOST RECENT BLOCK HEIGHT FOR HOST ZONE
+for zone in hostZones:
+  client = getClient
+  mostRecentBlock = client.height
```
riley-stride, line 212: again, might need to port this to clock time. hoping we can do it with block height. if we do it by block height, we prob wanna trigger a LC update before this line

proto/stakeibc/deposit_record.proto
```diff
@@ -6,13 +6,15 @@ option go_package = "github.com/Stride-Labs/stride/x/stakeibc/types";
...
+    // pending transfer to delegate account
+    TRANSFER = 0;
+    // pending staking on delegate account
+    STAKE = 1;
   }
-  Purpose purpose = 6; 
-  
+  Status status = 6; 
+
+  reserved 5;
```
riley-stride, line 20: I haven't been using reserved when I fiddle with other protos, do I need to go back and do this even if I delete *.pb.go's?

x/stakeibc/keeper/msg_server_liquid_stake.go
```diff
@@ -68,9 +68,9 @@ func (k msgServer) LiquidStake(goCtx context.Context, msg *types.MsgLiquidStake)
 		k.Logger(ctx).Info("failed to send tokens from Account to Module")
 		return nil, sdkerrors.Wrap(sdkerrors.ErrInvalidCoins, "failed to mint stAssets to user")
 	}
-	// create a deposit record of these tokens
-	depositRecord := types.NewDepositRecord(msg.Amount, msg.HostDenom, hostZone.ChainId,
-		sender.String(), types.DepositRecord_RECEIPT)
+	// create a deposit record of these tokens (pending transfer)
+	// TODO: increment amount on deposit record, add creation to end of epoch 
+	depositRecord := types.NewDepositRecord(msg.Amount, msg.HostDenom, hostZone.ChainId, types.DepositRecord_TRANSFER)
```
riley-stride: Why a new record? Thought we want to just add new amount to the current record's amount per https://github.com/Stride-Labs/stride/pull/46/commits/54705464e44b6472b915a49d4938e68b25e3a266#diff-e39538e7f27a7bf579cd9b85a103c0f0b86b60b788534295538d0301a9c5dce6R64

x/stakeibc/keeper/msg_server_liquid_stake.go
```diff
@@ -69,9 +70,19 @@ func (k msgServer) LiquidStake(goCtx context.Context, msg *types.MsgLiquidStake)
...
+		k.Logger(ctx).Info("failed to find epoch")
+		return nil, sdkerrors.Wrapf(types.ErrInvalidLengthEpochTracker, "no number for epoch (%s)", epochtypes.STRIDE_EPOCH)
+	}
+	// Does this use too much gas?
+	depositRecord, found := k.GetDepositRecordByEpochAndChain(ctx, strideEpochTracker.EpochNumber, hostZone.ChainId)
+	if !found {
+		k.Logger(ctx).Info("failed to find deposit record")
+		return nil, sdkerrors.Wrapf(types.ErrInvalidLengthEpochTracker, "no deposit record (%s)", strideEpochTracker.EpochNumber)
+	}
+	depositRecord.Amount += msg.Amount
```
riley-stride, line 84: Looks like you addressed my comments on the last commit here 👍

x/stakeibc/keeper/hooks.go
```diff
@@ -17,42 +17,64 @@ func (k Keeper) BeforeEpochStart(ctx sdk.Context, epochIdentifier string, epochN
...
+
+			// Create one new deposit record / host zone for the next epoch
+			createDepositRecords := func(index int64, zoneInfo types.HostZone) (stop bool) {
+				// create a deposit record / host zone
+				depositRecord := types.NewDepositRecord(0, zoneInfo.HostDenom, zoneInfo.ChainId, types.DepositRecord_TRANSFER, uint64(epochNumber))
+				k.AppendDepositRecord(ctx, *depositRecord)
+				return false
+			}
+			// Iterate the zones and apply icaReinvest
+			k.IterateHostZones(ctx, createDepositRecords)
```
riley-stride: to make sure I'm tracking: the intent is to create a deposit record per epoch, even if no deposits occur in that epoch?

x/stakeibc/keeper/hooks.go
```diff
@@ -58,27 +60,52 @@ func (k Keeper) BeforeEpochStart(ctx sdk.Context, epochIdentifier string, epochN
...
-					transferCoin := sdk.NewCoin(hostZone.GetIBCDenom(), sdk.NewInt(int64(depositRecord.Amount)))
-					goCtx := sdk.WrapSDKContext(ctx)
-	
-					msg := ibctypes.NewMsgTransfer("transfer", hostZone.TransferChannelId, transferCoin, addr, delegateAddress, timeoutHeight, 0)
-					_, err := k.transferKeeper.Transfer(goCtx, msg)
-					if err != nil {
-						pstr := fmt.Sprintf("\tERROR WITH DEPOSIT RECEIPT {%d}", depositRecord.Id)
-						k.Logger(ctx).Info(pstr)
-						panic(err)
+					if depositRecord.Status == types.DepositRecord_TRANSFER {
```
riley-stride: Why move the staking logic to hook rather than keep it in a separate function? Might be cleaner to separate it out, but I assume you have good reason to put it here

utils/utils.go
```diff
@@ -0,0 +1,20 @@
...
+func FilterDepositRecords(arr []stakeibctypes.DepositRecord, condition func(stakeibctypes.DepositRecord) bool) (ret []stakeibctypes.DepositRecord) {
+	for _, elem := range arr {
+		if condition(elem) {
+			ret = append(ret, elem)
+		}
+	}
+	return ret
+}
+
+func Int64ToCoinString(amount int64, denom string) string {
```
riley-stride, line 18: Haha great idea, I've been recreating int64=>string, string=>int64, int64=>coinstring all over the place. Will def use this and add to this file with a few other helpers

x/stakeibc/keeper/hooks.go
riley-stride: Nvm, looks like you addressed this in the next commit 👍

proto/stakeibc/deposit_record.proto
asalzmann: The short answer is that this will always work, while not reserving / updating fields live will sometimes break. Easier just to reserve fields!

If we start with a clean chain state, we don't need this. Once we're updating a live chain, we usually will.

notes.txt
asalzmann: These notes evolved through the commits, in it's current state all timeouts must be l.t. the epoch length

x/stakeibc/keeper/hooks.go
asalzmann: Yes that's right. We might need some edge case logic to handle empty records. We could create a record the first time a user liquid stakes in an epoch, but that would require us to change the way records are stored and use a deterministic key (right now the key for a new record is just an id, which is the deposit record count+1)

utils/utils.go
asalzmann: What error handling are you thinking we should add?

notes.txt
asalzmann: Will look into this - thanks for flagging

x/stakeibc/keeper/hooks.go
```diff
@@ -2,90 +2,82 @@ package keeper
...
+		}
+		epochTracker := types.EpochTracker{
+			EpochIdentifier: epochIdentifier,
+			EpochNumber: uint64(epochNumber),
+		}
+		// deposit records *must* exist for this epoch
+		k.SetEpochTracker(ctx, epochTracker)
+		k.Logger(ctx).Info("Triggering deposits")
+		// Create a new deposit record for each host zone for the upcoming epoch
+		k.CreateDepositRecordsForEpoch(ctx, epochNumber)
```
vish-stride, line 52: I think I'm just missing something, why do we need to create empty deposit records for each host zone in each epoch? is there a lookup that we want to make sure won't fail?

vish-stride (approved):
Overall looks good to me! 

one thing though, given that the record keeping touches a lot of code, I think it'd be good to write some test cases to make sure the records are being added properly. This doesn't necessarily have to be in this PR (esp given time constraints), but if not, we should make a ticket to add them in soon

x/stakeibc/client/cli/query.go
```diff
@@ -35,7 +35,9 @@ func GetQueryCmd(queryRoute string) *cobra.Command {
 	cmd.AddCommand(CmdShowDepositRecord())
 	cmd.AddCommand(CmdModuleAddress())
 	cmd.AddCommand(CmdShowInterchainAccount())
-	// this line is used by starport scaffolding # 1
+	cmd.AddCommand(CmdListEpochTracker())
+	cmd.AddCommand(CmdShowEpochTracker())
+// this line is used by starport scaffolding # 1
```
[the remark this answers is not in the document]
asalzmann: not sure why ignite did this, indented

x/stakeibc/genesis.go
```diff
@@ -28,7 +28,11 @@ func InitGenesis(ctx sdk.Context, k keeper.Keeper, genState types.GenesisState)
 
 	// Set depositRecord count
 	k.SetDepositRecordCount(ctx, genState.DepositRecordCount)
-	// this line is used by starport scaffolding # genesis/module/init
+	// Set all the epochTracker
+for _, elem := range genState.EpochTrackerList {
```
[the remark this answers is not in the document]
asalzmann: another ignite error, thx for flagging

x/stakeibc/keeper/hooks.go
```diff
@@ -2,90 +2,82 @@ package keeper
...
+		// by DepositInterval
+		if epochNumber < 0 {
+			panic("epochs must be positive")
+		}
+		epochTracker := types.EpochTracker{
+			EpochIdentifier: epochIdentifier,
+			EpochNumber: uint64(epochNumber),
+		}
+		// deposit records *must* exist for this epoch
+		k.SetEpochTracker(ctx, epochTracker)
```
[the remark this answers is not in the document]
asalzmann: I originally tried doing it this way, the issue is that in app.go, stakeibc's initialized keeper is an input to the epochs keeper (here: https://github.com/Stride-Labs/stride/blob/9891516fb1ceb5aa1d812bffc0d12c378bbf8e86/app/app.go#L455)

It's possible we could do some tricky logic where we initialize the stakeibc keeper, pass the hooks into epochs then reinitialize the stakeibc keeper with the epochs keeper, although I think this introduces a circular dependency

x/stakeibc/keeper/hooks.go
```diff
@@ -2,90 +2,82 @@ package keeper
...
+	if epochIdentifier == epochstypes.STRIDE_EPOCH {
 		k.Logger(ctx).Info(fmt.Sprintf("Stride Epoch %d", epochNumber))
+
+		// NOTE: We could nest this under `if epochNumber%depositInterval == 0 {`
+		// -- should we?
+		// e.g. CreateDepositRecordsForDepositInterval
+		// Imagine it will be slightly cleaner to track state by epoch, rather than
+		// by DepositInterval
+		if epochNumber < 0 {
+			panic("epochs must be positive")
```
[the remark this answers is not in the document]
asalzmann: Done.

x/stakeibc/keeper/hooks.go
```diff
@@ -115,3 +107,196 @@ func (h Hooks) BeforeEpochStart(ctx sdk.Context, epochIdentifier string, epochNu
...
+			timeoutHeight := clienttypes.NewHeight(0, 1000000)
+			transferCoin := sdk.NewCoin(hostZone.GetIBCDenom(), sdk.NewInt(int64(depositRecord.Amount)))
+			goCtx := sdk.WrapSDKContext(ctx)
+
+			msg := ibctypes.NewMsgTransfer("transfer", hostZone.TransferChannelId, transferCoin, addr, delegateAddress, timeoutHeight, 0)
+			_, err := k.transferKeeper.Transfer(goCtx, msg)
+			if err != nil {
+				pstr := fmt.Sprintf("\tERROR WITH DEPOSIT RECEIPT {%d}", depositRecord.Id)
+				k.Logger(ctx).Info(pstr)
+				panic(err)
```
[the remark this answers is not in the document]
asalzmann: Done.

notes.txt
asalzmann: Done.

notes.txt
asalzmann: I imagine it might actually be easier to do by clock time? Clock time should be consistent across Stride / other zones, so we wouldn't even have to update the LC! (might be missing something here though, don't have a super solid understanding of how clock time works in Cosmos)

x/stakeibc/keeper/hooks.go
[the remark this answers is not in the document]
asalzmann: No longer relevant! Moved everything back to epochs

asalzmann:
Adding more thorough description to this PR tomorrow

PR #47, opened 2022-06-17
(DEPRACATED) TEST-119

riley-stride:
## Summary
This PR updates redemptionRate every few epochs.

In `x/stakeibc/keeper/hooks.go` we 
1. grab `latestGaiaHeight`, the latest height of the gaia light client on Stride
2. grab the total stAsset supply and the module account's IBCdenom balances, save both of these values in `ControllerBalances` (a stakeibc struct) as values with key equal to the `latestGaiaHeight` we just read
3. issue daisy chained ICQ queries to read gaia's delegated, then delegated balances at height `latestGaiaHeight`. We chose to daisy chain bc if more than one tx write to hostZone in a single block, they overwrite each other.
4. when the second ICQ call returns, set the exchange rate if safety checks are satisfied 

Please note the code is very rough still and poorly documented & named. Planning to do lots of cleanup on Sunday. Also note that I scaffolded a new ignite map type, so when reviewing, it may make sense to step over that commit rather than review everything together. 

## Test plan
For now, examine the exchange rate updates manually by running:
`make init build=stride`
`sh ./scripts/tests/1.sh`
`sh ./scripts/tests/2.sh`

Watch the state of accounts, delegations and redemption rate update in `scripts/logs/accounts.log` by running `while true; do make init build=logs ; sleep 2 ; done`.

### Bugs
1. Most of the time, after ~150-200 blocks, the ICQ relayer throws a `panic: timed out after: 60000000000; timed out after waiting for tx to get included in the block` error. Why is this happening? Can we either address the root cause, or restart the relayer when this happens? If you watch `icq.log` when testing, you'll see this error. 
- Could the `MakeRequest` period affect this?
- Can we fix this by slowing the redemptionRate way down e.g. every 30 epochs
2. stAsset supply updates **before** latest height from the gaia LC in `hooks.go`, so we sometimes get redemptionRates below that are lower than lastRedemptionRate. Updating the LC height could fix this. Not sure how to do this yet

### Clean up (nice to have's)
- Various TODO(NOW)'s and TODO(TEST-119)'s in the comments
- [X] Rename `Balances` to `UndelegatedBalances` for clarity
- [X] Clean up function names in `msg_server_submit_tx.go` for clarity
- [X] Condense logic in  `x/interchainquery/keeper/keeper.go`, perhaps condense `QueryBalances` and `QueryDelegatorDelegations` into a single parametrized function, and consider moving event emission logic to endblocker in`abci.go`


scripts/logs/create_logs.sh
```diff
@@ -1,30 +1,17 @@
...
-
-rm $SCRIPT_DIR/gaia.log
-docker compose logs gaia1 >> $SCRIPT_DIR/gaia.log
-
-rm $SCRIPT_DIR/icq.log
-docker compose logs icq >> $SCRIPT_DIR/icq.log
-
-rm $SCRIPT_DIR/hermes.log
-docker compose logs hermes >> $SCRIPT_DIR/hermes.log
+TMP="temp"
```
vish-stride, line 4: maybe "logfiles"

x/interchainquery/keeper/abci.go
```diff
@@ -17,35 +15,5 @@ const (
 // EndBlocker of interchainquery module
 func (k Keeper) EndBlocker(ctx sdk.Context) {
 	defer telemetry.ModuleMeasureSince(types.ModuleName, time.Now(), telemetry.MetricKeyBeginBlocker)
-	_ = k.Logger(ctx)
```
vish-stride, line 20: what a tricky error... thank you for dealing with this!

x/interchainquery/keeper/keeper.go
```diff
@@ -117,32 +119,38 @@ func (k Keeper) QueryBalances(ctx sdk.Context, zone stakeibctypes.HostZone, cb C
...
-func (k Keeper) QueryDelegatorDelegations(ctx sdk.Context, zone stakeibctypes.HostZone, cb Callback, address string) error {
+func (k Keeper) QueryDelegatorDelegations(ctx sdk.Context, zone stakeibctypes.HostZone, cb Callback, address string, height int64) error {
 	connectionId := zone.ConnectionId
 	chainId := zone.ChainId
 
 	query_type := "cosmos.staking.v1beta1.Query/DelegatorDelegations"
 	// Get delegationAddress dynamically
 	delegationQuery := stakingtypes.QueryDelegatorDelegationsRequest{DelegatorAddr: address}
+	k.Logger(ctx).Info(fmt.Sprintf("\tabout to QueryDelegatorDelegations %s at height %d", address, height))
 	bz := k.cdc.MustMarshal(&delegationQuery)
```
vish-stride, line 155: nit: you can probably set `heightStr := strconv.FormatInt(height, 10)` and then reference that in the two calls below

x/stakeibc/client/cli/query.go
```diff
@@ -35,7 +35,9 @@ func GetQueryCmd(queryRoute string) *cobra.Command {
 	cmd.AddCommand(CmdShowDepositRecord())
 	cmd.AddCommand(CmdModuleAddress())
 	cmd.AddCommand(CmdShowInterchainAccount())
-	// this line is used by starport scaffolding # 1
+	cmd.AddCommand(CmdListControllerBalances())
+	cmd.AddCommand(CmdShowControllerBalances())
+// this line is used by starport scaffolding # 1
```
vish-stride, line 40: not sure, but we might have to indent this line for starport

x/stakeibc/client/cli/query_controller_balances_test.go
```diff
@@ -0,0 +1,164 @@
...
+	"google.golang.org/grpc/status"
+
+	"github.com/Stride-Labs/stride/testutil/network"
+	"github.com/Stride-Labs/stride/testutil/nullify"
+	"github.com/Stride-Labs/stride/x/stakeibc/client/cli"
+    "github.com/Stride-Labs/stride/x/stakeibc/types"
+)
+
+// Prevent strconv unused error
+var _ = strconv.IntSize
```
vish-stride, line 22: What does this line do? does the linter not pick up on the strconv references below?

x/stakeibc/genesis.go
```diff
@@ -28,7 +28,11 @@ func InitGenesis(ctx sdk.Context, k keeper.Keeper, genState types.GenesisState)
 
 	// Set depositRecord count
 	k.SetDepositRecordCount(ctx, genState.DepositRecordCount)
-	// this line is used by starport scaffolding # genesis/module/init
+	// Set all the controllerBalances
```
vish-stride, line 31: nit: these lines should be indented

x/stakeibc/genesis.go
```diff
@@ -60,7 +64,8 @@ func ExportGenesis(ctx sdk.Context, k keeper.Keeper) *types.GenesisState {
 	genesis.HostZoneCount = k.GetHostZoneCount(ctx)
 	genesis.DepositRecordList = k.GetAllDepositRecord(ctx)
 	genesis.DepositRecordCount = k.GetDepositRecordCount(ctx)
-	// this line is used by starport scaffolding # genesis/module/export
+	genesis.ControllerBalancesList = k.GetAllControllerBalances(ctx)
+// this line is used by starport scaffolding # genesis/module/export
```
vish-stride, line 68: nit: indent

x/stakeibc/keeper/hooks.go
```diff
@@ -53,20 +54,56 @@ func (k Keeper) BeforeEpochStart(ctx sdk.Context, epochIdentifier string, epochN
...
 		exchangeRateInterval := int64(k.GetParam(ctx, types.KeyExchangeRateInterval))
-		if epochNumber%exchangeRateInterval == 3 { // allow a few blocks from UpdateUndelegatedBal to avoid conflicts
-			// TODO(TEST-97) update only when balances, delegatedBalances and stAsset supply are results from the same block
-			k.ProcessUpdateDelegatedBalance(ctx)
-		}
-		if epochNumber%exchangeRateInterval == 6 && (epochNumber > 100) { // allow a few blocks from UpdateDelegatedBal to avoid conflicts & wait until chain has registered zones to calc exch rate
-			// TODO(TEST-97) update only when balances, delegatedBalances and stAsset supply are results from the same block
-			k.ProcessUpdateExchangeRate(ctx)
-		}
+		if epochNumber%exchangeRateInterval == 0 && (epochNumber > 100) { // allow a few blocks from UpdateUndelegatedBal to avoid conflicts
```
vish-stride, line 66: agreed, we should wrap this into a function in the keeper which is called here 

why wait until 100 epochs have passed? is this to allow the channels / connections to be setup properly?

x/stakeibc/keeper/hooks.go
```diff
@@ -53,20 +54,56 @@ func (k Keeper) BeforeEpochStart(ctx sdk.Context, epochIdentifier string, epochN
...
+			// GET LATEST HEIGHT
+			// TODO(NOW) wrap this into a function
+			var latestHeightGaia int64 // defaults to 0
+			// get light client's latest height
+			connectionID := "connection-0"
```
vish-stride, line 71: imo we should loop through all the registered host zones here, and then skip the host zone is delegation account isn't defined yet (which should get rid of the issue where we have to wait 100 epochs)

curious what you think though!

x/stakeibc/keeper/hooks.go
```diff
@@ -53,20 +54,56 @@ func (k Keeper) BeforeEpochStart(ctx sdk.Context, epochIdentifier string, epochN
...
+			conn, found := k.IBCKeeper.ConnectionKeeper.GetConnection(ctx, connectionID)
+			if !found {
+				k.Logger(ctx).Info(fmt.Sprintf("invalid connection id, \"%s\" not found", connectionID))
```
vish-stride, line 74: in the above "host zone loop" we should add a "continue" here, so we don't keep processing this hostzone

vish-stride (approved):
overall looks good to me! I see there are a few TODOs added here (e.g. generalize host zone, process only when updates happen in the same block, etc) that might be good to just throw into this PR 

small comments on the next steps: 
1. What is this ICQ panic 60000000000 error? That's pretty weird. 100% agree we should get to the bottom of this (doesn't have to be in this PR ofc), restarting the relayer feels more like a bandaid solution
2. What do you mean by `Before grabbing latest height from the gaia LC in hooks.go, update it!`

riley-stride:
> overall looks good to me! I see there are a few TODOs added here (e.g. generalize host zone, process only when updates happen in the same block, etc) that might be good to just throw into this PR
> 
> small comments on the next steps:
> 
> 1. What is this ICQ panic 60000000000 error? That's pretty weird. 100% agree we should get to the bottom of this (doesn't have to be in this PR ofc), restarting the relayer feels more like a bandaid solution
> 2. What do you mean by `Before grabbing latest height from the gaia LC in hooks.go, update it!`

1. Icq panic 60000000000 is "timed out waiting for tx to process"
2. I mean: update the state of the gaia light client on stride to the latest block height

x/stakeibc/keeper/hooks.go
```diff
@@ -53,20 +54,56 @@ func (k Keeper) BeforeEpochStart(ctx sdk.Context, epochIdentifier string, epochN
...
+			}
+			//TODO(TEST-112) make sure to update host LCs here!
+			clientState, found := k.IBCKeeper.ClientKeeper.GetClientState(ctx, conn.ClientId)
+			if !found {
+				k.Logger(ctx).Info(fmt.Sprintf("client id \"%s\" not found for connection \"%s\"", conn.ClientId, connectionID))
+				// latestHeightGaia = 0
+			} else {
+				// TODO(TEST-119) get stAsset supply at SAME time as gaia height
+				// TODO(TEST-112) check on safety of castng uint64 to int64
+				latestHeightGaia = int64(clientState.GetLatestHeight().GetRevisionHeight())
```
asalzmann: What's the revision height?

x/stakeibc/keeper/hooks.go
```diff
@@ -53,20 +54,56 @@ func (k Keeper) BeforeEpochStart(ctx sdk.Context, epochIdentifier string, epochN
...
 
+				// TODO(119) generalize to host_zones
+				// SET STASSETSUPPLY
+				hz, _ := k.GetHostZone(ctx, "GAIA")
+				currStSupply := k.bankKeeper.GetSupply(ctx, "st"+hz.HostDenom)
```
asalzmann, line 77: Use `StAssetDenomFromHostZoneDenom` here

x/stakeibc/keeper/hooks.go
```diff
@@ -53,20 +54,56 @@ func (k Keeper) BeforeEpochStart(ctx sdk.Context, epochIdentifier string, epochN
...
+				// GET MODULE ACCT BALANCE
+				addr := k.accountKeeper.GetModuleAccount(ctx, types.ModuleName).GetAddress()
+				modAcctBal := k.bankKeeper.GetBalance(ctx, addr, hz.IBCDenom)
+
+				ControllerBalancesRecord := types.ControllerBalances{
+					Index:             strconv.FormatInt(latestHeightGaia, 10),
```
asalzmann, line 83: Will this work with multiple host zones? E.g. we create a ControllerBalancesRecord for gaia at height 10 and the same for osmosis at height 10

asalzmann: You might need something like f"{zoneId}.{height}"

x/stakeibc/keeper/msg_server_submit_tx.go
```diff
@@ -181,49 +181,147 @@ func (k Keeper) ProcessUpdateUndelegatedBalance(ctx sdk.Context) {
...
+	}
+	updateBalances := func(ctx sdk.Context, index int64, zoneInfo types.HostZone) error {
+		k.Logger(ctx).Info(fmt.Sprintf("\tUpdating balances on %s", zoneInfo.ChainId))
+
+		delegationIca := zoneInfo.GetDelegationAccount()
+		if delegationIca == nil || delegationIca.Address == "" {
+			k.Logger(ctx).Error("Zone %s is missing a delegation address!", zoneInfo.ChainId)
+			return errors.New("Zone is missing a delegation address!")
+		}
+		var queryDBCBDaisyChain icqkeeper.Callback = func(icqk icqkeeper.Keeper, ctx sdk.Context, args []byte, query icqtypes.Query, h int64) error {
```
asalzmann, line 266: nit: expand this name - queryDBCBDaisyChain is hard to understand / read

x/stakeibc/keeper/msg_server_submit_tx.go
```diff
@@ -181,49 +181,147 @@ func (k Keeper) ProcessUpdateUndelegatedBalance(ctx sdk.Context) {
...
+
+			// Add up delegations across validators
+			delegatorSum := sdk.NewCoin(zoneInfo.HostDenom, sdk.ZeroInt())
+			for _, delegation := range response.DelegationResponses {
+				delegatorSum = delegatorSum.Add(delegation.Balance)
+				if err != nil {
+					return err
+				}
+			}
+			// --- Update Undelegated Balance ---
```
asalzmann, line 283: Should this read "Update Delegated Balance"? Possible I'm missing something but it seems like we're updating the delegated balance in the following lines

x/stakeibc/keeper/msg_server_submit_tx.go
```diff
@@ -181,49 +181,147 @@ func (k Keeper) ProcessUpdateUndelegatedBalance(ctx sdk.Context) {
...
+			hz, _ := k.GetHostZone(ctx, zoneInfo.ChainId)
+			da := hz.DelegationAccount
+			// TODO() make HostZone.DelegationAccount.Balance a Cosmos.Dec type (rather than int)
+			// only update height if this is the most updated query (ICQ msgresponses are not always FIFO)
+			if h >= da.HeightLastQueriedDelegatedBalance {
```
asalzmann, line 288: Why is this check necessary?

asalzmann (commented):
Looking good! Main comments are

- I think we should bundle generalizing this to work for multiple host zones into this PR
- We should review ProcessUpdateDelegatedBalance and ProcessUpdateUndelegatedBalance once the code is one notch cleaner / better documented and named. It's a bit tricky to follow the code path and it's quite sensitive!

Overall, the approach makes sense.

We should review the ICQ spec and verify that specifying a block height is included (wouldn't be the end of the world if we drift from the ICQ spec, but probably also not great long-term)

x/stakeibc/keeper/hooks.go
asalzmann: I think this is partially resolved by [45e6333](https://github.com/Stride-Labs/stride/pull/47/commits/45e6333c258b2dcfcc0ba273b301d3977ca1c17c), we still need to generalize this to multiple host zones though!

x/stakeibc/client/cli/query_controller_balances_test.go
riley-stride: not sure, it was scaffolded by ignite

riley-stride: I removed it and the proj builds OK

x/stakeibc/keeper/msg_server_submit_tx.go
```diff
@@ -164,13 +164,13 @@ func (k Keeper) ReinvestRewards(ctx sdk.Context, hostZone types.HostZone) error
 	// 1. query withdraw account balances using icq
 	// 2. transfer withdraw account balances to the delegation account in the cb
 	// 3. TODO: in the ICA ack upon transfer, reinvest those rewards and withdraw rewards
-	k.InterchainQueryKeeper.QueryBalances(ctx, hostZone, cb, withdrawAccount.Address)
+	k.InterchainQueryKeeper.QueryBalances(ctx, hostZone, cb, withdrawAccount.Address, 0)
```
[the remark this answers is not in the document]
riley-stride: it means latest height, added a comment to this effect

scripts/logs/create_logs.sh
riley-stride: it's already nested under logs

x/stakeibc/keeper/hooks.go
riley-stride: yes it allows setup

x/stakeibc/keeper/controller_balances.go
```diff
@@ -0,0 +1,63 @@
...
+func (k Keeper) SetControllerBalances(ctx sdk.Context, controllerBalances types.ControllerBalances) {
+	store :=  prefix.NewStore(ctx.KVStore(k.storeKey), types.KeyPrefix(types.ControllerBalancesKeyPrefix))
+	b := k.cdc.MustMarshal(&controllerBalances)
+	store.Set(types.ControllerBalancesKey(
+        controllerBalances.Index,
+    ), b)
+}
+
+// GetControllerBalances returns a controllerBalances from its index
+func (k Keeper) GetControllerBalances(
```
vish-stride, line 19: nit: this might just be my editor, but the spacing/indents in this file seem a bit off

x/stakeibc/keeper/hooks.go
```diff
@@ -53,20 +53,29 @@ func (k Keeper) BeforeEpochStart(ctx sdk.Context, epochIdentifier string, epochN
...
-		if epochNumber%exchangeRateInterval == 3 { // allow a few blocks from UpdateUndelegatedBal to avoid conflicts
-			// TODO(TEST-97) update only when balances, delegatedBalances and stAsset supply are results from the same block
-			k.ProcessUpdateDelegatedBalance(ctx)
-		}
-		if epochNumber%exchangeRateInterval == 6 && (epochNumber > 100) { // allow a few blocks from UpdateDelegatedBal to avoid conflicts & wait until chain has registered zones to calc exch rate
-			// TODO(TEST-97) update only when balances, delegatedBalances and stAsset supply are results from the same block
-			k.ProcessUpdateExchangeRate(ctx)
+		if epochNumber%exchangeRateInterval == 0 && (epochNumber > 50) { // allow a few blocks from UpdateUndelegatedBal to avoid conflicts
+			for _, hz := range k.GetAllHostZone(ctx) {
+				// TODO(NOW) update LC before getting latest height
```
vish-stride, line 66: do we need to add a check here that the host zone has a delegation account?

vish-stride (approved):
Overall, this looks good to me. Apologies for the many `nit` comments. Really cool to see this coming together, it's an incredible feature (and one that should be obviously useful to anyone in crypto)

One thing I haven't worked through in my head is what happens if the second leg of the daisy chain takes a persistently long time to go through? E.g. if all of the 1st legs come back before the previous 2nd leg finished. If it's helpful, I'm down to walk through any race conditions together (good chance you've already done this, if so please ignore!). Regardless, I don't think this should be an issue in prod, given the long epoch times, but might lead to some strange results in testing

PR #48, opened 2022-06-22
[TEST-95] Local Mode Development Option

sampocs:
**Summary**
* Added option to run Stride/Gaia/Hermes/Icq nodes locally to speed up dev cycles (i.e. eliminate excessive build time from docker overhead)
* This only runs 1 node for each chain. It is a less robust testing interface than docker, but it should be useful for tough debugging scenarios that require quick trial and error
* Each dependency is added a git submodule under `deps`
* The flow is separated into 3 steps
  1. `install`: download dependencies for each module 
  2. `init`: compile executables (configurable with build flag) and initialize chains, accounts, keys and other state
  4.  `start`: first starts Stride and Gaia, then starts Hermes and ICQ
* The relayer's listen to the logs of Stride and Gaia and launch intelligently once blocks have been confirmed (to cut down on time wasted from having a conservative `sleep` command)
* The flow from `init` through `start` should take ~2 min
* See README for run instructions
* First time flow should be:
```bash
git submodule update --init
make local-install
make local-init build=sghi
make stop
```
* Successive flow should generally be:
```bash
make local-init build={target_module}
make stop
```
* Optionally pass `cache=true` to restore from a backed-up set of state files
```
make local-init build={target_module} cache=true
make stop
```

**Remaining Work:**

- [X]  Host zone doesn't seem to be registering (hoping this is just a timing thing that needs to get fixed)
  * **EDIT: It was a timing issue. New commit delays startup until Hermes handshake finishes.**
- [X] Create local copies of docker testing scripts `1.sh`, `2.sh`, `3.sh`
- [X]  Considering restructuring the organizational layout of these scripts and relevant Makefile commands 
    * I intentionally isolated this code/commands to get something out there quick
    * My proposal is, if we find value with this approach, we can integrate it into the original test scripts as a flag 
    * **EDIT: Moved it into a `scripts-local` folder for now (slightly cleaner then having it nested)**
- [X] I think there's a chance we can restore the state from the point after relayer connections have been made on Stride and Gaia. This would cut down the dev cycle from 2min to <10 seconds. My initial attempt (keeping a copy of the state directory right after initialization) doesn't seem to be working, and if I can't fix it, I'll remove that option from the repo before merging.
   * **EDIT: This is working now! Cycles should be ~20 seconds when utilizing the cache.**
   
**Future Work**
* Add error checking (right now you have to check the logs to see if the chain or relayers have failed)

**Testing Plan**
* Follow the flow described above to setup
* Run testing scripts


scripts/local/init_gaia.sh
```diff
@@ -0,0 +1,55 @@
...
+## add the message types ICA should allow to the host chain
+ALLOW_MESSAGES='\"/cosmos.bank.v1beta1.MsgSend\", \"/cosmos.bank.v1beta1.MsgMultiSend\", \"/cosmos.staking.v1beta1.MsgDelegate\", \"/cosmos.staking.v1beta1.MsgUndelegate\", \"/cosmos.staking.v1beta1.MsgRedeemTokensforShares\", \"/cosmos.staking.v1beta1.MsgTokenizeShares\", \"/cosmos.distribution.v1beta1.MsgWithdrawDelegatorReward\", \"/cosmos.distribution.v1beta1.MsgSetWithdrawAddress\", \"/ibc.applications.transfer.v1.MsgTransfer\"'
+sed -i -E "s|\"allow_messages\": \[\]|\"allow_messages\": \[${ALLOW_MESSAGES}\]|g" "${STATE}/${GAIA_NODE_NAME}/config/genesis.json"
+
+# Update ports so they don't conflict with the stride chain
+sed -i -E "s|9090|9080|g" "${STATE}/${GAIA_NODE_NAME}/config/app.toml"
+sed -i -E "s|9091|9081|g" "${STATE}/${GAIA_NODE_NAME}/config/app.toml"
+sed -i -E "s|26657|26557|g" "${STATE}/${GAIA_NODE_NAME}/config/client.toml"
+sed -i -E "s|26656|26556|g" "${STATE}/${GAIA_NODE_NAME}/config/config.toml"
+sed -i -E "s|26657|26557|g" "${STATE}/${GAIA_NODE_NAME}/config/config.toml"
```
vish-stride, line 53: I think this line might be repeated?

scripts/local/start_network.sh
```diff
@@ -0,0 +1,65 @@
...
+ICQ_LOGS=$SCRIPT_DIR/logs/icq.log
+
+if [ "$cache" == "true" ]; then
+    echo "Restoring from cache..."
+    rm -rf $SCRIPT_DIR/state 
+    mv $SCRIPT_DIR/.state.backup $SCRIPT_DIR/state
+fi
+
+printf '%s' "Starting Stride and Gaia...   "
+nohup $STRIDE_CMD start | sed -r -u "s/\x1B\[([0-9]{1,3}(;[0-9]{1,2})?)?[mGK]//g" > $STRIDE_LOGS 2>&1 &
```
vish-stride, line 26: are these the lines that wait until blocks are committed? if so, very cool! 

what does the sed command do on this line? and does the `tail` line below wait until the grep command returns something? i would have thought this requires a while loop, v neat if not

scripts/local/start_network.sh
```diff
@@ -0,0 +1,65 @@
...
+nohup $ICQ_CMD run --local > $ICQ_LOGS 2>&1 &
+echo "Done"
+
+echo "Network is ready for transactions."
+cp -r $SCRIPT_DIR/state $SCRIPT_DIR/.state.backup
+
+echo "Creating host zone..."
+ATOM='uatom'
+IBCATOM='ibc/27394FB092D2ECCD56123C74F36E4C1F926001CEADA9CA97EA622B25F41E5EB2'
+$STRIDE_CMD tx stakeibc register-host-zone \
```
vish-stride, line 61: you said this is throwing an error? what do you get when you run this line?

sampocs:
@vish-stride I was expecting it to work as well! Stride and Gaia actually started up fine at their previous block heights. The issue was with hermes. I also thought it was possible hermes was storing state somewhere else but the only thing I could find is that it stores the keys in `~/.hermes`, which I don't think would be connected.
But yeah it would be great to walk through it with you!

scripts/local/init_gaia.sh
sampocs: one's in **client**.toml one's in **config**.toml (always trips me up as well haha)

scripts/local/start_network.sh
sampocs: I'm seeing this from the staking module in stride

```
Zone %s is missing a delegation address! module=x/stakeibc
```

PR #49, opened 2022-06-22
[TEST-104] Validator Selection

vish-stride:
**Summary**

This PR adds validator selection! A quick high level summary of features:
1. Ability to add, remove, and update weight of validators. Added "weight" feature to validators. These are all added as msgs.
2. Validator re-balance added under function `RebalanceValidators` in `keepers/msg_server_rebalance_validators`
3. Upon staking, we automatically stake the right amount for each validator. 

Strictly speaking, we don't _need_ the rebalance function (we can instead do it through some redelegation messages in the gov module), but I think slightly safer / cleaner to have. 

**Test plan**

Right now, I'm running scripts/tests/validator_selection and observing the output, but will migrate these to BATS now. There are a lot of edge cases we should add to the tests


x/stakeibc/keeper/msg_server_rebalance_validators.go
```diff
@@ -2,16 +2,92 @@ package keeper
...
+
+	for i := 1; i < 6; i++ {
+		underWeightElem := valDeltaList[underWeightIndex]
+		overWeightElem := valDeltaList[overWeightIndex]
+		if (underWeightElem.deltaAmt < 0) == (overWeightElem.deltaAmt < 0) {
+			// if the two elements have the same sign, we stop changing weights
+			break
+		}
+		if abs(underWeightElem.deltaAmt) > abs(overWeightElem.deltaAmt) {
+			// if the underweight element is more overweight than the overweight element
```
sampocs, line 66: nit: I think you mean "is more underweight" here?

Might be a little more clear to say something like: 
"If the underweight element requires more rebalancing than the overweight element"

x/stakeibc/keeper/msg_server_rebalance_validators.go
```diff
@@ -2,16 +2,92 @@ package keeper
...
+	sort.Slice(valDeltaList, lessFunc)
+	// now varDeltaList is sorted by deltaAmt
+	underWeightIndex := 0
+	overWeightIndex := len(valDeltaList) - 1
+
+	var msgs []sdk.Msg
+	delegatorAddressStr := hostZone.DelegationAccount.Address
+	delegatorAddress := sdk.AccAddress(delegatorAddressStr)
+
+	for i := 1; i < 6; i++ {
```
sampocs, line 58: nit: should the 6 be split out as a config variable?

x/stakeibc/keeper/msg_server_rebalance_validators.go
```diff
@@ -2,16 +2,92 @@ package keeper
...
 )
 
+func abs(n int64) int64 {
+	if n < 0 {
+		return -n
+	}
+	return n
+}
+
 func (k msgServer) RebalanceValidators(goCtx context.Context, msg *types.MsgRebalanceValidators) (*types.MsgRebalanceValidatorsResponse, error) {
```
sampocs, line 21: this is so clean! nicely done!!!

x/stakeibc/keeper/msg_server_rebalance_validators.go
```diff
@@ -2,16 +2,92 @@ package keeper
...
+			underWeightElem.deltaAmt += overWeightElem.deltaAmt
+			overWeightIndex -= 1
+			// issue an ICA call to the host zone to rebalance the validator
+			redelagateMsg := stakingTypes.NewMsgBeginRedelegate(
+				delegatorAddress,
+				underWeightElem.valAddr,
+				overWeightElem.valAddr,
+				sdk.NewInt64Coin(hostZone.HostDenom, abs(overWeightElem.deltaAmt)))
+			msgs = append(msgs, redelagateMsg)
+			overWeightElem.deltaAmt = 0 // can we just drop this line?
```
sampocs, line 76: i think you can drop it

x/stakeibc/keeper/validator_selection.go
```diff
@@ -0,0 +1,46 @@
...
+	valWeights := out[hostZone.ChainId]
+	totalWeight := int64(0)
+	totalDelegatedWeight := k.GetTotalValidatorWeight()
+	for _, weight := range valWeights.ValidatorWeights {
+		totalWeight += weight
+	}
+	var targetWeight map[string]int64
+	for valAddr, weight := range valWeights.ValidatorWeights {
+		targetWeight[valAddr] = int64(float64(weight*totalDelegatedWeight) / float64(totalWeight))
+	}
```
sampocs, line 50: I'm a little confused on "weight" vs "amt" (might just be me). Is "amt" the current weight and "weight" is the desired weight? If so, might not be a bad idea to rename accordingly

sampocs (commented):
Looks great, good stuff!!

x/stakeibc/keeper/msg_server_rebalance_validators.go
sampocs: (I meant the function as a whole)

x/stakeibc/keeper/msg_server_rebalance_validators.go
```diff
@@ -0,0 +1,99 @@
...
+	overWeightIndex := len(valDeltaList) - 1
+
+	var msgs []sdk.Msg
+	delegatorAddressStr := hostZone.DelegationAccount.Address
+	delegatorAddress := sdk.AccAddress(delegatorAddressStr)
+
+	for i := 1; i < 6; i++ {
+		underWeightElem := valDeltaList[underWeightIndex]
+		overWeightElem := valDeltaList[overWeightIndex]
+		if (underWeightElem.deltaAmt < 0) == (overWeightElem.deltaAmt < 0) {
```
asalzmann, line 61: I think there's another invariant here which is the underWeightElem *must* be <= 0 and the overWeightElem *must* be >= 0, wdyt of explicitly checking this?

x/stakeibc/keeper/msg_server_rebalance_validators.go
asalzmann: Think it's slightly safer to keep it, although we shouldn't need it

asalzmann (commented):
Looking awesome!! Added a few minor comments, but the approach overall seems sound. Two other comments are
- I think we had discussed adding a check to RebalanceValidators so that the function only executes if the weights are off by > X%, do you think this is worth adding?
- If we're not adding BATS / unittests, can you add some detail to Test Plan around the inputs / outputs you're using?

proto/stakeibc/tx.proto
```diff
@@ -18,6 +18,10 @@ service Msg {
       rpc SubmitTx(MsgSubmitTx) returns (MsgSubmitTxResponse);
       // TODO(TEST-53): Remove this pre-launch (no need for clients to create / interact with ICAs)
       rpc RegisterHostZone(MsgRegisterHostZone) returns (MsgRegisterHostZoneResponse);
+  rpc RebalanceValidators(MsgRebalanceValidators) returns (MsgRebalanceValidatorsResponse);
+  rpc AddValidator(MsgAddValidator) returns (MsgAddValidatorResponse);
```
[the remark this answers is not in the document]
vish-stride: i think we should keep them, but restrict them to the governance module or a multisig that we control

x/stakeibc/keeper/host_zone.go
```diff
@@ -92,6 +93,32 @@ func (k Keeper) GetAllHostZone(ctx sdk.Context) (list []types.HostZone) {
...
+
+func (k Keeper) RemoveValidatorFromHostZone(ctx sdk.Context, chainId string, validatorName string) (success bool) {
+	hostZone, found := k.GetHostZone(ctx, chainId)
+	if !found {
+		k.Logger(ctx).Error(fmt.Sprintf("HostZone not found %s", chainId))
+		return false
+	}
+	for i, val := range hostZone.Validators {
+		if val.Name == validatorName {
+			hostZone.Validators = append(hostZone.Validators[:i], hostZone.Validators[i+1:]...)
```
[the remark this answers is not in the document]
vish-stride: actually I think this shouldn't error! see here: https://go.dev/play/p/MFXqfUbriCm

x/stakeibc/keeper/msg_server_add_validator.go
```diff
@@ -0,0 +1,42 @@
...
+		k.Logger(ctx).Info(fmt.Sprintf("Host Zone not found %s", msg.HostZone))
+	}
+	validators := hostZone.Validators
+	// check that we don't already have this validator
+	for _, validator := range validators {
+		if validator.Address == msg.Address {
+			k.Logger(ctx).Info(fmt.Sprintf("Validator address %s already exists on Host Zone %s", msg.Address, msg.HostZone))
+			return nil, types.ErrValidatorAlreadyExists
+		}
+		if validator.Name == msg.Name {
```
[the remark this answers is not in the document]
vish-stride: We actually don't need validator name to be unique. The only place it's used right now is to remove validators from the host zone (which can easily be changed to remove by address instead). Do you think better to not enforce Name uniqueness?

x/stakeibc/keeper/validator_selection.go
```diff
@@ -0,0 +1,70 @@
...
+		k.Logger(ctx).Error(fmt.Sprintf("Error getting target weights for host zone %s", hostZone.ChainId))
+		return nil, err
+	}
+	for _, validator := range validators {
+		scaled_weights[validator.Address] = int64(target_weights[validator.Address]) - int64(validator.DelegationAmt)
+	}
+	return scaled_weights, nil
+}
+
+func (k Keeper) GetTargetValAmtsForHostZone(ctx sdk.Context, hostZone types.HostZone, finalDelegation uint64) (map[string]uint64, error) {
```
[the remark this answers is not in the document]
vish-stride: yeah the names here are definitely confusing

the weight is the sum of the "target weights", but finalDelegation is what we want the output of this function to sum up to. 

e.g. if we have 3 validators, with weights A:5, B:5, C:10, and finalDelegation=40, this function would return A:10, B:10, C:20

should we rename the variables?

x/stakeibc/keeper/validator_selection.go
```diff
@@ -0,0 +1,70 @@
...
+		return nil, types.ErrNoValidatorWeights
+	}
+	var targetWeight map[string]uint64
+	allocatedAmt := uint64(0)
+	for i, validator := range hostZone.Validators {
+		if i == len(hostZone.Validators)-1 {
+			// for the last element, we need to make sure that the allocatedAmt is equal to the finalDelegation
+			targetWeight[validator.Address] = finalDelegation - allocatedAmt
+		} else {
+			delegateAmt := uint64(float64(validator.Weight*finalDelegation) / float64(totalWeight))
```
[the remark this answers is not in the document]
vish-stride: Ah, great catch! On line 17, we should be fetching the total delegation amt, not the total weight

x/stakeibc/keeper/validator_selection.go
```diff
@@ -0,0 +1,70 @@
...
+	validators := hostZone.GetValidators()
+	scaled_weights := make(map[string]int64)
+	totalWeight := k.GetTotalValidatorWeight(ctx, hostZone)
+	target_weights, err := k.GetTargetValAmtsForHostZone(ctx, hostZone, totalWeight)
+	if err != nil {
+		k.Logger(ctx).Error(fmt.Sprintf("Error getting target weights for host zone %s", hostZone.ChainId))
+		return nil, err
+	}
+	for _, validator := range validators {
+		scaled_weights[validator.Address] = int64(target_weights[validator.Address]) - int64(validator.DelegationAmt)
```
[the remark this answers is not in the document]
vish-stride: no that's 100% right! the comment is totally wrong, I think I forgot to finish what I was trying to say

x/stakeibc/keeper/validator_selection.go
```diff
@@ -0,0 +1,70 @@
...
+	if finalDelegation == 0 {
+		k.Logger(ctx).Error(fmt.Sprintf("Cannot calculate target delegation if final amount is 0 %s", hostZone.ChainId))
+		return nil, types.ErrNoValidatorWeights
+	}
+	var targetWeight map[string]uint64
+	allocatedAmt := uint64(0)
+	for i, validator := range hostZone.Validators {
+		if i == len(hostZone.Validators)-1 {
+			// for the last element, we need to make sure that the allocatedAmt is equal to the finalDelegation
+			targetWeight[validator.Address] = finalDelegation - allocatedAmt
```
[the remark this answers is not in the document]
vish-stride: there just might be rounding issues with the way the weights are handled. E.g. if I have 3 validators with weights A:5, B:5, C:5, and need to allocate 100 tokens, I don't want to do 33, 33, 33, but instead 33, 33, 34 (the extra token could go to any of the validators, we're arbitrarily picking the last one)

x/stakeibc/keeper/msg_server_rebalance_validators.go
```diff
@@ -0,0 +1,99 @@
...
+				delegatorAddress,
+				underWeightElem.valAddr,
+				overWeightElem.valAddr,
+				sdk.NewInt64Coin(hostZone.HostDenom, abs(overWeightElem.deltaAmt)))
+			msgs = append(msgs, redelagateMsg)
+			overWeightElem.deltaAmt = 0 // can we just drop this line?
+		} else {
+			// if the overweight element is more overweight than the underweight element
+			overWeightElem.deltaAmt -= underWeightElem.deltaAmt
+			underWeightIndex += 1
```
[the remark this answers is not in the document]
vish-stride: i don't think we strictly need to do this (i believe what happens now is the next 2 loops of the rebalancing just don't actually move tokens around), but i agree we should add it in regardless

x/stakeibc/keeper/validator_selection.go
```diff
@@ -0,0 +1,46 @@
...
+	validators := hostZone.GetValidators()
+	scaled_weights := make(map[string]int64)
+	target_weights := k.GetTargetValAmtsForHostZone(ctx, hostZone)
+	for _, validator := range validators {
+		scaled_weights[validator.Address] = target_weights[validator.Address] - validator.DelegationAmt
+	}
+	return scaled_weights
+}
+
+func (k Keeper) GetTargetValAmtsForHostZone(ctx sdk.Context, hostZone types.HostZone) map[string]int64 {
```
riley-stride: let's add a TODO to comment with a reference to the algo we're using

x/stakeibc/keeper/validator_selection.go
riley-stride: good call, I agree we should at least comment to make this clear. i assume weight is a decimal \in[0,1] and amt=w*total

x/stakeibc/keeper/msg_server_rebalance_validators.go
riley-stride: agree with both

x/stakeibc/client/cli/tx_add_validator.go
```diff
@@ -0,0 +1,54 @@
...
+	"github.com/cosmos/cosmos-sdk/client"
+	"github.com/cosmos/cosmos-sdk/client/flags"
+	"github.com/cosmos/cosmos-sdk/client/tx"
+	"github.com/spf13/cast"
+	"github.com/spf13/cobra"
+)
+
+var _ = strconv.Itoa(0)
+
+func CmdAddValidator() *cobra.Command {
```
riley-stride, line 16: makes sense to expose this for testing -- should we remove this for testnet and mainnet?

riley-stride: or restrict to our account as caller, please add TODO for that

x/stakeibc/keeper/msg_server_submit_tx.go
```diff
@@ -81,12 +82,23 @@ func (k Keeper) DelegateOnHost(ctx sdk.Context, hostZone types.HostZone, amt sdk
...
+	targetDelegatedAmts, err := k.GetTargetValAmtsForHostZone(ctx, hostZone, amt.Amount.Uint64())
+	if err != nil {
+		k.Logger(ctx).Error(fmt.Sprintf("Error getting target delegation amounts for host zone %s", hostZone.ChainId))
+		return err
+	}
+	for _, validator := range hostZone.GetValidators() {
+		relAmt := sdk.NewCoin(amt.Denom, sdk.NewIntFromUint64(targetDelegatedAmts[validator.Address]))
+		msgs = append(msgs, &stakingTypes.MsgDelegate{
+			DelegatorAddress: delegationIca.GetAddress(),
+			ValidatorAddress: validator.GetAddress(),
```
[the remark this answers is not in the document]
asalzmann: Hmm, this is a bit unfortunate. Might be worth looking into - wouldn't be great for one validator going down to halt all delegations

x/stakeibc/handler.go
```diff
@@ -32,6 +32,18 @@ func NewHandler(k keeper.Keeper) sdk.Handler {
...
+			return sdk.WrapServiceResult(ctx, res, err)
+		case *types.MsgAddValidator:
+			res, err := msgServer.AddValidator(sdk.WrapSDKContext(ctx), msg)
+			return sdk.WrapServiceResult(ctx, res, err)
+		case *types.MsgChangeValidatorWeight:
+			res, err := msgServer.ChangeValidatorWeight(sdk.WrapSDKContext(ctx), msg)
+			return sdk.WrapServiceResult(ctx, res, err)
+		case *types.MsgDeleteValidator:
+			res, err := msgServer.DeleteValidator(sdk.WrapSDKContext(ctx), msg)
+			return sdk.WrapServiceResult(ctx, res, err)
```
riley-stride, line 46: add TODO to restrict all these to be callable only by our account

x/stakeibc/keeper/host_zone.go
```diff
@@ -92,6 +93,32 @@ func (k Keeper) GetAllHostZone(ctx sdk.Context) (list []types.HostZone) {
 	return
 }
 
+func (k Keeper) AddValidatorToHostZone(ctx sdk.Context, chainId string, val types.Validator) (success bool) {
+	hostZone, found := k.GetHostZone(ctx, chainId)
+	if !found {
+		k.Logger(ctx).Error(fmt.Sprintf("HostZone not found %s", chainId))
+		return false
+	}
+	hostZone.Validators = append(hostZone.Validators, &val)
```
riley-stride, line 102: should we check whether it already exists and handle that accordingly, maybe "editValidator" if we try to add some with address equal to one that already exists?

x/stakeibc/keeper/msg_server_add_validator.go
```diff
@@ -0,0 +1,42 @@
...
+			k.Logger(ctx).Info(fmt.Sprintf("Validator name %s already exists on Host Zone %s", msg.Name, msg.HostZone))
+			return nil, types.ErrValidatorAlreadyExists
+		}
+	}
+	// add the validator
+	hostZone.Validators = append(validators, &types.Validator{
```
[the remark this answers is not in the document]
riley-stride: looks like that happens below

x/stakeibc/keeper/msg_server_add_validator.go
```diff
@@ -0,0 +1,42 @@
...
+		Name:           msg.Name,
+		Address:        msg.Address,
+		Status:         types.Validator_Active,
```
riley-stride, line 35: Is this field intended to represent the `active` status from the host chain? if so, should we be getting the status from there (if the validator is not actually active we prob should not)

x/stakeibc/keeper/host_zone.go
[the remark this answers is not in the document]
asalzmann: Woah did not know you could save + link code snippets, that's awesome.

x/stakeibc/keeper/msg_server_change_validator_weight.go
```diff
@@ -0,0 +1,30 @@
...
+func (k msgServer) ChangeValidatorWeight(goCtx context.Context, msg *types.MsgChangeValidatorWeight) (*types.MsgChangeValidatorWeightResponse, error) {
+	ctx := sdk.UnwrapSDKContext(goCtx)
+
+	hostZone, found := k.GetHostZone(ctx, msg.HostZone)
+	if !found {
+		k.Logger(ctx).Error(fmt.Sprintf("Host Zone %s not found", msg.HostZone))
+		return nil, types.ErrInvalidHostZone
+	}
+	validators := hostZone.Validators
+	for _, validator := range validators {
```
riley-stride, line 20: Can / should we store validators in a map and index them with address, rather than iterate them every time? Might not a be a big difference, maybe best to add a TODO for later

x/stakeibc/keeper/msg_server_add_validator.go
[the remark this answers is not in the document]
asalzmann: agreed

x/stakeibc/keeper/msg_server_rebalance_validators.go
```diff
@@ -0,0 +1,128 @@
...
+	lessFunc := func(i, j int) bool {
+		return valDeltaList[i].deltaAmt < valDeltaList[j].deltaAmt
+	}
+	sort.Slice(valDeltaList, lessFunc)
+	// now varDeltaList is sorted by deltaAmt
+	underWeightIndex := 0
+	overWeightIndex := len(valDeltaList) - 1
+
+	var msgs []sdk.Msg
+	delegatorAddressStr := hostZone.DelegationAccount.Address
```
riley-stride, line 60: nit: getDelegationAccount() and getAddress() -- check for this throughout pls

x/stakeibc/keeper/msg_server_rebalance_validators.go
```diff
@@ -0,0 +1,128 @@
...
+	delegatorAddress := sdk.AccAddress(delegatorAddressStr)
+
+	// max_delta = utils.abs(valDeltaList[underWeightIndex].deltaAmt / )
```
riley-stride, line 63: nit: rm

x/stakeibc/keeper/msg_server_rebalance_validators.go
```diff
@@ -0,0 +1,128 @@
...
+			// if overWeightElem is negative, we're done rebalancing
+			break
+		}
+		if abs(underWeightElem.deltaAmt) > abs(overWeightElem.deltaAmt) {
+			// if the underweight element is more overweight than the overweight element
+			// we transfer stake from the underweight element to the overweight element
+			underWeightElem.deltaAmt += overWeightElem.deltaAmt
+			overWeightIndex -= 1
+			// issue an ICA call to the host zone to rebalance the validator
+			redelagateMsg := stakingTypes.NewMsgBeginRedelegate(
```
riley-stride, line 86: nit: redelagateMsg => redel**e**gateMsg

x/stakeibc/keeper/validator_selection.go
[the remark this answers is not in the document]
riley-stride: can we call it totalWeight

x/stakeibc/client/cli/tx_add_validator.go
vish-stride: added a TODO for this! and we already can only add validators that haven't been added

x/stakeibc/keeper/validator_selection.go
```diff
@@ -0,0 +1,46 @@
...
+	var out map[string]types.ValidatorWeights
+	k.paramstore.Get(ctx, types.KeyHostZoneValidatorWeights, &out)
+	valWeights := out[hostZone.ChainId]
+	totalWeight := int64(0)
+	totalDelegatedWeight := k.GetTotalValidatorWeight()
+	for _, weight := range valWeights.ValidatorWeights {
+		totalWeight += weight
+	}
```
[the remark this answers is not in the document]
vish-stride: it's actually not guaranteed to sum to 1, we just normalize by the total weight

PR #50, opened 2022-06-24
[TEST-121] Add IBC Middleware module (records)

asalzmann:
**Summary**
We need to add an IBC middleware in order to properly update records. [IBC middleware](https://ibc.cosmos.network/main/ibc/middleware/develop.html) modules can wrap core IBC modules and other middlewares. Specifically, this adds a middleware stack to app.go with the following structure: `records -> transfer`. All ibc packets routed to the `transfer` module will first pass through `records`, where we can apply custom logic (record keeping) before passing messages to the underlying `transfer` module.

Note:
- The middleware stack is added in app.go
- The custom handler logic is added in ibc_module.go by implementing the IBCModule interface

**Test plan**
Run `sh scripts/tests/1.sh && sh scripts/tests/2.sh`, which triggers an IBC transfer from `stakeibc` (using the standard `transfer` module).

**Next steps**
Merge #46 into this PR, which contains the core record keeping logic

Check logs
```
stride-stride3-1  | [90m8:38AM[0m [1m[31mERR[0m[0m This is where we're going to add DepositRecords!
```


app/app.go
```diff
@@ -471,23 +471,26 @@ func NewStrideApp(
...
 		app.GetSubspace(recordsmoduletypes.ModuleName),
-		app.IBCKeeper.ChannelKeeper,
-		&app.IBCKeeper.PortKeeper,
 		scopedRecordsKeeper,
 	)
 	recordsModule := recordsmodule.NewAppModule(appCodec, app.RecordsKeeper, app.AccountKeeper, app.BankKeeper)
 
+	// create IBC stacks by combining middleware with base application
+	// recordsStack contains records -> transfer
+	recordsStack := recordsmodule.NewIBCModule(app.RecordsKeeper, transferIBCModule)
```
riley-stride, line 480: makes sense based on https://ibc.cosmos.network/main/ibc/middleware/integration.html

riley-stride: we don't need to pass in a third arg because we are only setting a single scopedKeeper, right? per https://ibc.cosmos.network/main/ibc/middleware/integration.html#:~:text=customIBCModule1%20%3A%3D%20custom.NewIBCModule(customKeeper%2C%20scopedKeeperCustom1%2C%20%22portCustom1%22)

riley-stride:
This looks good to me, mainly based on checking against the tutorial [here](https://ibc.cosmos.network/main/ibc/middleware/integration.html)!

vish-stride (approved):
Looks good to me. No real comments, seems like it matches the tutorial we walked through. I didn't look at the scaffolding code though, but if this compiles + runs properly, let's do it

app/app.go
asalzmann: I think you can optionally pass in a default port for the IBCModule to use, but we're just using "Transfer" in this case

PR #55, opened 2022-06-26
[TEST-125] Unbonding Flow

vish-stride:
**Summary**

This PR implements the unbonding flow. We cannot merge it in yet, there are a handful of other tasks that need to be completed that this is blocked on. 

The "flow" of redemption is laid out in `stakeibc/keeper/hooks.go` under the BeginEpoch hook for `day`. 

Summary of changes:
1. Create `UserRedemptionRecord` that tracks an individual user's unbonding.
2. Create `EpochUnbondingRecord` that tracks all unbondings for an Epoch, by Host Zone. Please note, this is a _different_ epoch than `stride_epoch`, epochs here are days. The logic being we upsample unbondings to the nearest day. Then, not all host zones are redeemed daily. We add a `unbondingFrequency` to Host Zones that tracks how many days we want to wait before issuing an unbonding (because we can only have 7 unbonds pending at any time). Please note that this parameter must be passed in when creating a hostZone. 
3.  Added `QueryUnbondingDelegation` to interchainquery's keeper, in order to query what the status is of our pending unbondings
4. Added a `redemptionAccount` ICAAccount to Host Zone
5. Added `ClaimUndelegatedTokens` msg to stakeibc. This lets anyone transfer a certain number of unbonded tokens back to the corresponding users. 
6. Modify `RedeemStake` on stakeibc. This will now update the right records when a user calls unstake. 
8. Create `EpochTracker` that tracks all epochs based on their key. E.g. epochTracker stores day, minute, and stride_epoch

TODO:
1. Validator selection on unbonding. This is blocked on the validator selection PR. I (Vishal) am moving to that right now, and will come back and implement that validator selection here
2. Properly parsing the ICQ unbonding call. Right now the ICQ call is made in `keeper/unbonding_records.go:SweepAllUnbondedTokens`. However, we don't parse the output right now. We need to parse the output to determine which records are finished, sweep them, and then on the ack properly cleanup the records. 
3. Add a `ClaimUndelegatedTokensAddress` function that claims tokens just for one address. 
4. Burn stAssets on ack from unbonding

**Test plan**


x/stakeibc/keeper/msg_server_redeem_stake.go
```diff
@@ -13,73 +15,98 @@ import (
...
 	}
-	// StAssetDenom is valid
-	// Should we register stAssets somewhere and add an additional check here?
-	if types.IsStAsset(msg.StAssetDenom) != true {
-		return nil, sdkerrors.Wrapf(sdkerrors.ErrInvalidCoins, "denom is not a valid stAsset. found: %s", msg.StAssetDenom)
-	}
-	// Creator owns at least "amount" stAssets
-	balance := k.bankKeeper.GetBalance(ctx, sender, msg.StAssetDenom)
+	// 	- Creator owns at least "amount" stAssets
+	balance := k.bankKeeper.GetBalance(ctx, sender, hostZone.IBCDenom)
```
sampocs, line 44: Is this supposed to be "st" + hostZone.IBCDenom?

x/stakeibc/keeper/msg_server_register_host_zone.go
```diff
@@ -60,6 +63,21 @@ func (k Keeper) RegisterHostZone(goCtx context.Context, msg *types.MsgRegisterHo
 		return nil, err
 	}
```
sampocs, line 65: do we need to register the Redemption ICA here?

x/stakeibc/keeper/msg_server_register_host_zone.go
```diff
@@ -60,6 +63,21 @@ func (k Keeper) RegisterHostZone(goCtx context.Context, msg *types.MsgRegisterHo
...
 
+	// add this host zone to unbonding hostZones, otherwise users won't be able to unbond
+	// for this host zone until the following day
+	epochUnbondingRecord, found := k.recordsKeeper.GetLatestEpochUnbondingRecord(ctx)
```
sampocs, line 68: will this error out after genesis if there's no EpochUnbondingRecords? Or do we have a guarantee that an unbonding record will wind up being created before this could ever get called?

PR #56, opened 2022-06-27
TEST-102: redemptionRate working!

riley-stride:
**Summary**
Add working redemptionRate calculated at the beginning of the epoch (every redemptionRateInterval epochs) using depositRecords. RedemptionRate = (UB+SB+MA)/stSupply.

**Test plan**
Build locally, run 1.sh, then 2.sh, then 2.sh after about 10 seconds (to allow some rewards to accrue). Watch the redemptionRate increment as rewards are restaked and included in the new redemptionRate calc.


x/stakeibc/keeper/hooks.go
```diff
@@ -238,3 +246,66 @@ func (k Keeper) TransferExistingDepositsToHostZones(ctx sdk.Context, epochNumber
 		k.RecordsKeeper.RemoveDepositRecord(ctx, recordId)
 	}
 }
+
+func (k Keeper) UpdateRedemptionRates(ctx sdk.Context, epochNumber int64, depositRecords []recordstypes.DepositRecord) {
```
asalzmann, line 250: rm epochNumber

asalzmann (commented):
one minor comment otherwise LGTM

PR #58, opened 2022-06-28
[TEST-93] Updated Makefile

sampocs:
**Summary**
* Removed anything from the Makefile that wasn't used (we can always add it back later if we need to)
* Removed local install steps as it was not necessary to download packages and it would cause changes to the submodules that would be displayed in when running `git status`
* Added clean steps
* **Changed names of local dev commands**

**Test plan**
* Execute each make command (the test commands will fail until we fix the unit tests)


vish-stride:
Looks great to me!

PR #62, opened 2022-07-01
[TEST-130] Fix Outdated Unit Tests from Scaffolding

sampocs:
**Summary**
* Trimmed down ignite keeper tests (`testutil/keeper`) to leverage `test_helper.go` (which was borrowed from Osmosis)
* Fixed misc unit tests - mostly from misalignment between the expected usage from scaffolding and the actual usage of each data structure
* Removed queries that aren’t actually needed and were just a byproduct of scaffolding (e.g. `Delegation`, `MinValidatorRequirements`)
* Changed validator query to take a host zone as a parameter
* Removed some tests based on the following rational (please double check me on this!)
  * I *think* the query tests in the `client/cli` folder (e.g. `client/cil/query_host_zone_test.go`) were meant to test the genesis state, since each test file lined up with a variable that was initialized at genesis (e.g. ICQAccount, EpochTracker, HostZone, Validator). 
  * However, it is my understanding that:
    1. The HostZone data structure is actually supposed to be a map instead of a list
    2. The Validator data type is only used in the HostZone data structure (i.e., there is no list of validators that is stored outside of the host zone map)
  * As a result, I removed the empty host zone list that was initialized in the genesis state, and I removed the query test files under `client/cli` for the host zone and validator queries (`query_host_zone_test.go` and `query_validator_test.go`)

**Unit Test Setup**
* `test_helper.go` instantiates a new chain with an in-memory database
* `testutil/keeper/{module}.go` calls the setup function from `test_helper.go`, and grabs and returns the corresponding keeper
* `testutil/network/network.go` creates a new chain (similar to `test_helper.go` but with more specific configurations) 
  * TBD if this is even necessary / if it provides value beyond  `test_helper.go`. However, many tests leveraged it so I left it in for the time being
*  Each test in `keeper` leverages the mock keeper that's returned from `testutil/keeper/{module}.go`
*  Each test in `client` leverages the network from `testutil/keeper/{module}.go`  and calls mock CLI functions

**Test plan**
* Run `make test-unit`


x/epochs/genesis_test.go
```diff
@@ -12,8 +12,6 @@ import (
 
 func TestGenesis(t *testing.T) {
 	genesisState := types.GenesisState{
-		Params: types.DefaultParams(),
```
riley-stride, line 15: why removing params?

x/records/keeper/grpc_query_deposit_record_test.go
```diff
@@ -12,11 +12,21 @@ import (
 
 	keepertest "github.com/Stride-Labs/stride/testutil/keeper"
 	"github.com/Stride-Labs/stride/testutil/nullify"
+	"github.com/Stride-Labs/stride/x/records/keeper"
 	"github.com/Stride-Labs/stride/x/records/types"
 )
 
+func createNDepositRecord(keeper *keeper.Keeper, ctx sdk.Context, n int) []types.DepositRecord {
```
riley-stride, line 19: Looks like this was added to get the tests running, we must have deleted it at some point for no good reason?

x/stakeibc/genesis.go
```diff
@@ -49,8 +49,6 @@ func ExportGenesis(ctx sdk.Context, k keeper.Keeper) *types.GenesisState {
 	if found {
 		genesis.ICAAccount = &iCAAccount
 	}
-	genesis.HostZoneList = k.GetAllHostZone(ctx)
```
riley-stride, line 52: Why are we removing this?

x/stakeibc/keeper/host_zone_test.go
```diff
@@ -14,7 +15,10 @@ import (
 func createNHostZone(keeper *keeper.Keeper, ctx sdk.Context, n int) []types.HostZone {
 	items := make([]types.HostZone, n)
 	for i := range items {
-		items[i].Id = keeper.AppendHostZone(ctx, items[i])
+		items[i].ChainId = strconv.Itoa(i)
+		items[i].RedemptionRate = sdk.NewDec(0)
```
riley-stride, line 19: should start at 1. same for lastRedemptionRate

testutil/network/network.go
```diff
@@ -64,7 +63,7 @@ func DefaultConfig() network.Config {
 		},
 		GenesisState:    app.ModuleBasics.DefaultGenesis(encoding.Marshaler),
 		TimeoutCommit:   2 * time.Second,
-		ChainID:         "chain-" + tmrand.NewRand().Str(6),
+		ChainID:         "stride-" + tmrand.NewRand().Str(6),
```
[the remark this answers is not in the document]
sampocs: couldn't tell ya, it was boilerplate. My original thought was that it was to prevent any issues when running tests in parallel

x/epochs/genesis_test.go
sampocs: I dont think those params are set at genesis for the epochs module like they are in stakeibc

StakeIBC Ref
https://github.com/Stride-Labs/stride/blob/main/x/stakeibc/genesis.go#L37
https://github.com/Stride-Labs/stride/blob/main/x/stakeibc/genesis.go#L43

x/stakeibc/client/cli/query_delegation.go
```diff
@@ -1,36 +0,0 @@
...
-import (
-	"context"
-
-	"github.com/Stride-Labs/stride/x/stakeibc/types"
-	"github.com/cosmos/cosmos-sdk/client"
-	"github.com/cosmos/cosmos-sdk/client/flags"
-	"github.com/spf13/cobra"
-)
-
-func CmdShowDelegation() *cobra.Command {
```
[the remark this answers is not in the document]
sampocs: I think there were certain CLI tests that were scaffolded because ignite assumed we were going to initialize them at genesis (e.g. Delegation and MinValidatorRequirements data types). Since we don't actually need them at genesis, I think we can remove these tests

PR #64, opened 2022-07-02
Robustness improvements to local net

riley-stride:
## Summary
Prepare binary for poolparty testnet by fixing some bugs and getting things in order for release to public! Added instructions below on how to holistically test the Stride app!

### Contentful changes:
* Fix ibc-go v3.0.0 `marshalling error func() uint64` logging issue by upgrading go.mod dependency to ibc-go v3.0.1
* Pin interchain-queries relayer version to a specific commit (poolparty1 branch)
    * in interchain-queries relayer, re-order the blocks of code that grab trusted and header heights, so that we get trusted *before* header. this prevents the possibility of trusted advancing by a block during execution, which breaks proof verification (trusted must be <= header height) 
* Add test script `4.sh` to test claiming redeemed tokens from the redemption account 
* Reduce the load on ICA channel by preventing unnecessary calls to ICA when we know balances are 0 or no action needs to be taken (e.g. `callbacks.go`)
* Fix bug with `register_host_zones`: make default HostZoneUnbonding BONDED rather than UNBONDED 

### Params
* Hermes' trusting_period = 180s
* Gaia's unbonding_time = 200s
* Stride's DAY_EPOCH_LEN = 60s
* Stride's EPOCH_LEN = 10s
* Stride and Gaia's BLOCK_TIME = 10s

### Scripts
* In `build.sh`, 
    * add `-p` flag to `mkdir` to allow safe folder creation. If it already exists, we shouldn't error
    * nuke the state folder each time we re-build! Note this makes building slower but more safe and reliable
* Sleep longer in most scripts, for safety and reliability 

## Test plan
Run through the following scripts to test the Stride App. Observe balances and records being updated live in `scripts-local/logs/accounts.log`. 

Build with 
`make init-build build=sghi`
Once built, ibc some tokens from gaia to stride and register gaia's validator set with 
`sh ./scripts-local/tests/1.sh`
Then liquid stake some tokens on gaia. We will 60s to allow them to be delegated on the host zone...
`sh ./scripts-local/tests/2.sh`
Liquid stake some more tokens. Anytime a new delegation is made after the first, accrued rewards are automatically claimed and compounding begins!
`sh ./scripts-local/tests/2.sh`
Now you might want to redeem some of your staked atom from the cosmos hub. To being unbonding to address `cosmos1g6qdx6kdhpf000afvvpte7hp0vnpzapuyxp8uf` (you can choose where to unbond to with Stride), run 
`sh ./scripts-local/tests/3.sh`
You'll see unbonding-delegations in `accounts.log` and their `completion_time`. Once the completion time elapses, the tokens will be available to transfer to the address specified when redemption was invoked. Run
`sh ./scripts-local/tests/4.sh`
You should see the balance of `cosmos1g6qdx6kdhpf000afvvpte7hp0vnpzapuyxp8uf` increased! Everything is working as expected. 

A few more things to look for in `accounts.log`:
* The exchange rate should begin increasing (fairly rapidly) once rewards begin to compound
* At the bottom of the logfile, you'll see records being created. This is a window into Stride's internal accounting system.


scripts-local/build.sh
```diff
@@ -1,5 +1,8 @@
 #!/bin/bash
 
+rm -rf scripts-local/state/*
```
asalzmann, line 3:
```suggestion
rm -rf $SCRIPT_DIR/state/*
```

PR #70, opened 2022-07-07
Fix claim bug

riley-stride:
## Summary
1. Add input validation to redeem-stake
2. Rename whitelist => admin and make include both local-dev and testnet admin 
3. Fix claim bug
    i. Remove `claimableRecordIds` from `HostZone`
    ii. Remove records in ack `6d5da1`
    iii. Add `PendingClaims` data structure to enable ICA callbacks

### Note: 
- moving the admin addresses to params may not be a good solution because it would not allow us to check certain functions whose msg_servers are defined outside our custom modules (e.g. registerICA's msg server is defined in the ICAcontrollerkeeper). We can't access stakeibc params inside stakeibc/types, which is where ValidateBasic is defined, so we can't check sender addresses against admin param there. We'd need to do so in the message server. This works for the functions we defined, but not for those defined in other modules e.g. registerICA. 

### Input validation:
_[note: would ideally move all validation logic to ValidateBasic(), but we can't access the keeper from there, and we need the keeper to validate hostZone]_
- Add a `bech32prefix` field to the hostZone proto (necessary for validating foreign zone addresses from Stride) and add bech32prefix args to all registerHostZone messages, types and invocations
- Copy zone address verification code from cosmos-sdk into `utils.go` and parametrize it by hostZone so we can invoke it for any hostZone when we'd like to check the validity of an arbitrary address (note this assumes len 20 addresses, so will fail with ICA accounts; should be OK, we don't need to redeem to those accounts but maybe worth revisiting)
- In `msg_server_redeem_stake.go`, check the receiver address is a valid address on the specified host zone, after checking the host zone exists. 
- In `message_redeem_stake.go`, check amount is > 0

## Test plan
To test input validation, attempt to submit an incorrectly formatted redeem-stake message! 
The following redeem-stake command should succeed (be sure to run `1.sh` and `2.sh` beforehand as setup.
```
$STRIDE_CMD tx stakeibc redeem-stake 89 GAIA cosmos1g6qdx6kdhpf000afvvpte7hp0vnpzapuyxp8uf --from $STRIDE_VAL_ACCT --keyring-backend test --chain-id $STRIDE_CHAIN -y

```
The following should fail:
```
# stride prefix, nonexistent OK
$STRIDE_CMD tx stakeibc redeem-stake 89 GAIA stride1g6qdx6kdhpf000afvvpte7hp0vnpzapuyxp8uf --from $STRIDE_VAL_ACCT --keyring-backend test --chain-id $STRIDE_CHAIN -y
# stride prefix, exists on stride
$STRIDE_CMD tx stakeibc redeem-stake 89 GAIA stride159atdlc3ksl50g0659w5tq42wwer334ajl7xnq --from $STRIDE_VAL_ACCT --keyring-backend test --chain-id $STRIDE_CHAIN -y
# cosmos address with bad hash at end
$STRIDE_CMD tx stakeibc redeem-stake 89 GAIA cosmos1g6qdx6kdhpf000afvvpte7hp0vnpzapuyxp8ug --from $STRIDE_VAL_ACCT --keyring-backend test --chain-id $STRIDE_CHAIN -y
# cosmos addr too long
$STRIDE_CMD tx stakeibc redeem-stake 89 GAIA cosmos1g6qdx6kdhpf000afvvpte7hp0vnpzapuyxp8uff --from $STRIDE_VAL_ACCT --keyring-backend test --chain-id $STRIDE_CHAIN -y
# empty address 
$STRIDE_CMD tx stakeibc redeem-stake 89 GAIA " " --from $STRIDE_VAL_ACCT --keyring-backend test --chain-id $STRIDE_CHAIN -y
# too short
$STRIDE_CMD tx stakeibc redeem-stake 89 GAIA cosmos1g6qdx6kdhpf000afvvpte7hp0vnpzapuyxp8u --from $STRIDE_VAL_ACCT --keyring-backend test --chain-id $STRIDE_CHAIN -y
# bad host zone
$STRIDE_CMD tx stakeibc redeem-stake 89 GAIAA cosmos1g6qdx6kdhpf000afvvpte7hp0vnpzapuyxp8u --from $STRIDE_VAL_ACCT --keyring-backend test --chain-id $STRIDE_CHAIN -y
```

## Changes (claim bug fix)
- Scaffold `PendingClaims` to enable ICA callbacks (`ca71746`). This maps IBC packet sequence number to a claimable record id (note: for now, this maps to a single record, but this could be extended to process multiple claims in a single transaction)
- Updates logic in `ClaimUndelegatedTokens` (`x/stakeibc/keeper/msg_server_claim_undelegated_tokens.go`)
    - claim records one-by-one
    - store a mapping from IBC packet sequence number to record id (enabling callback logic in the ack)
    - add checks (record exists, is claimable, host zone exists)
- add logic to `HandleSend` in ack to delete `PendingClaims` on failure, and to delete both `PendingClaims` and `UserRedemptionRecords` on success

## Test plan (claim bug fix)
Run 1.sh, 2.sh, 3.sh then claim the first record with 4.sh
Run 3.sh twice more to create records two and three, claim record three with 4.sh
Run 4.sh one last time to claim record two


proto/stakeibc/host_zone.proto
```diff
@@ -38,7 +38,8 @@ message HostZone {
   ];
   // stores how many days we should wait before issuing unbondings
   uint64 unbondingFrequency = 14;
-  repeated string claimableRecordIds = 15;
   //TODO(TEST-101) int to dec
   int64 stakedBal = 13;
+
+  reserved 15;
```
riley-stride, line 45: ty for pointing out this best practice for mainnet

proto/stakeibc/tx.proto
```diff
@@ -93,8 +93,10 @@ message MsgRegisterHostZoneResponse{}
 
 message MsgClaimUndelegatedTokens {
   string creator = 1;
-  string hostZone = 2;
-  uint64 maxClaims = 3;
+  // UserUnbondingRecords are keyed on {chain_id}.{epoch}.{sender}
+  string hostZoneId = 2;
+  uint64 epoch = 3;
+  string sender = 4;
```
riley-stride, line 101: just to make sure I understand, the creator is the user who called redeem-stake, and the sender is the user who called claim-undelegated?

x/stakeibc/client/cli/tx_claim_undelegated_tokens.go
```diff
@@ -13,17 +13,24 @@ import (
 
 var _ = strconv.Itoa(0)
 
+// string creator = 1;
```
riley-stride, line 16: nit: rm

x/stakeibc/client/cli/tx_claim_undelegated_tokens.go
```diff
@@ -13,17 +13,24 @@ import (
...
+//   string hostZoneId = 2;
+//   uint64 epoch = 3;
+//   string sender = 4;
+// }
+
 func CmdClaimUndelegatedTokens() *cobra.Command {
 	cmd := &cobra.Command{
-		Use:   "claim-undelegated-tokens [host-zone] [max-claims]",
+		Use:   "claim-undelegated-tokens [host-zone] [epoch] [sender]",
```
riley-stride, line 24: should we parse the sender from the args? seems odd to allow the user calling this function to choose the sender arbitrarily. you can get the user with `clientCtx.GetFromAddress().String()` I believe, as in line 41

x/stakeibc/keeper/msg_server_submit_tx.go
```diff
@@ -171,30 +171,30 @@ func (k Keeper) UpdateWithdrawalBalance(ctx sdk.Context, zoneInfo types.HostZone
 }
 
 // SubmitTxs submits an ICA transaction containing multiple messages
-func (k Keeper) SubmitTxs(ctx sdk.Context, connectionId string, msgs []sdk.Msg, account types.ICAAccount) error {
+func (k Keeper) SubmitTxs(ctx sdk.Context, connectionId string, msgs []sdk.Msg, account types.ICAAccount) (uint64, error) {
 	chainId, err := k.GetChainID(ctx, connectionId)
 	if err != nil {
-		return err
+		return 0, err
```
riley-stride, line 177: I assume we're error checking everytime we call SubmitTxs so this return val of 0 shouldn't be an issue. Is there a safer val we can return here? Don't think so (negatives and nils won't work).

x/stakeibc/keeper/msg_server_redeem_stake.go
```diff
@@ -59,7 +59,7 @@ func (k Keeper) RedeemStake(goCtx context.Context, msg *types.MsgRedeemStake) (*
 		return nil, sdkerrors.Wrapf(types.ErrEpochNotFound, "epoch tracker found: %s", "day")
 	}
 	senderAddr := sender.String()
-	redemptionId := fmt.Sprintf("%s.%d.%s", hostZone.ChainId, epochTracker.EpochNumber, senderAddr) // {chain_id}.{epoch}.{sender}
+	redemptionId := recordstypes.UserRedemptionRecordKeyFormatter(hostZone.ChainId, epochTracker.EpochNumber, senderAddr)
```
riley-stride, line 69: Does this new redemptionId format render the checks on line 73-76 obsolete? What issues might that introduce, if a user could redeem multiple times per epoch?

x/stakeibc/keeper/msg_server_claim_undelegated_tokens.go
```diff
@@ -4,75 +4,75 @@ import (
...
 	bankTypes "github.com/cosmos/cosmos-sdk/x/bank/types"
 )
 
-func min(a int, b int) int {
-	if a < b {
-		return a
-	}
-	return b
-}
+var min = utils.Min
```
riley-stride, line 16: Looks like this is unused

x/stakeibc/keeper/msg_server_claim_undelegated_tokens.go
```diff
@@ -4,75 +4,75 @@ import (
...
+	if !userRedemptionRecord.GetIsClaimable() {
+		errMsg := fmt.Sprintf("User redemption record %s is not claimable", userRedemptionRecord.Id)
+		k.Logger(ctx).Error(errMsg)
+		return nil, sdkerrors.Wrapf(types.ErrInvalidUserRedemptionRecord, "user redemption record is not claimable: %s", userRedemptionRecordKey)
 	}
+
 	// grab necessary fields to construct ICA call
-	var msgs []sdk.Msg
-	redemptionAccount, found := k.GetRedemptionAccount(ctx, hostZone)
+	hostZone, found := k.GetHostZone(ctx, msg.HostZoneId)
```
riley-stride, line 38: Let's move this check to the top

x/stakeibc/keeper/msg_server_redeem_stake.go
asalzmann: It's just a formatter, it doesn't change any of the logic (unless I'm missing something)

x/stakeibc/keeper/msg_server_claim_undelegated_tokens.go
asalzmann: It's used below

x/stakeibc/keeper/msg_server_claim_undelegated_tokens.go
```diff
@@ -4,75 +4,75 @@ import (
...
+		return nil, sdkerrors.Wrap(types.ErrInvalidHostZone, errMsg)
 	}
-	// TODO we should do some error handling here, in case this call fails
-	err := k.SubmitTxs(ctx, hostZone.GetConnectionId(), msgs, *redemptionAccount)
+
+	var msgs []sdk.Msg
+	msgs = append(msgs, &bankTypes.MsgSend{
+		FromAddress: redemptionAccount.Address,
+		ToAddress:   userRedemptionRecord.Receiver,
+		Amount:      sdk.NewCoins(sdk.NewInt64Coin(userRedemptionRecord.Denom, int64(userRedemptionRecord.Amount))),
```
riley-stride, line 55: I've never used `NewInt64Coin`, looks like it's just a wrapper around `NewCoin`, should be fine https://github.com/cosmos/cosmos-sdk/blob/411c272d01df89cdf67c34afbba8d4188ab99320/types/coin.go#L14-L33

x/stakeibc/keeper/msg_server_claim_undelegated_tokens.go
asalzmann: Can you elaborate on this?

x/stakeibc/keeper/msg_server_claim_undelegated_tokens.go
```diff
@@ -4,75 +4,75 @@ import (
...
+	userRedemptionRecord.IsClaimable = false
+	k.RecordsKeeper.SetUserRedemptionRecord(ctx, userRedemptionRecord)
+
+	// Store the sequence number to record id mapping
+	pendingClaims := types.PendingClaims{
+		Sequence: string(sequence),
+		// NOTE: we could extend this to process multiple claims in the future, given this field is repeated
+		UserRedemptionRecordIds: []string{userRedemptionRecord.Id},
 	}
+	k.SetPendingClaims(ctx, pendingClaims)
```
riley-stride, line 74: nice clean logic in this function!

utils/utils.go
```diff
@@ -31,3 +31,10 @@ func ValidateWhitelistedAddress(address string) error {
 	}
 	return nil
 }
+
+func Min(a int, b int) int {
```
riley-stride, line 41: seems like this is unused. if so, let's rm?

x/stakeibc/keeper/ibc_handlers.go
```diff
@@ -24,6 +24,7 @@ import (
 // TODO(TEST-33): Scope out what to store on different acks (by function call, success/failure)
 func (k Keeper) HandleAcknowledgement(ctx sdk.Context, modulePacket channeltypes.Packet, acknowledgement []byte) error {
 	ack := channeltypes.Acknowledgement_Result{}
+	packetSequenceKey := string(modulePacket.Sequence)
```
riley-stride, line 27: can we add some error checking here e.g. if sequence is nil or empty, break? worried about reliability with these packet sequence #s

x/stakeibc/keeper/ibc_handlers.go
```diff
@@ -35,6 +36,11 @@ func (k Keeper) HandleAcknowledgement(ctx sdk.Context, modulePacket channeltypes
 	err := json.Unmarshal(acknowledgement, &ack)
 	if err != nil {
 		ackErr := channeltypes.Acknowledgement_Error{}
+		// Clean up any pending claims
+		_, found := k.GetPendingClaims(ctx, packetSequenceKey)
+		if found {
+			k.RemovePendingClaims(ctx, packetSequenceKey)
```
riley-stride, line 42: is the idea here to remove pending status on ack errors so that we can re-attempt to claim them in the next day epoch?

x/stakeibc/client/cli/tx_claim_undelegated_tokens.go
asalzmann: Done.

utils/utils.go
asalzmann: https://github.com/Stride-Labs/stride/pull/70#discussion_r917017476

It's used in the claim function

x/stakeibc/keeper/ibc_handlers.go
```diff
@@ -256,20 +262,31 @@ func (k *Keeper) HandleSend(ctx sdk.Context, msg sdk.Msg) error {
...
+		}
+		userRedemptionRecordKey := pendingClaims.UserRedemptionRecordIds[0]
+		_, found = k.RecordsKeeper.GetUserRedemptionRecord(ctx, userRedemptionRecordKey)
+		if !found {
+			errMsg := fmt.Sprintf("User redemption record %s not found on host zone", userRedemptionRecordKey)
+			k.Logger(ctx).Error(errMsg)
+			return sdkerrors.Wrapf(types.ErrInvalidUserRedemptionRecord, "could not get user redemption record: %s", userRedemptionRecordKey)
+		}
+		k.RecordsKeeper.RemoveUserRedemptionRecord(ctx, userRedemptionRecordKey)
+		k.RemovePendingClaims(ctx, sequence)
```
riley-stride, line 289: logic looks sound to me

x/stakeibc/keeper/ibc_handlers.go
```diff
@@ -256,20 +262,31 @@ func (k *Keeper) HandleSend(ctx sdk.Context, msg sdk.Msg) error {
...
 	} else {
 		k.Logger(ctx).Error("ACK - sendMsg.FromAddress != withdrawalAddress && sendMsg.FromAddress != delegationAddress")
```
riley-stride, line 291: nit: add redemption acct to the comment

x/stakeibc/keeper/ibc_handlers.go
asalzmann: Because it's an int field on a proto, it'll always come back as 0 or some positive int. 0 is technically valid (will just be the first packet), so not sure what the best way to check this is. Wdyt?

x/stakeibc/keeper/msg_server_claim_undelegated_tokens.go
riley-stride: must be a different commit, I don't see it here https://github.com/Stride-Labs/stride/blob/6d5da1b96a45f80c2fef4c439183b8fcd23060ad/x/stakeibc/keeper/msg_server_claim_undelegated_tokens.go

x/stakeibc/keeper/msg_server_claim_undelegated_tokens.go
riley-stride: part of my PR that does validation on inputs was to check the validity of hostzones in each msg_server that has a hostZone in its msg (like this msg_server). I always do that at the top of the msg_server. Mostly for consistency

x/stakeibc/keeper/msg_server_redeem_stake.go
riley-stride: Nevermind, I now see how this works 👍

x/stakeibc/keeper/msg_server_submit_tx.go
riley-stride: Either should be fine!

x/stakeibc/keeper/ibc_handlers.go
riley-stride: makes sense!

x/stakeibc/keeper/ibc_handlers.go
riley-stride: hmm fair points - no good way to validate it afaiu

x/stakeibc/keeper/msg_server_claim_undelegated_tokens.go
asalzmann: sorry, I did remove it! removing the function now

x/stakeibc/keeper/msg_server_claim_undelegated_tokens.go
asalzmann: got it - I think this should still fail if the host zone doesn't exist! there isn't any substantive logic above it, just checks on UserRedemptionRecords. slight preference to keep this here so the sections of the function are
1. Checks on UserRedemptionRecord
2. Checks on all fields required for the ICA
3. Constructing and sending the transaction, updating state

utils/utils.go
asalzmann: resolved above

utils/utils.go
```diff
@@ -31,3 +36,80 @@ func ValidateWhitelistedAddress(address string) error {
...
+
+	return nil
+}
+
+// AccAddress a wrapper around bytes meant to represent an account address.
+// When marshaled to a string or JSON, it uses Bech32.
+type AccAddress []byte
+
+// AccAddressFromBech32 creates an AccAddress from a Bech32 string.
+func AccAddressFromBech32(address string, bech32prefix string) (addr AccAddress, err error) {
```
asalzmann, line 105:
```suggestion
func AccAddressFromBech32WithPrefix(address string, bech32prefix string) (addr AccAddress, err error) {
```

x/stakeibc/keeper/msg_server_redeem_stake.go
```diff
@@ -19,14 +21,19 @@ func (k Keeper) RedeemStake(goCtx context.Context, msg *types.MsgRedeemStake) (*
...
-
-	// TODO(TEST-112) add safety check to validate the receiver address is a valid hostZone address
 	// then make sure host zone is valid
 	hostZone, found := k.GetHostZone(ctx, msg.HostZone)
 	if !found {
 		return nil, sdkerrors.Wrapf(types.ErrInvalidHostZone, "host zone is invalid: %s", msg.HostZone)
 	}
 
+	// ensure the recipient address is a valid bech32 address on the hostZone
+	// TODO(TEST-112) do we need to check the hostZone before this check? Would need access to keeper
```
asalzmann, line 31: What does this comment mean?

x/stakeibc/keeper/msg_server_redeem_stake.go
```diff
@@ -19,14 +21,19 @@ func (k Keeper) RedeemStake(goCtx context.Context, msg *types.MsgRedeemStake) (*
...
-	// TODO(TEST-112) add safety check to validate the receiver address is a valid hostZone address
 	// then make sure host zone is valid
 	hostZone, found := k.GetHostZone(ctx, msg.HostZone)
 	if !found {
 		return nil, sdkerrors.Wrapf(types.ErrInvalidHostZone, "host zone is invalid: %s", msg.HostZone)
 	}
 
+	// ensure the recipient address is a valid bech32 address on the hostZone
+	// TODO(TEST-112) do we need to check the hostZone before this check? Would need access to keeper 
+	_, err = utils.AccAddressFromBech32(msg.Receiver, hostZone.Bech32Prefix)
```
asalzmann, line 32: Should this live in `ValidateBasic`? I don't have a great sense of where to put these validity checks. On the one hand, it's logically cleaner to put it in `ValidateBasic`, but on the other we'd have to read in the host zone twice

x/stakeibc/types/message_register_host_zone.go
```diff
@@ -42,6 +43,7 @@ func (msg *MsgRegisterHostZone) GetSignBytes() []byte {
 	return sdk.MustSortJSON(bz)
 }
 
+// TODO(TEST-112) add validation on bech32prefix upon zone creation
```
asalzmann, line 47: What's this todo?

asalzmann (commented):
lgtm overall! few minor questions

x/stakeibc/keeper/msg_server_redeem_stake.go
asalzmann: Never mind, per the docs `ValidateBasic can include only stateless checks (the checks that do not require access to the state`

utils/utils.go
riley-stride: good suggestion, done

PR #71, opened 2022-07-08
add validateBasic() checks to most of x/stakeibc

riley-stride:
**Summary**
Add some basic input checking to most the stakeibc functions, with particular emphasis on those exposed to the public

**Test plan**
Run through the expected testing flow:

Observe balances and records being updated live in scripts-local/logs/accounts.log. Build with
`make init-build build=sghi`
Once built, ibc some tokens from gaia to stride and register gaia's validator set with
`sh ./scripts-local/tests/1.sh`
Then liquid stake some tokens on gaia. We will 60s to allow them to be delegated on the host zone...
`sh ./scripts-local/tests/2.sh`
Liquid stake some more tokens. Anytime a new delegation is made after the first, accrued rewards are automatically claimed and compounding begins!
`sh ./scripts-local/tests/2.sh`
Now you might want to redeem some of your staked atom from the cosmos hub. To being unbonding to address cosmos1g6qdx6kdhpf000afvvpte7hp0vnpzapuyxp8uf (you can choose where to unbond to with Stride), run
`sh ./scripts-local/tests/3.sh`
You'll see unbonding-delegations in accounts.log and their completion_time. Once the completion time elapses, the tokens will be available to transfer to the address specified when redemption was invoked. Run
`sh ./scripts-local/tests/4.sh`
You should see the balance of cosmos1g6qdx6kdhpf000afvvpte7hp0vnpzapuyxp8uf increased! Everything is working as expected.

A few more things to look for in accounts.log:

The exchange rate should begin increasing (fairly rapidly) once rewards begin to compound
At the bottom of the logfile, you'll see records being created. This is a window into Stride's internal accounting system.

### DIFF
--- a/x/stakeibc/types/message_add_validator.go
+++ b/x/stakeibc/types/message_add_validator.go
@@ -1,6 +1,7 @@
 package types
 
 import (
+	"fmt"
 	"github.com/Stride-Labs/stride/utils"
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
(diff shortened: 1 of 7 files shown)


x/stakeibc/types/message_add_validator.go
```diff
@@ -50,5 +51,21 @@ func (msg *MsgAddValidator) ValidateBasic() error {
 	if err := utils.ValidateWhitelistedAddress(msg.Creator); err != nil {
 		return err
 	}
+	// name validation
+	if len(msg.Name) == 0 {
+		return sdkerrors.Wrapf(sdkerrors.ErrInvalidRequest, "name is required")
+	}
+	// commission validation
+	if msg.Commission > 100 {
```
sampocs, line 59: is this guaranteed to be greater than 0 since it's an unsigned int? And if so, wouldn't the same apply to the weight validation below and maxclaims validation [here](https://github.com/Stride-Labs/stride/pull/71/files#diff-80eae880f9395eddcac85b86fc08e5f55f9f4b651ffd80ae9faf2ff775124f8fR47)

sampocs (approved):
Looks good!

vish-stride (approved):
lgtm!

scalalang2 (approved):

PR #74, opened 2022-07-11
Add integration tests to local net

asalzmann:
**Summary**
Add integration tests to local net to test the end-to-end liquid staking and redemption flow.

Specifically
- Add integration tests to local net
    - Add bats files to `scripts-local`
    - Add tests to `scripts-local/tests/integration_tests.bats`
- add BLOCK_SLEEP function
- Fix bug in transfer ack that causes hermes to crash (only process transfer acks if the transfer was initiated from the `stakeibc` module account)
- Add a query to get seconds remaining in an epoch (convenient for testing)
- Add `scripts-local/test-utils/run_any_cli_command.sh` to run one off commands that are helpful when testing

**Test plan**
```
➜  stride git:(integration-testing) ✗ make test-integration
sh scripts-local/tests/run_all_tests.sh
integration_tests.bats
 ✓ address names are correct
 ✓ ibc transfer updates all balances
 ✓ liquid stake mints stATOM
 ✓ host zone successfully registered
 ✓ tokens were transferred to GAIA after liquid staking
 ✓ tokens on GAIA were staked
 ✓ [NOT-IMPLEMENTED] rewards are being reinvested
 ✓ [NOT-IMPLEMENTED] redemption works
 ✓ [NOT-IMPLEMENTED] claimed tokens are returned to sender
 ✓ [NOT-IMPLEMENTED] exchange rate is updating

10 tests, 0 failures
```

**Up next**
These tests have been specced but not yet implemented (we should spec and add more tests too!)
```
# check that a second liquid staking call kicks off reinvestment
@test "rewards are being reinvested" {
  # check the rewards balance
  # wait a day
  # check the withdrawal account balance
  # wait a day
  # check that 90% of rewards were transferred to the delegation account
  # check that 10% of rewards were transferred to the rev EOA
  # wait a day
  # check that rewards were staked
}

# check that redemptions and claims work
@test "redemption works" {
  # call redeem-stake
  # check for an unbonding record
  # check that a UserRedemptionRecord was created with isClaimabled = false
  # wait for the unbonding period to pass
  # check that the tokens were transferred to the delegation account
  # wait for an epoch to pass
  # check that the tokens were transferred to the redemption account
}

@test "claimed tokens are returned to sender" {
  # check that the UserRedemptionRecord has isClaimable = true
  # claim the record
  # check that UserRedemptionRecord has isClaimable = false
  # check that the tokens were transferred to the sender account
  # check that the 
}

# check that exchange rate is updating
@test "exchange rate is updating" {
  # read the exchange rate
  # wait a day
  # check that the exchange rate has updated
}
```


scripts-local/tests/ica_redeem.sh
```diff
@@ -0,0 +1,53 @@
+#!/bin/bash
+
+echo "Register zone"
+# setup logic on controller zone
+$STRIDE1_EXEC tx stakeibc register-host-zone \
+    connection-0 uatom ibc/C4CFF46FD6DE35CA4CF4CE031E643C8FDC9BA4B99AE598E9B0ED98FE3A2319F9 channel-0 --chain-id \
```
vish-stride, line 6: is this the right IBC hash? i thought it was `ibc/27...`

scripts-local/tests/integration_tests.bats
```diff
@@ -0,0 +1,214 @@
+#!/usr/bin/env bats
```
vish-stride, line 1: this is a really good layout for this file

utils/utils.go
```diff
@@ -9,6 +9,7 @@ import (
 )
 
 var WHITELIST = map[string]bool{
+	"stride1uk4ze0x4nvh4fk0xm4jdud58eqn4yxhrt52vv7": true,
```
vish-stride, line 12: is this STRIDE_1 ? we should be careful about making sure this is removed before launching poolparty or mainnet, but both of those events are sufficiently rare where I don't think we need to change this

x/epochs/client/cli/query.go
```diff
@@ -106,3 +109,49 @@ func GetCmdCurrentEpoch() *cobra.Command {
 
 	return cmd
 }
+
+
+// GetCmdSecondsRemaining provides seconds-remaining by specified identifier
+func GetCmdSecondsRemaining() *cobra.Command {
```
vish-stride, line 115: very helpful query

vish-stride (approved):
good to merge imo!

scripts-local/tests/integration_tests.bats
```diff
@@ -0,0 +1,214 @@
...
+
+@test "liquid stake mints stATOM" {
+  # get module address 
+  MODADDR=$($STRIDE_CMD q stakeibc module-address stakeibc | awk '{print $NF}') 
+  # get initial balances
+  mod_balance_atom=$($STRIDE_CMD q bank balances $MODADDR --denom $IBC_ATOM_DENOM | GETBAL)
+  str1_balance_atom=$($STRIDE_CMD q bank balances $STRIDE_ADDRESS --denom $IBC_ATOM_DENOM | GETBAL)
+  str1_balance_statom=$($STRIDE_CMD q bank balances $STRIDE_ADDRESS --denom $STATOM_DENOM | GETBAL)
+  # liquid stake
+  $STRIDE_CMD tx stakeibc liquid-stake 1000 uatom --keyring-backend test --from val1 -y --chain-id $STRIDE_CHAIN
```
riley-stride, line 80: a few of the tests are failing for me because we did not register any validator before calling liquid-stake @asalzmann does that need to be added ?

utils/utils.go
riley-stride: Yeah it's local val1, we def need to remove it before mainnet, good flag

x/epochs/client/cli/query.go
riley-stride: +1

scripts-local/tests/integration_tests.bats
asalzmann: Ah ty, yes was just about to debug this

scripts-local/tests/ica_redeem.sh
asalzmann: This is an old file, going to remove it

PR #77, opened 2022-07-12
Custom Module Consensus Versions

sampocs:
We should start with consensus version 1 for our custom modules. It will change to 2 when we do our first soft upgrade


riley-stride (approved):
Looks straightforward. Only Q: do we need to do the same for interchainquery and epochs?

sampocs:
They're actually already at ConsensusVersion 1! All though more reason why we should change ours

riley-stride:
> They're actually already at ConsensusVersion 1! All though more reason why we should change ours

Sweet let's merge it in!

vish-stride (approved):
very contentious but i'll approve

PR #79, opened 2022-07-12
Add .circleci/config.yml

asalzmann:
# TODO
- add docker image with required deps
- add config file to build, init network and run bats
- maybe: make bats tests network agnostic (run on both docker and local net), use docker on circleci

**Summary**

**Test plan**


asalzmann:
will use gh actions

PR #85, opened 2022-07-14
Address validator comments

vish-stride:
**Summary**

Minor changes to address feedback from call today

**Test plan**


riley-stride (commented):
Lgtm! can you also address finding #1 from the google doc please? Will reply re: these changes in the channel

asalzmann (approved):
lgtm - do you mind also running the integration tests for this PR? I'm working on getting them on GH actions but until then we should manually run them if possible

PR #87, opened 2022-07-15
Unit Testing Framework (+ Liquid Stake and Redeem Stake Tests)

sampocs:
**Summary**
* Added unit testing framework 
* Migrate the epochs module to use this framework 
* Added tests for the Liquid Stake flow. 
* I've tried my best to describe everything below, but it's probably a little hard to follow just from reading so feel free to reach out if you'd prefer a walk though! 

**Testing Framework**
* `app/test_setup.go` defines a function that initializes a test blockchain instance
* `app/apptesting/test_helpers.go` defines an object `AppTestHelper` that extends a go testing package called `suite` and manages the app (`App`) and context (`Ctx`) as attributes
  * It also has a setup function that calls the chain init function (from `test_setup.go`) and creates sample accounts
  * It also has testing helper functions (e.g. `CreateSampleAccounts`, `FundAccount`, etc.)
* For each module, a `keeper_test.go` file should define a `KeeperTestSuite` object that extends the `AppTestHelper` and contains module-specific setup functions (e.g. initializing the message server)
  * It should be a part of a new package called `keeper_test` 
  * The purpose of extending the `suite` class is to group tests together, such that they share common setup, teardown functionality and allow them to run in parallel (although it's still unclear whether tests usually run in parallel by default?)
  * The `keeper_test.go` file must also have a function (shown below), that registers all the `suite` tests with the default go testing context
```
func TestKeeperTestSuite(t *testing.T) {
   suite.Run(t, new(KeeperTestSuite))
}
```
 * Finally, a new file can be created for each message/flow we want to test, (e.g. `msg_server_liquid_stake_test.go`)
   * This file will extend the `KeeperTestSuite` which will give it access to everything it needs including the keepers (`suite.App.{KeeperName}`) and the message server (`suite.msgServer`)
   * The combination of these allows you to call any message directly and test any account balance or state changes 

**Setting up a new Test**
1. Create the new test file in the module's keeper (e.g. `msg_server_liquid_stake_test.go`)
2. The file should be a part of the `keeper_test` package and each test function must extend the `KeeperTestSuite` 
3. Create a setup function for the group of tests. This function should be called before every test. This will initialize the state and balances for the group of tests (see `SetupLiquidStake`). I'd recommend returning a struct from this setup function that provides information about the initial state that can be used in each test function (e.g. initial account balances). The struct should also provide an example of a successful message, given the initial state setup. 
4. Define one or more functions to test a successful flow. For these tests, call the tested functionality and then compare the actual output state to the state that was expected. To compare values use, `suite.Require().Equals(expected, actual, description)`
5. Define functions to test out errors. For these, simply alter either the message or the state before calling the tested function, and then confirm the returned error lines up with what's expected using `suite.Require().ErrorEquals(...)`
6. Finally you can run the test! I'd recommend running directly from VSCode as you're debugging and moving to the command line at the end. In order for the play button to register in vscode, you'll have to import `suite` to that file 
```
import _ "github.com/stretchr/testify/suite"
```

**Unanswered Questions**
* How well this framework will work with IBC transactions 
* Whether we'll need to iterate blocks
  * I assume no, because all that would do is finalize the state; but, with this setup, the state is finalized immediately
  * If we find that we do need to iterate blocks, I'm confident it can be done (osmosis has an example of this)

**Remaining Work**
* ~Tests only work from VSCode, atm. Working through it now~ 
* **Edit**: This has been resolved. Run tests with `make test-unit`

**Future Work**
* I'm sure things could be organized more gracefully. But I would prefer to be redundant at the start, and then over time we can start combining functionality once we write more tests and it becomes clear what logic is test-specific and what can be generalized. I moved the functions that were obviously generalizable to `test_helpers.go` 
* Definitely not a priority - but we should migrate all the old keeper tests to this framework at some point, and eventually remove the `testutil` directory.
  * It should be fairly straightforward (just need to extend the KeeperTestSuite and change the `keeper` and `ctx` references to come from the suite attributes. 

**Test plan**
```
make test-unit
```
Note you can clear the test cache with `go clean -testcache`.
* ~Note: Some tests were already broken before starting this branch~
* **Edit**: Outstanding tests were resolved in latest commit


x/stakeibc/keeper/msg_server_liquid_stake.go
```diff
@@ -91,8 +83,8 @@ func (k msgServer) LiquidStake(goCtx context.Context, msg *types.MsgLiquidStake)
 func (k msgServer) MintStAsset(ctx sdk.Context, sender sdk.AccAddress, amount int64, denom string) error {
 	// repeat safety checks from LiquidStake in case MintStAsset is called from another site
 	// ensure Amount is positive, liquid staking 0 or less tokens is invalid
-	if amount < 1 {
-		k.Logger(ctx).Info("Amount to mint must be positive")
+	if amount <= 0 {
```
riley-stride, line 86: Noting this is a logic fix Sam and I made, unrelated to the testing in the PR. Looks correct to me, pls take a look @vish-stride @asalzmann

x/stakeibc/keeper/msg_server_liquid_stake.go
```diff
@@ -105,23 +97,23 @@ func (k msgServer) MintStAsset(ctx sdk.Context, sender sdk.AccAddress, amount in
...
 	}
 	// transfer those coins to the user
-	sdkerror = k.bankKeeper.SendCoinsFromModuleToAccount(ctx, types.ModuleName, sender, stCoins)
-	if sdkerror != nil {
-		k.Logger(ctx).Info("Failed to send coins from module to account")
-		panic(sdkerror)
+	err = k.bankKeeper.SendCoinsFromModuleToAccount(ctx, types.ModuleName, sender, stCoins)
+	if err != nil {
+		k.Logger(ctx).Error("Failed to send coins from module to account")
+		return sdkerrors.Wrapf(sdkerrors.ErrInvalidRequest, "Failed to send %s from module to account", stCoins.GetDenomByIndex(0))
```
riley-stride, line 116: Error cleanups look good!

app/apptesting/test_helpers.go
```diff
@@ -0,0 +1,67 @@
...
+func (s *AppTestHelper) FundModuleAccount(moduleName string, amount sdk.Coin) {
+	err := s.App.BankKeeper.MintCoins(s.Ctx, moduleName, sdk.NewCoins(amount))
+	s.Require().NoError(err)
+}
+
+func (s *AppTestHelper) FundAccount(acc sdk.AccAddress, amount sdk.Coin) {
+	amountCoins := sdk.NewCoins(amount)
+	err := s.App.BankKeeper.MintCoins(s.Ctx, minttypes.ModuleName, amountCoins)
+	s.Require().NoError(err)
+	err = s.App.BankKeeper.SendCoinsFromModuleToAccount(s.Ctx, minttypes.ModuleName, acc, amountCoins)
```
riley-stride, line 47: why minttypes?

x/stakeibc/keeper/msg_server_liquid_stake_test.go
```diff
@@ -0,0 +1,215 @@
...
+
+type TestCase struct {
+	user         Account
+	module       Account
+	initialState State
+	validMsg     stakeibc.MsgLiquidStake
+}
+
+func (suite *KeeperTestSuite) SetupLiquidStake() TestCase {
+	stakeAmount := int64(1_000_000)
```
riley-stride, line 33: love this convention, will use it going forward!

x/stakeibc/keeper/msg_server_liquid_stake_test.go
```diff
@@ -0,0 +1,215 @@
...
+	suite.CompareCoins(expectedUserAtomBalance, actualUserAtomBalance, "user ibc/uatom balance")
+	suite.CompareCoins(expectedModuleAtomBalance, actualModuleAtomBalance, "module ibc/uatom balance")
+
+	// Confirm deposit record adjustment
+	records := suite.App.RecordsKeeper.GetAllDepositRecord(suite.Ctx)
+	suite.Require().Len(records, 1, "number of deposit records")
+
+	expectedDepositRecordAmount := tc.initialState.depositRecordAmount + stakeAmount.Int64()
+	actualDepositRecordAmount := records[0].Amount
+	suite.Require().Equal(expectedDepositRecordAmount, actualDepositRecordAmount, "deposit record amount")
```
riley-stride, line 119: Nit: can we also check the deposit record "status" and "source" are correct?

app/apptesting/test_helpers.go
```diff
@@ -0,0 +1,67 @@
...
+
+	App         *app.StrideApp
+	Ctx         sdk.Context
+	QueryHelper *baseapp.QueryServiceTestHelper
+	TestAccs    []sdk.AccAddress
+}
+
+func (s *AppTestHelper) Setup() {
+	config := sdk.GetConfig()
+	config.SetBech32PrefixForAccount(Bech32Prefix, Bech32Prefix+sdk.PrefixPublic)
```
riley-stride, line 26: Awesome job powering through this, really was a confusing issue 💪 
The spreadsheet you made took some real dedication!

x/stakeibc/keeper/msg_server_liquid_stake_test.go
```diff
@@ -0,0 +1,215 @@
...
+func (suite *KeeperTestSuite) TestLiquidStakeIbcCoinParseError() {
+	tc := suite.SetupLiquidStake()
+	// Update hostzone with denom that can't be parsed
+	badHostZone := tc.initialState.hostZone
+	badHostZone.IBCDenom = "ibc.0atom"
+	suite.App.StakeibcKeeper.SetHostZone(suite.Ctx, badHostZone)
+	_, err := suite.msgServer.LiquidStake(sdk.WrapSDKContext(suite.Ctx), &tc.validMsg)
+
+	badCoin := fmt.Sprintf("%d%s", tc.validMsg.Amount, badHostZone.IBCDenom)
+	suite.Require().EqualError(err, fmt.Sprintf("failed to parse coin (%s): invalid decimal coin expression: %s", badCoin, badCoin))
```
riley-stride, line 170: where did you get the second argument to `EqualError`? Did you have to trigger the error to find this string?

riley-stride (approved):
Very impressive stuff -- could become the common unit testing framework for the SDK.  Left some nit comments and questions throughout. Your verbal explanation helped a lot to understand this; description is good too. Can you add testing instructions pls?

riley-stride:
One other nit comment: we should prob pull the definitions of `Account`, `State` and `TestCase` out of the liquid stake test to a more general file in `keeper_test` bc they'll be reused across tests (e.g. in RedeemStake's test)

app/apptesting/test_helpers.go
sampocs: Great question! It doesn't really matter what module we use here since the funds are immediately moved out of the module account after minting. But since this function is defined at the app level (not module level), I thought it made more sense to use a generic cosmos module instead of stakeibc - and mint seemed suitable. I believe I also saw one of the other repo's using mint

x/stakeibc/keeper/msg_server_liquid_stake_test.go
sampocs: great call, will add that in!

x/stakeibc/keeper/msg_server_liquid_stake_test.go
sampocs: You can technically do it though tracing the code. You'd see that the parse coin [error](https://github.com/Stride-Labs/stride/pull/87/files#diff-7af1afb39d0c4ae4d8e39395c899dfcb7e81a99901dd0f56213173e1e7d029afR36) returns "`failed to parse coin (%s)`". And that it also wraps the error from an upstream function "`invalid decimal coin expression: %s`". But yeah in practice it's a little easier to run the test once to find the message and copy it. 

I think for most of these I would copy the first half of the error message from the `msg_liquid_stake` file (since that's the message we really want to test), and then I'd get the other error message (the upstream error) by triggering the test once and viewing the output

sampocs:
> One other nit comment: we should prob pull the definitions of `Account`, `State` structs out of the liquid stake test to a more general file in `keeper_test` bc they'll be reused across tests (e.g. in RedeemStake's test)

Yeah I very much agree. But I this is one of those areas where I generally want to start off a little more redundant at the start and wait to see how much overlap we actually have before we start optimizing for organization. For instance, the fields in the state struct will differ depending on the test (e.g. you might not be testing the deposit records). And the Account struct might want to check balances from other coins.

x/stakeibc/keeper/msg_server_liquid_stake_test.go
sampocs: Actually, this function shouldn't change those fields, right?

sampocs: I think at this point, `Source` will be nil, and Status will be set to transfer by default, but isn't touched by the liquid stake function

x/stakeibc/keeper/msg_server_liquid_stake.go
```diff
@@ -91,8 +83,8 @@ func (k msgServer) LiquidStake(goCtx context.Context, msg *types.MsgLiquidStake)
 func (k msgServer) MintStAsset(ctx sdk.Context, sender sdk.AccAddress, amount int64, denom string) error {
```
asalzmann, line 83: I wonder if we should make this function signature 
```suggestion
func (k msgServer) MintStAsset(ctx sdk.Context, sender sdk.AccAddress, amount uint64, denom string) error {
```

and make `Msg.Amount` a `uint64`

buys us a little bit of safety through types

asalzmann (commented):
This is incredible! Thank you for writing this and enabling the rest of the team to write unit tests. I can't wait to get started on writing some tests

x/stakeibc/types/message_set_num_validators_test.go
```diff
@@ -3,28 +3,46 @@ package types
...
+				NumValidators: 1,
+			},
+		},
+		{
+			name: "not enough validators",
+			msg: MsgSetNumValidators{
+				Creator:       adminAddress,
+				NumValidators: 0,
+			},
+			err: sdkerrors.ErrInvalidRequest,
```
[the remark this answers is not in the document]
sampocs: Agreed! I think the `ValidateBasic`'s in general might warrant a lot of custom errors 

Will update!

x/stakeibc/keeper/msg_server_register_host_zone_test.go
```diff
@@ -0,0 +1,5 @@
+package keeper_test
+
+func (suite *KeeperTestSuite) TestRegisterHostZone() {
```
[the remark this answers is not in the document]
sampocs: Whoops didn't realize I included this file. I think that's a great idea! Although I think the scaffolding could be a flushed out a little more before we start adding tags (TestRegisterHostZone is pretty generic). 

I think I'll remove this file from the PR and we can run inventory on our tests later?

app/apptesting/test_helpers.go
```diff
@@ -0,0 +1,62 @@
...
+	testAddrs := make([]sdk.AccAddress, numAccts)
+	for i := 0; i < numAccts; i++ {
+		pk := ed25519.GenPrivKey().PubKey()
+		testAddrs[i] = sdk.AccAddress(pk.Address())
+	}
+
+	return testAddrs
+}
+
+func floatToFixed(num float64, precision int) float64 {
```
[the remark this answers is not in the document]
sampocs: I'm actually no longer using this function - going to remove it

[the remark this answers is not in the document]
sampocs: (also agreed that a function like that would make more sense in utils)

x/stakeibc/keeper/msg_server_liquid_stake.go
```diff
@@ -21,66 +21,58 @@ func (k msgServer) LiquidStake(goCtx context.Context, msg *types.MsgLiquidStake)
...
 	hostZone, err := k.GetHostZoneFromHostDenom(ctx, msg.HostDenom)
 	if err != nil {
-		k.Logger(ctx).Info("Host Zone not found for denom (%s)", msg.HostDenom)
-		return nil, sdkerrors.Wrap(types.ErrInvalidHostZone, "no host zone found for denom")
+		k.Logger(ctx).Error("Host Zone not found for denom (%s)", msg.HostDenom)
+		return nil, sdkerrors.Wrapf(types.ErrInvalidHostZone, "no host zone found for denom (%s)", msg.HostDenom)
 	}
 	// get the sender address
 	sender, err := sdk.AccAddressFromBech32(msg.Creator)
-	if err != nil {
```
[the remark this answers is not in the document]
sampocs: yeah it was removed from validatebasic. Agreed on changing to an underscore.

Riley and I were discussing this a little yesterday. I think I'm more in the camp of not duplicating code (it's just another place that we need coverage, but doesn't add much value). But I also see the case for be redundant

x/stakeibc/keeper/msg_server_liquid_stake_test.go
```diff
@@ -0,0 +1,226 @@
...
+
+import (
+	"fmt"
+
+	epochtypes "github.com/Stride-Labs/stride/x/epochs/types"
+	recordtypes "github.com/Stride-Labs/stride/x/records/types"
+	"github.com/Stride-Labs/stride/x/stakeibc/types"
+	stakeibc "github.com/Stride-Labs/stride/x/stakeibc/types"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	_ "github.com/stretchr/testify/suite"
```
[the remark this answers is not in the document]
sampocs: This line is so that the VSCode linter can identify that these are unit tests (even though they don't extend the golang `testing` package directly like most unit tests do). This adds a play button next to them so you can run them directly from VSCode. 
The underscore is so that the import line doesn't flag an error saying that `suite` is never used.

x/stakeibc/keeper/msg_server_liquid_stake.go
sampocs: I'm on board with that!

sampocs: might make sense to change all our other `amount`s as well (e.g. RedeemStake, DepositRecord), but I think I'll leave that for another PR

app/apptesting/test_helpers.go
asalzmann: Huh, weird. The only slight advantage of using the bank module is that it'll always be initialized, but using the mint module seems fine!

x/stakeibc/keeper/msg_server_register_host_zone_test.go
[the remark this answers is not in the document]
asalzmann: sgtm

app/apptesting/test_helpers.go
sampocs: I agree, I would prefer to use the bank module too. But hacking around it does make me mildly nervous. I think I'll keep it with the mint module cause I'm pretty confident that's how the other project did it, if I'm remembering correctly

PR #88, opened 2022-07-16
Basic integration tests

riley-stride:
## Summary
Adds new integration tests to test the basic user flow in the most basic case. Note:
- The tests are fairly superficial in their current state (e.g. sometimes they check a balance has increased rather than checking it has increased by a particular expected amount). 
- The tests are not optimized, particularly in redeem-stake and claim tests: flows were taking longer than expected to run so I resorted to sleeping extra days rather than the exact amount of time I would have expected the flows to take. We should come back and scrutinize/optimize these. 

## Details
1. Added new integration tests in `integration_tests.bats `(with helper functions for bash dec math as needed)
2. Made all the stakeibc param intervals `1` in `params.go` so that testing is fast + clear

## Test plan + Reviewer instructions
Please run the tests, take a look at the sleep timing in the last two tests (lmk if you can tell why we need to sleep so long, it was surprising to me). Tests take ~15 min (mostly because epochs/days/unbondingperiods are long -- you can shorten them when running the test to make things much faster!

```
make init-local build=sghi
make test-integration
```


scripts-local/tests/integration_tests.bats
```diff
@@ -141,78 +155,124 @@ setup() {
...
+  remaining_seconds=$($STRIDE_CMD q epochs seconds-remaining day)
+  sleep $remaining_seconds
+  remaining_seconds=$($STRIDE_CMD q epochs seconds-remaining day)
+  sleep $remaining_seconds
+
+  RR2=$($STRIDE_CMD q stakeibc list-host-zone | grep -Fiw 'RedemptionRate' | grep -Eo '[+-]?[0-9]+([.][0-9]+)?')
+
+  # check that the exchange rate has increased
+  MULT=1000000
+  RR_INCREASED=$(( $(FLOOR $(DECMUL $RR2 $MULT)) > $(FLOOR $(DECMUL $RR1 10000))))
```
sampocs, line 211: very impressive bash math!

scripts-local/tests/integration_tests.bats
```diff
@@ -141,78 +155,124 @@ setup() {
...
 
+
 # check that redemptions and claims work
-@test "[NOT-IMPLEMENTED] redemption works" {
+@test "[INTEGRATION-BASIC] redemption works" {
+  old_redemption_ica_bal=$($GAIA_CMD q bank balances $REDEMPTION_ICA_ADDR --denom uatom | GETBAL)
   # call redeem-stake
-  # check for an unbonding record
-  # check that a UserRedemptionRecord was created with isClaimabled = false
+  amt_to_redeem=89
```
sampocs, line 220: jc, any particular reason this is 89?

scripts-local/tests/integration_tests.bats
```diff
@@ -44,7 +58,7 @@ setup() {
 ######                              SETUP TESTS                                         ######
 ##############################################################################################
 
-@test "address names are correct" {
+@test "[INTEGRATION-BASIC] address names are correct" {
```
sampocs, line 61: nit: can we add the ICA addresses here as well and then reference them in "host zone successfully registered"?

scripts-local/tests/integration_tests.bats
```diff
@@ -141,78 +155,124 @@ setup() {
...
+
 # check that redemptions and claims work
-@test "[NOT-IMPLEMENTED] redemption works" {
+@test "[INTEGRATION-BASIC] redemption works" {
+  old_redemption_ica_bal=$($GAIA_CMD q bank balances $REDEMPTION_ICA_ADDR --denom uatom | GETBAL)
   # call redeem-stake
-  # check for an unbonding record
-  # check that a UserRedemptionRecord was created with isClaimabled = false
+  amt_to_redeem=89
+  $STRIDE_CMD tx stakeibc redeem-stake $amt_to_redeem GAIA cosmos1g6qdx6kdhpf000afvvpte7hp0vnpzapuyxp8uf \
```
sampocs, line 221: is this the ICQ relayer address? If so, why is that be used instead of one of the validator addresses?

scripts-local/tests/integration_tests.bats
```diff
@@ -141,78 +155,124 @@ setup() {
...
+  # wait for beginning of next day, then for ibc transaction time for the unbonding period to begin
+  remaining_seconds=$($STRIDE_CMD q epochs seconds-remaining day)
+  sleep $remaining_seconds
+  sleep $IBC_TX_WAIT_SECONDS
+  # TODO check for an unbonding record
+  # TODO check that a UserRedemptionRecord was created with isClaimabled = false
   # wait for the unbonding period to pass
-  # check that the tokens were transferred to the delegation account
-  # wait for an epoch to pass
+  UNBONDING_PERIOD=$($GAIA_CMD q staking params |  grep -o -E '[0-9]+' | tail -n 1)
```
sampocs, line 230: very clean having this parametrized!!

sampocs (commented):
Nicely done and very good documentation in each of the tests! Tests all passed for me. Added some nit comments below

asalzmann (commented):
Looking great. Just a few comments! Only major one is the genesis param change

scripts-local/tests/integration_tests.bats
asalzmann: I think this is from 3.sh, long story is we were testing at one point and started by redeeming 100 tokens, then 90 (to differentiate from the previous claim, without killing the network), then went to 89, then the test work. Short story is there's no reason it needs to be 89! :D

riley-stride: Agreed, no reason to be 89 :)

scripts-local/tests/integration_tests.bats
riley-stride: Added to vars.sh but assert_line won't let me pipe them in with `$variable` for some reason (syntax issue)? Will put this on the backlog as a nice to have.

scripts-local/tests/integration_tests.bats
riley-stride: this is a random address (might be the icq relayer, not sure) chosen at random simply to check the functionality of redeem-stake

scripts-local/tests/integration_tests.bats
```diff
@@ -55,15 +69,15 @@ setup() {
...
+@test "[INTEGRATION-BASIC] ibc transfer updates all balances" {
   # get initial balances
   str1_balance=$($STRIDE_CMD q bank balances $STRIDE_ADDRESS --denom ustrd | GETBAL)
   gaia1_balance=$($GAIA_CMD q bank balances $GAIA_ADDRESS --denom $IBC_STRD_DENOM | GETBAL)
   str1_balance_atom=$($STRIDE_CMD q bank balances $STRIDE_ADDRESS --denom $IBC_ATOM_DENOM | GETBAL)
   gaia1_balance_atom=$($GAIA_CMD q bank balances $GAIA_ADDRESS --denom uatom | GETBAL)
   # do IBC transfer
-  $STRIDE_CMD tx ibc-transfer transfer transfer channel-0 $GAIA_ADDRESS 10000ustrd --from val1 --chain-id STRIDE -y --keyring-backend test
-  $GAIA_CMD tx ibc-transfer transfer transfer channel-0 $STRIDE_ADDRESS 10000uatom --from gval1 --chain-id GAIA -y --keyring-backend test
+  $STRIDE_CMD tx ibc-transfer transfer transfer channel-0 $GAIA_ADDRESS 3000ustrd --from val1 --chain-id STRIDE -y --keyring-backend test
```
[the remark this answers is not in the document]
riley-stride: I made this change when I was optimizing for being able to run this test multiple times in sequence, so I reduced the number of tokens to transfer from the full amount in the balance to 1/3 the balance, to allow us to call the tests 3+ times. we could change this back to 10000 if we wanted to, but either should work

scripts-local/tests/integration_tests.bats
```diff
@@ -141,78 +155,124 @@ setup() {
...
+  OLD_STAKE=0
   # wait for another epoch to pass so that tokens are staked
   remaining_seconds=$($STRIDE_CMD q epochs seconds-remaining stride_epoch)
   sleep $remaining_seconds
   # sleep 30 seconds for the IBC calls to settle
   sleep $IBC_TX_WAIT_SECONDS
   # check staked tokens
   NEW_STAKE=$($GAIA_CMD q staking delegation $DELEGATION_ICA_ADDR $GAIA_DELEGATE_VAL | GETSTAKE)
-  assert_equal "$NEW_STAKE" "333"
+  diff=$(($NEW_STAKE - $OLD_STAKE))
```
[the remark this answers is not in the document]
riley-stride: good point, addressed

scripts-local/tests/integration_tests.bats
```diff
@@ -141,78 +155,124 @@ setup() {
...
+  remaining_seconds=$($STRIDE_CMD q epochs seconds-remaining day)
+  sleep $remaining_seconds
+  remaining_seconds=$($STRIDE_CMD q epochs seconds-remaining day)
+  sleep $remaining_seconds
+  remaining_seconds=$($STRIDE_CMD q epochs seconds-remaining day)
+  sleep $remaining_seconds
+
+  # simple check that number of tokens staked increases
+  new_staked_bal=$($GAIA_CMD q staking delegation $DELEGATION_ICA_ADDR $GAIA_DELEGATE_VAL | GETSTAKE)
+  STAKED_BAL_INCREASED=$((new_staked_bal > old_staked_bal))
```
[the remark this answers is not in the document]
riley-stride: Good point, further I was comparing string rather than values 😨  (was missing the `$`). fixed !

scripts-local/tests/integration_tests.bats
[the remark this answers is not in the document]
riley-stride: typo, fixed. good eye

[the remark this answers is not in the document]
riley-stride: the mult is to get them both into integer space for bash math to work smoothly

scripts-local/tests/integration_tests.bats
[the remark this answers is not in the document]
asalzmann: I'm not sure the code got pushed up 🤔 no rush on this - just a heads up!

scripts-local/tests/integration_tests.bats
sampocs: It might be because you're using single quotes in the register host zone test? A single quote in bash doesn't expand the variable. 
Ex: 
```
var="hello"

echo "$var"
>>> hello

echo '$var'
>>> $var
```

Also my comment was a little unclear but I was saying we should add those addresses to this "address names are correct" test as well! In the same way that we validate stride/gaia/icq addresses. 
I thought we ran into a case when testing at one point where our tests were broke cause these addresses got changed, so I was thinking it'd be nice to double check em here so it doesn't happen again!

riley-stride: Great point. Added!

scripts-local/tests/integration_tests.bats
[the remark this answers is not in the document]
riley-stride: pushed!

asalzmann:
I get the following running the tests @riley-stride 
```
Creating host zone...
Error: invalid creator address (stride1uk4ze0x4nvh4fk0xm4jdud58eqn4yxhrt52vv7): invalid address
Usage:
  strided tx stakeibc register-host-zone [connection-id] [host-denom] [bech32prefix] [ibc-denom] [channel-id] [unbonding-frequency] [flags]
```

riley-stride:
> 

Yes, you need to do this every time we develop in local mode, until we merge in Sam's suggestion for putting this address into an env variable

asalzmann:
@vish-stride had mentioned at one point that putting the seed phrase in a local variable was challenging because github actions / our GCP nodes would also need access to it. Is this no longer the case?

Sorry, might have been discussed at standup this morning

riley-stride:
> 

@sampocs @vish-stride did we reach resolution on best way to handle the local admin address? I don't recall

sampocs:
My suggestion was to place the mnemonic as an environment variable in our .zshrc and then restore from that mnemonic. 
This would only be for local mode - our GCP nodes don't need to create this account (since it doesn't call register-host-zone). However, passing a variable like this is easy with github actions secrets, if we ever did want to do that

sampocs:
@riley-stride let's pull this discussion out of this PR and merge this bad boy

PR #90, opened 2022-07-18
Add default vscode settings, lint stride

asalzmann:
**Summary**
Add .vscode/settings.json
Lint project

**Test plan**


app/app.go
```diff
@@ -466,7 +466,10 @@ func NewStrideApp(
 	stakeibcModule := stakeibcmodule.NewAppModule(appCodec, app.StakeibcKeeper, app.AccountKeeper, app.BankKeeper)
 	stakeibcIBCModule := stakeibcmodule.NewIBCModule(app.StakeibcKeeper)
 
-	app.InterchainqueryKeeper.SetCallbackHandler(stakeibcmoduletypes.ModuleName, app.StakeibcKeeper.CallbackHandler())
+	err := app.InterchainqueryKeeper.SetCallbackHandler(stakeibcmoduletypes.ModuleName, app.StakeibcKeeper.CallbackHandler())
+	if err != nil {
+		return nil
```
sampocs, line 471: should this be a panic instead?

x/stakeibc/types/message_add_validator.go
```diff
@@ -59,13 +59,6 @@ func (msg *MsgAddValidator) ValidateBasic() error {
 	if msg.Commission > 100 {
 		return sdkerrors.Wrapf(sdkerrors.ErrInvalidRequest, "commission must be between 0 and 100")
 	}
-	if msg.Commission > 10 {
-		fmt.Sprintf("WARNING: commission is %d (greater than 10pct)", msg.Commission)
-	}
-	// weight validation
-	if msg.Weight < 0 {
```
sampocs, line 66: is this covered already since the variable type is unsigned int?

sampocs (commented):
Lgtm!

sampocs (approved):

x/stakeibc/types/message_add_validator.go
riley-stride: yes but nice to have the checks explicitly for uints IMO in case we mindlessly change the types, I could see that happening with the developer not realizing that the type was acting as a validation of sorts. redundancy here is worth the safety, IMO. wdyt?

app/app.go
riley-stride: probably is worth panicing in app.go if these necessary setups fail. wdyt @asalzmann?

riley-stride:
linting looks good. I don't see the vscode settings

x/stakeibc/types/message_add_validator.go
asalzmann: The linter disallows the check, but it's possible we can add a rule so that the linter ignores this

app/app.go
asalzmann: I'm not sure - I think this goes back to the question of when to use panics vs. errors. In the original thread, the err advocate was saying app.go should not panic in this case, the argument for a panic was that something at a higher level will panic if you err.

I have a slight preference to keep this PR as minimal / lint focused as possible though, so leaving as an err for now (the app won't initialize if nil is returned)

asalzmann:
Sorry - I should have made this a draft PR. It's not quite finished, still need to fix 5-10 lint errors

asalzmann:
I'll try and add this to GH actions as well

sampocs:
I think we still want to keep .vscode in the git ignore for when people open subdirectories with vscode. I haven't tested these but I think you could do either

```
# might ignore folder if it occurs outside of the root directory?
*/.vscode
```
or
```
# might ignore the folder except for this one file
.vscode
!.vscode/settings.json
```

asalzmann:
Do you have an example of a use-case the requires a sub-directory in .vscode?

sampocs:
I meant more that when you open from a subdirectory, vscode automatically creates the `.vscode` directory for you (which we want to be ignored). I was advocating for keeping this main settings.json that you added - while making sure to ignore any other instances of `.vscode`

riley-stride:
LGTM! ty for adding vscode settings. integration tests passing ✅

riley-stride (approved):

asalzmann:
@sampocs makes sense - added .vscode back and added !.vscode/settings.json to track that one file

PR #92, opened 2022-07-18
add IBC transfer timeout

asalzmann:
**Summary**
Add ibc transfer timeout that is relative to the current LC height. Previously, we were timing out IBC transfers at 1_000_000 blocks, which is obviously wrong (we just added this as a stopgap before we could easily pull the LC height!). Concretely,
- Timeout at a relative block height in hooks.go
- Add `DefaultIbcTimeoutBlocks` param

**Open questions**
* How stale could the LC be when this is called? The "real" timeout is the timeout - staleness, so the LC is 20 min stale, the "real" timeout will happen 10 min after the transfer is called. Should we raise the timeout so that this is safer? Higher timeouts don't really matter to Stride

**Test plan**
TODO: run integration tests


x/stakeibc/keeper/hooks.go
```diff
@@ -270,9 +271,15 @@ func (k Keeper) TransferExistingDepositsToHostZones(ctx sdk.Context, epochNumber
 				continue
 			}
 			delegateAddress := delegateAccount.GetAddress()
-			// TODO(TEST-89): Set NewHeight relative to the most recent known gaia height (based on the LC)
 			// TODO(TEST-90): why do we have two gaia LCs?
-			timeoutHeight := clienttypes.NewHeight(0, 1000000)
+			blockHeight, found := k.GetLightClientHeightSafely(ctx, hostZone.ConnectionId)
```
riley-stride, line 275: Looking through `GetLightClientHeightSafely`: do we need to `return 0, false` in `msg_server_submit_tx.go: GetLightClientHeightSafely:259`?

x/stakeibc/keeper/hooks.go
```diff
@@ -270,9 +271,15 @@ func (k Keeper) TransferExistingDepositsToHostZones(ctx sdk.Context, epochNumber
...
+			if !found {
+				k.Logger(ctx).Error(fmt.Sprintf("Could not find blockHeight for host zone %s", hostZone.ConnectionId))
```
riley-stride, line 277: Can we be a little more verbose here, something like 
```suggestion
				k.Logger(ctx).Error(fmt.Sprintf("Could not find blockHeight for host zone %s, aborting transfers to host zone this epoch", hostZone.ConnectionId))
```

riley-stride:
LGTM! Please run integration tests before merging

riley-stride (approved):
LGTM! Just please hit the error return comment on GetLightClientHeightSafely() please. In integration tests pass let's merge!

PR #94, opened 2022-07-18
added colour on the three types of epochs. epochs module documentatio…

viroot27:
…n done

Let me know what you think. Am I correct in saying that we may change the stride epoch length  in the future?


sampocs:
Looks good! You can remove "week" from the docs as that will soon be deleted from the code. 

I'd have to defer to someone else on the question of whether we'll ever change the length of the epoch. My guess is that it's pretty unlikely we'd ever change it after launch.

PR #95, opened 2022-07-19
[Depracated] Gas spam defense

riley-stride:
## Summary
Adding in logic to support stopping a spam attack against the network. If admin observes a spam attack underway, they can set a flag that will charge gas (in ustrd) to users who are liquid staking small amounts (less than 10 units of host zone currency). The logic is that spam attackers would not be economically feasible in they liquid staked large amounts. 

The default fee size is 10_000 ustrd = 0.01 str. The fee is turned off by default at genesis.

_Note_: this logic ONLY stops spam attacks launched through txs of type `liquidStake()`. No other message types will be charges gas fees due to this attack response. Those will need to be built out separately. 

## Details
- We scaffold a message to set the safety gas fee status (boolean). This is it's own commit, for clarity. Please review separately for clarity. We add a struct in `safety_gas_fee_flag.proto` that can be set only by the admin (gated at validateBasic()) through the scaffolded message `set-safety-gas-fee-flag`. It holds a single boolean: `Enabled`.
- In `x/stakeibc/keeper.go` we write a function `ConsumeSafetyGasFee()` which charges the user `SafetyGasFee` gas (denominated in ustrd). This function is triggered from msg_server_liquid_stake, only if the conditions are met
- In `msg_server_liquid_stake.go`, we add conditions to check the `SafetyGasFeeFlag` param, and invoke `ConsumerSafetyGasFee` iff the param is enabled and the amount liquid staked is LT 10 units of the token being liq staked (e.g. 10atom). We hardcode the threshold for now but could move it to governance as a param if desired
- We add a parameter in stakeibc's `params.go` which decides how much gas to charge users staking low amounts when the SafetyGasFee is turned on. We set this param to 1ustrd.
- We modify the test script `2.sh` to reduce the amount liquid staked to LT the gas fee threshold (for testing)

## Test plan
Build the chain locally and uncomment the test address in `utils.go`, then:
1. verify you can liquid stake amounts LT 10atom gaslessly with the standard flow. 
Setup with:
`sh scripts-local/test-utils/1.sh`
Drain all ustrd from the val1 account with
`build/strided --home ./scripts-local/state/stride tx bank send val1 stride1ft20pydau82pgesyl9huhhux307s9h3078692y 499000000000ustrd --chain-id STRIDE`
Liquid stake from val1, which has no ustrd (this should work, proving you can liq stake gaslessly
`sh scripts-local/test-utils/2.sh`
Then, suppose there's a spam attack on the network! We turn on the safety gas fee flag
`build/strided --home ./scripts-local/state/stride tx stakeibc set-safety-gas-fee-flag true --from val1 --chain-id STRIDE --keyring-backend test`
Now try liquid staking again gaslessly from the same account which has 0ustrd
`sh scripts-local/test-utils/2.sh`
This should fail! 
You can fund the account with some gas and try again, now the tx should succeed
If you'd like you can set the flag back to false to verify we can send messages again after disabling the flag.


asalzmann:
What's the status of this PR?

riley-stride:
Closing since this approach does not work

asalzmann:
For posterity, is the correct approach to add a function that can set a chain-wide `min_gas_price`? (either through governance or a permissioned function call)

PR #97, opened 2022-07-19
Local Mode Admin

sampocs:
## Summary
* Created local mode admin key. Mnemonic must be set in the zshrc with `export STRIDE_ADMIN_MNEMONIC='seed phrase here ...'`. 
 
I still think there's a better way we could do this, but I think this is marginally better than the constant comment/uncomment approach. Let's try and think of something better by launch.
* Moved post-chain startup logic to separate script (e.g. register host zone, create validators)
* Updated the client configs for local mode so we don't have to specify the chain-id and keyring-backend in every command
* **Question for y'all**: Does it make more sense to have a separate address for local mode or should we re-use the testnet one?


riley-stride:
test

sampocs:
ping

asalzmann:
test

riley-stride (approved):
lgtm! tests passed, updated description to describe how to set the env var

sampocs:
great call, ty!

Yaykut:
great

PR #105, opened 2022-07-22
[PLAT-2] Restore ICA Channel

sampocs:
### Summary
* Added message `RestoreInterchainAccount` to restore an ICA Account after the channel has gone down
* Below I've included the testing plan I went through. It's a little convoluted, so I'll leave it up to the descretion of the reviewer whether they want to run through this again.

#### Setup
* Replace L139 in `msg_server_submit_tx.go` with the following (this will cause delegations between blocks 75 and 150 to timeout):
```
if ctx.BlockHeader().Height > 75 && ctx.BlockHeader().Height < 150 {
	_, err = k.SubmitTxs(ctx, connectionId, msgs, *delegationIca, uint64(ctx.BlockTime().UnixNano()+1_000_000_000))
} else {
	_, err = k.SubmitTxsStrideEpoch(ctx, connectionId, msgs, *delegationIca)
}
```
* Startup local mode
* Liquid stake using the normal setup scripts (try to submit before block 65)
* Confirm the liquid stake registered in the GAIA host zone 
	* `stakedBalance` field listed http://localhost:1317/Stride-Labs/stride/stakeibc/host_zone 
* After block 75, issue another liquid stake
* Confirm channel closes (takes ~1 min) 
	* http://localhost:1317/ibc/core/channel/v1/channels/channel-1/ports/icacontroller-GAIA.DELEGATION
* Confirm the staking balance does not reflect the most recent liquid stakes
	
#### Success Case
* Wait until block 150 and then restore the channel
```
build/strided --home scripts-local/state/stride tx stakeibc restore-interchain-account GAIA DELEGATION --from admin
```
* Wait ~1 min and then check the total staked balance again and it should reflect the stakes that occured while the channel was down
* Confirm a second channel was created for the delegation ICA
```
build/strided --home scripts-local/state/stride q ibc channel channels
```
#### Failure Cases
* Try to restore the account again and check the tx logs (it should fail with an error saying the channel already exists)
* Try to restore the account, but this time include a typo in DELEGATION (it should fail immediately)
```
build/strided --home scripts-local/state/stride tx stakeibc restore-interchain-account GAIA DEL --from admin
```
* Try to restore the account, but this time include a typo in GAIA (it should saying host zone doesn't exist)
```
build/strided --home scripts-local/state/stride tx stakeibc restore-interchain-account G DELEGATION --from admin
```
* Finally, try from a different address (it should fail with invalid address)
```
build/strided --home scripts-local/state/stride tx stakeibc restore-interchain-account GAIA DEL --from val1
```


x/stakeibc/types/message_restore_interchain_account.go
```diff
@@ -0,0 +1,51 @@
...
+}
+
+func (msg *MsgRestoreInterchainAccount) ValidateBasic() error {
+	_, err := sdk.AccAddressFromBech32(msg.Creator)
+	if err != nil {
+		return sdkerrors.Wrapf(sdkerrors.ErrInvalidAddress, "invalid creator address (%s)", err)
+	}
+	if err := utils.ValidateAdminAddress(msg.Creator); err != nil {
+		return err
+	}
```
riley-stride, line 49: Can/should we add a check on the accountType here as well? Could rip some of the checks from `message_restore_interchain_account.go`

riley-stride:
side note: incredible approach to add links to localhost in testing instructions in the PR description, SO much easier to verify/review that way. Awesome idea

riley-stride (approved):
IMO, to date the most beautiful PR description Stride Labs has pushed. I ran through the flow and everything worked as expected. Let's 🚢 it!

x/stakeibc/module_simulation.go
```diff
@@ -23,6 +23,14 @@ var (
 	_ = baseapp.Paramspace
 )
 
+const (
```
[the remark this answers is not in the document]
sampocs: scaffolded!

x/stakeibc/types/message_restore_interchain_account.go
```diff
@@ -0,0 +1,51 @@
...
+	bz := ModuleCdc.MustMarshalJSON(msg)
+	return sdk.MustSortJSON(bz)
+}
+
+func (msg *MsgRestoreInterchainAccount) ValidateBasic() error {
+	_, err := sdk.AccAddressFromBech32(msg.Creator)
+	if err != nil {
+		return sdkerrors.Wrapf(sdkerrors.ErrInvalidAddress, "invalid creator address (%s)", err)
+	}
+	if err := utils.ValidateAdminAddress(msg.Creator); err != nil {
```
asalzmann, line 47: I think we should allow anyone to restore this - there's no reason to restrict it (afaict) and I think we should remove ADMINS altogether before launch

x/stakeibc/types/params.go
```diff
@@ -19,8 +19,8 @@ var (
 	DefaultStrideCommission              uint64 = 10
 	DefaultValidatorRebalancingThreshold uint64 = 100 // divide by 10,000, so 100 = 1%
 	// 10 minutes
-	DefaultICATimeoutNanos uint64 = 600000000000
-	DefaultBufferSize      uint64 = 20
+	DefaultICATimeoutNanos  uint64 = 600000000000
```
asalzmann, line 22: nit: revert (just trying to minimize LOC changed!)

x/stakeibc/types/codec.go
```diff
@@ -45,6 +46,9 @@ func RegisterInterfaces(registry cdctypes.InterfaceRegistry) {
 	registry.RegisterImplementations((*sdk.Msg)(nil),
 		&MsgDeleteValidator{},
 	)
+	registry.RegisterImplementations((*sdk.Msg)(nil),
```
asalzmann, line 49: nit: move into above `RegisterImplementations` (we should just consolidate all of these I think)

x/stakeibc/types/message_restore_interchain_account.go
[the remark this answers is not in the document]
asalzmann: +1

x/records/keeper/keeper.go
```diff
@@ -16,11 +16,11 @@ import (
 type (
 	Keeper struct {
 		// *cosmosibckeeper.Keeper
-		Cdc          codec.BinaryCodec
-		storeKey     sdk.StoreKey
-		memKey       sdk.StoreKey
-		paramstore   paramtypes.Subspace
-		scopedKeeper capabilitykeeper.ScopedKeeper
+		Cdc           codec.BinaryCodec
```
asalzmann, line 19: nit: revert (to minimize LOC changed)

x/stakeibc/client/cli/tx_restore_interchain_account.go
```diff
@@ -0,0 +1,51 @@
...
+	cmd := &cobra.Command{
+		Use:   "restore-interchain-account [chain-id] [account-type]",
+		Short: "Broadcast message restore-interchain-account",
+		Args:  cobra.ExactArgs(2),
+		RunE: func(cmd *cobra.Command, args []string) (err error) {
+			argChainId := args[0]
+			argAccountType := args[1]
+
+			// QUESTION: Does it make more sense for this to be in validate basic?
+			accountType, found := types.ICAAccountType_value[argAccountType]
```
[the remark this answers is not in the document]
asalzmann: I think it should be in both - putting it in the CLI prevents users from sending txs that will fail, `ValidateBasic` does the real safety checks

x/stakeibc/keeper/msg_server_restore_interchain_account.go
```diff
@@ -0,0 +1,27 @@
...
+	ctx := sdk.UnwrapSDKContext(goCtx)
+
+	hostZone, found := k.GetHostZone(ctx, msg.ChainId)
+	if !found {
+		k.Logger(ctx).Error(fmt.Sprintf("Host Zone not found: %s", msg.ChainId))
+		return nil, types.ErrInvalidHostZone
+	}
+
+	owner := types.FormatICAAccountOwner(msg.ChainId, msg.AccountType)
+	if err := k.ICAControllerKeeper.RegisterInterchainAccount(ctx, hostZone.ConnectionId, owner); err != nil {
```
asalzmann, line 21: note: we're making an assumption that `ConnectionId` stored on the host zone matches what's in the IBC store. If we ever need to upgrade a connection (unsure if / why we'd have to do so), we'll need to also update the host zone.

I think we make this assumption in a lot of other places as well, just something for us to keep in mind

asalzmann (approved):
Looks awesome! Thank you for making this change

x/stakeibc/client/cli/tx_restore_interchain_account.go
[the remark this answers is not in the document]
sampocs: `ValidateBasic` takes type `MsgRestoreInterchainAccount` as a parameter - which has the enum as a field in the struct. So I don't think it's possible to get into validate basic with an invalid account type.

Does that sound right to you?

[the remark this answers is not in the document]
asalzmann: Ah yep totally - good catch

x/stakeibc/types/params.go
sampocs: woah, not even sure how this got changed

x/stakeibc/types/message_restore_interchain_account.go
riley-stride: @asalzmann can you auth with axolo so that these messages show up from your account , just slightly more clear [](edited-with-Axolo)

sampocs: just pushed up a commit with the PR changes address, let me know if I have your blessing [](edited-with-Axolo)

sampocs: running the integration tests now too [](edited-with-Axolo)

asalzmann:
lgtm!

asalzmann:
axolo test

x/stakeibc/types/message_restore_interchain_account.go
riley-stride: I've found the axolo channels useful so far for chatting about specific PRs internally with other reviewers , and only putting code comments on the PR itself on GH [](edited-with-Axolo)

asalzmann: I'm down to try axolo! Have just noticed that tools outside of GH tend to get used less over time (maybe slack is integrated enough though that we'll use this) [](edited-with-Axolo)

riley-stride: yeah down to test drive it and ditch it if it gets to use [](edited-with-Axolo)

sampocs: agreed. fwiw, I've liked it a lot so far. I'm much more up to date on the status of these PRs then I used to be

riley-stride: I am too, nice to have the ping on slack rather than having to swap between slack and email [](edited-with-Axolo)

asalzmann: the other minor comment I'll make is that I think we should try to push convos to GH for posterity (ik we can unarchive slack, but it's less ideal - if we lose access to slack or if someone else wants to understand why we made certain decisions, it's less transparent) [](edited-with-Axolo)

PR #106, opened 2022-07-22
[PLAT-1] Soft-Upgrade: Governance Parameters

sampocs:
### Summary
* Added handlers for governance parameter upgrades
* Extended local mode setup to run upgrades automatically with cosmovisor and helper scripts
* Please see further details and run instructions in `scripts-local/upgrades/README.md`
* Main entrypoint for testing upgrades is in `scripts-local/start_upgrade.sh`
* I realize we could probably do a better job of combining this upgrade script with the regular local mode scripts (maybe always run local mode with cosmovisor?) but for the sake of time and priorities, I took the quicker approach of just making a separate script
* The relevant lines of `start_upgrades.sh` (i.e. the lines that were not already in `start_networks.sh`) are:
  * [L7-9](https://github.com/Stride-Labs/stride/pull/106/files#diff-df90b08290500c9c68dd5a3f2556edec7b537cd7d7457e42e86012638f10c29bR7-R9)
  * [L31-51](https://github.com/Stride-Labs/stride/pull/106/files#diff-df90b08290500c9c68dd5a3f2556edec7b537cd7d7457e42e86012638f10c29bR31-R51)
  * [L90-END](https://github.com/Stride-Labs/stride/pull/106/files#diff-df90b08290500c9c68dd5a3f2556edec7b537cd7d7457e42e86012638f10c29bR90-R98)
* **NOTE**: I set up the PR with `PLAT-2` as the base branch instead of main (since that's what I branched off), but this PR is meant to be merged into main.

### Remaining Work

- [X] I wasn't sure what parameters and values we want so I'll fill those in once we decide 
- [X]  I overrode the `STRIDE_CMD` variable in `init_stride.sh` to get this to run (which would break local mode). But before merging, I'll think of a better way to change that variable based on whether we're running normally or in upgrade mode

### Test plan
* Follow instructions in `scripts-local/upgrades/README.md`
* Test queries to confirm upgrade: 
```
# Before upgrade 
scripts-local/upgrades/binaries/strided1 --home scripts-local/state/stride q gov params

# After upgrade
scripts-local/upgrades/binaries/strided2 --home scripts-local/state/stride q gov params
```


scripts-local/vars.sh
```diff
@@ -25,6 +25,8 @@ DAY_EPOCH_LEN="100s"
 STRIDE_EPOCH_INDEX=2
 STRIDE_EPOCH_LEN="40s"
 IBC_TX_WAIT_SECONDS=30
+MAX_DEPOSIT_PERIOD="30s"
```
riley-stride, line 28: Why 30s instead of 60m?

sampocs: This is just the initial setup so that you run locally, submit a proposal, and test the upgrade. These settings are generalized across all local mode upgrade runs - they are not connected to the actual upgrade logic (which in this case happened to also be updating these parameters)

riley-stride (approved):
Discussed and tested this one live. Let's ship!

PR #124, opened 2022-07-24
fix gosec issues and add gosec action

asalzmann:
# Summary
- Add gosec to actions, a golang security checker for the cosmos sdk built by informal systems https://github.com/informalsystems/gosec
- Fix a number of issues raised by gosec
- Add a `gosec` and `ci` target to the Makefile

### gosec fixes
There were two main types of changes raised by gosec

#### Second, non-deterministic map iteration was fixed
In Go, iterating over maps is intentionally non-deterministic as the runtime defines. Unfortunately for us, in the Cosmos-SDK, we encountered an issue with non-deterministic upgrades in [Issue cosmos-sdk#10188](https://github.com/cosmos/cosmos-sdk/issues/10188) [PR cosmos-sdk#10189](https://github.com/cosmos/cosmos-sdk/pull/10189) that resulted from exactly this non-deterministic iteration. To ensure determinism, we only permit an iteration to retrieve the map keys and then those keys can then be sorted, so instead of
```go
for k, v := range m {
    // Do something with key and value.
    _, _ = k, v
}
```

the requested pattern is
```go
keys := make([]string, 0, len(m))
for k := range m {
    keys = append(keys, k)
}
sort.Strings(keys)

for _, key := range keys {
    // Use the value.
    _ = m[key]
}
```

go doesn't support generics, so I added a number of files in utils to handle different map types

# Test plan
Install gosec locally with `go install github.com/securego/gosec/v2/cmd/gosec@latest`
Then, make sure gosec action passes: `make gosec`


utils/utils.go
```diff
@@ -48,6 +49,42 @@ func Min(a int, b int) int {
...
+func HostZoneUnbondingKeys(m map[string]*recordstypes.HostZoneUnbonding) []string {
+	keys := make([]string, 0, len(m))
+	for k := range m {
+		keys = append(keys, k)
+	}
+	sort.Strings(keys)
+	return keys
+}
+
+func StringToIntMapKeys(m map[string]int64) []string {
```
riley-stride, line 61: I think you can collapse these functions into one using this approach
https://go.dev/play/p/g8BXwgQq14S

x/epochs/client/cli/query.go
```diff
@@ -138,7 +139,7 @@ func GetCmdSecondsRemaining() *cobra.Command {
 			}
 
 			// duration: seconds
-			duration := int64(res.Epoch.Duration.Seconds())
+			duration := cast.ToInt64(res.Epoch.Duration.Seconds())
```
riley-stride, line 142: These casts are great. Should we also be replacing instances of `strconv.FormatInt` with casts, e.g. line 150?

x/interchainquery/keeper/abci.go
```diff
@@ -54,7 +55,7 @@ func (k Keeper) EndBlocker(ctx sdk.Context) {
 			// query was removed; delete datapoint
 			k.DeleteDatapoint(ctx, dp.Id)
 		} else {
-			if dp.LocalHeight.Int64()+int64(q.Ttl) > ctx.BlockHeader().Height {
+			if dp.LocalHeight.Int64()+cast.ToInt64(q.Ttl) > ctx.BlockHeader().Height {
```
riley-stride, line 58: Nit: looks like we're now using casts where we previously were using `int64(X)` but not where we are using `X.Int64()`. Is your thinking that in the second case, perhaps the type has a custom `Int64()` conversion method?

x/stakeibc/module_simulation.go
```diff
@@ -24,7 +25,7 @@ var (
 )
 
 const (
-	opWeightMsgRestoreInterchainAccount = "op_weight_msg_register_interchain_account"
+	opWeightMsgRestoreInterchainAccount = "op_weight_msg_register_interchain_account" // #nosec
```
riley-stride, line 28: why nosec here?

riley-stride (commented):
Looks good and the tests passed for me! Added the gosec install instructions to the PR description and a few comments throughout. Two main Qs:
- should we be using `cast` in more places than the changes shown in this PR?
- nit: can we condense the utils.go by using interfaces (see the go playground example in my comment)

riley-stride:
Please check whether we need to sort app.go:849 before merging

x/stakeibc/module_simulation.go
asalzmann: it was flagging it as a private key for some reason

x/epochs/client/cli/query.go
asalzmann: This just returns strings, which I think can be arbitrarily long (no overflow risk) so I think using `strconv.FormatInt` is safe

x/interchainquery/keeper/abci.go
asalzmann: I just fixed what was flagged by gosec

It looks like `Int64.(...)` is a method defined on `sdk.Int`, so I imagine it's safe!

utils/utils.go
asalzmann: I think this does work, but the linter doesn't like it

asalzmann:
I don't think we need to sort in app.go:849 because it returns another map (the order of which is non-deterministic anyway)

riley-stride (approved):
lgtm, discussed in person!

PR #128, opened 2022-07-25
Distribution

riley-stride:
## Summary
We point to a new version of the cosmos-sdk that has updates to the distribution module which blacklist a chosen validator (and all who delegate to it) from receiving staking rewards, redirecting those rewards to other validators proportionally. These changes occur in the distribution module, which we have modified in our fork of the cosmos-sdk repo, linked below. This PR to the stride repo simply replaces the cosmos-sdk repo we import and adds logic to stand up a second validator node on Stride, in order to test the changes to the distribution module. 

The distribution module logic changes live in the [associated pull request in the Stride cosmos-sdk repo](https://github.com/Stride-Labs/cosmos-sdk/pull/1).  

## Details
- changed `go.mod` to point to the updated `cosmos-sdk` repo which contains changes to `x/distribution`
- added scripts to spin up two stride validators (ty vishal!) for testing

## Testing
To test, init locally and inspect the accrued rewards on each validator. Those on the `stride1uk...` will only ever increase by the proposer reward rate, which is very small. The two validators' rewards should diverge meaningfully!

To test that the rewards that would have accrued to the blocked validator instead accrue to the other, you can check the rate at which rewards accrue to the non-blocked validator. It seems to be intervals of `0.12STRD` (you can refresh the block explorer after ~30 sec to see the accrued rewards update). Then, if you tweak the validator address coded in `x/distribution/keeper/allocation.go` (e.g. add a few random characters at the end) so that the blocking is disabled, point stride to that updated code (push a tag and update go.mod's replace to point to the updated tag), you should see the rate of increase in accrued rewards is roughly equal across the two validators at `0.06STRD`. When blocking is enabled, the `0.06STRD` that would have accrued to `stride1uk...` instead accrues to the other validator!

## Before Mainnet
We'll need to change the blacklisted address before mainnet, potentially even create several so that we can distribute stake amongst them and thus keep the network decentralized.


asalzmann:
couple qqs
- What commits should we diff to review this?
- Once ready, should we add this to an sdk fork?

vish-stride (approved):
Do you know where the line is that adds the blacklisted validators? 

Main question on my mind is how can we update that set after launch? My prior was that we won't run any validator infra, but rather would delegate to other professional validators. Can we add to / subtract from that set after launch? I assume so (at the very least, we can software upgrade to update our Cosmos SDK)

On a BD note, we should start reaching out to validators to ask them if they can set up an extra validator for us + investors

asalzmann:
Sorry for the delay - will review this tomorrow!

asalzmann (commented):
Nice! I reviewed this as well https://github.com/Stride-Labs/cosmos-sdk/pull/1#pullrequestreview-1059498138

riley-stride:
> Do you know where the line is that adds the blacklisted validators?
> 
> Main question on my mind is how can we update that set after launch? My prior was that we won't run any validator infra, but rather would delegate to other professional validators. Can we add to / subtract from that set after launch? I assume so (at the very least, we can software upgrade to update our Cosmos SDK)
> 
> On a BD note, we should start reaching out to validators to ask them if they can set up an extra validator for us + investors

Good Qs -- blacklsited val list is a param in distr module now, updateable by governance. See [here](https://github.com/Stride-Labs/cosmos-sdk/pull/1)

asalzmann:
What's the status of this PR? (not urgent, just want to make sure we don't lose the thread)

riley-stride:
Please add firehose test results to this before merging
* Instructions on how to run them [here](https://github.com/Stride-Labs/firehose/pull/1)
* Example of including them in a Stride PR [here](https://github.com/Stride-Labs/stride/pull/190#:~:text=2022/08/13%2022%3A54%3A25%20TOTAL%3A%20100/100(100.00)%20Success)

asalzmann:
https://github.com/Stride-Labs/stride/pull/253

PR #131, opened 2022-07-25
Removed v2 upgrade

sampocs:
**Summary**
* Removed v2 upgrade (since this will be a chain reset) but kept boilerplat


.gitignore
```diff
@@ -44,5 +44,5 @@ vue/*
 .terraform
 *.pem
 
-.vscode/*
+.vscode
```
asalzmann, line 47: Out of curiosity, what's the diff here?

asalzmann (approved):
lgtm!

tip: you can request @Stride-Labs/core-chain-eng as a reviewer to ping everyone now!

.gitignore
sampocs: New one will look recursively into each directory, old one only looks at the root level

asalzmann: nice! sorry, think you may have mentioned this on an older PR and I missed it

sampocs: No worries! It’s pretty unintuitive, I still don’t really understand why it works like that

PR #136, opened 2022-07-30
IBC-02 | Missing Important Validation Step

asalzmann:

Add validation step to ibc handler

> In L227 in file x/stakeibc/keeper/ibc_handler.go, epochUnbondingRecord.Id is used to compared to the dayEpochTracker.EpochNumber. According to the design, it should be variable epochUnbondingRecord.UnbondingEpochNumber rather than epochUnbondingRecord.Id

## Brief Changelog

- Check that `undelegateAmt` is g.t. 0

## Testing and Verifying

- CI

### DIFF
--- a/x/stakeibc/keeper/ibc_handlers.go
+++ b/x/stakeibc/keeper/ibc_handlers.go
@@ -364,7 +364,8 @@ func (k Keeper) HandleUndelegate(ctx sdk.Context, msg sdk.Msg, completionTime ti
 
 	
 	undelegateAmt := undelegateMsg.Amount.Amount.Int64()
-	if undelegateAmt < 0 {
+	if undelegateAmt <= 0 {
+
 		return sdkerrors.Wrapf(sdkerrors.ErrLogic, "Undelegate amount was negative")
 	}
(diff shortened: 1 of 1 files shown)


x/stakeibc/keeper/ibc_handlers.go
```diff
@@ -362,7 +362,11 @@ func (k Keeper) HandleUndelegate(ctx sdk.Context, msg sdk.Msg, completionTime ti
 		return sdkerrors.Wrapf(sdkerrors.ErrLogic, "Undelegate message was not for a delegation account")
 	}
 
+	
 	undelegateAmt := undelegateMsg.Amount.Amount.Int64()
+	if undelegateAmt < 0 {
```
riley-stride, line 367:
```suggestion
	if undelegateAmt <= 0 {
```

riley-stride: We probably want to fail it at 0, right? Or should/do we stop that case earlier so that the 0-balance redemption is never issued in the first place?

riley-stride (approved):
Can you check my (low confidence) suggestion before merging?

PR #142, opened 2022-07-30
HOS-02 : registerHostZone duplicate denom check

riley-stride:

> The only place hostZones are created is in msg_register_host_zone, so we check for already-existing hostZones with the same denom as that which is being registered. If one is found, we error

## Brief Changelog

  - Add the check in `x/stakeibc/keeper/msg_server_register_host_zone.go`.

## Testing and Verifying
Run the integration and unit tests


asalzmann:
lgtm - should we also check that
- connectionId
- bech32prefix
are unique?

asalzmann (approved):

PR #147, opened 2022-07-31
finding a bug

snipeTR:
find a bug in stride's core repo (must be a non-trivial bug)


riley-stride:
Hey! What's the bug?

snipeTR:
> thanks for diving into the code! can you explain what bug you're flagging?

./... invalid file system referance.

PR #148, opened 2022-08-01
Update hermes.toml

karborantr:
trusting_period fixed to 8 hours

### DIFF
--- a/scripts/config/hermes.toml
+++ b/scripts/config/hermes.toml
@@ -42,7 +42,7 @@ max_msg_num = 30
 max_tx_size = 2097152
 clock_drift = '5s'
 max_block_time = '10s'
-trusting_period = '14days'
+trusting_period = '8hrs'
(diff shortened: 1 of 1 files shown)


asalzmann:
Can you explain this fix?

karborantr:
> Can you explain this fix?

Ensuring the IBC Client Isn't Expired
In the event that Tendermint consensus fails (if >1/3 of validators produce a conflicting block), and proof of this consensus failure is submitted on-chain, the IBC client becomes frozen with a frozen_height that is nonzero.The output of strided query ibc channel client-state confirms the client status and you know the IBC client is not expired.

For Tendermint clients, the frequency at which Hermes refreshes them is 2/3 of their trusting period (e.g., refresh every ~9 days if the trusting period is 14 days).
Stride unbounding period is 10hours 30 minutes. So we set te trusting period is 8 Hours.

asalzmann:
Gotcha! We don't use this script for testnet, so going to close this PR.

PR #139, opened 2022-07-30
HOS-01 | Lack of Sanity Check - Remove Validator

asalzmann:
## What is the purpose of the change
When removing a validator with non-zero delegation amount from the set, those funds will be stranded until we re-add this validator.

## Brief Changelog

- ensure that Validator.DelegationAmt and Validator.Weight is ZERO before removing the target validator from a host zone.
- when rebalanding validators, sort ascending to avoid last-validator precision amount issue. See this example code: https://go.dev/play/p/3Zd_cVKZVyU

## Testing and verifying

CI + integration

## Documentation and Release Note

  - Does this pull request introduce a new feature or user-facing behavior changes? no
  - Is a relevant changelog entry added to the `Unreleased` section in `CHANGELOG.md`? no
  - How is the feature or change documented? audit

PR #164, opened 2022-08-07
Revert "Revert "HOS-01 | Lack of Sanity Check - Remove Validator""

asalzmann:
adds back https://github.com/Stride-Labs/stride/pull/139

When removing a validator with non-zero delegation amount from the set, those funds will be stranded until we re-add this validator.

## Brief Changelog

- ensure that Validator.DelegationAmt and Validator.Weight is ZERO before removing the target validator from a host zone.
- when rebalanding validators, sort ascending to avoid last-validator precision amount issue. See this example code: https://go.dev/play/p/3Zd_cVKZVyU

## Testing and verifying

CI + integration


asalzmann:
(this was previously approved)

PR #170, opened 2022-08-09
certik 8.9.22

riley-stride:
See skyharbor for description, everything is pretty nit

### DIFF
--- a/x/interchainquery/keeper/abci.go
+++ b/x/interchainquery/keeper/abci.go
@@ -11,10 +11,6 @@ import (
 	"github.com/Stride-Labs/stride/x/interchainquery/types"
 )
 
-const (
-	RetryInterval = 25
-)
-
 // EndBlocker of interchainquery module
 func (k Keeper) EndBlocker(ctx sdk.Context) {
 	defer telemetry.ModuleMeasureSince(types.ModuleName, time.Now(), telemetry.MetricKeyBeginBlocker)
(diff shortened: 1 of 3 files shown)


x/records/keeper/deposit_record.go
```diff
@@ -75,6 +76,12 @@ func (k Keeper) GetDepositRecord(ctx sdk.Context, id uint64) (val types.DepositR
 func (k Keeper) RemoveDepositRecord(ctx sdk.Context, id uint64) {
 	store := prefix.NewStore(ctx.KVStore(k.storeKey), types.KeyPrefix(types.DepositRecordKey))
 	store.Delete(GetDepositRecordIDBytes(id))
+	count := k.GetDepositRecordCount(ctx)
+	if count > 0 {
```
asalzmann, line 80: flagging as a bug (per offline discussion) that we should not merge in

asalzmann (changes requested):

x/records/keeper/deposit_record.go
riley-stride: reverted, ty

x/interchainquery/keeper/abci.go
```diff
@@ -11,10 +11,6 @@ import (
 	"github.com/Stride-Labs/stride/x/interchainquery/types"
 )
 
-const (
-	RetryInterval = 25
```
asalzmann, line 15: Is this just an unused variable? If so, good to merge

asalzmann (approved):

x/interchainquery/keeper/abci.go
riley-stride: yes

PR #178, opened 2022-08-11
use sprintf to fix error messages

asalzmann:

The logger throws an error if a string is formatted incorrectly (e.g. Logger.Info("This should be an int: %s", 1)). This pollutes our error logs, a better pattern to enforce is to use `k.Logger(ctx).Info(fmt.Sprintf("... %s", stringToLog))`

Also, logging without Sprintf fails to log objects (e.g. `k.Logger(ctx).Error("this is an error object %s", object)` doesn't work (it's possible `Errorf` fixes this, but Sprintf is a core golang library so I have a slight preference to use that and pipe strings into `Error`)

## Brief Changelog

- Update logging to use `Sprintf` throughout the codebase

  - How is the feature or change documented? an

## Additional work
- open a cosmos SDK issue
- add informalsystems gosec rule to ensure the Logger isn't trying to replace variables


x/icacallbacks/keeper/keeper.go
```diff
@@ -105,15 +105,15 @@ func (k Keeper) CallRegisteredICACallback(ctx sdk.Context, modulePacket channelt
...
+		k.Logger(ctx).Error(fmt.Sprintf("CallRegisteredICACallback %s", err.Error()))
 		return err
 	}
 	// call the callback
 	if callbackHandler.HasICACallback(callbackData.CallbackId) {
 		// if acknowledgement is empty, then it is a timeout
 		err := callbackHandler.CallICACallback(ctx, callbackData.CallbackId, modulePacket, acknowledgement, callbackData.CallbackArgs)
 		if err != nil {
-			k.Logger(ctx).Info("CallRegisteredICACallback", "err", err)
+			k.Logger(ctx).Error(fmt.Sprintf("CallRegisteredICACallback %s", err.Error()))
```
sampocs, line 116: nit: should we put a more specific error here? Something like "no callback handler found for module %s" and "error executing {callbackId} callback"

x/interchainquery/keeper/msg_server.go
```diff
@@ -59,22 +59,22 @@ func (k msgServer) SubmitQueryResponse(goCtx context.Context, msg *types.MsgSubm
 
 			tmclientstate, ok := clientState.(*tmclienttypes.ClientState)
 			if !ok {
-				k.Logger(ctx).Error("error unmarshaling client state", "cs", clientState)
+				k.Logger(ctx).Error(fmt.Sprintf("error unmarshaling client state %v", clientState))
```
sampocs, line 62: do we know if the keys get printed here? I think I've seen both cases in the logs. This is a pretty big struct so if the keys don't show up, it might make sense to print them as well

x/records/module_ibc.go
```diff
@@ -187,8 +187,7 @@ func (im IBCModule) OnAcknowledgementPacket(
 	if data.Sender == im.keeper.AccountKeeper.GetModuleAddress(stakeibctypes.ModuleName).String() {
 		switch resp := ack.Response.(type) {
 		case *channeltypes.Acknowledgement_Result:
-			log := fmt.Sprintf("\t [IBC-TRANSFER] Acknowledgement_Result {%s}", string(resp.Result))
-			im.keeper.Logger(ctx).Info(log)
+			im.keeper.Logger(ctx).Info(fmt.Sprintf("\t [IBC-TRANSFER] Acknowledgement_Result {%s}", string(resp.Result)))
```
sampocs, line 190: I think this prints gibberish
Screen Shot 2022-08-13 at 5 33 11 PM

x/interchainquery/keeper/msg_server.go
riley-stride, line 62: was there as issue at some point with `%v` prints?

riley-stride (approved):
LGTM. 

One things to check before merging: was there an issue with `%v` values in `Sprtinf` causing errors/panics? I recall some talk of this. I see several %v's in the PR (and generally throughout the codebase)

x/stakeibc/keeper/hooks.go
```diff
@@ -262,8 +261,7 @@ func (k Keeper) TransferExistingDepositsToHostZones(ctx sdk.Context, epochNumber
 	var emptyRecords []uint64
 	for _, depositRecord := range transferDepositRecords {
 		if depositRecord.DepositEpochNumber < cast.ToUint64(epochNumber) {
-			pstr := fmt.Sprintf("\t[TransferExistingDepositsToHostZones] Processing deposits {%d} {%s} {%d}", depositRecord.Id, depositRecord.Denom, depositRecord.Amount)
-			k.Logger(ctx).Info(pstr)
+			k.Logger(ctx).Info(fmt.Sprintf("\t[TransferExistingDepositsToHostZones] Processing deposits {%d} {%s} {%d}", depositRecord.Id, depositRecord.Denom, depositRecord.Amount))
```
sampocs, line 264: nit: The brackets here are a little confusing, might be a little cleaner to do
```
"\t[TransferExistingDepositsToHostZones] Processing deposits ID: %d, Amount: %d%s", depositRecord.Id, depositRecord.Amount, depositRecord.Denom)
```

x/stakeibc/keeper/hooks.go
```diff
@@ -274,12 +272,12 @@ func (k Keeper) TransferExistingDepositsToHostZones(ctx sdk.Context, epochNumber
 
 			hostZone, hostZoneFound := k.GetHostZone(ctx, depositRecord.HostZoneId)
 			if !hostZoneFound {
-				k.Logger(ctx).Error("[TransferExistingDepositsToHostZones] Host zone not found for deposit record {%d}", depositRecord.Id)
+				k.Logger(ctx).Error(fmt.Sprintf("[TransferExistingDepositsToHostZones] Host zone not found for deposit record {%d}", depositRecord.Id))
```
sampocs, line 275: nit: Add "ID" to log

sampocs (commented):
lgtm!

x/interchainquery/keeper/msg_server.go
sampocs: It's also not clear to me whether %v's prints the key's or not. I think I've seen both cases. Do you know @riley-stride ?

riley-stride: Not 100% sure, we should prob just be safe by not using %v [](edited-with-Axolo)

asalzmann: %v uses the default print function for the type (e.g. %s for strings) and won't crash. the Error logger from the sdk crashed using %d (unclear why). 

to @sampocs point, %v won't print keys (just values), but I think that's fine (we can fix it in a future PR)

asalzmann:
%v within Sprintf should be safe!

x/interchainquery/keeper/msg_server.go
sampocs: That's what I thought too, but then I saw this in the logs
```
log = fmt.Sprintf("\t [IBC-TRANSFER] Deposit record updated: {%v}", record)
```
Screen Shot 2022-08-13 at 8 59 10 PM

riley-stride: cant see the screenshot [](edited-with-Axolo)

sampocs: check github [](edited-with-Axolo)

riley-stride: :+1: [](edited-with-Axolo)

asalzmann: oh, nice! so it does print keys

x/icacallbacks/keeper/keeper.go
asalzmann: done

x/records/module_ibc.go
asalzmann: it just prints the bytes from the ack (possibly useful for debugging)

x/stakeibc/keeper/hooks.go
asalzmann: nice! done

x/interchainquery/keeper/msg_server.go
asalzmann: this data type comes from a pb.go file, so I think it will print the keys!

asalzmann:
> One things to check before merging: was there an issue with %v values in Sprtinf causing errors/panics? I recall some talk of this. I see several %v's in the PR (and generally throughout the codebase)

for some reason %d was causing an error when used in combination with `sdk.Error(...)`, but %v should never throw an error for `fmt.Sprintf`!

PR #182, opened 2022-08-11
add reinvestment callback

asalzmann:

After rewards are withdrawn to the withdrawal account, they are queried and transferred to the delegation account for staking. The way this works is that after a `MsgSend` from the withdrawal account to the delegation account happens, we create a `DepositRecord` for the tokens.

This PR migrates that logic to use callbacks.

## Brief Changelog

- Add `ReinvestCallback` arguments
- Store callback args, call `SubmitTxsStrideEpoch` with those args
- Migrate callback logic from `ibc_handlers.go` to `icacallbacks.go`
- Register "reinvest" callback

## Testing and Verifying

integration tests
```
integration_tests.bats
 ✓ [INTEGRATION-BASIC] ibc transfer updates all balances
 ✓ [INTEGRATION-BASIC] host zone successfully registered
 ✓ [INTEGRATION-BASIC] liquid stake mints stATOM
 ✓ [INTEGRATION-BASIC] tokens were transferred to GAIA after liquid staking
 ✓ [INTEGRATION-BASIC] tokens on GAIA were staked
 ✓ [INTEGRATION-BASIC] redemption works
 ✓ [INTEGRATION-BASIC] claimed tokens are properly distributed
 ✓ [INTEGRATION-BASIC] rewards are being reinvested (delegated balance increasing)
 ✓ [INTEGRATION-BASIC] exchange rate is updating

11 tests, 0 failures
```

  - How is the feature or change documented? na


x/stakeibc/keeper/icacallbacks.go
```diff
@@ -116,3 +120,60 @@ func DelegateCallback(k Keeper, ctx sdk.Context, packet channeltypes.Packet, ack
 	k.RecordsKeeper.RemoveDepositRecord(ctx, cast.ToUint64(recordId))
 	return nil
 }
+
+// ----------------------------------- Reinvest Callback ----------------------------------- //
+func (k Keeper) MarshalReinvestCallbackArgs(ctx sdk.Context, delegateCallback types.ReinvestCallback) ([]byte, error) {
```
sampocs, line 125: var name should be `reinvestCallback`

x/stakeibc/keeper/icacallbacks.go
```diff
@@ -116,3 +120,60 @@ func DelegateCallback(k Keeper, ctx sdk.Context, packet channeltypes.Packet, ack
...
+	out, err := proto.Marshal(&delegateCallback)
+	if err != nil {
+		k.Logger(ctx).Error(fmt.Sprintf("MarshalReinvestCallbackArgs %v", err.Error()))
+		return nil, err
+	}
+	return out, nil
+}
+
+func (k Keeper) UnmarshalReinvestCallbackArgs(ctx sdk.Context, delegateCallback []byte) (*types.ReinvestCallback, error) {
```
sampocs, line 134: var name should be reinvestCallback

x/stakeibc/keeper/icacallbacks.go
```diff
@@ -116,3 +120,60 @@ func DelegateCallback(k Keeper, ctx sdk.Context, packet channeltypes.Packet, ack
...
+	unmarshalledReinvestCallback := types.ReinvestCallback{}
+	if err := proto.Unmarshal(delegateCallback, &unmarshalledReinvestCallback); err != nil {
+        k.Logger(ctx).Error(fmt.Sprintf("UnmarshalReinvestCallbackArgs %v", err.Error()))
+		return nil, err
+	}
+	return &unmarshalledReinvestCallback, nil
+}
+
+func ReinvestCallback(k Keeper, ctx sdk.Context, packet channeltypes.Packet, ack *channeltypes.Acknowledgement_Result, args []byte) error {
+	// invariant: sendMsg.FromAddress == withdrawalAddress && sendMsg.ToAddress == delegationAddress
```
sampocs, line 144: is this something we should confirm here?

x/stakeibc/keeper/icacallbacks.go
```diff
@@ -116,3 +120,60 @@ func DelegateCallback(k Keeper, ctx sdk.Context, packet channeltypes.Packet, ack
...
+	k.Logger(ctx).Info("ReinvestCallback executing", "packet", packet)
```
sampocs, line 145: will packet be printed legibly here?

x/stakeibc/keeper/icacallbacks.go
```diff
@@ -116,3 +120,60 @@ func DelegateCallback(k Keeper, ctx sdk.Context, packet channeltypes.Packet, ack
...
+	strideEpochTracker, found := k.GetEpochTracker(ctx, epochtypes.STRIDE_EPOCH)
+	if !found {
+		k.Logger(ctx).Error("failed to find epoch")
+		return sdkerrors.Wrapf(types.ErrInvalidLengthEpochTracker, "no number for epoch (%s)", epochtypes.STRIDE_EPOCH)
+	}
+	epochNumber := strideEpochTracker.EpochNumber
+	// create a new record so that rewards are reinvested
+	record := recordstypes.DepositRecord{
+		Id:                 0,
+		Amount:             cast.ToInt64(reinvestCallback.GetAmount()),
```
sampocs, line 169: ToInt64E?

x/stakeibc/keeper/ibc_handlers.go
```diff
@@ -159,39 +158,16 @@ func (k *Keeper) HandleSend(ctx sdk.Context, msg sdk.Msg, sequence string) error
...
 	if err != nil {
 		return err
 	}
 
 	// Check to and from addresses
-	withdrawalAddress := zone.GetWithdrawalAccount().GetAddress()
 	delegationAddress := zone.GetDelegationAccount().GetAddress()
 	redemptionAddress := zone.GetRedemptionAccount().GetAddress()
 
-	// Process bank sends that reinvest user rewards
```
sampocs, line 173: I love watching `ibc_handlers` slowly evaporate

x/stakeibc/keeper/callbacks.go
```diff
@@ -9,28 +9,28 @@ import (
...
-	"github.com/Stride-Labs/stride/x/interchainquery/types"
 	icqtypes "github.com/Stride-Labs/stride/x/interchainquery/types"
+	"github.com/Stride-Labs/stride/x/stakeibc/types"
 )
 
 // ___________________________________________________________________________________________________
 
 // Callbacks wrapper struct for interchainstaking keeper
-type Callback func(Keeper, sdk.Context, []byte, types.Query) error
+type Callback func(Keeper, sdk.Context, []byte, icqtypes.Query) error
```
sampocs, line 19: wdyt about renaming this file to `icqcallbacks` and renaming this struct accordingly?

x/stakeibc/keeper/icacallbacks.go
```diff
@@ -46,7 +48,9 @@ func (c ICACallbacks) AddICACallback(id string, fn interface{}) icacallbackstype
 }
 
 func (c ICACallbacks) RegisterICACallbacks() icacallbackstypes.ICACallbackHandler {
-	a := c.AddICACallback("delegate", ICACallback(DelegateCallback))
+	a := c.
+		AddICACallback("delegate", ICACallback(DelegateCallback)).
+		AddICACallback("reinvest", ICACallback(ReinvestCallback))
 	return a.(ICACallbacks)
```
sampocs, line 54:
```suggestion
	return c.
		AddICACallback("delegate", ICACallback(DelegateCallback)).
		AddICACallback("reinvest", ICACallback(ReinvestCallback))
```

x/stakeibc/keeper/icacallbacks.go
```diff
@@ -46,7 +48,9 @@ func (c ICACallbacks) AddICACallback(id string, fn interface{}) icacallbackstype
 }
 
 func (c ICACallbacks) RegisterICACallbacks() icacallbackstypes.ICACallbackHandler {
-	a := c.AddICACallback("delegate", ICACallback(DelegateCallback))
+	a := c.
+		AddICACallback("delegate", ICACallback(DelegateCallback)).
```
sampocs, line 52: should we store the callback key's as constants at the top of this file so we can access them when we call SubmitTx?

x/stakeibc/keeper/callbacks.go
```diff
@@ -39,12 +39,12 @@ func (c Callbacks) Has(id string) bool {
...
 }
 
-func (c Callbacks) AddCallback(id string, fn interface{}) types.QueryCallbacks {
+func (c Callbacks) AddCallback(id string, fn interface{}) icqtypes.QueryCallbacks {
 	c.callbacks[id] = fn.(Callback)
 	return c
 }
 
-func (c Callbacks) RegisterCallbacks() types.QueryCallbacks {
+func (c Callbacks) RegisterCallbacks() icqtypes.QueryCallbacks {
```
sampocs, line 47: Just to make sure I'm following the flow, it goes:

1. Query withdrawal balance with ICQ (initiated from hooks) which gives us the amount of outstanding rewards
2. When the query returns, ICQ callback `callbacks.go` is executed 
3. The ICQ callback initiates an ICA transfer of 90% of that rewards amount from the withdrawal account to the delegation account
4. When the ICA returns, the Reinvestment callback is executed, which creates a new deposit record so the rewards can be swept up and staked in the next epoch

Is that right?

x/stakeibc/keeper/icacallbacks.go
```diff
@@ -116,3 +120,60 @@ func DelegateCallback(k Keeper, ctx sdk.Context, packet channeltypes.Packet, ack
...
+	// fetch epoch
+	strideEpochTracker, found := k.GetEpochTracker(ctx, epochtypes.STRIDE_EPOCH)
+	if !found {
+		k.Logger(ctx).Error("failed to find epoch")
+		return sdkerrors.Wrapf(types.ErrInvalidLengthEpochTracker, "no number for epoch (%s)", epochtypes.STRIDE_EPOCH)
+	}
+	epochNumber := strideEpochTracker.EpochNumber
+	// create a new record so that rewards are reinvested
+	record := recordstypes.DepositRecord{
+		Id:                 0,
```
sampocs, line 168: Does the ID field not matter here cause we treat it as a list? If so, should we remove this field?

sampocs (commented):
Callbacks strike again!! Looks good! Couple comments/questions below

proto/stakeibc/callbacks.proto
```diff
@@ -13,3 +13,10 @@ message DelegateCallback {
   uint64 depositRecordId = 2;
   repeated SplitDelegation splitDelegations = 3;
 }
+
+// ---------------------- Reinvest Callback ---------------------- //
+message ReinvestCallback {
```
riley-stride, line 18: isn't it really a `bankSendCallback` that's used as part of the reinvestment cycle? Depending on how generalizable we'd like to make these protos, might make sense to rename

proto/stakeibc/callbacks.proto
```diff
@@ -13,3 +13,10 @@ message DelegateCallback {
...
+  string denom = 1;
```
riley-stride, line 19: nit: should we replace denom and amount with sdk.Coin type?

e.g.
```
  string amount           = 2 [
    (gogoproto.moretags)   = "yaml:\"stake\"",
    (gogoproto.nullable)   = false
  ];
  ```

File: x/stakeibc/keeper/ibc_handlers.go (hunk 6 above)
Comment (riley-stride) [reply x/stakeibc/keeper/ibc_handlers.go:173]: hahaha it's beautiful

File: x/stakeibc/keeper/icacallbacks.go [hunk 14]
```diff
@@ -116,3 +120,60 @@ func DelegateCallback(k Keeper, ctx sdk.Context, packet channeltypes.Packet, ack
...
+		k.Logger(ctx).Error(fmt.Sprintf("MarshalReinvestCallbackArgs %v", err.Error()))
+		return nil, err
+	}
+	return out, nil
+}
+
+func (k Keeper) UnmarshalReinvestCallbackArgs(ctx sdk.Context, delegateCallback []byte) (*types.ReinvestCallback, error) {
+	unmarshalledReinvestCallback := types.ReinvestCallback{}
+	if err := proto.Unmarshal(delegateCallback, &unmarshalledReinvestCallback); err != nil {
+        k.Logger(ctx).Error(fmt.Sprintf("UnmarshalReinvestCallbackArgs %v", err.Error()))
```
riley-stride, line 137: nano nit
```suggestion
        k.Logger(ctx).Error(fmt.Sprintf("UnmarshalReinvestCallbackArgs %s", err.Error()))
```

x/stakeibc/keeper/icacallbacks.go
riley-stride: +1: the [tx you submit with this callback](https://github.com/Stride-Labs/stride/pull/182/files?diff=split&w=0#diff-b6597a6df3c7e86ae53cb25938eab9ea9e920ea686410592068c6deaf4cae116R163) contains two messages : one to the revenue acct and one to the delegation acct. where do we distinguish between one and the other in the CB?

x/stakeibc/keeper/icacallbacks.go
```diff
@@ -116,3 +120,60 @@ func DelegateCallback(k Keeper, ctx sdk.Context, packet channeltypes.Packet, ack
...
+}
+
+func ReinvestCallback(k Keeper, ctx sdk.Context, packet channeltypes.Packet, ack *channeltypes.Acknowledgement_Result, args []byte) error {
+	// invariant: sendMsg.FromAddress == withdrawalAddress && sendMsg.ToAddress == delegationAddress
+	k.Logger(ctx).Info("ReinvestCallback executing", "packet", packet)
+
+	if ack == nil {
+		// transaction on the host chain failed
+		// don't create a record
+		return sdkerrors.Wrap(sdkerrors.ErrInvalidRequest, "ack is nil")
```
riley-stride, line 150: Is there any error information in the packet that we could log here?

x/stakeibc/keeper/icacallbacks.go
riley-stride: Afaiu it gets overwritten by the latest ID when we `append` the record, so we could set it to anything here. If we can omit, we should

riley-stride (approved):
Callbacks are so cool. 

One high level Q [here](https://github.com/Stride-Labs/stride/pull/182/files?diff=split&w=0#diff-581a35bf2c17a8ec60baf5ca2cde154543e34219feeb3a6a2d0744da004b8f55R144)

One suggestion -- should we rename the reinvestCB proto type to bankSendCB so that it can be flexibly generalized later? I think it makes sense to keep specificity with the CB functions, but maybe the protos should generalize?

x/stakeibc/keeper/icacallbacks.go
sampocs: ah yes, good point. Maybe makes sense to just not include an Id here (and let it default to nil) to avoid confusion

x/stakeibc/keeper/icacallbacks.go
asalzmann: done

x/stakeibc/keeper/icacallbacks.go
asalzmann: The callback executes at the transaction level, so you don't need to distinguish between messages (both succeed or fail atomically as part of the transaction)

We have so much sensitive logic in callbacks that I'm not sure how much it would add to check the invariant

x/stakeibc/keeper/icacallbacks.go
asalzmann: it should be
packet struct
```
type Packet struct {
	// number corresponds to the order of sends and receives, where a Packet
	// with an earlier sequence number must be sent and received before a Packet
	// with a later sequence number.
	Sequence uint64 `protobuf:"varint,1,opt,name=sequence,proto3" json:"sequence,omitempty"`
	// identifies the port on the sending chain.
	SourcePort string `protobuf:"bytes,2,opt,name=source_port,json=sourcePort,proto3" json:"source_port,omitempty" yaml:"source_port"`
	// identifies the channel end on the sending chain.
	SourceChannel string `protobuf:"bytes,3,opt,name=source_channel,json=sourceChannel,proto3" json:"source_channel,omitempty" yaml:"source_channel"`
	// identifies the port on the receiving chain.
	DestinationPort string `protobuf:"bytes,4,opt,name=destination_port,json=destinationPort,proto3" json:"destination_port,omitempty" yaml:"destination_port"`
	// identifies the channel end on the receiving chain.
	DestinationChannel string `protobuf:"bytes,5,opt,name=destination_channel,json=destinationChannel,proto3" json:"destination_channel,omitempty" yaml:"destination_channel"`
	// actual opaque bytes transferred directly to the application module
	Data []byte `protobuf:"bytes,6,opt,name=data,proto3" json:"data,omitempty"`
	// block height after which the packet times out
	TimeoutHeight types.Height `protobuf:"bytes,7,opt,name=timeout_height,json=timeoutHeight,proto3" json:"timeout_height" yaml:"timeout_height"`
	// block timestamp (in nanoseconds) after which the packet times out
	TimeoutTimestamp uint64 `protobuf:"varint,8,opt,name=timeout_timestamp,json=timeoutTimestamp,proto3" json:"timeout_timestamp,omitempty" yaml:"timeout_timestamp"`
}
```

x/stakeibc/keeper/icacallbacks.go
asalzmann: done

x/stakeibc/keeper/icacallbacks.go
asalzmann: done

x/stakeibc/keeper/callbacks.go
asalzmann: that's right!

x/stakeibc/keeper/icacallbacks.go
asalzmann: done

x/stakeibc/keeper/icacallbacks.go
asalzmann: logging the packet can't hurt! done.

x/stakeibc/keeper/icacallbacks.go
asalzmann: done

proto/stakeibc/callbacks.proto
asalzmann: I don't think we should generalize them - callbacks are currently designed to be at the level of abstraction of a single transaction "type", where the "type" of the transaction has logic that's specific to the application. Bank sends aren't specific to the application logic, so those callbacks could be handled in `OnAcknowledgementPacket`. The protos are just a convenient way to store function arguments for the different "types" of callbacks (e.g. you'd write a callback for a specific bank send that's relevant to some logic in a module, rather than a callback that's a catch-all for all bank sends)

x/stakeibc/keeper/icacallbacks.go
asalzmann: done

proto/stakeibc/callbacks.proto
asalzmann: done

x/stakeibc/keeper/callbacks.go
asalzmann: it's a good idea - going to leave it out of this PR in the interest of launching tomorrow (but added a note to do this EOD tomorrow)

asalzmann:
> One suggestion -- should we rename the reinvestCB proto type to bankSendCB so that it can be flexibly generalized later? I think it makes sense to keep specificity with the CB functions, but maybe the protos should generalize?

commented above, but the protos are just a convenient way to serialize / deserialize the arguments to the callbacks, so I think they should be defined at the same level of specificity as the callbacks themselves

proto/stakeibc/callbacks.proto
riley-stride: makes sense, ty for the writeup [](edited-with-Axolo)

PR #184, opened 2022-08-12
Update pull_request_template.md

asalzmann:

Updates the pull request template to cover proto changes (and the downstream effects).

## Brief Changelog

Add two checklist items to pull_request_template.md

  - How is the feature or change documented? no

### DIFF
--- a/pull_request_template.md
+++ b/pull_request_template.md
@@ -48,4 +48,6 @@ This change added tests and can be verified as follows:
   - Does this pull request introduce a new feature or user-facing behavior changes? (yes / no)
   - Is a relevant changelog entry added to the `Unreleased` section in `CHANGELOG.md`? (yes / no)
   - How is the feature or change documented? (not applicable   /   specification (`x/<module>/spec/`)  / README.md  /   not documented)
-  - Does this pull request change protos and require a migration? (yes / no)
+  - Does this pull request update existing proto field values and require a migration? (yes / no)
+  - Does this pull request change existing proto field names and require frontend changes? (yes / no)
+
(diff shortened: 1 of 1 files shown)


pull_request_template.md
```diff
@@ -48,3 +48,6 @@ This change added tests and can be verified as follows:
   - Does this pull request introduce a new feature or user-facing behavior changes? (yes / no)
   - Is a relevant changelog entry added to the `Unreleased` section in `CHANGELOG.md`? (yes / no)
   - How is the feature or change documented? (not applicable   /   specification (`x/<module>/spec/`)  / README.md  /   not documented)
+  - Does this pull request update existing proto field values and require a migration? (yes / no)
```
sampocs, line 51: nit: but if the field names change, that could also require a migration. Maybe just generalize it to "Does this pull request update existing proto data types and require a migration?"

pull_request_template.md
```diff
@@ -48,3 +48,6 @@ This change added tests and can be verified as follows:
...
+  - Does this pull request change existing proto field names and require frontend changes? (yes / no)
```
sampocs, line 52: If the data types change, that would also impact the front end right?

pull_request_template.md
asalzmann: why would a field name change require a migration?

pull_request_template.md
asalzmann: Good point! Yeah that would impact the frontend as well

pull_request_template.md
sampocs: ah yeah I guess if just the name changed, it wouldn't matter. But if you added a field, it could require a migration (if you have to backfill)

sampocs: I think how you have it is fine! I was interpreting value as the stored value in state - not the "proto value" (i.e. the index)

sampocs (approved):

sampocs (approved):

PR #194, opened 2022-08-16
Fixed Acknowledgement Error Catching

sampocs:
* The ICACallbacks acknowledgment can be returned as either an `Acknowledgement_Result` type or `Acknowledgement_Error` type. If it is returned as an error type, it is still possible to unmarshal as a result type, yet the fields will be empty. This was causing ack error's to appear as successes.
* The fix, is to switch on the datatype to correctly differentiate between success and error acks


asalzmann (approved):
Thank you for tracking this down! This was a really tricky issue that we could have easily missed

PR #196, opened 2022-08-16
Dockernet (v2)

sampocs:
Closes: #XXX

* Dockernet is back! The consensus error and newly added host zones warranted another look at our docker setup, as it makes it easier to run multiple nodes and validators, and aligns more closely with our production setup.
* After some changes, the startup time with docker should only be ~10 sec longer than the local mode startup. And with some additional changes to the setup scripts, this dockernet build is now **2 minutes faster** than the previous localnet build (which was ~5min)
  * Startup time includes time to build + start + register host zone and validators
  * Comparison was between running dockernet with 3 stride and 3 gaia nodes, vs running localnet with 1 stride and 2 gaia

## Brief Changelog
- Modified dockerfiles to make better use of caching
- Created a generalized/parameterized chain init script that should make it easier to add new host zones (notice `init_stride` and `init_gaia` are almost identical). I’ll actually separate this out as separate script in a different PR to prevent this one from getting too bloated.
- You can now toggle the number of nodes with a single parameter in `vars.sh` (`STRIDE_NUM_NODES`/`GAIA_NUM_NODES`). This is currently capped at 5 but very easy to add more if needed.
- The new docker setup lets you still issue commands locally (i.e. with the local binaries), but it will point to docker behind the scenes. All you have to do is specify the new home directory. I'd recommend setting an alias.
```
build/strided --home scripts/state/stride1
build/gaiad --home scripts/state/gaia1
```
- Laid groundwork for switch to go v2 relayer (to be done finished in another PR)
- **Changed makefiles commands**

## Next Steps (in separate PR)
- Split out `init_gaia` into a more generalized `init_chain` script and add Osmosis and Juno
- Add go relayer
- Reorganize project structure (lower priority - will do after launch)

## Instructions
- Start
```
make start-docker build=sghi 
# and then `build=s` for successive builds
```
- Stop
```
make stop-docker
```

## Testing and Verifying
- Once all the outstanding callback PRs have been merged in, I'll merge them into here and rerun these
- [ ] Integration tests
- [ ] Firehose

  - Does this pull request introduce a new feature or user-facing behavior changes? (yes / no) no
  - Is a relevant changelog entry added to the `Unreleased` section in `CHANGELOG.md`? (yes / no) no
  - How is the feature or change documented? (not applicable   /   specification (`x/<module>/spec/`)  / README.md  /   not documented) N/A
  - Does this pull request update existing proto field values (and require a backend and frontend migration)? (yes / no) no
  - Does this pull request change existing proto field names (and require a frontend migration)? (yes / no) no


Dockerfile.icq
```diff
@@ -1,24 +1,21 @@
...
-COPY . /src/app/stride_ref/stride
-WORKDIR /src/app/stride_ref/stride
-WORKDIR /src/app
-RUN go mod download
-RUN go build
 
-RUN ln -s /src/app/interchain-queries /usr/local/bin
-RUN adduser --system --home /icq --disabled-password --disabled-login icq -U 1000
-USER icq
+RUN git clone https://github.com/Stride-Labs/interchain-queries.git \
```
riley-stride, line 7: Should we add a hash or tag here?

Dockerfile.icq
```diff
@@ -1,24 +1,21 @@
...
+FROM alpine:3.15
 
+COPY --from=icq-builder /src/app/interchain-queries /usr/local/bin/
+
+RUN adduser -S -h /home/icq -D icq -u 1000
+USER icq
 
+WORKDIR /home/icq
 
+CMD ["interchain-queries", "run"]
```
riley-stride, line 21: How are hot restarts going to be handled here? Local mode has a while loop, cates has auto restart (afaiu). Is there a similar mechanism here, or could we add one? Not super important, let's only do it if it's a <10 min task

Makefile
```diff
@@ -54,14 +54,22 @@ test-integration-docker:
 ###                                DockerNet                                ###
 ###############################################################################
 
-init-docker:
-	sh scripts/init_main.sh -${build}
+build-docker: 
+	@sh scripts/build.sh -${build} ${BUILDDIR}
+	
+start-docker: build-docker
```
riley-stride, line 60: Should start-docker run stop-docker as its first command? Seems like `start_network.sh` runs `docker-compose down` under the hood, so maybe already handled. Flagging just in case!

scripts/build.sh
```diff
@@ -0,0 +1,55 @@
...
+
+BUILDDIR="$2"
+
+# build docker images and local binaries
+while getopts sghir flag; do
+    case "${flag}" in
+        s) echo "Building Stride Docker...  ";
+           docker build --tag stridezone:stride -f Dockerfile.stride . ;
+
+           printf '%s' "Building Stride Locally...  ";
```
riley-stride, line 15: Why do we build both on Docker _and_ Locally? Because local gets copied over to docker?

scripts/init_relayers.sh
```diff
@@ -0,0 +1,27 @@
...
+
+cp ${SCRIPT_DIR}/config/icq_config.yaml $STATE/icq/config.yaml
+cp ${SCRIPT_DIR}/config/hermes_config.toml $STATE/hermes/config.toml
+cp ${SCRIPT_DIR}/config/relayer_config.yaml $STATE/relayer/config/config.yaml
+
+echo "Adding Hermes keys"
+$HERMES_CMD keys restore --name rly1 --mnemonic "$HERMES_STRIDE_MNEMONIC" $STRIDE_CHAIN_ID 
+$HERMES_CMD keys restore --name rly2 --mnemonic "$HERMES_GAIA_MNEMONIC" $GAIA_CHAIN_ID
+
+# echo "Adding Relayer keys"
```
riley-stride, line 20: Not using ibcgo relayer?

scripts/vars.sh
```diff
@@ -3,79 +3,97 @@
...
+DAY_EPOCH_INDEX=1
+DAY_EPOCH_DURATION="100s"
+STRIDE_EPOCH_INDEX=2
+STRIDE_EPOCH_DURATION="40s"
+UNBONDING_TIME="200s"
+MAX_DEPOSIT_PERIOD="30s"
+VOTING_PERIOD="30s"
+
+VAL_TOKENS=5000000000000
+STAKE_TOKENS=3000000000000
```
riley-stride, line 26: nice, this will keep the RR increases in check. Might even suggest raising it to ~90% of VAL_TOKENS

scripts/vars.sh
```diff
@@ -3,79 +3,97 @@
...
+IBC_STRD_DENOM='ibc/FF6C2E86490C1C4FBBD24F55032831D2415B9D7882F85C3CC9C2401D79362BEA'
+IBC_ATOM_DENOM='ibc/27394FB092D2ECCD56123C74F36E4C1F926001CEADA9CA97EA622B25F41E5EB2'
+ATOM_DENOM='uatom'
+STRD_DENOM='ustrd'
+STATOM_DENOM="stuatom"
 
-STRIDE_CHAIN=STRIDE
-STRIDE_NODE_NAMES=(stride1 stride2 stride3)
-STRIDE_MAIN_NODE=${STRIDE_NODE_NAMES[$MAIN_ID]}
+BLOCK_TIME='5s'
```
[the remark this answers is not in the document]
riley-stride: Addressed offline

riley-stride (approved):
Really good work, was able to build with no problems and interact with stride and gaia using the aliases provided. Added some questions, but they're not merge-blocking!

scripts/init_relayers.sh
riley-stride: Nvm, I see your comments about "laying the groundwork"

Dockerfile.icq
sampocs: Yeah probably a good call! Do you think ICQ is in a pretty stable state? For awhile it seemed like there were a lot of changes to main which is why I just kept this pointed there. I think we've since added a hash to the gitmodule though so maybe I'll just use that

Dockerfile.icq
sampocs: This bad boy! https://github.com/Stride-Labs/stride/blob/docker-net-v2/docker-compose.yml#L75

Makefile
sampocs: Good call!

scripts/build.sh
sampocs: Docker just provides the runtime environment, but all commands are executed locally through local binaries. And then the state is bridged with docker volumes which is a way to share the file system between a local machine and the docker container.
So for instance, when running locally, we point `--home` to `scripts/state/stride1`, and when inside the docker container, we point `--home` to `~/.stride` - but those two folders are actually mirror images of each other

scripts/vars.sh
sampocs: done

Dockerfile.icq
sampocs: Actually just noticed our submodule is pointed to main as well (as is our testnet)

PR #197, opened 2022-08-16
add fees

asalzmann:

Add a mechanism to collect fees. Rather than hardcoding a "cosmos..." address, we use a single "stride..." address, which receives fees from host zones. Fees are collected on the `FeeAccount` on host zones, then IBC transferred back to the "stride..." account (triggered by a message).

## Brief Changelog

- Remove "cosmos..." rev account
- `MsgSend` fees to the fee ICA
- Expose a messages `ClearBalance`, which sends fees back 

## Testing and Verifying

Liquid stake twice, in order to trigger a withdrawal of rewards to the withdrawal account.
```
balances:
- amount: "2967"
  denom: uatom
pagination:
  next_key: null
  total: "0"
```
run
```
$STRIDE_CMD tx stakeibc clear-balance GAIA 66 channel-0 --keyring-backend test --from val1 --chain-id $STRIDE_CHAIN
```
check balances again (on gaia)
```
balances:
- amount: "2901"
  denom: uatom
pagination:
  next_key: null
  total: "0"
```
verify the stride rev account balance increased
```
balances:
- amount: "66"
  denom: ibc/27394FB092D2ECCD56123C74F36E4C1F926001CEADA9CA97EA622B25F41E5EB2
pagination:
  next_key: null
  total: "0"
```

I also tried passing in the wrong channel
```
$STRIDE_CMD tx stakeibc clear-balance GAIA 66 channel-3 --keyring-backend test --from val1 --chain-id $STRIDE_CHAIN
```
which failed
```
balances:
- amount: "2967"
  denom: uatom
pagination:
  next_key: null
  total: "0"
```
and did _not_ take down the ICA

## Open questions
- where should we store the rev account?
- should we parameterize the revenue account?
- should the channel be an argument to the function?

  - Does this pull request introduce a new feature or user-facing behavior changes? yes
  - How is the feature or change documented? not documented
  - Does this pull request update existing proto field values (and require a backend and frontend migration)? no
  - Does this pull request change existing proto field names (and require a frontend migration)? no


x/stakeibc/keeper/msg_server_clear_balance.go
```diff
@@ -0,0 +1,62 @@
...
+	sourceChannel := msg.Channel
+	coinString := cast.ToString(msg.Amount) + zone.GetHostDenom()
+	tokens, err := sdk.ParseCoinNormalized(coinString)
+	if err != nil {
+		k.Logger(ctx).Error(fmt.Sprintf("failed to parse coin (%s)", coinString))
+		return nil, sdkerrors.Wrapf(err, "failed to parse coin (%s)", coinString)
+	}
+	sender := feeAccount.GetAddress()
+	// where to store this?
+	receiver := "stride19uvw0azm9u0k6vqe4e22cga6kteskdqq3ulj6q"
```
vish-stride, line 40: which address is this? we should make a note or reminder to update this prior to launch

x/stakeibc/keeper/msg_server_clear_balance.go
```diff
@@ -0,0 +1,62 @@
...
+	zone, found := k.GetHostZone(ctx, msg.ChainId)
+	if !found {
+		return nil, sdkerrors.Wrapf(types.ErrInvalidHostZone, "chainId: %s", msg.ChainId)
+	}
+	feeAccount := zone.GetFeeAccount()
+	if feeAccount == nil {
+		return nil, sdkerrors.Wrapf(types.ErrFeeAccountNotRegistered, "chainId: %s", msg.ChainId)
+	}
+
+	// should this be a param? the transfer port _should_ always be the same across zones
```
vish-stride, line 27: agreed, probably don't need to make this a variable

x/stakeibc/types/message_clear_balance.go
```diff
@@ -0,0 +1,53 @@
...
+	bz := ModuleCdc.MustMarshalJSON(msg)
+	return sdk.MustSortJSON(bz)
+}
+
+func (msg *MsgClearBalance) ValidateBasic() error {
+	_, err := sdk.AccAddressFromBech32(msg.Creator)
+	if err != nil {
+		return sdkerrors.Wrapf(sdkerrors.ErrInvalidAddress, "invalid creator address (%s)", err)
+	}
+	// TODO: add validation
```
vish-stride, line 47: I know it doesn't really matter (not sure what the attack vector would be for this function), but feels _slightly_ safer to restrict this as an admin message, what do you think? 

The specific thing I'm paranoid about is if someone can time the message to cause an ICA timeout (or something else funny) that would cause the transaction to fail

The downside is it's fairly cumbersome to make function calls from the admin address (cold storage), and we might want to frequently call this function

Maybe we should make a "ValidateTrustedAddress" checker that has a slightly broader scope of people who can call it? And that has access to this function (maybe something like rebalance validators too?)

vish-stride (approved):
looks great to me! 

some very minor points. I think it's probably slightly better to restrict ClearBalance, just because there's a very small chance the ICA callbacks might have some potential vulnerability (to be clear, the expected and likely case is that it does not)

scripts-local/test-utils/run_any_cli_command.sh
```diff
@@ -4,20 +4,28 @@ SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
 source ${SCRIPT_DIR}/../account_vars.sh
 
 
-# $GAIA_CMD tx ibc-transfer transfer transfer channel-0 stride1uk4ze0x4nvh4fk0xm4jdud58eqn4yxhrt52vv7 100000uatom --from gval1 --chain-id GAIA -y --keyring-backend test
```
riley-stride, line 7: should we rm this file? seems like scratch

x/stakeibc/types/message_clear_balance.go
asalzmann: Per offline discussion, risk/reward of exposing this is not in our favor. Going to gate it for now

x/stakeibc/types/message_clear_balance.go
```diff
@@ -0,0 +1,53 @@
...
+	// Should we let anyone call this?
```
riley-stride: Add a few checks
1. channel contains the string "channel" 
    `!strings.HasPrefix(msg.ChannelId, "channel")`
3. amount GT 0

x/stakeibc/types/message_clear_balance.go
```diff
@@ -0,0 +1,53 @@
...
+	// if err := utils.ValidateAdminAddress(msg.Creator); err != nil {
```
riley-stride: nit rm comment

x/stakeibc/types/params.go
```diff
@@ -22,6 +22,8 @@ var (
 	DefaultICATimeoutNanos  uint64 = 600000000000
 	DefaultBufferSize       uint64 = 5   // 1/5=20% of the epoch
 	DefaultIbcTimeoutBlocks uint64 = 300 // 300 blocks ~= 30 minutes
+	DefaultFeeTransferTimeoutNanos  uint64 = 600000000000 // 10 minutes
```
riley-stride, line 25: What scenario are you envisioning might prompt us to change this through governance?

x/stakeibc/keeper/msg_server_clear_balance.go
riley-stride, line 40: maybe a const in `types`

x/stakeibc/keeper/msg_server_clear_balance.go
```diff
@@ -0,0 +1,62 @@
...
+			TimeoutTimestamp: timeoutTimestamp,
+		},
+	}
+
+	connectionId := zone.GetConnectionId()
+	
+	icaTimeoutNanos := k.GetParam(ctx, types.KeyICATimeoutNanos)
+	icaTimeoutNanos = cast.ToUint64(ctx.BlockTime().UnixNano()) + icaTimeoutNanos
+
+	k.SubmitTxs(ctx, connectionId, msgs, *feeAccount, icaTimeoutNanos, "", nil)
```
riley-stride, line 59: Can you explain how the timeouts work here? Seems like we're using both `KeyFeeTransferTimeoutNanos` and `KeyICATimeoutNanos`

x/stakeibc/keeper/msg_server_clear_balance.go
```diff
@@ -0,0 +1,62 @@
...
+	if !found {
+		return nil, sdkerrors.Wrapf(types.ErrInvalidHostZone, "chainId: %s", msg.ChainId)
+	}
+	feeAccount := zone.GetFeeAccount()
+	if feeAccount == nil {
+		return nil, sdkerrors.Wrapf(types.ErrFeeAccountNotRegistered, "chainId: %s", msg.ChainId)
+	}
+
+	// should this be a param? the transfer port _should_ always be the same across zones
+	sourcePort := "transfer"
```
riley-stride, line 28: can we get this dynamically? thinking about sdk upgrades

asalzmann:
Added a few more checks to `ValidateBasic`

riley-stride (changes requested):
Left a few small comments , overall looks good!

Main Q is how are timeouts working ? Seems like you have two params in play for ica timeouts on clearbalance

scripts-local/test-utils/run_any_cli_command.sh
asalzmann: it's useful imo and low overhead to keeping it

x/stakeibc/types/message_clear_balance.go
asalzmann: added these in the most recent commit

x/stakeibc/types/message_clear_balance.go
asalzmann: done

x/stakeibc/types/params.go
asalzmann: just gives us a bit more flexibility, it's possible we find calls are timing out frequently if we only allow for a 10 min buffer (can't think of a reason not to parameterize, although lmk if you're thinking of one)

x/stakeibc/keeper/msg_server_clear_balance.go
asalzmann: `KeyICATimeoutNanos` are for our Stride ICA calls, `KeyFeeTransferTimeoutNanos` is for the IBC transfer

x/stakeibc/keeper/msg_server_clear_balance.go
```diff
@@ -0,0 +1,62 @@
...
+			Sender: sender,
+			Receiver: receiver,
+			TimeoutTimestamp: timeoutTimestamp,
+		},
+	}
+
+	connectionId := zone.GetConnectionId()
+	
+	icaTimeoutNanos := k.GetParam(ctx, types.KeyICATimeoutNanos)
+	icaTimeoutNanos = cast.ToUint64(ctx.BlockTime().UnixNano()) + icaTimeoutNanos
```
sampocs, line 57: do we need safe case here or does it not matter since the param is fixed?

x/stakeibc/keeper/msg_server_clear_balance.go
asalzmann: good call! done

x/stakeibc/keeper/msg_server_clear_balance.go
asalzmann: added a comment

x/stakeibc/keeper/msg_server_clear_balance.go
sampocs, line 27: fwiw, I _think_ the "transfer" port is defined in the default genesis on new chains, so I think it should always be the same
```
    "transfer": {
      "port_id": "transfer",
      "denom_traces": [],
      "params": {
        "send_enabled": true,
        "receive_enabled": true
      }
    },
```

x/stakeibc/keeper/msg_server_clear_balance.go
riley-stride: makes sense!

x/stakeibc/keeper/msg_server_clear_balance.go
asalzmann: good call, done!

x/stakeibc/keeper/msg_server_clear_balance.go
asalzmann: I checked the ICS-20 spec and couldn't find anything about port names, but in ibc-go it should always be transfer. probably safe for now! https://github.com/cosmos/ibc/blob/main/spec/app/ics-020-fungible-token-transfer/README.md

x/stakeibc/keeper/msg_server_clear_balance.go
asalzmann: the param is a uint, and I figure there are probably checks that `BlockTime` is positive deeper in the sdk

riley-stride (approved):

PR #204, opened 2022-08-20
Icq ttl

riley-stride:

Our ICQ queries currently don't time out at all! That's bad, because it means that if they get hung up in transit, they could come back much later, perhaps overlapping with ICAs! When ICAs and ICQs overlap, we run the risk of race conditions, as ICQs could read a balance, then ICAs could update that balance before the ICQ response is received back on Stride, making the ICQ response stale!

This PR adds timeouts for ICQs, using the `ttl` field on the `SetQuery` method on Stride. How does it work? When we kick off a query with `MakeRequest` on Stride, we set a time we'd like the response to expire at in a `ttl` field. When the query response comes back from the host zone, Stride retrieves this `ttl` and aborts the query is the current time is later than the ttl. If the query is expired, Stride also deletes the query metadata from Stride's interchainquery module store.

The `ttl` has `UnixNano` units. 

Practically, we always set the `ttl` to the time at the end of the next epoch, because that's the time at which we want queries to expire. 

Which queries does this change impact?
- **Reinvestment Query**: We set the timeout to the same internal as our ICAs. Empirically these queries come back in ~2-3 seconds using the current icq relayer
- **Slashing Query**: here, the ttl can make a big difference because these queries are submitted ad-hoc throughout the epoch -- they could be submitted even 1 second before the epoch end. In this case, the ttl is a safety mechanism that aborts such last-minute queries.

## Brief Changelog

- all icq's are submitted from `msg_server_submit_tx.go`, so I added logic there to grab the next epoch start time (using a new helper fn `GetStartTimeNextEpoch`) and pass it into the query submission 
- all icq responses are submitted to `interchainquery/msg_server.go`, so I added logic there to check the ttl time against the stride block time at the time stride receives the response, and abort & delete the query accordingly
- I cleaned up some legacy unused interchainquery module methods e.g. `SetDatapoint`, `GetDatapoint` and various derivatives thereof.

## Testing and Verifying

I recommend two reviews for this PR

1. Review up to commit [2261c31](https://github.com/Stride-Labs/stride/pull/204/commits/2261c310a88f84e66d83ca481c6876f85378958c), at which point the ttl are implemented and working. Please test the ttls! See below

To test the case where the ttl is below current Stride block time, therefore the query response is discarded, se tthe ttl when you submit the `UpdateWithdrawalBalance` query in `msg_server_submit_tx.go` to the current time plus one nanosecond, with `ttl  = uint64(ctx.BlockHeader().Time.UnixNano()) + 1`. Then observe the Stride logs, you should see aborted query responses, like this:

```
4:33PM INF [ICQ Resp] query 1a98035e9fac86fc300945707bc1d79a8adc8e7aee146e13d137c6a1bfdc0fe3 with ttl: 1661027575034691001, resp time: 1661027616216014000. module=x/interchainquery
4:33PM INF [ICQ Resp] aborting query callback due to ttl expiry! ttl is 1661027575034691001, time now 1661027616216014000 for query of type store/bank/key with id OSMO, on chain 1a98035e9fac86fc300945707bc1d79a8adc8e7aee146e13d137c6a1bfdc0fe3 module=x/interchainquery
```

To speed up the above test, you may want to set stakeibc's`DefaultReinvestInterval` param to `1`.

2. Review the remaining commits, where I cleaned up some unused logic from `x/interchainquery`. In particular please check that`SetDatapoint`, `GetDatapoint`, `IterateDatapoint` and various derivatives thereof that I removed were in fact not being used in any way that would impact Stride
- `IterateDatapoint` was being used in `interchainquery/abci.go`, where at the end of each block it would look for stale datapoints and delete them. This was empty logic afaiu because we were never setting any datapoints in the first place!
- `SetDatapointForId` was in `msg_server.go` but we were never traversing that codepath: it was used to queries having nonzero ttl (we didn't have any!) as a `datapoint` so they could be re-tried later. We do not want this behavior, so we don't need this logic. 

  - How is the feature or change documented? not applicable
  - Does this pull request update existing proto field values (and require a backend and frontend migration)? no
  - Does this pull request change existing proto field names (and require a frontend migration)? no


x/interchainquery/keeper/msg_server.go
```diff
@@ -31,8 +32,26 @@ var _ types.MsgServer = msgServer{}
 func (k msgServer) SubmitQueryResponse(goCtx context.Context, msg *types.MsgSubmitQueryResponse) (*types.MsgSubmitQueryResponseResponse, error) {
 	ctx := sdk.UnwrapSDKContext(goCtx)
 	q, found := k.GetQuery(ctx, msg.QueryId)
+
 	if found {
+
+		// ABORT PROCESSING QUERY RESPONSE IF WE EXCEEDED THE TTL
+		k.Logger(ctx).Info(fmt.Sprintf("[ICQ Resp] query %s with ttl: %d, resp time: %d.", msg.QueryId, q.Ttl, ctx.BlockHeader().Time.UnixNano()))
+		curT, err := cast.ToUint64E(ctx.BlockHeader().Time.UnixNano())
```
asalzmann, line 40:
```suggestion
		curT, err := cast.ToUint64E(ctx.BlockTime().UnixNano())
```

asalzmann (approved):
LGTM! I'm a little less familiar with ICQ, so might be good to get another set of eyes on this (although the change seems pretty straightforward)

Thanks for writing the detailed description!

asalzmann:
Per offline discussion, sounds like we should give `Reinvestment Query` the same timeout as our ICAs

x/interchainquery/keeper/msg_server.go
```diff
@@ -31,8 +32,27 @@ var _ types.MsgServer = msgServer{}
 func (k msgServer) SubmitQueryResponse(goCtx context.Context, msg *types.MsgSubmitQueryResponse) (*types.MsgSubmitQueryResponseResponse, error) {
 	ctx := sdk.UnwrapSDKContext(goCtx)
 	q, found := k.GetQuery(ctx, msg.QueryId)
+
 	if found {
+
+		// ABORT PROCESSING QUERY RESPONSE IF WE EXCEEDED THE TTL
+		k.Logger(ctx).Info(fmt.Sprintf("[ICQ Resp] query %s with ttl: %d, resp time: %d.", msg.QueryId, q.Ttl, ctx.BlockHeader().Time.UnixNano()))
+		curT, err := cast.ToUint64E(ctx.BlockTime().UnixNano())
```
sampocs, line 40: Are these two time's supposed to be the same?
```
ctx.BlockHeader().Time.UnixNano()
ctx.BlockTime().UnixNano()
```

x/interchainquery/keeper/msg_server.go
```diff
@@ -31,8 +32,27 @@ var _ types.MsgServer = msgServer{}
...
+
+		if err != nil {
+			return nil, err
+		}
+		if q.Ttl < curT {
+			errMsg := fmt.Sprintf("[ICQ Resp] aborting query callback due to ttl expiry! ttl is %d, time now %d for query of type %s with id %s, on chain %s", q.Ttl, ctx.BlockHeader().Time.UnixNano(), q.QueryType, q.ChainId, msg.QueryId)
+			k.DeleteQuery(ctx, msg.QueryId)
+			k.Logger(ctx).Info(errMsg)
```
sampocs, line 48: nit: log as Error, not Info

sampocs (commented):
lgtm!

sampocs:
Just ran the testing flow described in the description and confirmed the query was timed out

PR #205, opened 2022-08-20
Add staticcheck static analyzer

riley-stride:

Add [staticcheck static analyzer](https://staticcheck.io/) to the project. It caught some minor stuff that our linter did not, minor stuff, fixed in this PR.

Install locally with `go install honnef.co/go/tools/cmd/staticcheck@2022.1.3`

I couldn't figure out how to make staticcheck run on the whole project, so I added a short bash script to run staticcheck in each relevant directory.

## How to run it

`make staticcheck`


app/app.go
```diff
@@ -80,7 +80,6 @@ import (
 	ibcclient "github.com/cosmos/ibc-go/v3/modules/core/02-client"
 	ibcclientclient "github.com/cosmos/ibc-go/v3/modules/core/02-client/client"
 	ibcclienttypes "github.com/cosmos/ibc-go/v3/modules/core/02-client/types"
-	ibcporttypes "github.com/cosmos/ibc-go/v3/modules/core/05-port/types"
```
asalzmann, line 83: nit: can we keep `ibcporttypes` around and remove `porttypes`?

asalzmann (commented):
nice! would it be possible to add this as a required check using github actions?

app/app.go
riley-stride: done

asalzmann:
This looks like the right approach, but the action is failing for some reason  https://github.com/Stride-Labs/stride/runs/7941585193?check_suite_focus=true

It might be a path issue? 
```
staticcheck /home/runner/work/stride/stride/scripts-local/../x
./scripts-local/staticcheck.sh: line 13: staticcheck: command not found
```

riley-stride:
Updated, GH CI staticcheck is working now! The check _fails_, but that's OK, I made it non-required since it is a very strict linter and we cannot and should not make all the suggested changes today, rather strive to make them over time.

PR #209, opened 2022-08-21
UNIT TEST: update redemption rate

riley-stride:
added unit test for update redemption rate

## reviewers:
can you think of cases we miss here?

### DIFF
--- a/x/stakeibc/keeper/update_redemption_rates_test.go
+++ b/x/stakeibc/keeper/update_redemption_rates_test.go
@@ -15,13 +15,8 @@ import (
 )
 
 type UpdateRedemptionRatesTestCase struct {
-	hostZone         stakeibc.HostZone
-	redemptionRateT0 sdk.Dec
-	stakedBal        uint64
-	undelegatedBal   uint64
-	justDepositedBal uint64
-	stSupply         uint64
-	allRecords       []recordtypes.DepositRecord
+	hostZone   stakeibc.HostZone
+	allRecords []recordtypes.DepositRecord
 }
 
 func (s *KeeperTestSuite) SetupUpdateRedemptionRates(
@@ -67,13 +62,8 @@ func (s *KeeperTestSuite) SetupUpdateRedemptionRates(
 	s.App.StakeibcKeeper.SetHostZone(s.Ctx, hostZone)
 
 	return UpdateRedemptionRatesTestCase{
-		hostZone:         hostZone,
-		redemptionRateT0: redemptionRateT0,
-		stakedBal:        stakedBal,
-		undelegatedBal:   undelegatedBal,
-		justDepositedBal: justDepositedBal,
-		stSupply:         stSupply,
-		allRecords:       []recordtypes.DepositRecord{toBeStakedDepositRecord, toBeTransferedDepositRecord},
+		hostZone:   hostZone,
+		allRecords: []recordtypes.DepositRecord{toBeStakedDepositRecord, toBeTransferedDepositRecord},
 	}
 }
 
@@ -88,18 +78,17 @@ func (s *KeeperTestSuite) TestUpdateRedemptionRatesSuccessful() {
 	tc := s.SetupUpdateRedemptionRates(stakedBal, undelegatedBal, justDepositedBal, stSupply, redemptionRateT0)
 
 	// sanity check on inputs (check redemptionRate at genesis is 1)
-	s.Require().Equal(tc.redemptionRateT0, sdk.NewDec(1))
+	s.Require().Equal(redemptionRateT0, sdk.NewDec(1), "to rr")
 
 	records := tc.allRecords
 	s.App.StakeibcKeeper.UpdateRedemptionRates(s.Ctx, records)
 
 	hz, found := s.App.StakeibcKeeper.GetHostZone(s.Ctx, tc.hostZone.ChainId)
-	s.Require().True(found)
+	s.Require().True(found, "hz found")
 	rrNew := hz.RedemptionRate
 
 	expectedNewRate := sdk.NewDec(5 + 3 + 3).Quo(sdk.NewDec(10))
-	s.Require().Equal(rrNew, expectedNewRate)
-	s.Require().Equal(rrNew, sdk.NewDec(11).Quo(sdk.NewDec(10)))
+	s.Require().Equal(rrNew, expectedNewRate, "rr as expected")
 }
 
 func (s *KeeperTestSuite) TestUpdateRedemptionRatesRandomized() {
@@ -123,13 +112,13 @@ func (s *KeeperTestSuite) TestUpdateRedemptionRatesRandomized() {
 	tc := s.SetupUpdateRedemptionRates(stakedBal, undelegatedBal, justDepositedBal, stSupply, redemptionRateT0)
 
 	// sanity check on inputs (check redemptionRate at genesis is 1)
-	s.Require().Equal(tc.redemptionRateT0, sdk.NewDec(1))
+	s.Require().Equal(redemptionRateT0, sdk.NewDec(1), "t0 rr")
 
 	records := tc.allRecords
 	s.App.StakeibcKeeper.UpdateRedemptionRates(s.Ctx, records)
 
 	hz, found := s.App.StakeibcKeeper.GetHostZone(s.Ctx, tc.hostZone.ChainId)
-	s.Require().True(found)
+	s.Require().True(found, "hz found")
 	rrNew := hz.RedemptionRate
 
 	numerator := int64(stakedBal) + int64(undelegatedBal) + int64(justDepositedBal)
@@ -157,17 +146,17 @@ func (s *KeeperTestSuite) TestUpdateRedemptionRateZeroStAssets() {
 	tc := s.SetupUpdateRedemptionRates(stakedBal, undelegatedBal, justDepositedBal, stSupply, redemptionRateT0)
 
 	// sanity check on inputs (check redemptionRate at genesis is 1)
-	s.Require().Equal(tc.redemptionRateT0, sdk.NewDec(1))
+	s.Require().Equal(redemptionRateT0, sdk.NewDec(1), "t0 rr")
 
 	records := tc.allRecords
 	s.App.StakeibcKeeper.UpdateRedemptionRates(s.Ctx, records)
 
 	hz, found := s.App.StakeibcKeeper.GetHostZone(s.Ctx, tc.hostZone.ChainId)
-	s.Require().True(found)
+	s.Require().True(found, "hz found")
 	rrNew := hz.RedemptionRate
 
 	// RR should be unchanged
-	s.Require().Equal(rrNew, sdk.NewDec(1))
+	s.Require().Equal(rrNew, sdk.NewDec(1), "rr as expected")
 }
 
 func (s *KeeperTestSuite) TestUpdateRedemptionRateZeroNativeAssets() {
@@ -181,17 +170,17 @@ func (s *KeeperTestSuite) TestUpdateRedemptionRateZeroNativeAssets() {
 	tc := s.SetupUpdateRedemptionRates(stakedBal, undelegatedBal, justDepositedBal, stSupply, redemptionRateT0)
 
 	// sanity check on inputs (check redemptionRate at genesis is 1)
-	s.Require().Equal(tc.redemptionRateT0, sdk.NewDec(1))
+	s.Require().Equal(redemptionRateT0, sdk.NewDec(1), "t0 rr")
 
 	records := tc.allRecords
 	s.App.StakeibcKeeper.UpdateRedemptionRates(s.Ctx, records)
 
 	hz, found := s.App.StakeibcKeeper.GetHostZone(s.Ctx, tc.hostZone.ChainId)
-	s.Require().True(found)
+	s.Require().True(found, "hz found")
 	rrNew := hz.RedemptionRate
 
 	// RR should be 0
-	s.Require().Equal(rrNew, sdk.NewDec(0))
+	s.Require().Equal(rrNew, sdk.NewDec(0), "rr as expected")
 }
 
 func (s *KeeperTestSuite) TestUpdateRedemptionRateNoModuleAccountRecords() {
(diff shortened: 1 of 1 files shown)


x/stakeibc/keeper/update_redemption_rates_test.go
```diff
@@ -0,0 +1,272 @@
...
+	just_deposited_bal uint64
+	st_supply          uint64
+}
+
+func (s *KeeperTestSuite) SetupUpdateRedemptionRates(
+	STAKED_BAL uint64,
+	UNDELEGATED_BAL uint64,
+	JUST_DEPOSITED_BAL uint64,
+	ST_SUPPLY uint64,
+	REDEMPTION_RATE_T0 sdk.Dec,
```
sampocs, line 30: looks like this parameter only ever equals 1 in these tests. If that's the case, I would just hard code it in this setup function and remove this parameter from the function definition

x/stakeibc/keeper/update_redemption_rates_test.go
```diff
@@ -0,0 +1,272 @@
...
+
+	records := s.App.RecordsKeeper.GetAllDepositRecord(s.Ctx)
+	s.App.StakeibcKeeper.UpdateRedemptionRates(s.Ctx, records)
+
+	hz, found := s.App.StakeibcKeeper.GetHostZone(s.Ctx, tc.host_zone.ChainId)
+	s.Require().True(found)
+	rrNew := hz.RedemptionRate
+
+	s.Require().Equal(rrNew, sdk.NewDec(int64(STAKED_BAL)+int64(UNDELEGATED_BAL)+int64(JUST_DEPOSITED_BAL)).Quo(sdk.NewDec(int64(ST_SUPPLY))))
+	s.Require().Equal(rrNew, sdk.NewDec(11).Quo(sdk.NewDec(10)))
```
sampocs, line 98: i'd be careful about including a copy of the code that's being tested, in the test itself (i.e. this equation). I see you have the hard coded case below though!

Also do you have to do `sdk.NewDec(11).Quo(sdk.NewDec(10))` because `NewDec` takes an int as a parameter?

x/stakeibc/keeper/update_redemption_rates_test.go
```diff
@@ -0,0 +1,272 @@
...
+	genRandUintBelowMax := func(MAX int) uint64 {
+		MIN := int(0)
+		n := 0 + rand.Intn(MAX-MIN+1)
+		return uint64(n)
+	}
+
+	MAX := 1_000_000_000
+	STAKED_BAL := genRandUintBelowMax(MAX)
+	UNDELEGATED_BAL := genRandUintBelowMax(MAX)
+	JUST_DEPOSITED_BAL := genRandUintBelowMax(MAX)
```
sampocs, line 114: nit: camel case

x/stakeibc/keeper/update_redemption_rates_test.go
```diff
@@ -0,0 +1,272 @@
...
+	s.App.StakeibcKeeper.UpdateRedemptionRates(s.Ctx, records)
+
+	hz, found := s.App.StakeibcKeeper.GetHostZone(s.Ctx, tc.host_zone.ChainId)
+	s.Require().True(found)
+	rrNew := hz.RedemptionRate
+
+	numerator := int64(STAKED_BAL) + int64(UNDELEGATED_BAL) + int64(JUST_DEPOSITED_BAL)
+	denominator := int64(ST_SUPPLY)
+	expectedNewRate := sdk.NewDec(numerator).Quo(sdk.NewDec(denominator))
+	s.Require().Equal(rrNew, expectedNewRate)
```
sampocs, line 135: since these inputs are randomized, I'd log them here by adding the description argument to `.Equal`, otherwise if this fails, you wont know the values that caused it to fail

x/stakeibc/keeper/update_redemption_rates_test.go
```diff
@@ -0,0 +1,272 @@
...
+	ST_SUPPLY := genRandUintBelowMax(MAX)
+
+	// s.Require().ElementsMatch([]int{0, 0, 0, 0}, []int{int(STAKED_BAL), int(UNDELEGATED_BAL), int(JUST_DEPOSITED_BAL), int(ST_SUPPLY)}) //
+	REDEMPTION_RATE_T0 := sdk.NewDec(1)
+	tc := s.SetupUpdateRedemptionRates(STAKED_BAL, UNDELEGATED_BAL, JUST_DEPOSITED_BAL, ST_SUPPLY, REDEMPTION_RATE_T0)
+
+	// sanity check on inputs (check redemptionRate at genesis is 1)
+	s.Require().Equal(tc.redemption_rate_t0, sdk.NewDec(1))
+
+	records := s.App.RecordsKeeper.GetAllDepositRecord(s.Ctx)
```
sampocs, line 125: Might make sense to return these records from the setup function so this test does not rely on GetAllDepositRecords

x/stakeibc/keeper/update_redemption_rates_test.go
```diff
@@ -0,0 +1,272 @@
...
+	s.App.StakeibcKeeper.UpdateRedemptionRates(s.Ctx, records)
+
+	hz, found := s.App.StakeibcKeeper.GetHostZone(s.Ctx, tc.host_zone.ChainId)
+	s.Require().True(found)
+	rrNew := hz.RedemptionRate
+
+	numerator := int64(STAKED_BAL) + int64(UNDELEGATED_BAL)
+	denominator := int64(ST_SUPPLY)
+	expectedNewRate := sdk.NewDec(numerator).Quo(sdk.NewDec(denominator))
+	s.Require().Equal(rrNew, expectedNewRate)
```
sampocs, line 218: same point above, I'd hard code these values instead of repeating the equation

x/stakeibc/keeper/update_redemption_rates_test.go
```diff
@@ -0,0 +1,272 @@
...
+	rrNew := hz.RedemptionRate
+
+	numerator := int64(STAKED_BAL) + int64(UNDELEGATED_BAL) + int64(JUST_DEPOSITED_BAL)
+	denominator := int64(ST_SUPPLY)
+	expectedNewRate := sdk.NewDec(numerator).Quo(sdk.NewDec(denominator))
+	s.Require().Equal(rrNew, expectedNewRate)
+}
+
+func (s *KeeperTestSuite) TestUpdateRedemptionRatesRandomized_MultipleRuns() {
+	for i := 0; i < 100; i++ {
```
sampocs, line 139: how long does this take?

x/stakeibc/keeper/update_redemption_rates_test.go
```diff
@@ -0,0 +1,272 @@
...
+
+	hz, found := s.App.StakeibcKeeper.GetHostZone(s.Ctx, tc.host_zone.ChainId)
+	s.Require().True(found)
+	rrNew := hz.RedemptionRate
+
+	// RR should be 0
+	s.Require().Equal(rrNew, sdk.NewDec(0))
+}
+
+func (s *KeeperTestSuite) TestUpdateRedemptionRate_NoModuleAccountRecords() {
```
sampocs, line 194: nit: some function names use the underscore and some don't - I think either is fine, but just keep it consistent

sampocs (commented):
looks good! Couple small comments below.

I can't think of any other test cases. Also, good thinking on the randomized inputs!!

x/stakeibc/keeper/update_redemption_rates_test.go
```diff
@@ -0,0 +1,272 @@
...
+	_ "github.com/stretchr/testify/suite"
+
+	recordtypes "github.com/Stride-Labs/stride/x/records/types"
+
+	// "github.com/Stride-Labs/stride/x/stakeibc/types"
+	stakeibc "github.com/Stride-Labs/stride/x/stakeibc/types"
+)
+
+type UpdateRedemptionRatesTestCase struct {
+	host_zone          stakeibc.HostZone
```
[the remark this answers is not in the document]
riley-stride: done

x/stakeibc/keeper/update_redemption_rates_test.go
```diff
@@ -0,0 +1,272 @@
...
+	redemption_rate_t0 sdk.Dec
+	staked_bal         uint64
+	undelegated_bal    uint64
+	just_deposited_bal uint64
+	st_supply          uint64
+}
+
+func (s *KeeperTestSuite) SetupUpdateRedemptionRates(
+	STAKED_BAL uint64,
```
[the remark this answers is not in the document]
riley-stride: done

x/stakeibc/keeper/update_redemption_rates_test.go
riley-stride: good catch, just added a test that starts with a redemption rate that's not 1

x/stakeibc/keeper/update_redemption_rates_test.go
riley-stride: right, NewDec only takes ints so afaiu the best way to get 1.1 is using 11 Quo 10

Any ideas on how to test randomized input tests without "including a copy of the code that's being tested, in the test itself (i.e. this equation)" ? @sampocs

x/stakeibc/keeper/update_redemption_rates_test.go
riley-stride: done

x/stakeibc/keeper/update_redemption_rates_test.go
riley-stride: good call, done

x/stakeibc/keeper/update_redemption_rates_test.go
riley-stride: good idea, done

x/stakeibc/keeper/update_redemption_rates_test.go
riley-stride: 0.2 seconds

x/stakeibc/keeper/update_redemption_rates_test.go
riley-stride: done

x/stakeibc/keeper/update_redemption_rates_test.go
riley-stride: done

x/stakeibc/keeper/update_redemption_rates_test.go
sampocs: I think it's fine to use the equation for that! Especially since you will already have plenty of coverage with hard coded numbers from the other tests

x/stakeibc/keeper/update_redemption_rates_test.go
sampocs: haha beautiful

x/stakeibc/keeper/update_redemption_rates_test.go
```diff
@@ -14,27 +15,28 @@ import (
...
-	staked_bal         uint64
-	undelegated_bal    uint64
-	just_deposited_bal uint64
-	st_supply          uint64
+	hostZone         stakeibc.HostZone
+	redemptionRateT0 sdk.Dec
+	stakedBal        uint64
+	undelegatedBal   uint64
+	justDepositedBal uint64
+	stSupply         uint64
```
sampocs, line 23: Changes look good! Also you can remove these from this struct since they're defined in each function (and thus are already accessible as normal variables). 

Also, super nit, but you only have to return the hostzone id, not the whole object. 

```
type UpdateRedemptionRatesTestCase struct {
    hostZoneId  string
    allRecords  []recordtypes.DepositRecord
}
```

x/stakeibc/keeper/update_redemption_rates_test.go
```diff
@@ -43,96 +45,97 @@ func (s *KeeperTestSuite) SetupUpdateRedemptionRates(
...
 
-	hz, found := s.App.StakeibcKeeper.GetHostZone(s.Ctx, tc.host_zone.ChainId)
+	hz, found := s.App.StakeibcKeeper.GetHostZone(s.Ctx, tc.hostZone.ChainId)
 	s.Require().True(found)
 	rrNew := hz.RedemptionRate
 
-	s.Require().Equal(rrNew, sdk.NewDec(int64(STAKED_BAL)+int64(UNDELEGATED_BAL)+int64(JUST_DEPOSITED_BAL)).Quo(sdk.NewDec(int64(ST_SUPPLY))))
+	expectedNewRate := sdk.NewDec(5 + 3 + 3).Quo(sdk.NewDec(10))
+	s.Require().Equal(rrNew, expectedNewRate)
 	s.Require().Equal(rrNew, sdk.NewDec(11).Quo(sdk.NewDec(10)))
```
sampocs, line 102: nit: no need to keep both of these, you can remove the second

x/stakeibc/keeper/update_redemption_rates_test.go
```diff
@@ -43,96 +45,97 @@ func (s *KeeperTestSuite) SetupUpdateRedemptionRates(
...
 	s.App.StakeibcKeeper.UpdateRedemptionRates(s.Ctx, records)
 
-	hz, found := s.App.StakeibcKeeper.GetHostZone(s.Ctx, tc.host_zone.ChainId)
+	hz, found := s.App.StakeibcKeeper.GetHostZone(s.Ctx, tc.hostZone.ChainId)
 	s.Require().True(found)
 	rrNew := hz.RedemptionRate
 
-	s.Require().Equal(rrNew, sdk.NewDec(int64(STAKED_BAL)+int64(UNDELEGATED_BAL)+int64(JUST_DEPOSITED_BAL)).Quo(sdk.NewDec(int64(ST_SUPPLY))))
+	expectedNewRate := sdk.NewDec(5 + 3 + 3).Quo(sdk.NewDec(10))
+	s.Require().Equal(rrNew, expectedNewRate)
```
sampocs, line 101: add descriptions to all the requires. Doesn't need to be super descriptive, just enough to point you to the corresponding line. Can be as simple as:

```
s.Require().True(found, "host zone found")

s.Require().Equal(rrNew, expectedNewRate, "redemption rate")
```

x/stakeibc/keeper/update_redemption_rates_test.go
riley-stride: good suggestion, done

x/stakeibc/keeper/update_redemption_rates_test.go
riley-stride: done

x/stakeibc/keeper/update_redemption_rates_test.go
riley-stride: good one, added

x/stakeibc/keeper/update_redemption_rates_test.go
```diff
@@ -88,18 +78,17 @@ func (s *KeeperTestSuite) TestUpdateRedemptionRatesSuccessful() {
 	tc := s.SetupUpdateRedemptionRates(stakedBal, undelegatedBal, justDepositedBal, stSupply, redemptionRateT0)
 
 	// sanity check on inputs (check redemptionRate at genesis is 1)
-	s.Require().Equal(tc.redemptionRateT0, sdk.NewDec(1))
+	s.Require().Equal(redemptionRateT0, sdk.NewDec(1), "to rr")
```
sampocs, line 81: is this supposed to be "to" or "t0" (as in time zero)?
Maybe just rename it to `initialRedemptionRate`?

riley-stride: done

sampocs (approved):

sampocs:
Beautiful!

PR #218, opened 2022-08-23
Testnet upgrade support

sampocs:
Closes: #XXX

There's currently no easy way to upgrade our testnet. This PR introduces a script that we can run from each node to create the proper cosmovisor structure for the upgrade. 

## Brief Changelog
- Added upgrade script
- Added alias in dockerfile so `strided` always points to the current binary in use

## Testing and Verifying
- Commented out the stateful lines from the dockerfile (since those depend on a GH actions build) and built the image locally

  - Does this pull request introduce a new feature or user-facing behavior changes? (yes / no) no
  - Is a relevant changelog entry added to the `Unreleased` section in `CHANGELOG.md`? (yes / no) no
  - How is the feature or change documented? (not applicable   /   specification (`x/<module>/spec/`)  / README.md  /   not documented) n/a
  - Does this pull request update existing proto field values (and require a backend and frontend migration)? (yes / no) no
  - Does this pull request change existing proto field names (and require a frontend migration)? (yes / no) no

### DIFF
--- a/scripts/testnet/upgrade.sh
+++ b/scripts/testnet/upgrade.sh
@@ -1,0 +1,29 @@
+#!/bin/bash
+
+set -eu 
+
+UPGRADE_NAME="$1"
+STRIDE_COMMIT_HASH="$2"
+COSMOVISOR_HOME=/stride/.stride/cosmovisor/upgrades
+
+echo "Upgrade Name: $UPGRADE_NAME"
+printf "Stride Commit Hash: $STRIDE_COMMIT_HASH\n\n"
+
+while true; do
+    read -p "Continue? [y/n]" yn
+    case $yn in
+        [Yy]* ) echo ""; break;;
+        [Nn]* ) exit ;;
+        * ) printf "Please answer yes or no.\n";;
+    esac
+done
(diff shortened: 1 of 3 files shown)


scripts/testnet/upgrade.sh
```diff
@@ -0,0 +1,30 @@
...
+    case $yn in
+        [Yy]* ) echo ""; break;;
+        [Nn]* ) exit ;;
+        * ) printf "Please answer yes or no.\n";;
+    esac
+done
+
+UPGRADE_DIR=${COSMOVISOR_HOME}/upgrades/${UPGRADE_NAME}/bin
+
+mkdir -p $UPGRADE_DIR
```
asalzmann, line 24: does this script assume the upgrade is being run from a particular directory?

asalzmann (approved):
lgtm!

scripts/testnet/upgrade.sh
sampocs: Yeah this is run from inside the docker image. Its sitting in the home directory after you exec into the container

PR #221, opened 2022-08-25
[SAFETY] set cap number of deposit records to be staked per epoch

riley-stride:
Closes #134 

We limit the number of ICA calls that could process per epoch in order to avoid a potential scenario in which, after the channel having gone down for some time, the pending number of ICAs to be submitted fills the BeginBlocker gas limit on Stride.

_Note: we choose to send only one ICA per `MsgDelegate`, so that we don't exceed the gas limit on host zones (the transactions would just be included in the next block)._ 

## Brief Changelog

  - *Added `MaxIcaCallsPerEpoch` param in `x/stakeibc`*
  - *Added logic to cap number of ICA txs processed each epoch in `hooks.go:StakeExistingDepositsOnHostZones`*
  - *Modified `scripts-local/2.sh` to manually test limiting the number of DRs processed. 

## Testing and Verifying

For this test, set `stakeibc/types/params.go`'s `DefaultDelegateInterval` to `10`, set its `DefaultDepositInterval` to `1`, and its `DefaultMaxICACallsPerEpoch` to `2`. STAKEs should happen less frequently than TRANSFERs, so we can see a pile-up of STAKE records and observe the cap on STAKE records processed (more or less simulating what would happen if `MsgDelegate` functionality were to go down for some time)

Set the params in `init_stride.sh`:
```
    jq '.app_state.stakeibc.params.delegate_interval = $interval' --arg interval 10 $main_config > json.tmp && mv json.tmp $main_config
    jq '.app_state.stakeibc.params.deposit_interval = $interval' --arg interval 1 $main_config > json.tmp && mv json.tmp $main_config

```
To run the test: `make init-local build=s`, then run gaia integration tests to fund the account, then run `2.sh`. The script `2.sh` adds more than `MaxIcaCallsPerEpoch` DepositRecords. At the next epoch, only `MaxIcaCallsPerEpoch` of these should be processed! Check this in `accounts.log` manually.

For my example test, all DRs between epoch numbers 13-18 were type STAKE, and at the 20th epoch, only 13 and 14 were staked (bc I had a `MaxIcaCallsPerEpoch` limit of 2!)


proto/stakeibc/params.proto
```diff
@@ -28,4 +28,5 @@ message Params {
   uint64 buffer_size = 10;
   uint64 ibc_timeout_blocks = 11;
   uint64 fee_transfer_timeout_nanos = 12;
+  uint64 max_ica_calls_per_epoch = 13;
```
asalzmann, line 31: I think this PR and your RR check PR will have a merge conflict here, just flagging

scripts-local/test-utils/2.sh
```diff
@@ -7,3 +7,11 @@ $STRIDE_CMD tx stakeibc liquid-stake 10000 $ATOM_DENOM --keyring-backend test --
 echo "Waiting for liquid staked tokens to be delegated..."
 CSLEEP 60
 # $STRIDE_CMD tx stakeibc liquid-stake 10000 $ATOM_DENOM --keyring-backend test --from $STRIDE_VAL_ACCT -y --chain-id $STRIDE_CHAIN
+
+# liquid stake many times in sequence to flood the deposit record queue, in order to test the max staking ICA calls limit
```
asalzmann, line 11: nit: put in a different file

x/stakeibc/types/params.go
```diff
@@ -19,11 +19,11 @@ var (
...
-	DefaultICATimeoutNanos  uint64 = 600000000000
-	DefaultBufferSize       uint64 = 5   // 1/5=20% of the epoch
-	DefaultIbcTimeoutBlocks uint64 = 300 // 300 blocks ~= 30 minutes
-	DefaultFeeTransferTimeoutNanos  uint64 = 600000000000 // 10 minutes
-
+	DefaultICATimeoutNanos         uint64 = 600000000000
+	DefaultBufferSize              uint64 = 5            // 1/5=20% of the epoch
+	DefaultIbcTimeoutBlocks        uint64 = 300          // 300 blocks ~= 30 minutes
+	DefaultFeeTransferTimeoutNanos uint64 = 600000000000 // 10 minutes
+	DefaultMaxICACallsPerEpoch     uint64 = 100
```
asalzmann, line 26:
```suggestion
	DefaultMaxICAMsgDelegateCallsPerEpoch     uint64 = 100
```
this only limits the staking calls afaict

asalzmann (commented):
One minor comment on the description of the PR - we send one ICA per `MsgDelegate`, so we shouldn't exceed the gas limit on host zones (the transactions would just be included in the next block). I think the issue here is that we could hit the `BeginBlocker` gas limit on _stride_

nit: in PR description, add "closes #134"

Is it possible to add a unittest for this?

scripts-local/test-utils/2.sh
riley-stride: done

x/stakeibc/types/params.go
riley-stride: renamed, ty!

riley-stride: renamed, ty!

riley-stride:
@asalzmann I think this is ready to merge

riley-stride:
> @riley-stride Is it possible to add a unittest for this?
> 
> Sorry, might have missed this if you responded to it somewhere

My bad, should have replied here re: unit test. Scaffolding the unit test is involved, so I added a note to include a test for this on the sheet on Sam's unit test for this. Will make sure to review that one. 

Tagging for visibility @asalzmann @sampocs

asalzmann:
Reference https://docs.google.com/spreadsheets/d/1TN6l6B-gIcPJvreBpopIK9hubml4tGHas-spdzP6MdM/edit?disco=AAAAeyNpTIo

asalzmann (approved):

PR #228, opened 2022-08-26
update pull request template

riley-stride:
What we have is good but could be even better:
- sometimes we're not running integration tests
- cosmos-sdk repo has a groovy PR template, I borrowed from some of their items
- checkmarks are nice and easy + they make clearly visible what is/isn't complete
- added reviewer checklist

### DIFF
--- a/pull_request_template.md
+++ b/pull_request_template.md
@@ -27,10 +27,7 @@ Closes: #XXX
 
 I have...
 
-- [ ] Run and PASSED locally all integration tests on...
-    - [ ] GAIA
-    - [ ] JUNO
-    - [ ] OSMO
+- [ ] Run and PASSED locally all GAIA integration tests
 - [ ] If the change is contentful, I either:
     - [ ] Added a new unit test OR 
     - [ ] Added test cases to existinig unit tests
--- a/pull_request_template.md
+++ b/pull_request_template.md
@@ -7,7 +7,7 @@ v    If your PR doesn't close an issue, that's OK!  Just remove the Closes: #XXX
 
 Closes: #XXX
 
-## What is the purpose of the change
+## Context and purpose of the change
 
 <!-- Add a description of the overall background and high level changes that this PR introduces
 
@@ -34,6 +34,7 @@ I have...
 - [ ] If the change is contentful, I either:
     - [ ] Added a new unit test OR 
     - [ ] Added test cases to existinig unit tests
+- [ ] OR this change is a trivial rework / code cleanup without any test coverage
 
 If skipped any of the tests above, explain.
 <!-- *(example:)*
@@ -42,7 +43,7 @@ If skipped any of the tests above, explain.
   - *Extended integration test for ...*
   - *Manually verified the change by ...* -->
 
-### Reviewers Checklist
+## Reviewers Checklist
 
 *All items are required. Please add a note if the item is not applicable and please add
 your handle next to the items reviewed if you only reviewed selected items.*
@@ -67,4 +68,4 @@ I have...
       - [ ] jira ticket `XXX` 
       - [ ] specification (`x/<module>/spec/`) 
       - [ ] README.md 
-      - [ ] not documented
+      - [ ] not documented <!-- because ... EXPLAIN WHY! -->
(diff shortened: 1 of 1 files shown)


sampocs (approved):
Beautiful!

I grant you permission to skip the integration tests for this PR :)

pull_request_template.md
```diff
@@ -9,45 +9,62 @@ Closes: #XXX
...
-*(or)*
+If skipped any of the tests above, explain.
+<!-- *(example:)*
+  - *Added unit test that validates ...*
+  - *Added integration tests for end-to-end deployment with ...*
+  - *Extended integration test for ...*
+  - *Manually verified the change by ...* -->
 
-This change is already covered by existing tests, such as *(please describe tests)*.
+### Reviewers Checklist
```
asalzmann, line 45:
```suggestion
## Reviewers Checklist
```

riley-stride:
pls add more / edit, I prob missed a bunch of items and maybe added some we don't need

pull_request_template.md
```diff
@@ -9,45 +9,62 @@ Closes: #XXX
...
+  - *Daemons retrieve the RPC data from the blob cache* -->
 
 
-## Testing and Verifying
+## Author's Checklist
 
-*(Please pick one of the following options)*
+I have...
 
-This change is a trivial rework / code cleanup without any test coverage.
```
asalzmann, line 30: Given these take ~30 min right now, it might be a bit unrealistic to run these for every PR

There are also trivial PRs where we wouldn't have to run tests (e.g. changes that don't change core chain logic), so we might want to keep this line around

pull_request_template.md
```diff
@@ -9,45 +9,62 @@ Closes: #XXX
 
 ## What is the purpose of the change
```
asalzmann, line 10:
```suggestion
## Context and purpose of the change
```

It seems like lots of PRs have recently been adding additional sections explaining things, let's change this title and consolidate those sections here

pull_request_template.md
```diff
@@ -9,45 +9,62 @@ Closes: #XXX
...
+  - [ ] Does this pull request introduce a new feature or user-facing behavior changes? 
+  - [ ] Is a relevant changelog entry added to the `Unreleased` section in `CHANGELOG.md`?
+  - [ ] This pull request updates existing proto field values (and require a backend and frontend migration)? 
+  - [ ] Does this pull request change existing proto field names (and require a frontend migration)?
+  How is the feature or change documented? 
+      - [ ] not applicable
+      - [ ] jira ticket `XXX` 
+      - [ ] specification (`x/<module>/spec/`) 
+      - [ ] README.md 
+      - [ ] not documented
```
asalzmann, line 70: if so, explain why

asalzmann (approved):
lgtm! thanks for adding this

pull_request_template.md
riley-stride: done

pull_request_template.md
riley-stride: good call, done

pull_request_template.md
riley-stride: done

pull_request_template.md
sampocs: As annoying as it is, I'd advocate for running the integration tests before each merge. I wasted half the day yesterday fixing things that broke across multiple PRs. It would have been much easier to fix each at the time of the change. And most of them wound up being changes that you would not have expected to break the tests (e.g. removing a log).

I think we should up the priority of getting these running in GH actions so we can offload the burden and context switch to a different ticket. I'm happy to tackle this next week

riley-stride: Agree re pushing these into GH actions!

riley-stride: @asalzmann what's your take on the priority level of getting those into GH actions?

asalzmann: If we run these after every PR, I think we're going to slow down our dev cycle pretty significantly, although I hear you on having to go back and fix the tests, and that being really annoying (also shouldn't fall on you - if integration tests are broken, we should put it in #development and figure out why together).

What if we compromised and ran the gaia tests after every PR, until we can get these running in actions? That feels like the 80/20

PR #231, opened 2022-08-27
[UNIT-TEST] Deposit Records

sampocs:
Closes: #XXX

* Added unit tests for the deposit records functionality in hooks

## Brief Changelog
- Added tests for `CreateDepositRecordsForEpoch`, `TransferExistingDepositsToHostZones`, and `StakeExistingDepositsOnHostZones`
- Moved functions into their own `deposit_records.go` file. 
   - **This was done in the last commit. As a reviewing, it might be easier to first review the [diffs](https://github.com/Stride-Labs/stride/compare/main...0e2aa2be2a9cf866285f770c7a5bfa594e6a25ac) up until the last commit**
- Added function to `AppTestHelper` to determine get the IBC denom given a channel and host denom
- Changed some nit things in other unit tests so we have better references points (since most of the boilerplate gets duplicated when new tests are written)
 

## Author's Checklist

I have...
- [X] If the change is contentful, I either:

## Reviewers Checklist

I have...

      - [ ] not applicable
      - [ ] jira ticket `XXX` 
      - [ ] README.md 
      - [X] not documented  (Unit tests)


asalzmann:
(for other reviewers) there's a bug with the github UI so you can't "hold shift + select commits" if you need to scroll, this link should show the diff for the relevant commits though! https://github.com/Stride-Labs/stride/pull/231/files/0109497b0f121da5db6fe0d63c4cb5da09d94487..0e2aa2be2a9cf866285f770c7a5bfa594e6a25ac

x/stakeibc/keeper/hooks_deposit_records_test.go
```diff
@@ -0,0 +1,527 @@
...
+	}
+}
+
+func (s *KeeperTestSuite) SetupDepositRecords() DepositRecordsTestCase {
+	delegationAccountOwner := fmt.Sprintf("%s.%s", HostChainId, "DELEGATION")
+	delegationChannelID := s.CreateICAChannel(delegationAccountOwner)
+	delegationAddress := s.IcaAddresses[delegationAccountOwner]
+
+	ibcDenomTrace := s.GetIBCDenom(Atom) // we need a true IBC denom here
+	hostModuleAddress := stakeibctypes.NewZoneAddress(HostChainId)
```
asalzmann: nit: thoughts on `hostZoneModuleAddress` or `zoneModuleAddress`? I initially thought this account was on the host chain

x/stakeibc/keeper/hooks_deposit_records_test.go
```diff
@@ -0,0 +1,527 @@
...
+	s.App.TransferKeeper.SetDenomTrace(s.Ctx(), ibcDenomTrace)
+
+	initialModuleAccountBalance := sdk.NewCoin(ibcDenomTrace.IBCDenom(), sdk.NewInt(15_000))
```
asalzmann: where does this 15_000 come from?

x/stakeibc/keeper/hooks_deposit_records_test.go
```diff
@@ -0,0 +1,527 @@
...
+		initialModuleAccountBalance: initialModuleAccountBalance,
+		hostZone:                    hostZone,
+		hostModuleAddress:           hostModuleAddress,
+		epochNumber:                 currentEpoch,
+		TransferChannel: Channel{
+			PortID:    ibctesting.TransferPort,
+			ChannelID: ibctesting.FirstChannelID,
+		},
+		DelegationChannel: Channel{
+			PortID:    "icacontroller-" + delegationAccountOwner,
```
asalzmann: nit: this is defined in the ICA module as `KeyPrefix` I believe

x/stakeibc/keeper/hooks_deposit_records_test.go
```diff
@@ -0,0 +1,527 @@
...
+	// Confirm tx sequence was incremented
+	numTransferAttempts := len(tc.initialDepositRecords.recordsToBeTransfered)
+	numSuccessfulTransfers := uint64(numTransferAttempts - numTransfersFailed)
+
+	endSequence, found := s.App.IBCKeeper.ChannelKeeper.GetNextSequenceSend(s.Ctx(), transferPortID, transferChannelID)
+	s.Require().True(found, "sequence number not found after transfer")
+	s.Require().Equal(startSequence+numSuccessfulTransfers, endSequence, "tx sequence number after transfer")
+
+	// Confirm the callback data was stored for each transfer packet EXCLUDING the failed packets
+	numCallbacks := uint64(len(s.App.IcacallbacksKeeper.GetAllCallbackData(s.Ctx())))
```
asalzmann: Nice!

x/stakeibc/keeper/hooks_deposit_records_test.go
```diff
@@ -0,0 +1,527 @@
...
+	for _, emptyRecord := range tc.initialDepositRecords.emptyRecords {
+		_, found := s.App.RecordsKeeper.GetDepositRecord(s.Ctx(), emptyRecord.Id)
+		s.Require().False(found, "empty deposit record (%d) should have been removed", emptyRecord.Id)
+	}
+}
+
+func (s *KeeperTestSuite) TestTransferDepositRecords_Successful() {
+	tc := s.SetupDepositRecords()
+
+	numFailures := 0
```
asalzmann: clean

asalzmann (commented):
I think this is our best unittest yet

Just one question about deleting the deposit records while iterating them, other than that lgtm!

app/apptesting/test_helpers.go
```diff
@@ -280,3 +281,12 @@ func  (s *AppTestHelper) ICAPacketAcknowledgement(msgs []sdk.Msg) channeltypes.A
 	ack := channeltypes.NewResultAcknowledgement(marshalledTxMsgData)
 	return ack
 }
+
+// Get an IBC denom from it's native host denom
+// This assumes the transfer channel is channel-0
+func (s *AppTestHelper) GetIBCDenom(denom string) transfertypes.DenomTrace {
+	sourcePrefix := transfertypes.GetDenomPrefix(ibctesting.TransferPort, ibctesting.FirstChannelID)
```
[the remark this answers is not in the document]
sampocs: This function is really just doing string formatting (i.e. port/channelId/denom) and storing it as a trace type. It's not checking anything from state.

That said, maybe I could name this function a little more clearly? Fwiw, I originally had this return the actual IBC hash, but then changed it to return just the denom object since that's what you need to store it in state, and you can easy get the hash with `.IBCDenom()`

x/stakeibc/keeper/msg_server_add_validator_test.go
```diff
@@ -5,23 +5,22 @@ import (
 
 	sdk "github.com/cosmos/cosmos-sdk/types"
 
-	"github.com/Stride-Labs/stride/x/stakeibc/types"
-	stakeibc "github.com/Stride-Labs/stride/x/stakeibc/types"
+	stakeibctypes "github.com/Stride-Labs/stride/x/stakeibc/types"
 )
 
 type AddValidatorTestCase struct {
-	hostZone           types.HostZone
```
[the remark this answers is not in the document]
sampocs: I don't have a strong preference either! This was mostly be solving the case that both types were imported in each file, with one of them commented out.

I think I lean towards using `types` in the actual keeper functions, but `{module}types` in the tests. Reason being that the tests feel a little more global in the sense that we're generally operating at the app level, so it becomes a little less obvious what's the module in question. Wdyt?

x/stakeibc/keeper/hooks.go
```diff
@@ -239,107 +237,106 @@ func (k Keeper) SetWithdrawalAddress(ctx sdk.Context) {
...
+				sdk.EventTypeMessage,
+				sdk.NewAttribute("hostZone", hostZone.ChainId),
+				sdk.NewAttribute("newAmountStaked", strconv.FormatInt(depositRecord.Amount, 10)),
+			),
+		)
 	}
 }
 
 func (k Keeper) TransferExistingDepositsToHostZones(ctx sdk.Context, epochNumber uint64, depositRecords []recordstypes.DepositRecord) {
 	transferDepositRecords := utils.FilterDepositRecords(depositRecords, func(record recordstypes.DepositRecord) (condition bool) {
```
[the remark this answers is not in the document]
sampocs: Great call out, I hadn't thought about that. 

I think `depositRecords` is a copy because the state doesn't actually store a list data type. `GetAllDepositRecords` is iterating through every item in the store, unmarshalling the deposit record into a new object, and building a list from those objects. So by the time the list is returned, the actual state should be no longer referenced, and I think the list should just be a copy.  Which would mean, when we remove elements from the underlying store, it doesn't impact the list. 

I think the case that would be problematic would be if in hooks we were using our own iterator and iterating at the start of each for loop.

x/stakeibc/keeper/hooks_deposit_records_test.go
```diff
@@ -0,0 +1,527 @@
...
+	recordsToBeTransfered []recordstypes.DepositRecord
+	recordsToBeStaked     []recordstypes.DepositRecord
+	recordsInCurrentEpoch []recordstypes.DepositRecord
+	transferAmount        sdk.Int
+	stakeAmount           sdk.Int
+}
+
+func (r *TestDepositRecords) GetAllRecords() []recordstypes.DepositRecord {
+	allDepositRecords := []recordstypes.DepositRecord{}
+	allDepositRecords = append(allDepositRecords, r.emptyRecords...)
```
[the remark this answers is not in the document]
sampocs: I was looking for something like this! But it doesn't look like it's working for me. I get `unexpected c, expecting )` after the first spread.

x/stakeibc/keeper/hooks_deposit_records_test.go
sampocs: happy to rename it! Do you generally use zone when discussing something host-zone-related, but living on Stride? I had kind of though zone and hostZone were synonymous

x/stakeibc/keeper/hooks_deposit_records_test.go
sampocs: just an arbitrary number that doesn't matter much since balance changes are checked using the diff rather than the absolute value

x/stakeibc/keeper/hooks_deposit_records_test.go
sampocs: Nice, I was wondering where to find it! Updated

x/stakeibc/keeper/msg_server_add_validator_test.go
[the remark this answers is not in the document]
asalzmann: makes sense!

x/stakeibc/keeper/hooks_deposit_records_test.go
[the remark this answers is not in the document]
asalzmann: oh, maybe you can't chain spreads 🤔

asalzmann (approved):
lgtm

x/stakeibc/keeper/hooks_deposit_records_test.go
asalzmann: I think we do use them interchangeably, it would might be best to use host zone because that's the name of the message (so implies the data lives on stride)

PR #235, opened 2022-08-28
add claim callback unittest

asalzmann:

Add unittests to `ClaimCallback`

## Brief Changelog

- Add success case test and failure case tests for `ClaimCallback`
- Fix bug in `ClaimCallback` where the `UserRedemptionRecord` was not set back to claimable after a timeout

## Reviewers Checklist

I have...

      - [X] tests


sampocs (approved):
Lgtm! Great comments throughout the test! My only pieces of feedback are:

* Add a (super short) description to each `Require` statement so its easier to debug when tests fail 
* Might be worth revisiting the error messages in the claim callback function. For instance, the unmarshalClaimCallback error is a little cryptic without being wrapped with more context.

That last point brings up a larger question of whether we need to log and return every single error (and the answer is probably no, since everything gets propagated and wrapped). But I think it's easier to just do both today logging and returning today and then we can trim it down in the future

asalzmann:
> Add a (super short) description to each Require statement so its easier to debug when tests fail

Added

>  But I think it's easier to just do both today logging and returning today and then we can trim it down in the future

Agreed, we should figure out a consistent way to do this that is most helpful for debugging

PR #245, opened 2022-08-30
use blocktime to compute ibc transfer timeouts

asalzmann:

When ibc transferring tokens from Stride -> host zones (epochly), we use the block height on the LC to compute the timeout (we estimate 30 min in blocks and set timeoutHeight = lightClient.height + 30_min_in_blocks).There is a simple [pre-check](https://github.com/cosmos/ibc-go/blob/f49b8bc3c4e9ee2ada83f6c7529fc2e475ccb46e/modules/core/04-channel/keeper/packet.go#L80) done on the LC before sending the packet to ensure the timeout is not in the past (relative to the LC), but the final check that determines whether a packet will timeout happens in [`OnRecvPacket` on the destination chain](https://github.com/cosmos/ibc-go/blob/f49b8bc3c4e9ee2ada83f6c7529fc2e475ccb46e/modules/core/04-channel/keeper/packet.go#L203).

If the light client is more than 30 min stale, transfers will immediately timeout when they reach the host chain. In general, if the LC is X minutes stale, our timeout is 30 - X minutes, which is not desirable.

This PR updates our epochly ibc transfer logic so that transfers timeout 30 minutes after they were sent, regardless of the status of the LC.

Note: the reason this works is that tendermint chains come to consensus on time, based on machine clocks. https://github.com/tendermint/tendermint/blob/v0.34.x/spec/consensus/bft-time.md

If we onboard chains that don't use tendermint for consensus and are subject to clock drift, we'll need to update this.

## Brief Changelog

- Update `TransferExistingDepositsToHostZones` to use `ctx.BlockTime().UnixNano()` + a delta as the timeout for ibc transfers 
- Remove `TestTransferDepositRecords_HostBlockHeightNotFound` unittest which we no longer need

## Author's Checklist

I have...

Integration tests
```
```

## Reviewers Checklist

I have...

      - [ ] not applicable
      - [ ] jira ticket `XXX` 
      - [ ] README.md 
      - [ ] not documented


sampocs (approved):
Looks good! One question:

* Are we confident the tendermint time consensus is guaranteed _across chains_ and not just _across nodes_ on the same chain? My understanding was that it's the latter (as consensus is reached through validator votes). 
   * I think it's safe to assume that if each node is in consensus on a single chain using it's machine time, then that implies that each chain is also in sync. The only way this could not be true, is if every machine only validated one chain and one chain happened to be composed of machines that all had the exact same drift haha. Either way, just want to double check/clarify our assumptions here!

x/stakeibc/keeper/deposit_records.go
```diff
@@ -66,23 +68,19 @@ func (k Keeper) TransferExistingDepositsToHostZones(ctx sdk.Context, epochNumber
...
 		transferCoin := sdk.NewCoin(hostZone.GetIBCDenom(), sdk.NewInt(depositRecord.Amount))
-		// QUESTION: Is there a good place to store "tranfer" as a constant?
-		msg := ibctypes.NewMsgTransfer("transfer", hostZone.TransferChannelId, transferCoin, hostZoneModuleAddress, delegateAddress, timeoutHeight, 0)
+		// timeout 30 min in the future
+		// NOTE: this assumes no clock drift between chains, which tendermint guarantees
+		// if we onboard non-tendermint chains, we need to use the time on the host chain to
+		// calculate the timeout
+		// https://github.com/tendermint/tendermint/blob/v0.34.x/spec/consensus/bft-time.md
+		timeoutTimestamp := ctx.BlockTime().UnixNano() + int64(feeTransferTimeoutNanos)
+		msg := ibctypes.NewMsgTransfer(ibctransfertypes.PortID, hostZone.TransferChannelId, transferCoin, hostZoneModuleAddress, delegateAddress, clienttypes.Height{}, uint64(timeoutTimestamp))
```
vish-stride, line 78: _extremely_ nit, so I apologize in advance, but might be better to cast `ctx.BlockTime().UnixNano()` in line 77 to uint64 rather than casting on line 78 (1 cast vs 2 casts)

x/stakeibc/keeper/deposit_records.go
```diff
@@ -39,7 +41,7 @@ func (k Keeper) TransferExistingDepositsToHostZones(ctx sdk.Context, epochNumber
 		return isTransferRecord && isBeforeCurrentEpoch
 	})
 
-	ibcTimeoutBlocks := k.GetParam(ctx, types.KeyIbcTimeoutBlocks)
+	feeTransferTimeoutNanos := k.GetParam(ctx, types.KeyFeeTransferTimeoutNanos)
```
vish-stride, line 44: do you think we should raise the timeout here from 10 minutes? Given epochs are 6 hours, it might okay to make the timeout 30-60m?

vish-stride (approved):
Looks great, and very clever idea on your end to do it this way! 

What a neat feature of tendermint that you can compare block times across chains

x/stakeibc/keeper/deposit_records.go
asalzmann: great call! done

x/stakeibc/keeper/deposit_records.go
asalzmann: definitely! no reason to make this timeout so low. done.

x/stakeibc/keeper/deposit_records.go
```diff
@@ -66,23 +68,19 @@ func (k Keeper) TransferExistingDepositsToHostZones(ctx sdk.Context, epochNumber
...
+		// calculate the timeout
+		// https://github.com/tendermint/tendermint/blob/v0.34.x/spec/consensus/bft-time.md
+		timeoutTimestamp := ctx.BlockTime().UnixNano() + int64(feeTransferTimeoutNanos)
+		msg := ibctypes.NewMsgTransfer(ibctransfertypes.PortID, hostZone.TransferChannelId, transferCoin, hostZoneModuleAddress, delegateAddress, clienttypes.Height{}, uint64(timeoutTimestamp))
 		k.Logger(ctx).Info(fmt.Sprintf("TransferExistingDepositsToHostZones msg %v", msg))
 
 		err := k.RecordsKeeper.Transfer(ctx, msg, depositRecord.Id)
 		if err != nil {
-			k.Logger(ctx).Error(fmt.Sprintf("\t[TransferExistingDepositsToHostZones] ERROR WITH DEPOSIT RECEIPT %s %v %s %s %v", hostZone.TransferChannelId, transferCoin, hostZoneModuleAddress, delegateAddress, timeoutHeight))
+			k.Logger(ctx).Error(fmt.Sprintf("\t[TransferExistingDepositsToHostZones] ERROR WITH DEPOSIT RECEIPT %s %v %s %s %v", hostZone.TransferChannelId, transferCoin, hostZoneModuleAddress, delegateAddress, timeoutTimestamp))
```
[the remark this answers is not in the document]
asalzmann: done

x/stakeibc/keeper/deposit_records.go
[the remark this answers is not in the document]
asalzmann: this is the same timeout we use for our fee IBC transfer (which is triggered via an ICA)

I'll add a separate param

asalzmann:
@sampocs time is voted on as a weighted median in tendermint https://github.com/tendermint/tendermint/blob/v0.34.x/spec/consensus/bft-time.md

malicious validator sets could manipulate time, but this would show up in the light client anyway

PR #249, opened 2022-08-30
make claims more efficient

asalzmann:

Originally, we had considered adding a minimum amount to redeem to prevent long iterations in our callbacks. This takes a different approach, by removing the long iteration alltogether.

Setting `isClaimable` on all `UserRedemptionRecord`s is inefficient and unneccessary. The current behavior is as follows
- aggregate unbondings by epoch, by host zone on host zone unbondings
- each time a user redeems, update the host zone unbonding and create a `UserRedemptionRecord`
- each day, batch transfer unbonded tokens from the delegation ICA to the redemption ICA
- in the ack from this transfer, set all `UserRedemptionRecords` associated with the host zone unbonding to `isClaimable=True`

If we have lots of records, this winds up being an expensive ack. We can simplify this by simply defaulting `isClaimable=True` on all records and adding an additional check before a user claims to verify that the host zone unbonding has been transferred. The checks are now
1) Is `hzu.Status == TRANSFERRED`? If so, that plot of tokens has been transferred and corresponding `UserRedemptionRecord`s are ready to be claimed
2) Is `isClaimable == True`? If so, allow the user to claim their tokens

The reason we still need `isClaimable` is to prevent double claims; previously, `isClaimable` was overloaded and represented both whether tokens had been transferred and whether a claim was currently pending. We've now separated those checks out. To summarize, the reason this approach is more efficient is that we no longer iterate over all `UserRedemptionRecord`s on a host zone unbonding in ICA acks.

Question: Can we remove the user redemption record index on host zone unbondings entirely now?

## Brief Changelog
- Add `chainId` and `epochNumber` to `ClaimCallback`, so that we can fetch the associated host zone unbonding in the callback
- Add `GetAllPreviousEpochUnbondingRecord` helper function
- Default `userRedemptionRecord.IsClaimable=True` when we create them
- Update the logic to remove old `EpochUnbondingRecord`s
    - Update logic in hooks.go to delete _all_ `EpochUnbondingRecord`s from previous epochs without active host zone unbondings (all host zone unbondings have 0 native tokens)
    - Add `DecrementHostZoneUnbonding` to the claim callback
- Add test coverage

## Author's Checklist

I have...

## Reviewers Checklist

I have...

      - [ ] not applicable
      - [ ] jira ticket `XXX` 
      - [ ] README.md 
      - [x] not documented


x/stakeibc/keeper/icacallbacks_claim_test.go
```diff
@@ -71,6 +125,30 @@ func (s *KeeperTestSuite) TestClaimCallback_Successful() {
...
+	s.Require().True(found, "epoch unbonding record found")
+
+	// fetch the hzus
+	hzu1 := epochUnbondingRecord1.HostZoneUnbondings[0]
+	hzu2 := epochUnbondingRecord1.HostZoneUnbondings[1]
+	hzu3 := epochUnbondingRecord2.HostZoneUnbondings[0]
+	hzu4 := epochUnbondingRecord2.HostZoneUnbondings[1]
+
+	// check that hzu1 has a decremented amount
+	s.Require().Equal(hzu1.NativeTokenAmount-tc.initialState.decrementAmount, hzu1.NativeTokenAmount, "hzu1 amount decremented")
```
vish-stride, line 142: am I misunderstanding this test? is it just testing that `tc.initialSTate.decrementAmount` is 0? 

same question with the below tests too!

x/stakeibc/keeper/msg_server_claim_undelegated_tokens.go
```diff
@@ -68,10 +71,23 @@ func (k Keeper) GetClaimableRedemptionRecord(ctx sdk.Context, msg *types.MsgClai
 	}
 
 	// check that the record is claimable
-	if !userRedemptionRecord.GetIsClaimable() {
-		errMsg := fmt.Sprintf("User redemption record %s is not claimable", userRedemptionRecord.Id)
+	hostZoneUnbonding, found := k.RecordsKeeper.GetHostZoneUnbondingByChainId(ctx, userRedemptionRecord.EpochNumber, msg.HostZoneId)
+	if !found {
+		errMsg := fmt.Sprintf("Host zone unbonding record %s not found on host zone %s", userRedemptionRecordKey, msg.HostZoneId)
+		k.Logger(ctx).Error(errMsg)
+		return nil, sdkerrors.Wrapf(types.ErrInvalidUserRedemptionRecord, "could not get host zone unbonding record: %s", userRedemptionRecordKey)
```
vish-stride, line 78: sorry to be very nit, but could we wrap this in fmt.Sprintf ? Another easy option is to just return errMsg

x/stakeibc/keeper/msg_server_claim_undelegated_tokens.go
```diff
@@ -68,10 +71,23 @@ func (k Keeper) GetClaimableRedemptionRecord(ctx sdk.Context, msg *types.MsgClai
...
+	}
+	// records associated with host zone unbondings are claimable after the host zone unbonding tokens have been transferred to the redemption account
+	if hostZoneUnbonding.Status != recordstypes.HostZoneUnbonding_TRANSFERRED {
+		errMsg := fmt.Sprintf("User redemption record %s is not claimable, host zone unbonding has status: %s, requires status TRANSFERRED", userRedemptionRecord.Id, hostZoneUnbonding.Status)
+		k.Logger(ctx).Error(errMsg)
+		return nil, sdkerrors.Wrapf(types.ErrInvalidUserRedemptionRecord, "user redemption record is not claimable: %s, host zone unbonding has status: %s, requires status TRANSFERRED", userRedemptionRecordKey, hostZoneUnbonding.Status)
```
vish-stride, line 84: same, I think we should just return errMsg here

vish-stride (approved):
looks good to me! overall really like this approach, it's clean and scalable (and makes our earlier approach seem silly in comparison)

one thing though, we should make sure the integration tests pass here (I strongly suspect they will, but just to be sure), and then it might also be good to try to redeem from two different accounts, and make sure they can't claim records for the other user. Strongly suspect this will work, but just out of paranoia (this might be covered by firehose already!)

sampocs (commented):
Looks good! Much cleaner than the original!

What do you think about changing the `isClaimable` field to `alreadyClaimed` or something of the like? `isClaimable` now feels a little misleading

x/stakeibc/keeper/unbonding_records.go
```diff
@@ -241,21 +241,24 @@ func (k Keeper) InitiateAllHostZoneUnbondings(ctx sdk.Context, dayNumber uint64)
...
-		shouldDeleteRecord := true
+		shouldDeleteEpochUnbondingRecord := true
 		hostZoneUnbondings := epochUnbondingRecord.GetHostZoneUnbondings()
 		for _, hostZoneUnbonding := range hostZoneUnbondings {
 			k.Logger(ctx).Info(fmt.Sprintf("processing hostZoneUnbonding %v", hostZoneUnbonding))
-			if (hostZoneUnbonding.Status != recordstypes.HostZoneUnbonding_TRANSFERRED) && (hostZoneUnbonding.GetNativeTokenAmount() != 0) {
-				shouldDeleteRecord = false
+			// if an EpochUnbondingRecord has any HostZoneUnbonding with non-zero balances, we don't delete the EpochUnbondingRecord
+			// because it has outstanding tokens that need to be claimed
+			if hostZoneUnbonding.GetNativeTokenAmount() != 0 {
```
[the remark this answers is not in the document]
asalzmann: possible - although anyone can process claims, so theoretically we could clear them out. if this becomes an issue, an easy fix would be to add a min claim amount! (although I don't think it should cause too much state bloat - this state scales linearly with the number of transactions on the chain)

asalzmann:
Thanks for the review! I'll address the comments soon

On the naming, `alreadyClaimed` is also a bit misleading, because, after a successful claim, the record is deleted. I think `claimIsPending` is most accurate, but we'd need to flip all of the boolean logic (`claimIsPending == !isClaimable`) which is a tad annoying

wdyt?

sampocs:
I like `claimIsPending`! Maybe try and do a find & replace on `IsClaimable` and if it looks like it'll take more than a few minutes, just abort it. My gut tells me it wont be too bad!

sampocs:
I do think `claimIsPending` is much stronger than `isClaimable`

x/stakeibc/keeper/msg_server_claim_undelegated_tokens.go
asalzmann: done

x/stakeibc/keeper/msg_server_claim_undelegated_tokens.go
asalzmann: done

x/stakeibc/keeper/icacallbacks_claim_test.go
asalzmann: Good catch! Decrement amount wasn't being set correctly. Updated

x/records/keeper/epoch_unbonding_record.go
```diff
@@ -49,6 +49,24 @@ func (k Keeper) GetAllEpochUnbondingRecord(ctx sdk.Context) (list []types.EpochU
 	return
 }
 
+// GetAllPreviousEpochUnbondingRecord returns all epochUnbondingRecords prior to a given epoch
+func (k Keeper) GetAllPreviousEpochUnbondingRecord(ctx sdk.Context, epochNumber uint64) (list []types.EpochUnbondingRecord) {
```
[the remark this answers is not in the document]
asalzmann: I don't think they're guaranteed to be ordered - added that comment in!

x/stakeibc/keeper/icacallbacks_claim_test.go
```diff
@@ -71,6 +125,30 @@ func (s *KeeperTestSuite) TestClaimCallback_Successful() {
...
+	s.Require().Equal(hzu1.Status, recordtypes.HostZoneUnbonding_TRANSFERRED, "hzu1 status set to transferred")
+	// verify the other hzus are unchanged
+	s.Require().Equal(hzu2.NativeTokenAmount, hzu2.NativeTokenAmount, "hzu2 amount unchanged")
+	s.Require().Equal(hzu2.Status, recordtypes.HostZoneUnbonding_UNBONDED, "hzu2 status set to transferred")
+	s.Require().Equal(hzu3.NativeTokenAmount, hzu3.NativeTokenAmount, "hzu3 amount unchanged")
+	s.Require().Equal(hzu3.Status, recordtypes.HostZoneUnbonding_UNBONDED, "hzu3 status set to transferred")
+	s.Require().Equal(hzu4.NativeTokenAmount, hzu4.NativeTokenAmount, "hzu4 amount unchanged")
+	s.Require().Equal(hzu4.Status, recordtypes.HostZoneUnbonding_UNBONDED, "hzu4 status set to transferred")
```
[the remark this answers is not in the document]
asalzmann: They check that the callback data isn't modifying any other records- I just added them so that if any code in callbacks _does_ start modifying other records, the unittest will fail

x/stakeibc/keeper/msg_server_claim_undelegated_tokens_test.go
```diff
@@ -194,3 +197,22 @@ func (s *KeeperTestSuite) TestClaimUndelegatedTokens_NoEpochTracker() {
...
+	// update the hzu status to not transferred
+	epochUnbondingRecord, found := s.App.RecordsKeeper.GetEpochUnbondingRecord(s.Ctx(), tc.validMsg.Epoch)
+	s.Require().True(found, "epoch unbonding record found")
+	updatedHzu := epochUnbondingRecord.HostZoneUnbondings[0]
+	updatedHzu.Status = recordtypes.HostZoneUnbonding_UNBONDED
+	newEpochUnbondingRecord, success := s.App.RecordsKeeper.AddHostZoneToEpochUnbondingRecord(s.Ctx(), tc.validMsg.Epoch, tc.validMsg.HostZoneId, updatedHzu)
+	s.Require().True(success, "epoch unbonding record updated")
+	s.App.RecordsKeeper.SetEpochUnbondingRecord(s.Ctx(), *newEpochUnbondingRecord)
+
+	s.App.StakeibcKeeper.RemoveEpochTracker(s.Ctx(), epochtypes.STRIDE_EPOCH)
```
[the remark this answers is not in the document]
asalzmann: whoops, removed

x/stakeibc/keeper/msg_server_claim_undelegated_tokens_test.go
```diff
@@ -194,3 +197,22 @@ func (s *KeeperTestSuite) TestClaimUndelegatedTokens_NoEpochTracker() {
 	expectedErr += "Epoch tracker not found for epoch stride_epoch: epoch not found"
 	s.Require().EqualError(err, expectedErr)
 }
+
+func (s *KeeperTestSuite) TestClaimUndelegatedTokens_HzuNotStatusTransferred() {
+	// Remove epoch tracker
```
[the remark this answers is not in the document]
asalzmann: removed

x/records/keeper/epoch_unbonding_record.go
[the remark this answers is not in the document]
asalzmann: done

x/stakeibc/keeper/icacallbacks_claim_test.go
```diff
@@ -30,28 +32,80 @@ type ClaimCallbackTestCase struct {
...
+	s.App.RecordsKeeper.SetUserRedemptionRecord(s.Ctx(), userRedemptionRecord3)
+	// the hzu that we'll claim from
+	hostZoneUnbonding1 := recordtypes.HostZoneUnbonding{
+		HostZoneId:            HostChainId,
+		Status:                recordtypes.HostZoneUnbonding_TRANSFERRED,
+		UserRedemptionRecords: []string{recordId1, recordId2},
+		NativeTokenAmount:     uint64(1_000_000),
+	}
+	hostZoneUnbonding2 := recordtypes.HostZoneUnbonding{
+		HostZoneId:            HostChainId,
```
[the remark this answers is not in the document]
asalzmann: good catch, done

x/stakeibc/keeper/msg_server_claim_undelegated_tokens.go
```diff
@@ -28,7 +28,8 @@ func (k msgServer) ClaimUndelegatedTokens(goCtx context.Context, msg *types.MsgC
 	k.Logger(ctx).Info(fmt.Sprintf("ClaimUndelegatedTokens %v", msg))
 	userRedemptionRecord, err := k.GetClaimableRedemptionRecord(ctx, msg)
 	if err != nil {
-		return nil, sdkerrors.Wrapf(err, "unable to find claimable redemption record")
+		k.Logger(ctx).Error("unable to find claimable redemption record")
+		return nil, err
```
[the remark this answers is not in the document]
asalzmann: done

asalzmann:
Thanks both for the comments!

@sampocs I flipped `isClaimable` to `claimIsPending`, can you please take a look at https://github.com/Stride-Labs/stride/pull/249/commits/98689ddacc494c35c0aa973649384a7b04edf1a8?

x/stakeibc/keeper/icacallbacks_claim_test.go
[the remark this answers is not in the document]
sampocs: I think it's better to keep the tests lean now and extend them if those fields ever became relevant in the calling function. Imo, this isn't that different than checking a deposit record field here or any other element from the state that isn't touched by this function atm.

Not to make too big a deal about this very nit thing haha, but I do think keeping the tests focused to the function in question is an important principle with these unit tests. Wdyt?

x/stakeibc/keeper/msg_server_claim_undelegated_tokens.go
```diff
@@ -84,8 +84,8 @@ func (k Keeper) GetClaimableRedemptionRecord(ctx sdk.Context, msg *types.MsgClai
 		k.Logger(ctx).Error(errMsg)
 		return nil, sdkerrors.Wrapf(types.ErrInvalidUserRedemptionRecord, errMsg)
 	}
-	// records that have isClaimable set to False have already been claimed (and are pending an ack)
-	if !userRedemptionRecord.IsClaimable {
+	// records that have claimIsPending set to False have already been claimed (and are pending an ack)
```
sampocs, line 87: "set to **True**"

x/stakeibc/keeper/msg_server_claim_undelegated_tokens_test.go
```diff
@@ -115,7 +115,7 @@ func (s *KeeperTestSuite) TestClaimUndelegatedTokens_Successful() {
 
 	actualRedemptionRecord, found := s.App.RecordsKeeper.GetUserRedemptionRecord(s.Ctx(), redemptionRecordId)
 	s.Require().True(found, "redemption record found")
-	s.Require().False(actualRedemptionRecord.IsClaimable, "redemption record should not be claimable")
+	s.Require().True(actualRedemptionRecord.claimIsPending, "redemption record should not be claimable")
```
sampocs: "redemption record should be pending"

x/stakeibc/keeper/msg_server_redeem_stake.go
```diff
@@ -94,8 +94,8 @@ func (k msgServer) RedeemStake(goCtx context.Context, msg *types.MsgRedeemStake)
 		Denom:       hostZone.HostDenom,
 		HostZoneId:  hostZone.ChainId,
 		EpochNumber: epochTracker.EpochNumber,
-		// isClaimable represents whether a redemption is claimable, contingent on the host zone unbonding having status TRANSFERRED
-		IsClaimable: true,
+		// claimIsPending represents whether a redemption is claimable, contingent on the host zone unbonding having status TRANSFERRED
```
sampocs, line 97: nit: comment

sampocs:
Looks good! Just need to adjust a couple comments

x/stakeibc/keeper/msg_server_claim_undelegated_tokens.go
asalzmann: great catch

x/stakeibc/keeper/msg_server_claim_undelegated_tokens_test.go
asalzmann: done

x/stakeibc/keeper/msg_server_redeem_stake_test.go
```diff
@@ -153,8 +153,8 @@ func (s *KeeperTestSuite) TestRedeemStake_Successful() {
 	s.Require().Equal(msg.Receiver, userRedemptionRecord.Receiver, "redemption record receiver")
 	// check host zone
 	s.Require().Equal(msg.HostZone, userRedemptionRecord.HostZoneId, "redemption record host zone")
-	// check isClaimable
-	s.Require().True(userRedemptionRecord.IsClaimable, "redemption record should be marked claimable")
+	// check claimIsPending
+	s.Require().False(userRedemptionRecord.claimIsPending, "redemption record should be marked claimable")
```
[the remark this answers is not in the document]
asalzmann: done

x/stakeibc/keeper/msg_server_redeem_stake.go
asalzmann: done

asalzmann:
```
gaia_tests.bats
 ✓ [INTEGRATION-BASIC] host zones successfully registered
 ✓ [INTEGRATION-BASIC-GAIA] liquid stake mints stATOM
 ✓ [INTEGRATION-BASIC-GAIA] tokens were transferred to GAIA after liquid staking

10 tests, 0 failures
```

PR #254, opened 2022-09-01
[SAFETY] Add RateLimiting to BeginBlocker

vish-stride:
This short PR adds rate limiting to the BeginBlocker

Main question is: should we add any other checks here? I think redemption rate is the most critical, but if anything else comes to mind, we should add it in. 

Other question: I left the invariants.go file in, in case we want to add epochly invariants in the future. Open to removing this if yall think it's cleaner. 

## Brief Changelog

- Removed functionality from stakeibc/keeper/invariants.go
- Added RR check in BeginBlocker
- Modified RR params

## Author's Checklist

I have...

## Reviewers Checklist

I have...

      - [ ] not applicable
      - [ ] jira ticket `XXX` 
      - [ ] README.md 
      - [ ] not documented


x/stakeibc/abci.go
```diff
@@ -14,4 +15,12 @@ import (
 // BeginBlocker of stakeibc module
 func BeginBlocker(ctx sdk.Context, k keeper.Keeper, bk types.BankKeeper, ak types.AccountKeeper) {
 	defer telemetry.ModuleMeasureSince(types.ModuleName, time.Now(), telemetry.MetricKeyBeginBlocker)
+
+	// Iterate over all host zones and verify redemption rate
+	for _, hz := range k.GetAllHostZone(ctx) {
+		rrSafe, _ := k.IsRedemptionRateWithinSafetyBounds(ctx, hz)
```
riley-stride, line 21: can we populate the err and flow it into the log below? will help debug in case this ever triggers

riley-stride (changes requested):
Looks good! Two nit things before merging this
1. Can you revert `app.go:363`'s 2nd arg from `100` to `invCheckPeriod`? We don't need frequent invariant checks now
2. Can you do a quick sanity check that your changes work, by setting the default exchRate in registerHostZone to above and below the bounds?

x/stakeibc/types/params.go
```diff
@@ -23,7 +23,7 @@ var (
 	DefaultIbcTimeoutBlocks                 uint64 = 300           // 300 blocks ~= 30 minutes
 	DefaultFeeTransferTimeoutNanos          uint64 = 1800000000000 // 30 minutes
 	DefaultSafetyMinRedemptionRateThreshold uint64 = 90            // divide by 100, so 90 = 0.9
-	DefaultSafetyMaxRedemptionRateThreshold uint64 = 300           // divide by 100, so 300 = 3
+	DefaultSafetyMaxRedemptionRateThreshold uint64 = 150           // divide by 100, so 150 = 1.5
```
asalzmann, line 26: This is a good first measure I think, two possible issues here
- we might hit this pretty quickly for high inflation chains (~1-2 months for chains like Evmos)
- there might be a real scenario where this changes for a chain, e.g. Terra hyperinflation - had we been supporting Terra, would we have wanted to halt the chain? I think probably not, but maybe we would have wanted to halt functionality for Terra

In one of our first upgrades, I think we should rework this param so that it's defined chain by chain, and make it a dynamic range that changes through time (i.e. define an inflation param for a chain up front and re-compute the bounds every n blocks)

x/stakeibc/abci.go
```diff
@@ -14,4 +15,12 @@ import (
 // BeginBlocker of stakeibc module
 func BeginBlocker(ctx sdk.Context, k keeper.Keeper, bk types.BankKeeper, ak types.AccountKeeper) {
 	defer telemetry.ModuleMeasureSince(types.ModuleName, time.Now(), telemetry.MetricKeyBeginBlocker)
+
+	// Iterate over all host zones and verify redemption rate
+	for _, hz := range k.GetAllHostZone(ctx) {
```
asalzmann, line 20: nit: rm

asalzmann (approved):
Nice! I do think this introduces some tech debt, and we'll need to rework this shortly after launch, but this should give us some additional certainty around launch!

x/stakeibc/abci.go
riley-stride: discussed offline, keeping

riley-stride (approved):

Issue #275, opened 2022-09-13
more composable/interoperable protos


pyramation:
Hello! I'm working on tooling to generate clients, and here to help with packaging protos in a canonical way that other chains are doing

For example,

- [./proto/stargaze](https://github.com/public-awesome/stargaze/tree/main/proto/stargaze)
- [./proto/osmosis](https://github.com/osmosis-labs/osmosis/tree/main/proto/osmosis)
- [./proto/akash](https://github.com/ovrclk/akash/tree/master/proto/akash)
- [./proto/cosmos](https://github.com/cosmos/cosmos-sdk/tree/main/proto)
- [./proto/juno](https://github.com/CosmosContracts/juno/tree/main/proto)

To make this more interoperable, can we move 

```
proto/epochs            -> proto/stride/epochs
proto/icacallbacks      -> proto/stride/icacallbacks
proto/interchainquery   -> proto/stride/interchainquery
proto/mint              -> proto/stride/mint
proto/records           -> proto/stride/records
proto/stakeibc          -> proto/stride/stakeibc
```

It looks like a pretty small change, but would be HUGE in terms of interoperability.

You may also want to consider the package names also, I think there are two packages: `stride` and `stridelabs`, not sure if it's good to consolidate or not depending on the use case.

cc @nghuyenthevinh2000

pyramation:
Regarding package names, this type of thing will actually cause naming collisions (one is lowercase and one is pascal case, but both are global):

```proto
package stridelabs.stride.icacallbacks;
package Stridelabs.stride.epochs;
```

this is the source of https://github.com/cosmology-tech/stridejs/issues/1

pyramation:
so let's look at a few packages and their names:

```proto
package Stridelabs.stride.stakeibc;
package stride.mint.v1beta1;
package stride.interchainquery;
package stridelabs.stride.icacallbacks;
package Stridelabs.stride.epochs;
```

May I suggest we do something like this:

```proto
package stride.stakeibc;
package stride.mint.v1beta1;
package stride.interchainquery;
package stride.icacallbacks;
package stride.epochs;
```

nghuyenthevinh2000:
I have code for it but it is in my local. I am debugging script for buf generation.

pyramation:
> I have code for it but it is in my local. I am debugging script for buf generation.

please see my PR https://github.com/Stride-Labs/stride/pull/282

do you think you can include the same changes I did for LCD endpoints? Those were also affected by using starport...

[...]

PR #282, opened 2022-09-18
update Protos and LCD endpoints

pyramation:
Closes: #275 

## the issue

The current protos create multiple global namespaces, which is harder to organize (and also deal with APIs, codegen, documentation etc)

Three are for stride:

- `Stridelabs`
- `stridelabs` (which conflicts with the above
- `stride`

And then the rest:

- `epochs`
- `icacallabacks`
- `interchainquery`
- `mint`
- `records`
- `stakeibc`

## the solution

This solution organizes all the protobuffers into a nice `stride` module

- [x] move all the protos into a single `stride` root module 
- [x] clean up LCD URLs
- [x] update package paths inside of proto files 

## TODO

I'm not a Go dev, so this PR likely needs review and possible more requirements before it could get merged

cc @nghuyenthevinh2000


pyramation:
Do the go packages have to move? I'm starting to think not necessary.

I just realized maybe these (as well as the `/x` folder) can stay where they are, it's just the protos and proto imports that need to be updated.


e.g.

```
option go_package = "github.com/Stride-Labs/stride/x/stride/stakeibc/types";
```

even if we update protos, could stay as this?

```
option go_package = "github.com/Stride-Labs/stride/x/stakeibc/types";
```

nghuyenthevinh2000:
> Do the go packages have to move? I'm starting to think not necessary.
> 
> I just realized maybe these (as well as the `/x` folder) can stay where they are, it's just the protos and proto imports that need to be updated.
> 
> e.g.
> 
> ```
> option go_package = "github.com/Stride-Labs/stride/x/stride/stakeibc/types";
> ```
> 
> even if we update protos, could stay as this?
> 
> ```
> option go_package = "github.com/Stride-Labs/stride/x/stakeibc/types";
> ```

no need to change that go_package. It can still generate proto to correct x folders.

PR #283, opened 2022-09-19
Protobuf rework

nghuyenthevinh2000:
Closes: #275 

1. Change proto folder structure to comply to standard
2. Add newest protobuf generation, lint, format script
3. Linting

This change will allow developers to build application more easily on Stride.

WILL PRODUCE BREAKING CHANGES

## Author's Checklist

I have...

- [x] OR this change is a trivial rework / code cleanup without any test coverage

## Reviewers Checklist

I have...

      - [ ] not applicable
      - [ ] jira ticket `XXX` 
      - [ ] README.md 
      - [ ] not documented


nghuyenthevinh2000:
@asalzmann
@sampocs

Can you have a look at this pr?

Makefile
```diff
@@ -147,3 +149,28 @@ stop-local:
...
+
+containerProtoVer=v0.7
+containerProtoImage=tendermintdev/sdk-proto-gen:$(containerProtoVer)
+containerProtoGen=cosmos-sdk-proto-gen-$(containerProtoVer)
+containerProtoGenSwagger=cosmos-sdk-proto-gen-swagger-$(containerProtoVer)
+containerProtoFmt=cosmos-sdk-proto-fmt-$(containerProtoVer)
+
+proto-all: proto-format proto-lint proto-gen
+
+proto-gen:
```
asalzmann, line 164: Thanks for adding this - we were previously using the ignite cli. This seems much cleaner.

go.mod
```diff
@@ -4,7 +4,8 @@ go 1.18
 
 require (
 	github.com/cosmos/cosmos-proto v1.0.0-alpha7
-	github.com/cosmos/cosmos-sdk v0.45.5
+	github.com/cosmos/cosmos-sdk v0.46.1
```
asalzmann, line 7: Not sure we can switch to 46 here

asalzmann (commented):
lgtm! thank you for tackling this large PR, this is super helpful. I need to verify that we can upgrade to 46 in go.mod, otherwise looks good!

asalzmann:
I was chatting with @sampocs, and the one change I think we need to make here is to revert the enum changes. We have a few external services that rely on the enum names as they exist (those we can update, so if you think the enum names in this PR are significantly cleaner, we can keep them). The string of the ICA Account type enum however is used to construct the ICA, so we can't change those, because if an ICA channel closes (see: [ORDERED channels](enum ChannelOrder {
  ORDERED,
  UNORDERED,
  ORDERED_ALLOW_TIMEOUT,
})), we need to re-register the ICA with the same name to restore it.

go.mod
nghuyenthevinh2000: oh, it just jumps when I do go mod tidy, will change it back to 0.45.5

nghuyenthevinh2000:
> I was chatting with @sampocs, and the one change I think we need to make here is to revert the enum changes. We have a few external services that rely on the enum names as they exist (those we can update, so if you think the enum names in this PR are significantly cleaner, we can keep them). The string of the ICA Account type enum however is used to construct the ICA, so we can't change those, because if an ICA channel closes (see: [ORDERED channels](https://github.com/cosmos/ibc/blob/main/spec/core/ics-004-channel-and-packet-semantics/README.md)), we need to re-register the ICA with the same name to restore it.

If so, I will change it back. The name is not that important.

go.mod
asalzmann: Ah I see. I'll be upgrading this to 0.45.8 soon, which should fix this

sampocs:
Hey @nghuyenthevinh2000, thanks for taking care of this and apologies for jumping in late here!

Would it be possible to separate out the generated go files into a separate PR? Considering this is a large PR and there might be many hidden breaking changes, it'd be nice to first merge in just the formatted/restructured proto files and the makefile commands. And then we can generate the go files in a new PR and review that with a little more scrutiny.

Also can you rebase this branch?

nghuyenthevinh2000:
> Hey @nghuyenthevinh2000, thanks for taking care of this and apologies for jumping in late here!
> 
> Would it be possible to separate out the generated go files into a separate PR? Considering this is a large PR and there might be many hidden breaking changes, it'd be nice to first merge in just the formatted/restructured proto files and the makefile commands. And then we can generate the go files in a new PR and review that with a little more scrutiny.
> 
> Also can you rebase this branch? 

Sure, OK, let's me do it

nghuyenthevinh2000:
I have added proto files only in #328

PR #299, opened 2022-10-11
Validator Selection Governance v0

riley-stride:
Closes: #XXX

## Context and purpose of the change

Add governance routes and handlers for host zone validator selection. Allows STRD holders to vote to add or remove a validator from the host set on a host zone. New validators will have weight equal to the min weight of existing validators in that host zone set. 

## Brief Changelog

- Add gov routes and create + register proposal types for `AddValidatorProposal` and `DeleteValidatorProposal`
- When `DeleteValidatorProposal`s are executed, we refrain from deleting the validator, rather we set its weight to 0 (for protocol safety)
- When `AddValidatorProposal`s are executed, we 
    - Abort if the host zone already has a full validator set (based on new stakeibc module param `SafetyMaxValidators`). This is a temporary limit to allow batched ICA messages to safely fit into a single block on host zones with a restrictive max gas level (e.g. anything under ~20M). 
    - Set the weight to the min weight of the existing validators on the host zone

## Author's Checklist

I have...

- [ ] Run and PASSED locally all GAIA integration tests
- [ ] If the change is contentful, I either:
    - [ ] Added a new unit test OR 
    - [ ] Added test cases to existing unit tests
- [ ] OR this change is a trivial rework / code cleanup without any test coverage

If skipped any of the tests above, explain.

## Reviewers Checklist

*All items are required. Please add a note if the item is not applicable and please add
your handle next to the items reviewed if you only reviewed selected items.*

I have...

- [ ] reviewed state machine logic
- [ ] reviewed API design and naming
- [ ] manually tested (if applicable)
- [ ] confirmed the author wrote unit tests for new logic
- [ ] reviewed documentation exists and is accurate

## Documentation and Release Note

  - [ ] Does this pull request introduce a new feature or user-facing behavior changes? 
  - [ ] Is a relevant changelog entry added to the `Unreleased` section in `CHANGELOG.md`?
  - [ ] This pull request updates existing proto field values (and require a backend and frontend migration)? 
  - [ ] Does this pull request change existing proto field names (and require a frontend migration)?
  How is the feature or change documented? 
      - [ ] not applicable
      - [ ] jira ticket `XXX` 
      - [ ] specification (`x/<module>/spec/`) 
      - [ ] README.md 
      - [ ] not documented

PR #305, opened 2022-10-18
Add Validator through Governance

sampocs:
Closes: #XXX

**Most of this is ported over from Riley's [PR](https://github.com/Stride-Labs/stride/pull/299) - thank you for doing the majority of the work!**

Add governance routes and handlers for host zone validator selection. Allows STRD holders to vote to add a validator from the host set on a host zone. New validators will have weight equal to the min weight of existing validators in that host zone set. 

We decided to not include validator removal as a gov function since the actual implementation would only consist of a) checking that there were no delegations and b) then set the weight to 0. However, the only way for there to be no delegations is if the weight is already 0.

## Brief Changelog

- Add gov routes and create + register proposal type for `AddValidatorProposal` 
- When `AddValidatorProposal`s are executed, we 
    - Abort if the host zone already has a full validator set (based on new stakeibc module param `SafetyMaxValidators`). This is a temporary limit to allow batched ICA messages to safely fit into a single block on host zones with a restrictive max gas level (e.g. anything under ~20M). 
    - Set the weight to the min weight of the existing validators on the host zone, excluding zero-weight validators
- When changing validator weights, fail any zero => nonzero weight changes that increase the non-zero-wgt valset size above `SafetyMaxValidators`

### Setup
* Disable all hosts but GAIA in dockernet (remove other hosts from array in `scripts/start_network.sh`
* Set the max number of validators to 5 [here](https://github.com/Stride-Labs/stride/pull/305/files#diff-50de9d0087b725c2a5f237cb915237a29606d28304b2a56803fca3d8c1dbf403R29)
* Start dockernet
```bash
make start-docker build=sghr
```
* Confirm there are 2 validator's present
```bash
```
### Add Validators Success
* Add a validator through admin (with weight 3)
```bash
strided tx stakeibc add-validator GAIA val-from-admin cosmosvaloper196ax4vc0lwpxndu9dyhvca7jhxp70rmcvrj90c 10 3 --from admin 
```
* Confirm validator was added
```bash
```
* Create an add-validator proposal file `proposal.json` in your current directory. You will modify this file for the following governance add-validator proposals in this testing run. The contents of `proposal.json` are:
```
{
    "description": "Proposal to add Imperator because they contribute in XYZ ways!",
    "hostZone": "GAIA",
    "validatorName": "Imperator",
    "validatorAddress": "cosmosvaloper1v5y0tg0jllvxf5c3afml8s3awue0ymju89frut",
    "deposit": "64000000ustrd"
}
```
* Add a validator through governance (**you have 30 seconds to get the votes in!!**)
```bash
strided tx gov submit-proposal add-validator proposal.json --from val1 
```
* After the voting period has passed (30 seconds), confirm the validator was added. Note that the weight is equal to 3 (the lowest weight)
```bash
```
### Add Validators Failure
* Rerun the same add validator functions and query the transaction hash after - they should both fail with an error saying the validator already exists
```bash
strided tx stakeibc add-validator GAIA val-from-admin cosmosvaloper196ax4vc0lwpxndu9dyhvca7jhxp70rmcvrj90c 10 3 --from admin 
strided tx gov submit-proposal add-validator proposal.json 10000001ustrd --from val1 
```
* Try to add a validator with a new address, but existing name. They should both fail with an error saying the name already exists. Change the address in `proposal.json` to `cosmosvaloper1grgelyng2v6v3t8z87wu3sxgt9m5s03xfytvz7` but leave the name unchanged.
```
strided tx stakeibc add-validator GAIA val-from-admin cosmosvaloper1grgelyng2v6v3t8z87wu3sxgt9m5s03xfytvz7 10 3 --from admin 
strided tx gov submit-proposal add-validator proposal.json --from val1 
```
* Try with a host zone that does not exist - it should also fail
```
strided tx stakeibc add-validator FAKE val-from-admin cosmosvaloper196ax4vc0lwpxndu9dyhvca7jhxp70rmcvrj90c 10 3 --from admin 
```
* Finally, test that we can't add more than 5 validators
```bash
# add a 5th validator 
strided tx stakeibc add-validator GAIA numba5 cosmosvaloper1ey69r37gfxvxg62sh4r0ktpuc46pzjrm873ae8 10 5 --from admin 

### Check that we can't add a 6th (query the tx hash after to confirm failure)
Set the name of the validator in `proposal.json` to `no-way-jose` and the address to `cosmosvaloper1tflk30mq5vgqjdly92kkhhq3raev2hnz6eete3`. Then:
strided tx stakeibc add-validator GAIA no-way-jose cosmosvaloper1tflk30mq5vgqjdly92kkhhq3raev2hnz6eete3 10 5 --from admin 
strided tx gov submit-proposal add-validator proposal.json --from val1 
```

### Check that when adding a new validator, the weight takes the min existing weight, excluding zero-weight validators
* Add a validator through admin (with weight 0)
```bash
strided tx stakeibc add-validator GAIA val-wgt-0 cosmosvaloper1pcag0cj4ttxg8l7pcg0q4ksuglswuuedadj7ne 10 0 --from admin 
```
* Confirm validator was added
```bash
```
* Add a validator through governance (**you have 30 seconds to get the votes in!!**)
* Set the `proposal.json` name to `val-from-governance-2` and address to `cosmosvaloper133lfs9gcpxqj6er3kx605e3v9lqp2pg5syhvsz`. Then:
```bash
strided tx gov submit-proposal add-validator proposal.json --from val1 
```
* After the voting period has passed (30 seconds), confirm the validator was added and has non-zero weight (the lowest non-zero weight)
```bash
```

## Author's Checklist
* Integration tests
```
gaia_tests.bats
 ✓ [INTEGRATION-BASIC] host zones successfully registered
 ✓ [INTEGRATION-BASIC-GAIA] liquid stake mints stATOM
 ✓ [INTEGRATION-BASIC-GAIA] tokens were transferred to GAIA after liquid staking

10 tests, 0 failures

juno_tests.bats
 ✓ [INTEGRATION-BASIC] host zones successfully registered
 ✓ [INTEGRATION-BASIC-JUNO] ibc transfer updates all balances
 ✓ [INTEGRATION-BASIC-JUNO] liquid stake mints stJUNO
 ✓ [INTEGRATION-BASIC-JUNO] tokens were transferred to JUNO after liquid staking
 ✓ [INTEGRATION-BASIC-JUNO] tokens on JUNO were staked
 ✓ [INTEGRATION-BASIC-JUNO] redemption works
 ✓ [INTEGRATION-BASIC-JUNO] claimed tokens are properly distributed
 ✓ [INTEGRATION-BASIC-JUNO] rewards are being reinvested, exchange rate updating

10 tests, 0 failures

osmo_tests.bats
 ✓ [INTEGRATION-BASIC] host zones successfully registered
 ✓ [INTEGRATION-BASIC-OSMO] ibc transfer updates all balances
 ✓ [INTEGRATION-BASIC-OSMO] liquid stake mints stOSMO
 ✓ [INTEGRATION-BASIC-OSMO] tokens were transferred to OSMO after liquid staking
 ✓ [INTEGRATION-BASIC-OSMO] tokens on OSMO were staked
 ✓ [INTEGRATION-BASIC-OSMO] redemption works
 ✓ [INTEGRATION-BASIC-OSMO] claimed tokens are properly distributed
 ✓ [INTEGRATION-BASIC-OSMO] rewards are being reinvested, exchange rate updating

10 tests, 0 failures

stars_tests.bats
 ✓ [INTEGRATION-BASIC] host zones successfully registered
 ✓ [INTEGRATION-BASIC-STARS] ibc transfer updates all balances
 ✓ [INTEGRATION-BASIC-STARS] liquid stake mints stSTARS
 ✓ [INTEGRATION-BASIC-STARS] tokens were transferred to STARS after liquid staking
 ✓ [INTEGRATION-BASIC-STARS] tokens on STARS were staked
 ✓ [INTEGRATION-BASIC-STARS] redemption works
 ✓ [INTEGRATION-BASIC-STARS] claimed tokens are properly distributed
 ✓ [INTEGRATION-BASIC-STARS] rewards are being reinvested, exchange rate updating

9 tests, 0 failures
```

## Reviewers Checklist

I have...

      - [ ] not applicable
      - [ ] jira ticket `XXX` 
      - [ ] README.md 
      - [ ] not documented


app/app.go
```diff
@@ -142,6 +143,7 @@ func getGovProposalHandlers() []govclient.ProposalHandler {
 		upgradeclient.CancelProposalHandler,
 		ibcclientclient.UpdateClientProposalHandler,
 		ibcclientclient.UpgradeProposalHandler,
+		stakeibcclient.AddValidatorProposalHandler,
```
vish-stride, line 146: Should we add a GovProposal to change validator weight? 

This isn't necessarily urgent, and we can easily structure it in a way where this is done through signaling proposals for now, and then the foundation takes care of executing the messages. But, it is something to think about imo

x/stakeibc/client/rest/tx.go
```diff
@@ -0,0 +1,21 @@
+package rest
+
+import (
+	"net/http"
+
+	"github.com/cosmos/cosmos-sdk/client"
+	govrest "github.com/cosmos/cosmos-sdk/x/gov/client/rest"
+)
+
+// TODO: Do we need to implement these?
```
vish-stride, line 10: I'm assuming the answer is no?

x/stakeibc/keeper/host_zone.go
```diff
@@ -124,6 +125,69 @@ func (k Keeper) AddDelegationToValidator(ctx sdk.Context, hostZone types.HostZon
...
+	}
+
+	// Get max number of validators and confirm we won't exceed it
+	maxNumVals, err := cast.ToIntE(k.GetParam(ctx, types.KeySafetyNumValidators))
+	if err != nil {
+		errMsg := fmt.Sprintf("Error getting safety max num validators | err: %s", err.Error())
+		k.Logger(ctx).Error(errMsg)
+		return sdkerrors.Wrap(types.ErrMaxNumValidators, errMsg)
+	}
+	if len(hostZone.Validators) >= maxNumVals {
```
vish-stride, line 144: In "maxNumVals" do we only care about validators with 0 weight? If so, we should change this check! 

The specific use-case I'm imagining is we change one validator's weight to 0 so we can add another. As far as I know, we shouldn't run into any gas issues if this is the case

x/stakeibc/types/gov.go
```diff
@@ -0,0 +1,67 @@
...
+	return ProposalTypeAddValidator
+}
+
+func (p *AddValidatorProposal) ValidateBasic() error {
+	err := govtypes.ValidateAbstract(p)
+	if err != nil {
+		return err
+	}
+
+	// TODO: Add more checks?
```
vish-stride, line 48: not sure what else we would add? maybe verify that "valoper" shows up in validator address?

x/stakeibc/types/params.go
```diff
@@ -26,6 +26,7 @@ var (
 	DefaultSafetyMaxRedemptionRateThreshold uint64 = 150           // divide by 100, so 150 = 1.5
 	DefaultMaxStakeICACallsPerEpoch         uint64 = 100
 	DefaultIBCTransferTimeoutNanos          uint64 = 1800000000000 // 30 minutes
+	DefaultSafetyNumValidators              uint64 = 30
```
vish-stride, line 29: iirc we're currently at 35 for the Hub!

vish-stride (approved):
Overall looks great! 2 main comments
1. Should we add proposal functionality to change weight?
2. I think we should only count non-zero-weight validators when counting validators. This change would also require us to add a check on the ChangeValidatorWeight function to make sure we don't exceed the max

x/stakeibc/keeper/host_zone.go
sampocs: great point! We'll also want to ignore 0 weights when finding the min weight!

app/app.go
riley-stride: resolved offline

x/stakeibc/keeper/host_zone.go
riley-stride: addressed

x/stakeibc/types/params.go
riley-stride: I count 32. Good catch. Setting this to 35

x/stakeibc/keeper/keeper.go
```diff
@@ -260,3 +260,35 @@ func (k Keeper) IsRedemptionRateWithinSafetyBounds(ctx sdk.Context, zone types.H
...
+	// count up the number of validators with non-zero weights
+	numNonzeroWgtValidators := 0
+	for _, validator := range validators {
+		if validator.Weight > 0 {
+			numNonzeroWgtValidators++
+		}
+	}
+
+	// check if the number of validators with non-zero weights is below than the max
+	if numNonzeroWgtValidators >= maxNumVals {
```
sampocs, line 287: Since an error is thrown here, there's no need to return a boolean as well. And if you return only an error, then the calling functions only have to check if `err != nil` (i.e., no need to check the `hasSpace` variable as well)

My suggestion would be to remove the boolean and then call this something like `ConfirmValSetHasSpace`

sampocs (commented):
thanks for taking care of these!

riley-stride (approved):
LGMT! Please test my added logic using test in the docs. Main Q is whether we should change the submission to JSON style for commonwealth

x/stakeibc/keeper/keeper.go
riley-stride: Much cleaner approach, nice suggestion! Done.

x/stakeibc/client/rest/tx.go
sampocs: I'm in the camp of keeping it out unless there's an obvious reason why we need this. I saw other repo's did not implement it beyond this

x/stakeibc/client/cli/tx_add_validator_proposal.go
```diff
@@ -0,0 +1,62 @@
...
+	govtypes "github.com/cosmos/cosmos-sdk/x/gov/types"
+	"github.com/spf13/cobra"
+
+	"github.com/Stride-Labs/stride/x/stakeibc/types"
+)
+
+func CmdAddValidatorProposal() *cobra.Command {
+
+	cmd := &cobra.Command{
+		Use:   "add-validator [host-zone] [name] [address] [deposit]",
```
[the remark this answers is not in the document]
riley-stride: changed to json

x/stakeibc/client/cli/tx_add_validator_proposal.go
```diff
@@ -2,50 +2,83 @@ package cli
...
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	"github.com/cosmos/cosmos-sdk/version"
+	govcli "github.com/cosmos/cosmos-sdk/x/gov/client/cli"
 	govtypes "github.com/cosmos/cosmos-sdk/x/gov/types"
 	"github.com/spf13/cobra"
 
 	"github.com/Stride-Labs/stride/x/stakeibc/types"
 )
 
+func parseAddValidatorProposalWithDepositWithDeposit(cdc codec.JSONCodec, proposalFile string) (types.AddValidatorProposalWithDeposit, error) {
```
sampocs, line 21:
```suggestion
func parseAddValidatorProposalFile(cdc codec.JSONCodec, proposalFile string) (types.AddValidatorProposalWithDeposit, error) {
```

proto/stakeibc/gov.proto
```diff
@@ -15,3 +15,19 @@ message AddValidatorProposal {
   string validator_name = 4; 
   string validator_address = 5 [ (cosmos_proto.scalar) = "cosmos.AddressString" ];
 }
+
+
+message AddValidatorProposalWithDeposit {
```
sampocs, line 20: Can we remove the old struct above and rename this to `AddValidatorProposal`?

riley-stride: We need both (the one without deposit is used in line 81). This is following convention from community-pool-spend proposals

x/stakeibc/client/cli/tx_add_validator_proposal.go
riley-stride: done

PR #313, opened 2022-10-21
change to utils

asalzmann:
Test the admin change github action

changing utils
Screen Shot 2022-10-21 at 8 38 27 AM
.go only, the check passes

the check fails once we change admins.go
Screen Shot 2022-10-21 at 8 42 00 AM


asalzmann:
@sampocs no action required, just fyi

PR #330, opened 2022-11-07
Dockernet Slash Test Configuration

sampocs:
Closes: #XXX

We often need to test out functionality related to slashing (e.g. updating validator weights after a slash). This PR adds configuration settings to dockernet to harshen parameters related to slashing, enabling a quicker dev cycle.

## Brief Changelog
* Adds a boolean variable to `vars.sh` called `HARSHEN_SLASHING`, which, if enabled to true:
  * Reduces the number of missed blocks before a slash
  * Reduces the jail duration
  * Increases the magnitude of the slash
* **Question for reviewer**: Would it make more sense to just always use these settings (rather than making them configurable)? There's probably irrelevant normally since dockernet validators don't miss blocks unless you pause the process

## Author's Checklist

I have...

## Reviewers Checklist

I have...

      - [ ] not applicable
      - [ ] jira ticket `XXX` 
      - [ ] README.md 
      - [x] not documented


scripts/vars.sh
```diff
@@ -65,6 +65,7 @@ HOST_WEEK_EPOCH_DURATION="60s"
 UNBONDING_TIME="120s"
 MAX_DEPOSIT_PERIOD="30s"
 VOTING_PERIOD="30s"
+HARSHEN_SLASHING=false
```
asalzmann, line 68: can you add a slightly longer comment describing what this does?

asalzmann (approved):

scripts/init_chain.sh
```diff
@@ -56,6 +56,14 @@ set_host_genesis() {
     jq "del(.app_state.interchain_accounts)" $genesis_config > json.tmp && mv json.tmp $genesis_config
     interchain_accts=$(cat $SCRIPT_DIR/config/ica.json)
     jq ".app_state += $interchain_accts" $genesis_config > json.tmp && mv json.tmp $genesis_config
+
+    # Optionally harshen slashing parameters
+    if [[ "$HARSHEN_SLASHING" == "true" ]]; then
+        sed -i -E 's|"signed_blocks_window": "100"|"signed_blocks_window": "10"|g' $genesis_config
+        sed -i -E 's|"min_signed_per_window": ""0.500000000000000000"|"min_signed_per_window": ""1.000000000000000000"|g' $genesis_config
```
vish-stride, line 63: 1 seems a bit too harsh (will we get any accidental slashes + jailing if a container gets too slow?), should we do 0.85 instead? Requires them to be down for 2 blocks which is still quite fast!

scripts/vars.sh
sampocs: I can - although, do you think we should just enable these by default? I think you have to manually kill containers for any of these configurations to matter

vish-stride (approved):
LGTM! Although I think we can slightly lessen the harshness, and agreed with Aidan that a comment would be helpful

scripts/init_chain.sh
sampocs: Good point, I can make it a little more conservative! I also just realized our blocks are 1s. 

However, I do think we should either keep it strict and keep it optional, or relax it a tad and enable these by default. I'm not really sure how often slow containers cause missed blocks. 

Maybe we leave the min_signed_per_window at 0.5 and do signed_blocks_window as either 10 or 20?
Then I think they have to miss 5 or 10 blocks which is probably both unlikely to occur by accident and fast enough when it occurs intentionall?

PR #339, opened 2022-11-11
Refactor `GetHostZoneUnbondingMsgs` to sub funcs

hieuvubk:
Closes: #329 #417 

Refer from #329 there a lot of logic in `GetHostZoneUnbondingMsgs` that makes it difficult to read.

## Brief Changelog

Create sub funcs with separate processing logic:

- `GetUnbondingAmountAndRecords`: get total unbonding amount & all unbonding record of hostzone
- `DistributeUnbondingAmountToValidators`: distribute unbonding amount to all validators
- `SplitDelegationMsg`: wrap to msgs 

## Author's Checklist

I have...

## Future plan

- [ ] add test for new funcs
- [ ] refactor `GetHostZoneUnbondingMsgs` test to table-driven & add more edge case


x/stakeibc/keeper/unbonding_records_get_host_zone_unbondings_msgs_test.go
```diff
@@ -58,7 +63,22 @@ func (s *KeeperTestSuite) SetupGetHostZoneUnbondingMsgs() GetHostZoneUnbondingMs
 			DelegationAmt: amtVal2,
 			Weight:        wgtVal2,
 		},
-	}
+	},
+	totalWeight: 5,
+	expectPass: true,
+}
+
+func (s *KeeperTestSuite) SetupGetHostZoneUnbondingMsgs(tc GetHostZoneUnbondingMsgsTestCase) (uint64, types.HostZone) {
```
sampocs, line 71: I like the approach of passing the TestCase as a parameter, so you don't have to use keeper functions to handle the different cases! 

However, I find having functions for each test case more digestible/maintainable than having the map of test cases. With the map, it's hard to diagnose and debug when each test is ran in a loop rather than individually. I know the map is often the standard in cosmos, but the other tests I've seen have been much smaller (and require less state) than some of these Stride tests.

What do you think / what was your rationale for using the map?

x/stakeibc/keeper/unbonding_records.go
```diff
@@ -60,24 +92,19 @@ func (k Keeper) GetHostZoneUnbondingMsgs(ctx sdk.Context, hostZone types.HostZon
...
-	// we distribute the unbonding based on our target weights
 	newUnbondingToValidator, err := k.GetTargetValAmtsForHostZone(ctx, hostZone, totalAmtToUnbond)
 	if err != nil {
 		errMsg := fmt.Sprintf("Error getting target val amts for host zone %s %d: %s", hostZone.ChainId, totalAmtToUnbond, err)
 		k.Logger(ctx).Error(errMsg)
-		return nil, 0, nil, nil, sdkerrors.Wrap(types.ErrNoValidatorAmts, errMsg)
+		return nil, sdkerrors.Wrap(types.ErrNoValidatorAmts, errMsg)
 	}
-	valAddrToUnbondAmt := make(map[string]int64)
+	// valAddrToUnbondAmt := make(map[string]int64)
```
sampocs, line 107: nit: remove

x/stakeibc/keeper/unbonding_records.go
```diff
@@ -60,24 +92,19 @@ func (k Keeper) GetHostZoneUnbondingMsgs(ctx sdk.Context, hostZone types.HostZon
...
-	delegationAccount := hostZone.GetDelegationAccount()
-	if delegationAccount == nil || delegationAccount.GetAddress() == "" {
-		errMsg := fmt.Sprintf("Zone %s is missing a delegation address!", hostZone.ChainId)
-		k.Logger(ctx).Error(errMsg)
-		return nil, 0, nil, nil, sdkerrors.Wrap(types.ErrHostZoneICAAccountNotFound, errMsg)
-	}
+	return totalAmtToUnbond, epochUnbondingRecordIds
+}
+
+func (k Keeper) DistributeUnbondingAmountToValidators(ctx sdk.Context, hostZone types.HostZone, totalAmtToUnbond uint64) (map[string]int64, error) {
```
sampocs, line 98: can you use named return parameters here?

sampocs (commented):
This looks so much better! Thanks for taking care of this, and so sorry for the delay in reviewing!

A couple small comments below with a large comment around testing.

Also, now that this has been broken up into smaller functions, can you add unit tests for the smaller functions (`GetUnbondingAmountAndRecords`, `DistributeUnbondingAmountToValidators`, `SplitDelegationMsg`). That's one of the strongest benefits we get from increasing the modularity - it allows us to have more granular test coverage :)

x/stakeibc/keeper/unbonding_records_get_host_zone_unbondings_msgs_test.go
hieuvubk: There are some benefits of using table-driven tests for me:

- Easy to set up. For each change in the origin func, no need to set-up for all test funcs, just change the params
- Easy to look, other dev can see the difference of test-case from table
- Each test case run in its own context, if there're any err we can track which case threw them easily. Show example below
Ảnh chụp Màn hình 2022-11-22 lúc 12 02 22
Ảnh chụp Màn hình 2022-11-22 lúc 12 02 57

hieuvubk:
@sampocs I just formated code, add unit-tests for sub funcs, could u pls check it out

x/stakeibc/keeper/unbonding_records_get_host_zone_unbondings_msgs_test.go
sampocs: I think it brings a couple downsides with it though:
* While it saves ~1-2 lines of code per test by avoiding repeating the setup call, it adds a lot of duplicate code by repeating the struct definitions across each test (e.g. `epochUnbondingRecords` and `validators` are duplicated across each test). This makes it slightly harder to see what's being tested in each test.
* It makes it harder to debug because you have to run the full array of tests each time you make a change (i.e. there's no easy way to test one specific test case through an IDE). With separate functions, you can easily run each test individually (see the play button and "run test" option in the image below)
 Screen Shot 2022-11-22 at 11 47 20 AM
* It prevents the use of custom logic in each test (irrelevant for these tests, but we have other tests in the repo surrounding ICAs where we want to do things like close a channel in the middle of a test)
* It's also worth mentioning, that with the function setup, each test also runs in parallel in it's own context.

---

I think a good middle ground between what was originally in the code base and what you provided, would be to pass the test case as an argument, and provide functions for if there's any checks that seem to be repeated. 

that way the error tests will look something like this:

```go
var defaultUnbondingTestCase = ...

func (s *KeeperTestSuite) SetupGetHostZoneUnbondingMsgs(tc GetHostZoneUnbondingMsgsTestCase) (uint64, stakeibc.HostZone) {
   ...
}

func (s *KeeperTestSuite) TestGetHostZoneUnbondingMsgs_WrongChainId() {
   invalidState := defaultUnbondingTestCase
   invalidState.ChainId := "nonExistentChainId"

   resp, err := s.App.StakeibcKeeper.SomeFunctionBeingTested(s.Ctx(), invalidState)
   s.Require().EqualError(...)
}
```

This, in my opinion, makes it much easier to see what's going on in the test and makes the test file more modular and digestible overall.

And if you saw a lot of repeated logic, you could also always wrap the main function call + validation in it's own function so it looks like this: 
```go
func (s *KeeperTestSuite) VerifyTestCase(tc GetHostZoneUnbondingMsgsTestCase, success bool, err error) {
   resp, err := s.App.StakeibcKeeper.SomeFunctionBeingTested(s.Ctx(), invalidState)
   if success {
        ...
   }
   if err != nil {
        ...
   }
}

func (s *KeeperTestSuite) TestGetHostZoneUnbondingMsgs_WrongChainId() {
   invalidState := defaultUnbondingTestCase
   invalidState.ChainId := "nonExistentChainId"

   VerifyTestCase(invalidState)
}

```

What do you think?

hieuvubk: Agree with the idea of ​​using 1 object and then modifying it for each test case instead of using multiple.

hieuvubk: The design LGTM

x/stakeibc/keeper/unbonding_records_get_host_zone_unbondings_msgs_test.go
```diff
@@ -100,108 +109,349 @@ func (s *KeeperTestSuite) SetupGetHostZoneUnbondingMsgs() GetHostZoneUnbondingMs
...
-	s.Require().NoError(err)
+	s.VerifyTestCase_TestGetHostZoneUnbondingMsgs(name, test)
+}
 
-	// verify the callback attributes are as expected
-	actualCallbackResult, err := s.App.StakeibcKeeper.UnmarshalUndelegateCallbackArgs(s.Ctx(), actualCallbackArgs)
-	s.Require().NoError(err, "could unmarshal undelegation callback args")
-	s.Require().Equal(len(tc.hostZone.Validators), len(actualCallbackResult.SplitDelegations), "number of split delegations in success unbonding case")
-	s.Require().Equal(tc.hostZone.ChainId, actualCallbackResult.HostZoneId, "host zone id in success unbonding case")
+func (s *KeeperTestSuite) VerifyTestCase_TestGetUnbondingAmountAndRecords(name string, tc GetHostZoneUnbondingMsgsTestCase) {
```
sampocs, line 233: for `GetUnbondingAmountAndRecords` we want to test that we're only summing up records with status UNBONDING_QUEUE. So you can add to the default case some epoch unbonding records in various status's and then this should check that we are only summing some of them

x/stakeibc/keeper/unbonding_records_get_host_zone_unbondings_msgs_test.go
```diff
@@ -100,108 +109,349 @@ func (s *KeeperTestSuite) SetupGetHostZoneUnbondingMsgs() GetHostZoneUnbondingMs
...
 
-	val1Unbonded := strings.Contains(actualUnbondMsg1, val1UnbondAmt)
-	val2Unbonded := strings.Contains(actualUnbondMsg2, val2UnbondAmt)
+	s.VerifyTestCase_TestGetUnbondingAmountAndRecords(name, test)
 
-	// there's rounding in the logic that distributes stake amongst validators, so one or the other of the balances will be correct, depending on the rounding
-	// at least one will be correct, and the other will be off by 1 by rounding, so we check and OR condition
-	s.Require().True(val1Unbonded || val2Unbonded, "unbonding amt should be the correct amount")
 }
+func (s *KeeperTestSuite) VerifyTestCase_TestDistributeUnbondingAmountToValidators(name string, tc GetHostZoneUnbondingMsgsTestCase) {
```
sampocs, line 275: For `DistributeUnbondingAmountToValidators`, we also want to add a test case where a validator does not have enough at stake to handle their share of the unbonding and we traverse [this branch](https://github.com/Stride-Labs/stride/blob/main/x/stakeibc/keeper/unbonding_records.go#L100)

You can test this by setting the delegation amount of one of the validators to 0 (a validator that has non-zero weight) and then checking that the portion that was supposed to be undelegated from that validator was actually split among the others.

sampocs (commented):
This refactor looks great! Please let me know if you have any questions on the comments below

x/stakeibc/keeper/unbonding_records_get_host_zone_unbondings_msgs_test.go
```diff
@@ -100,108 +109,349 @@ func (s *KeeperTestSuite) SetupGetHostZoneUnbondingMsgs() GetHostZoneUnbondingMs
...
+			Address:       hostVal2Addr,
+			DelegationAmt: amtVal2,
+			Weight:        0,
+		},
+		{
+			Address:       hostVal3Addr,
+			DelegationAmt: amtVal2,
+			Weight:        0,
+		},
+	}
```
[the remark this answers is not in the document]
DongLieu: Sorry this is written by me, I was a trainee during training so there are many mistakes. Thank you so much for giving me a lesson for the next time.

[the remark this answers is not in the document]
sampocs: No need to apologize!! You did great, I appreciate your help!

riley-stride:
Friendly bump on this PR @hieuvubk, looks like there are some outstanding comments/recommendations from Sam, and you pushed new commits since his comments. Do you feel you've addressed them? 

I'm also happy to review this if it would be helpful. Just let me know @sampocs.

asalzmann:
open

PR #344, opened 2022-11-14
Reset TTL Upon Duplicate Query Submission

sampocs:
Closes: #XXX

* Psuedo code of current setup with description below...
```
def makeQueryRequest():
    queryId = hashOf(chainId, queryType) // this ID will always be the same across each epoch 
    callbackData = {ConnectionId, chainId, queryType, address, ..., ttl} // TTL is the only thing different across epochs

    if queryId not already in state:                   // we go down this branch during the first epoch
        k.SetValue(key=queryID, value=callbackData)   
    else:                                              // we'll go down this branch every epoch after that until we handle a response
        do nothing

def handleQueryResponse():
    callbackData = k.GetValue(key=queryId)
    if currentTime > callbackData.ttl:
        query has expired!
    else:
        handle response

    k.DeleteValue(key=queryId)
```
For a given host zone, each query type will share the same queryID (e.g. all delegation account balance queries for cosmoshub have the same query Id - regardless of the epoch that they're submitted from). However, the ttl field which is stored in the query object, is different across epochs. This introduces an edge case where if the same query is re-requested multiple times before any of them are handled, the query object is never updated. Which means that once one of the queries is finally handled, it will use the old TTL field and be marked as expired.
In other words, if a query wasn't handled after a given epoch, we'd have to wait an extra epoch before it was handled properly...
* Epoch 1) query (1) is requested but response is never relayed. Query TTL is defined in the store as roughly 2 hours before epoch 2.
* Epoch 2) - query (2) requested and response is relayed, but it has the same ID as query (1). This means it grabs the old query object from the store and marks this query as expired since it’s using the old TTL. The query is then deleted from the store
* Epoch 3) - query (3) is requested and response is relayed successfully

## Brief Changelog
* Added check to reset the TTL when a duplicate request is submitted
* Returned errors for a few edge cases where only errors were logged but the function continued
* Cleaned up the MakeRequest function in general

## Author's Checklist

I have...

## Reviewers Checklist

I have...

      - [ ] not applicable
      - [ ] jira ticket `XXX` 
      - [ ] README.md 
      - [ ] not documented

### DIFF
--- a/x/interchainquery/keeper/keeper.go
+++ b/x/interchainquery/keeper/keeper.go
@@ -105,8 +105,8 @@ func (k *Keeper) MakeRequest(ctx sdk.Context, connection_id string, chain_id str
 		k.SetQuery(ctx, *newQuery)
 
 	} else {
-		// a re-request of an existing query triggers resetting of height to trigger immediately.
-		existingQuery.LastHeight = sdk.ZeroInt()
+		// a re-request of an existing query should reset the TTL
+		existingQuery.Ttl = ttl
 		k.SetQuery(ctx, existingQuery)
 	}
 	return nil
--- a/x/interchainquery/keeper/keeper.go
+++ b/x/interchainquery/keeper/keeper.go
@@ -108,6 +108,7 @@ func (k *Keeper) MakeRequest(ctx sdk.Context, connectionId string, chainId strin
 		// Otherwise, if the same query is re-requested - reset the TTL
 		k.Logger(ctx).Info("Query already exists - resetting TTL")
 		query.Ttl = ttl
+		query.LastHeight = sdk.ZeroInt() // allows query to be emitted again
 	}
 
 	// Save the query to the store
--- a/x/interchainquery/keeper/keeper.go
+++ b/x/interchainquery/keeper/keeper.go
@@ -45,69 +45,62 @@ func (k Keeper) Logger(ctx sdk.Context) log.Logger {
 	return ctx.Logger().With("module", fmt.Sprintf("x/%s", types.ModuleName))
 }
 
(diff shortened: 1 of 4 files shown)


x/interchainquery/keeper/keeper.go
```diff
@@ -45,69 +45,72 @@ func (k Keeper) Logger(ctx sdk.Context) log.Logger {
...
+			return sdkerrors.Wrap(sdkerrors.ErrInvalidRequest, "no callback handler registered for module")
+		}
+	}
+
+	// Check to see if the query already exists
+	key := GenerateQueryHash(connectionId, chainId, queryType, request, module, height)
+	query, found := k.GetQuery(ctx, key)
 
+	// If this is a new query, build the query object
+	if !found {
```
asalzmann, line 105: Could we remove this block entirely?

asalzmann (approved):
lgtm!

x/interchainquery/keeper/keeper.go
sampocs: Can you add a GH suggestion for what you're envisioning?

asalzmann: Ah nevermind - the thought was to remove the if-else block of code but I realized the height is also different (not just the TTL)

asalzmann (approved):

x/interchainquery/keeper/keeper.go
sampocs: The height is actually the same each time (it's 0).

The idea here is we check if we already have a query in the store - and if we do, we just update the TTL. But if we don't, then we make a new query object and store that one.

If there's a way to format the code that would make that a little more obvious, please lmk!

asalzmann (approved):

Issue #347, opened 2022-11-16
Upgrade ibc-go to v5.1.0


sampocs:
## Background
* We are currently on ibc-go v3. Upgrading to v5 will bring some improvements to ICA and it will enable us to switch to the main version of the go relayer (we are currently using a side branch that has ibc v3 and new features backported)

## Suggested Design
* Update the package definition in go.mod, update the imports, and then make code changes as required

## Acceptance Criteria
* Unit and integration tests pass

PR #349, opened 2022-11-17
Sdk v0.46.* and ibc-go v5.1.0

hieuvubk:
Closes: #347 

This PR updates cosmos-sdk to v0.46.*.  It shouldn't be considered ready for production until the release of v0.46.7, which deprecates all prior sdk releases. 

## Brief Changelog

* update cosmos-sdk
* update ibc-go

## Author's Checklist

I have...

## Reviewers Checklist

I have...

      - [ ] not applicable
      - [ ] jira ticket `XXX` 
      - [ ] README.md 
      - [ ] not documented


asalzmann:
Is upgrade code also required here?

asalzmann:
open

asalzmann:
@hieuvubk can you fix the linting errors here please?

hieuvubk:
> @hieuvubk can you fix the linting errors here please?

Yes sure, but this will change all `sdk.Int` to `math.Int` so many changes. Are you sure you want to do it?

hieuvubk:
@asalzmann lint passed

asalzmann:
Thanks!

x/stakeibc/keeper/msg_server_restore_interchain_account.go
```diff
@@ -37,7 +37,15 @@ func (k msgServer) RestoreInterchainAccount(goCtx context.Context, msg *types.Ms
...
-	if err := k.ICAControllerKeeper.RegisterInterchainAccount(ctx, hostZone.ConnectionId, owner); err != nil {
+	appVerSion := string(icatypes.ModuleCdc.MustMarshalJSON(&icatypes.Metadata{
+		Version:                icatypes.Version,
+		ControllerConnectionId: hostZone.ConnectionId,
+		HostConnectionId:       hostZone.ConnectionId,
+		Encoding:               icatypes.EncodingProtobuf,
+		TxType:                 icatypes.TxTypeSDKMultiMsg,
+	}))
+
+	if err := k.ICAControllerKeeper.RegisterInterchainAccount(ctx, hostZone.ConnectionId, owner, appVerSion); err != nil {
```
sampocs, line 48:
```suggestion
	if err := k.ICAControllerKeeper.RegisterInterchainAccount(ctx, hostZone.ConnectionId, owner, apptesting.TestIcaVersion); err != nil {
```

x/records/module_ibc.go
```diff
@@ -78,7 +80,7 @@ func (im IBCModule) OnChanOpenTry(
 ) (string, error) {
 	// doCustomLogic()
 	// core/04-channel/types contains a helper function to split middleware and underlying app version
-	_, cpAppVersion := channeltypes.SplitChannelVersion(counterpartyVersion)
+	_, cpAppVersion := SplitChannelVersion(counterpartyVersion)
```
sampocs, line 83: I'm not positive but I think we pass don't need to call this and instead pass `counterpartyVersion` through to `im.app.OnChanOpenTry`

Definitely double check me on this though

x/records/module_ibc.go
```diff
@@ -349,3 +351,14 @@ func (am AppModule) NegotiateAppVersion(
 ) (version string, err error) {
 	return proposedVersion, nil
 }
+
+func SplitChannelVersion(version string) (middlewareVersion, appVersion string) {
```
sampocs, line 355: Related to early comment:
> I'm not positive but I think we pass don't need to call this and instead pass counterpartyVersion through to im.app.OnChanOpenTry

If that's correct, you can remove this function

app/test_setup.go
```diff
@@ -61,6 +71,80 @@ func InitStrideTestApp(initChain bool) *StrideApp {
 	return app
 }
 
+func GenesisStateWithValSet(app *StrideApp) GenesisState {
```
sampocs, line 74: what was the reason for adding this? / were there some tests that broke until the full genesis state was mocked out?

sampocs (commented):
Really impressive work here, this was a big one! Left a couple small comments / questions below.

Also, in the sdk/ibc-go/evmos, I've seen them import "cosmossdk.io/math" as `sdkmath` so it might be good to follow their lead and keep it consistent.

sampocs:
Also, do we need to add this [upgrade handler for ibc denoms](https://ibc.cosmos.network/main/migrations/v3-to-v4.html#migration-to-fix-support-for-base-denoms-with-slashes)?

sampocs:
One more important comment, we'll have to update how we handle acks (example [here](https://ibc.cosmos.network/main/migrations/v4-to-v5.html#cosmos-sdk-message-handler-responses-in-packet-acknowledgement))

app/apptesting/test_helpers.go
```diff
@@ -203,11 +203,11 @@ func (s *AppTestHelper) RegisterInterchainAccount(endpoint *ibctesting.Endpoint,
 	// Get the next channel available and register the ICA
 	channelSequence := s.App.IBCKeeper.ChannelKeeper.GetNextChannelSequence(s.Ctx)
 
-	err = s.App.ICAControllerKeeper.RegisterInterchainAccount(s.Ctx, endpoint.ConnectionID, owner)
+	err = s.App.ICAControllerKeeper.RegisterInterchainAccount(s.Ctx, endpoint.ConnectionID, owner, TestIcaVersion)
 	s.Require().NoError(err, "register interchain account error")
 
 	// Commit the state
-	endpoint.Chain.App.Commit()
+	// endpoint.Chain.App.Commit()
```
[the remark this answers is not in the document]
hieuvubk: Old code doesn't do enough process to end 1 block and start new block.
If u look into `NextBlock`, there are steps beside `chain.App.Commit()`

x/stakeibc/keeper/msg_server_register_host_zone.go
```diff
@@ -81,34 +83,42 @@ func (k msgServer) RegisterHostZone(goCtx context.Context, msg *types.MsgRegiste
 	// write the zone back to the store
 	k.SetHostZone(ctx, zone)
 
+	appVerSion := string(icatypes.ModuleCdc.MustMarshalJSON(&icatypes.Metadata{
```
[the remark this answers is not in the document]
hieuvubk: I calculated appVersion base on what we did in apptesting but it depends on `hostZone.ConnectionId`. So in my opinion we shouldn't hardcode the value here

app/app.go
```diff
@@ -360,13 +365,14 @@ func NewStrideApp(
 
 	// add keepers
 	app.AccountKeeper = authkeeper.NewAccountKeeper(
-		appCodec, keys[authtypes.StoreKey], app.GetSubspace(authtypes.ModuleName), authtypes.ProtoBaseAccount, maccPerms,
+		appCodec, keys[authtypes.StoreKey], app.GetSubspace(authtypes.ModuleName), authtypes.ProtoBaseAccount, maccPerms, sdk.Bech32MainPrefix,
```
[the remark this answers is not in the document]
hieuvubk: Yes it should be `AccountAddressPrefix`

x/records/module_ibc.go
hieuvubk: Oh I see, thank u

app/test_setup.go
hieuvubk: Since sdk 46, we can not init with an empty valset. That's why I use this function

asalzmann:
> Also, do we need to add this [upgrade handler for ibc denoms](https://ibc.cosmos.network/main/migrations/v3-to-v4.html#migration-to-fix-support-for-base-denoms-with-slashes)?

As far as I know none of our base denoms contain slashes, so we should be ok
```
stuatom
stujuno
stustars
stuosmo
strd
```

x/claim/keeper/hooks.go
```diff
@@ -21,8 +21,10 @@ func (k Keeper) AfterDelegationModified(ctx sdk.Context, delAddr sdk.AccAddress,
 			write()
 		} else {
 			k.Logger(ctx).Error(fmt.Sprintf("airdrop claim failure for %s on delegation hook: %s", delAddr.String(), err.Error()))
+			return err
 		}
 	}
+	return nil
```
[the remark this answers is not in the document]
asalzmann: Yes, I think returning an error is actually the right behavior. Checked the `StakingKeeper`, and `AfterDelegationModified` does handle/return an error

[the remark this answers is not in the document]
asalzmann: `AfterEpochEnd` and `AfterLiquidStake` do not, however

app/apptesting/test_helpers.go
```diff
@@ -268,7 +268,7 @@ func (s *AppTestHelper) ICAPacketAcknowledgement(msgs []sdk.Msg, msgResponse *pr
 		} else {
 			data = []byte("msg_response")
 		}
-		txMsgData.Data[i] = &sdk.MsgData{
+		txMsgData.Data[i] = &sdk.MsgData{ //nolint:staticcheck
```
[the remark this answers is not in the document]
asalzmann: Looks like this is because sdk.MsgData is deprecated

hieuvubk:
> One more important comment, we'll have to update how we handle acks (example [here](https://ibc.cosmos.network/main/migrations/v4-to-v5.html#cosmos-sdk-message-handler-responses-in-packet-acknowledgement))

Yes I did it in below commit

hieuvubk:
Hi @sampocs @asalzmann I addressed all your upper comments, could u guys take a look again? Also thank for leaving them

asalzmann:
Taking a look now

x/records/module_ibc.go
```diff
@@ -62,7 +60,7 @@ func (im IBCModule) OnChanOpenInit(
 		channelID,
 		channelCap,
 		counterparty,
-		appVersion, // note we only pass app version here
+		version, // note we only pass app version here
```
sampocs, line 63: you can remove the comment here as well

x/records/module_ibc.go
```diff
@@ -91,13 +87,13 @@ func (im IBCModule) OnChanOpenTry(
 		channelID,
 		chanCap,
 		counterparty,
-		cpAppVersion, // note we only pass counterparty app version here
+		counterpartyVersion, // note we only pass counterparty app version here
```
sampocs, line 90: same here

x/stakeibc/keeper/msg_server_register_host_zone.go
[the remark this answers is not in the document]
sampocs: There aren't any unit tests where we create a connection other than `connection-0` (in fact the connection ID is hard coded into the Setup function) so I think it's okay to use the hard coded app version as well. It's good to keep the tests as slim as possible

x/stakeibc/keeper/icacallbacks_undelegate_test.go
```diff
@@ -260,21 +261,14 @@ func (s *KeeperTestSuite) TestGetLatestCompletionTime_Success() {
 	// Construct TxMsgData
 	firstCompletionTime := time.Now().Add(time.Second * time.Duration(10))
 	secondCompletionTime := time.Now().Add(time.Second * time.Duration(20))
+	var err error
 	txMsgData := &sdk.TxMsgData{
-		Data: make([]*sdk.MsgData, 2), //nolint:staticcheck
+		MsgResponses: make([]*codectypes.Any, 2), //nolint:staticcheck
```
sampocs: I think you should be able to remove the `//nolint:staticcheck` now

x/stakeibc/keeper/icacallbacks_undelegate_test.go
```diff
@@ -284,7 +278,7 @@ func (s *KeeperTestSuite) TestGetLatestCompletionTime_Success() {
 func (s *KeeperTestSuite) TestGetLatestCompletionTime_Failure() {
 	_ = s.SetupUndelegateCallback()
 	txMsgData := &sdk.TxMsgData{
-		Data: make([]*sdk.MsgData, 2), //nolint:staticcheck
+		MsgResponses: make([]*codectypes.Any, 2), //nolint:staticcheck
```
sampocs: same here

app/test_setup.go
sampocs: ah got it. Thanks for adding this!

sampocs:
@asalzmann 
> As far as I know none of our base denoms contain slashes, so we should be ok

I think you're conflating base denom's with native denoms. Base denom's are the underlying denom after you remove the stride channel/port prefix. But just checked and doesn't look like we have any with slashes.

```
>>> stridedm q ibc-transfer denom-traces | grep base_denom
- base_denom: stuatom
- base_denom: uatom
- base_denom: uatom
- base_denom: ustars
- base_denom: uatom
- base_denom: uscrt
- base_denom: ujuno
- base_denom: uluna
- base_denom: uatolo
- base_denom: uxki
- base_denom: aevmos
- base_denom: uregen
- base_denom: ukuji
- base_denom: ustars
- base_denom: ujuno
- base_denom: uatom
- base_denom: uumee
- base_denom: utori
- base_denom: ucmdx
- base_denom: uscrt
- base_denom: utori
- base_denom: swth
- base_denom: ucmdx
- base_denom: stuatom
- base_denom: uatom
- base_denom: ustrd
- base_denom: stuosmo
- base_denom: ujuno
- base_denom: basecro
- base_denom: ustars
- base_denom: uosmo
- base_denom: orai
- base_denom: ucre
- base_denom: ustrd
- base_denom: aevmos
```

asalzmann:
@hieuvubk I think we'll need branching logic which is not addressed in `198f5ac1b4c36e8c2dd9168dedf045cf6a05b8fe`. Specifically, like so
```
var ack channeltypes.Acknowledgement
if err := channeltypes.SubModuleCdc.UnmarshalJSON(acknowledgement, &ack); err != nil {
   return err
}

var txMsgData sdk.TxMsgData
if err := proto.Unmarshal(ack.GetResult(), txMsgData); err != nil {
   return err
}

switch len(txMsgData.Data) {
case 0: // for SDK 0.46 and above
   for _, msgResponse := range txMsgData.MsgResponses {
      // unmarshall msgResponse and execute logic based on the response 
   }
   return nil
default: // for SDK 0.45 and below
   for _, msgData := range txMsgData.Data {
      // unmarshall msgData and execute logic based on the response 
   }
}
```

I'll do this now

sampocs:
Thanks for the updates Hieu! As far as the ack's I think you need to update module_ibc.go in icacallbacks/records to this:
```go
var ack channeltypes.Acknowledgement
if err := channeltypes.SubModuleCdc.UnmarshalJSON(acknowledgement, &ack); err != nil {
   return err
}

var txMsgData sdk.TxMsgData
if err := proto.Unmarshal(ack.GetResult(), txMsgData); err != nil {
   return err
}

switch len(txMsgData.Data) {
case 0: // for SDK 0.46 and above
   for _, msgResponse := range txMsgData.MsgResponses {
      // unmarshall msgResponse and execute logic based on the response 
   }
   return nil
default: // for SDK 0.45 and below
   for _, msgData := range txMsgData.Data {
      // unmarshall msgData and execute logic based on the response 
   }
}
```

x/stakeibc/keeper/icacallbacks_undelegate_test.go
hieuvubk: I checked in filechanges, they r all resolved

hieuvubk:
> @hieuvubk I think we'll need branching logic which is not addressed in `198f5ac1b4c36e8c2dd9168dedf045cf6a05b8fe`. Specifically, like so
> 
> ```
> var ack channeltypes.Acknowledgement
> if err := channeltypes.SubModuleCdc.UnmarshalJSON(acknowledgement, &ack); err != nil {
>    return err
> }
> 
> var txMsgData sdk.TxMsgData
> if err := proto.Unmarshal(ack.GetResult(), txMsgData); err != nil {
>    return err
> }
> 
> switch len(txMsgData.Data) {
> case 0: // for SDK 0.46 and above
>    for _, msgResponse := range txMsgData.MsgResponses {
>       // unmarshall msgResponse and execute logic based on the response 
>    }
>    return nil
> default: // for SDK 0.45 and below
>    for _, msgData := range txMsgData.Data {
>       // unmarshall msgData and execute logic based on the response 
>    }
> }
> ```
> 
> I'll do this now

Hey I addressed them in all callback files

sampocs:
> Hey I addressed them in all callback files
I believe you need to update the OnAcknowledgementPacket in module_ibc.go - but looks like Aidan's taking care of it

sampocs:
And then also bumping this 
> In the sdk/ibc-go/evmos, I've seen them import "cosmossdk.io/math" as sdkmath so it might be good to follow their lead and keep it consistent. Also, do we need to update the protos since they have an sdk.Int annotation?

Wdyt @asalzmann and @hieuvubk  ?

hieuvubk:
https://github.com/Stride-Labs/stride/pull/349#discussion_r1063476869
@sampocs about this one, I'm not sure about your idea cause its not a test file here. `appVersion` should depend on `msg.ConnectionId` that user pass through msg

hieuvubk:
> 

No need to update proto, I see evmos they still use `github.com/cosmos/cosmos-sdk/types.Int` in their proto

hieuvubk:
> And then also bumping this
> 
> > In the sdk/ibc-go/evmos, I've seen them import "cosmossdk.io/math" as sdkmath so it might be good to follow their lead and keep it consistent. Also, do we need to update the protos since they have an sdk.Int annotation?
> 
> Wdyt @asalzmann and @hieuvubk ?


No need to update proto, I see evmos they still use github.com/cosmos/cosmos-sdk/types.Int in their proto

hieuvubk:
> 

renamed all

hieuvubk:
> > No need to update proto, I see evmos they still use github.com/cosmos/cosmos-sdk/types.Int in their proto
> 
> Sounds good. And what about `cosmosmath` → `sdkmath`?

renamed all

sampocs:
Awesome. Can you also update the relayer to a new commit that uses ibc-v5? 

1. Update submodule
```
cd deps/relayer
git checkout 7f78a4d608329b2ae07a6512381bbadffee66be6
cd ../..
git add deps/relayer
```

2. Update [.gitmodules](https://github.com/notional-labs/stride/blob/upgrade_ibc_v5/.gitmodules#L5)
```
[submodule "deps/relayer"]
	# Commit: 7f78a4d608329b2ae07a6512381bbadffee66be6
	# Branch: andrew/client_icq
	path = deps/relayer
	url = https://github.com/cosmos/relayer.git
```

3. Update the relayer [dockerfile](https://github.com/notional-labs/stride/blob/upgrade_ibc_v5/dockernet/dockerfiles/Dockerfile.relayer#L6-L9)
```
ENV COMMIT_HASH=7f78a4d608329b2ae07a6512381bbadffee66be6

RUN apk add --update git make gcc linux-headers libc-dev eudev-dev 
RUN git clone https://github.com/cosmos/relayer.git \
```

hieuvubk:
> Awesome. Can you also update the relayer to a new commit that uses ibc-v5?
> 
> 1. Update submodule
> 
> ```
> cd deps/relayer
> git checkout 7f78a4d608329b2ae07a6512381bbadffee66be6
> cd ../..
> git add deps/relayer
> ```
> 
> 2. Update [.gitmodules](https://github.com/notional-labs/stride/blob/upgrade_ibc_v5/.gitmodules#L5)
> 
> ```
> [submodule "deps/relayer"]
> 	# Commit: 7f78a4d608329b2ae07a6512381bbadffee66be6
> 	# Branch: andrew/client_icq
> 	path = deps/relayer
> 	url = https://github.com/cosmos/relayer.git
> ```
> 
> 3. Update the relayer [dockerfile](https://github.com/notional-labs/stride/blob/upgrade_ibc_v5/dockernet/dockerfiles/Dockerfile.relayer#L6-L9)
> 
> ```
> ENV COMMIT_HASH=7f78a4d608329b2ae07a6512381bbadffee66be6
> 
> RUN apk add --update git make gcc linux-headers libc-dev eudev-dev 
> RUN git clone https://github.com/cosmos/relayer.git \
> ```

Where u get the commit hash `7f78a4d608329b2ae07a6512381bbadffee66be6 `?

sampocs:
> Where u get the commit hash 7f78a4d608329b2ae07a6512381bbadffee66be6 ?

I'm in contact with the Strangelove team who maintains the relayer. They just pushed the v5 update and I grabbed the commit from that.

Branch: https://github.com/cosmos/relayer/tree/andrew/client_icq
Last Commit: https://github.com/cosmos/relayer/commit/7f78a4d608329b2ae07a6512381bbadffee66be6

x/icacallbacks/module_ibc.go
```diff
@@ -92,7 +94,11 @@ func (im IBCModule) NegotiateAppVersion(
 }
 
 // GetTxMsgData returns the msgs from an ICA transaction and can be reused across authentication modules
-func GetTxMsgData(ctx sdk.Context, ack channeltypes.Acknowledgement, logger log.Logger) (*sdk.TxMsgData, error) {
+func GetTxMsgData(ctx sdk.Context, ack channeltypes.Acknowledgement, logger log.Logger) (*types.ICATxResponse, error) {
+	// if ack is nil, a timeout occurred
+	if &ack == nil {
```
sampocs, line 99: I think this is supposed to be just `ack`

x/icacallbacks/module_ibc.go
```diff
@@ -104,11 +110,13 @@ func GetTxMsgData(ctx sdk.Context, ack channeltypes.Acknowledgement, logger log.
 			logger.Error(fmt.Sprintf("cannot unmarshal ICS-27 tx message data: %s", err.Error()))
 			return nil, sdkerrors.Wrapf(sdkerrors.ErrUnknownRequest, "cannot unmarshal ICS-27 tx message data: %s", err.Error())
 		}
-		return txMsgData, nil
+		// NOTE: Is this safe to assume? According to the ics-27 spec, errors on host chains should be reflected in the acknowledgement type
```
sampocs, line 113: I think we can add the following under this branch:

```
	switch len(txMsgData.Data) {
	case 0: // for SDK 0.46 and above
		return  &types.ICATxResponse{Status: types.TIMEOUT, Data: txMsgData.MsgResponses}, nil 
	default: // for SDK 0.45 and below
		return  &types.ICATxResponse{Status: types.TIMEOUT, Data: txMsgData.Data}, nil
```

x/icacallbacks/module_ibc.go
```diff
@@ -92,7 +94,11 @@ func (im IBCModule) NegotiateAppVersion(
 }
 
 // GetTxMsgData returns the msgs from an ICA transaction and can be reused across authentication modules
-func GetTxMsgData(ctx sdk.Context, ack channeltypes.Acknowledgement, logger log.Logger) (*sdk.TxMsgData, error) {
+func GetTxMsgData(ctx sdk.Context, ack channeltypes.Acknowledgement, logger log.Logger) (*types.ICATxResponse, error) {
```
sampocs, line 97: If you're on board with moving the ack unmarshaling out, I think we should add it to the top of this function and rename it to something more generic like `UnpackICAAcknowledgement`(I'm sure you can think of a much better name)

And then each callback would take types.ICATxResponse as the argument instead of the Ack packet type

x/stakeibc/keeper/icacallbacks_undelegate.go
```diff
@@ -155,22 +154,44 @@ func (k Keeper) UpdateDelegationBalances(ctx sdk.Context, zone types.HostZone, u
...
+			}
+			err := proto.Unmarshal(msgResponseBytes.Data, &undelegateResponse)
+			if err != nil {
+				errMsg := fmt.Sprintf("Unable to unmarshal undelegation tx response | %s", err)
+				k.Logger(ctx).Error(errMsg)
+				return nil, sdkerrors.Wrapf(types.ErrUnmarshalFailure, errMsg)
+			}
+			if undelegateResponse.CompletionTime.After(latestCompletionTime) {
+				latestCompletionTime = undelegateResponse.CompletionTime
+			}
```
sampocs, line 191: So now you can just refactor this into:
```go
func (k Keeper) GetLatestCompletionTime(ctx sdk.Context, data []byte) (*time.Time, error) {
    // Update the completion time using the latest completion time across each message within the transaction
   latestCompletionTime := time.Time{}
   if data == nil {
	return nil, sdkerrors.Wrap(types.ErrTxMsgDataInvalid, "msgResponseBytes or msgResponseBytes.Data is nil")
   }
  err := proto.Unmarshal(msgResponseBytes.Data, &undelegateResponse)
  if err != nil {
       errMsg := fmt.Sprintf("Unable to unmarshal undelegation tx response | %s", err)
	k.Logger(ctx).Error(errMsg)
	return nil, sdkerrors.Wrapf(types.ErrUnmarshalFailure, errMsg)
  }
  if undelegateResponse.CompletionTime.After(latestCompletionTime) {
	latestCompletionTime = undelegateResponse.CompletionTime
  }

}
```

x/icacallbacks/types/ICATxResponseStatus.go
```diff
@@ -0,0 +1,15 @@
...
+
+type ICATxResponseStatus int
+const (
+    SUCCESS ICATxResponseStatus = iota
+    TIMEOUT
+    FAILURE
+)
+type ICATxResponse struct {
+	Status ICATxResponseStatus // enum of SUCCESS, TIMEOUT, FAILURE
+	Data sdk.TxMsgData // TxMsgData
```
sampocs, line 14: And then I think you want data to be a byte array here

x/icacallbacks/module_ibc.go
asalzmann: Might be misunderstanding your proposal though

x/icacallbacks/module_ibc.go
```diff
@@ -93,30 +94,70 @@ func (im IBCModule) NegotiateAppVersion(
...
-		// NOTE: Is this safe to assume? According to the ics-27 spec, errors on host chains should be reflected in the acknowledgement type
-		// so Acknowledgement_Result should only be returned if the transaction was successful
-		// However, I recall seeing a case where the acknowledgement type was Acknowledgement_Result but the transaction failed
-		return &types.ICATxResponse{Status: types.SUCCESS, Data: *txMsgData}, nil
+
+		// Unpack all the message responses based on the sdk version (determined from the length of txMsgData.Data)
+		switch len(txMsgData.Data) {
+		case 0:
+			// for SDK 0.46 and above
+			msgResponses := make([][]byte, len(txMsgData.Data))
```
asalzmann, line 145:
```suggestion
			msgResponses := make([][]byte, len(txMsgData.MsgResponses))
```
shouldn't this use the length of `txMsgData.MsgResponses`?

x/icacallbacks/module_ibc.go
```diff
@@ -93,30 +94,70 @@ func (im IBCModule) NegotiateAppVersion(
...
+	// so Acknowledgement_Result should only be returned if the transaction was successful
+	// However, I recall seeing a case where the acknowledgement type was Acknowledgement_Result but the transaction failed
+	//
+	// -> If my memory's correct, I think a failure on the host could come back as either an AcknowledgementError
+	//    or it could be an AcknowledgementResult with data length 0
+	//    It's a little concering that the switch statement now just assumes the data length 0 means it's sdk 46
+	//
+	// It's also unclear to me whether this timeout case below is possible
+	// In any case, we should just test this thoughly with the following cases:
+	//   Success
```
asalzmann, line 109: I don't think this sequence is possible, because the host chain should process either a timeout or an error, but not both

x/icacallbacks/module_ibc.go
```diff
@@ -93,30 +94,70 @@ func (im IBCModule) NegotiateAppVersion(
...
 
-// GetTxMsgData returns the msgs from an ICA transaction and can be reused across authentication modules
-func GetTxMsgData(ctx sdk.Context, ack channeltypes.Acknowledgement, logger log.Logger) (*types.ICATxResponse, error) {
+// UnpackAcknowledgementResponse returns the msgs from an ICA transaction and can be reused across authentication modules
+func UnpackAcknowledgementResponse(ctx sdk.Context, ack []byte, logger log.Logger) (*types.ICATxResponse, error) {
+	// NOTE: Is this safe to assume? According to the ics-27 spec, errors on host chains should be reflected in the acknowledgement type
+	// so Acknowledgement_Result should only be returned if the transaction was successful
+	// However, I recall seeing a case where the acknowledgement type was Acknowledgement_Result but the transaction failed
+	//
+	// -> If my memory's correct, I think a failure on the host could come back as either an AcknowledgementError
```
asalzmann, line 103: also recall this - I think the way we tested it (as you mentioned below) was delegating to a non-existent validator?

x/icacallbacks/types/ica_tx_response.go
```diff
@@ -0,0 +1,17 @@
+package types
+
+// QUESTION: Should we add this as a proto?
```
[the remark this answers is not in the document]
sampocs: Yeah good point

x/icacallbacks/module_ibc.go
sampocs: ah yes good catch!

x/icacallbacks/module_ibc.go
```diff
@@ -93,30 +94,70 @@ func (im IBCModule) NegotiateAppVersion(
...
+
+	// Unmarshal the raw ack response
+	var acknowledgement channeltypes.Acknowledgement
+	if err := ibctransfertypes.ModuleCdc.UnmarshalJSON(ack, &acknowledgement); err != nil {
+		return nil, sdkerrors.Wrapf(sdkerrors.ErrUnknownRequest, "cannot unmarshal ICS-20 transfer packet acknowledgement: %s", err.Error())
+	}
+
 	// if ack is nil, a timeout occurred
-	if &ack == nil {
+	if &acknowledgement == nil {
```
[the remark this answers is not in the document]
sampocs: yup, I was thinking that as well. Might make sense to add them in after testing though cause it's still a little unclear to me which branch will wind up being timeout, error, success, etc.
I'll add a TODO for the time being

sampocs:
For testing, we can add logging at each step in the Unpack function.

And then we can test the following cases:
* Transfer success
* Transfer failure (hard code wrong delegation address in transfer deposit records step)
* ICA sucess
* ICA failure (hard code wrong validator address)
* ICA timeout (hard code TTL that's 1 second after current block time)

@asalzmann  What am I forgetting? I think there's another failure case - maybe from a tx that's rejected in validate basic? Maybe we send a Msg type that doesn't actually exist on the host or send a message with missing args?

x/icacallbacks/module_ibc.go
asalzmann: We identify it with a nil ack out of convenience here https://github.com/notional-labs/stride/blob/acd8c57188652ce42c53a09c82a2d5bbe49b83af/x/stakeibc/module_ibc.go#L174

The timeout actually flows down a different code path (`OnTimeout`), so it probably makes sense to stop using a nil ack to identify timeouts and instead pass in an `ICATxResponse` with a timeout status. Sorry, should have caught this in the review - I'll push this up now.

asalzmann:
The above test plan makes sense - thoughts on doing this in a separate PR? This one is getting a bit hairy

x/icacallbacks/module_ibc.go
sampocs: Ah that's right, good call out!

sampocs:
> The above test plan makes sense - thoughts on doing this in a separate PR? This one is getting a bit hairy

Agreed, I don't think we should push any changes related to those tests, I think we can just do it locally. But I think we should avoid merging this to main until the integration tests pass. Maybe we get the acks working for the current v3 hosts, merge, then test a v5 host (evmos) and make relevant changes in another PR?

> I think we should probably write integration tests for these, wdyt?

I think as far as the individual success/failure test cases, we should just determine the Ack packet format and then make unit tests. And then wrt integration tests, I think adding a v5 host is adequate.

sampocs:
@asalzmann Integration tests are working. I'm trying to test restoring the channel but looks like there's some bugs in the relayer we might have to solve first.
If the new changes look good, I think we should merge this.

x/icacallbacks/icacallbacks.go
```diff
@@ -0,0 +1,84 @@
...
+		return msgResponses, nil
+	}
+}
+
+// UnpackAcknowledgementResponse unmarshals IBC Acknowledgements, determines the status of the acknowledgement (success or failure)
+//     and, if applicable, assembles the message responses
+// ICA transactions have associated messages responses. IBC transfer do not.
+// With ICA transactions, the schema of the response differs depending on the version of ibc-go used,
+//     however, this function unifies the format into a common response (a slice of byte arrays)
+func UnpackAcknowledgementResponse(ctx sdk.Context, logger log.Logger, ack []byte, isICA bool) (*types.AcknowledgementResponse, error) {
```
asalzmann, line 50: Not needed here, in the future we might want to encode the ack type as an enum (ICA, Transfer, etc.) so that we can support more IBC apps

asalzmann:
LGTM

x/icacallbacks/icacallbacks.go
sampocs: That's a great idea, much cleaner! I'll actually make a quick PR for that soon

PR #351, opened 2022-11-17
revert DELEGATION_IN_PROGRESS records to DELEGATION_QUEUE

asalzmann:
Closes: #XXX

`OnTimeoutPacket` reverts a single record in a pending state; we need to revert multiple pending records in some cases. This PR adds the ability to reset batches of records from pending -> queue state.

Note: the controller module in ibcv3 does not call `OnChanCloseConfirm`, so while this would be the most natural place to put this logic, it is not an option unless we fork or upgrade ibc (to a different version). For now, I put this logic in `RestoreInterchainAccount` (which should also simplify remedying future issues), although we could also put this logic in `OnTimeoutPacket`.

## Brief Changelog

- revert records to queue state when restoring an ICA

## Testing

Configure the go relayer to only relay transfer packets like so
```
paths:
  stride-gaia:
    src:
      chain-id: STRIDE
    dst:
      chain-id: GAIA
    src-channel-filter:
      rule: "allowlist"
      channel-list: [channel-0]
```

Once the chains are up and running, pause hermes, then liquid stake a few times. This will cause records to build up in the DELEGATION_PENDING state.

Then, unpause hermes. This will cause a timeout and channel closure.

Finally, restore the ICA like so
```
$STRIDE_MAIN_CMD tx stakeibc restore-interchain-account GAIA DELEGATION --keyring-backend test --from val1 -y --chain-id $STRIDE_CHAIN_ID
```
This reverts the pending records to the queue state.

## Author's Checklist

I have...

## Reviewers Checklist

I have...

      - [ ] not applicable
      - [ ] jira ticket `XXX` 
      - [ ] README.md 
      - [ ] not documented


x/stakeibc/module_ibc.go
```diff
@@ -184,6 +186,41 @@ func (im IBCModule) OnChanCloseConfirm(
...
+	im.keeper.Logger(ctx).Info(fmt.Sprintf("OnChanCloseConfirm: portID %s, channelID %s", portID, channelID))
+	controllerConnectionId, err := im.keeper.GetConnectionId(ctx, portID)
+	if err != nil {
+		ctx.Logger().Error(fmt.Sprintf("Unable to get connection for port: %s", portID))
+	}
+	// get host chain id from connection
+	hostChainId, err := im.keeper.GetChainID(ctx, controllerConnectionId)
+	if err != nil {
+		ctx.Logger().Error(fmt.Sprintf("Unable to obtain counterparty chain for connection: %s, port: %s, err: %s", controllerConnectionId, portID, err.Error()))
+		return nil
```
sampocs, line 198: return error here?

x/stakeibc/module_ibc.go
```diff
@@ -184,6 +186,41 @@ func (im IBCModule) OnChanCloseConfirm(
...
+	}
+	if portID != delegationAddress {
+		ctx.Logger().Info(fmt.Sprintf("Port %s is not a delegation port, skipping", portID))
+		return nil
+	}
+	//  get zone info
+	hostZone, found := im.keeper.GetHostZone(ctx, hostChainId)
+	if !found {
+		ctx.Logger().Error(fmt.Sprintf("Expected to find zone info for %v", hostChainId))
+		return nil
```
sampocs, line 212: return error here?

x/stakeibc/module_ibc.go
```diff
@@ -184,6 +186,41 @@ func (im IBCModule) OnChanCloseConfirm(
...
+	hostChainId, err := im.keeper.GetChainID(ctx, controllerConnectionId)
+	if err != nil {
+		ctx.Logger().Error(fmt.Sprintf("Unable to obtain counterparty chain for connection: %s, port: %s, err: %s", controllerConnectionId, portID, err.Error()))
+		return nil
+	}
+	delegationAddress, err := icatypes.NewControllerPortID(types.FormatICAAccountOwner(hostChainId, types.ICAAccountType_DELEGATION))
+	if err != nil {
+		return err
+	}
+	if portID != delegationAddress {
```
sampocs, line 204: superrr nit, but might be a tad clearer if you changed this to an equality and nested everything else below - considering this is still supposed to be a generic OnChanClose function, and the delegation account closing is just one possible branch

```go
if portID == delegationAddress {
    ...
}
```

sampocs (commented):
Looks great, thanks for taking care of this!

scripts/scratch.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
```
sampocs, line 1: can we keep scratch files out of the repo? There's also `scripts/test-util/1.sh` that does this exact liquid stake

sampocs (commented):
I like this approach soooo much better! Can you add unit tests to `msg_server_restore_interchain_account.test`? I'm thinking something like:
* `TestRestoreInterchainAccount_RevertDepositRecords_Success`
  * Follow a similar setup as `TestRestoreInterchainAccount_Success` but with PENDING deposit records present in the store and check that they were reverted
* `TestRestoreInterchainAccount_RevertDepositRecords_Failure`
  * Try the same thing as the above test, except *do not* close the delegation channel, and confirm that it fails when trying to register and the deposit records never get reverted

sampocs:
maybe one more test case where there are PENDING records present, but the restore function was called with a different channel - and then confirming that the deposit records are untouched

x/stakeibc/keeper/icacallbacks_delegate.go
```diff
@@ -56,8 +56,6 @@ func DelegateCallback(k Keeper, ctx sdk.Context, packet channeltypes.Packet, ack
 
 	if ack == nil {
 		// timeout
-		depositRecord.Status = recordstypes.DepositRecord_DELEGATION_QUEUE
```
vish-stride, line 59: Do we not need to handle this logic here, because a timeout will always close the channel, which would trigger `restore_interchain_account`?

x/stakeibc/module_ibc.go
```diff
@@ -184,6 +184,12 @@ func (im IBCModule) OnChanCloseConfirm(
 	portID,
 	channelID string,
 ) error {
+
+	// WARNING: For some reason, in IBCv3 the ICA controller module does not call the underlying OnChanCloseConfirm (this function)
```
vish-stride, line 188: So strange!

vish-stride (approved):
Looks great! Left one small question, but this will be very nice to have in

scripts/scratch.sh
asalzmann: Meant to add this to .gitignore, thx for pointing this out

x/stakeibc/keeper/icacallbacks_delegate.go
asalzmann: Yep!

asalzmann:
@sampocs added unittests, thanks for the suggestions!

x/stakeibc/keeper/msg_server_restore_interchain_account_test.go
```diff
@@ -105,3 +124,66 @@ func (s *KeeperTestSuite) TestRestoreInterchainAccount_FailsIfAccountExists() {
...
+	newChannelActive := false
+	for _, channel := range channels {
+		// The new channel should have the same port, a new channel ID and be in state INIT
+		if channel.PortId == portID && channel.ChannelId != channelID && channel.State == channeltypes.INIT {
+			newChannelActive = true
+		}
+	}
+	s.Require().True(newChannelActive, "a new channel should have been created")
+
+	// Verify the deposit record state was reverted
```
sampocs, line 183: *Verify the deposit record state was NOT reverted, because a non-delegation channel was restored

x/stakeibc/keeper/msg_server_restore_interchain_account_test.go
```diff
@@ -105,3 +124,66 @@ func (s *KeeperTestSuite) TestRestoreInterchainAccount_FailsIfAccountExists() {
...
+
+func (s *KeeperTestSuite) TestRestoreInterchainAccount_NoRecordChange_Success() {
+	tc := s.SetupRestoreInterchainAccount()
+	owner := "GAIA.WITHDRAWAL"
+	channelID := s.CreateICAChannel(owner)
+	portID := icatypes.PortPrefix + owner
+
+	// Confirm there are two channels originally
+	channels := s.App.IBCKeeper.ChannelKeeper.GetAllChannels(s.Ctx())
+	s.Require().Len(channels, 2, "there should be 2 channels initially (transfer + delegate)")
```
sampocs, line 155: *transfer + **withdrawal**

x/stakeibc/keeper/msg_server_restore_interchain_account_test.go
```diff
@@ -105,3 +124,66 @@ func (s *KeeperTestSuite) TestRestoreInterchainAccount_FailsIfAccountExists() {
...
+
+	// Close the delegation channel
```
sampocs, line 157: Close the **withdrawal** channel

x/stakeibc/keeper/msg_server_restore_interchain_account_test.go
```diff
@@ -105,3 +124,66 @@ func (s *KeeperTestSuite) TestRestoreInterchainAccount_FailsIfAccountExists() {
...
+	// Verify the deposit record state was reverted
+	for i := 0; i < 2; i++ {
+		depositRecord, found := s.App.RecordsKeeper.GetDepositRecord(s.Ctx(), uint64(i))
+		s.Require().True(found, "deposit record found")
+		s.Require().Equal(recordtypes.DepositRecord_DELEGATION_IN_PROGRESS, depositRecord.Status, "deposit record status should NOT nsg be reverted")
+	}
+}
+
+func (s *KeeperTestSuite) TestRestoreInterchainAccount_NoRecordChange_Success() {
+	tc := s.SetupRestoreInterchainAccount()
```
sampocs, line 148: i'd add a comment at the top of this test that says we're closing and restoring the withdrawal channel and checking that the deposit records do not get reverted

sampocs (approved):
Thanks for adding these tests! Couple nit guys

PR #359, opened 2022-11-18
Feature/airdrop

antstalepresh:
According to the frontend design, we are in need of defining a query msg which returns user vestings as a response and user spendable amount.
This update is for introducing a new query msg in airdrop module for returning all periods associated to the user vesting account.
And also, the update includes some changes on protobuf folder structure for claim and vesting module.
As we now have all protobuf files in proto/stride, I did same for claim and vesting protobuf files. 
And claim protobuf folder doesn't have the subfolder for versioning anymore.

## Brief Changelog
  - Moved proto/claim[vesting] folder into proto/stride/ folder.
  - Added new query msg for fetching user vestings and user spendable amount.
  - Added a corresponding CLI.

## Some information for tests
Here is the example CLI for UserVestings query
```
strided query claim user-vestings stride1vumz4z0ryrpjjt2wpge4dd0l4hzzlphgwufwp7
```


asalzmann (approved):
Thanks for making this change!

PR #373, opened 2022-11-19
Interchainquery Proto Cleanup

sampocs:
Closes: #XXX

The interchainquery module has unneeded functionality, which has lead to bloated / slightly confusing code. This PR trims down the protos and query submission logic. 

**Question for Reviewer**: Should we include upgrade logic to update any proto's that are already in the store? My thinking is that it's unnecessary because 1) it's unlikely that we'll have a query in the store at the time of an upgrade (it would mean there's a query that was never relayed) and 2) even if there is a pending query at the time of the update, we only have to wait until the next epoch for it to get re-requested 

## Brief Changelog
* Removed the `Height` field from the query proto since we always query at the latest height
* Removed the `Period` and `LastHeight` fields from the query proto since they're for used for issuing queries on a specified block-level cadence, and we already issue them epochly
* Added `RequestSent` field to track queries that are in flight
* Changed the query's ID to include the callback ID in the hash
* Build and store a new query object every time a query is requested (so if a query isn't handled and get's re-requested, the query object will get replaced with an updated TTL)

## Author's Checklist

I have...

## Reviewers Checklist

I have...

      - [ ] not applicable
      - [ ] jira ticket `XXX` 
      - [ ] README.md 
      - [ ] not documented


asalzmann (approved):
This is incredible! Looks great to me. Much easier to understand without the unused period / height concepts 

W.r.t migrating the store - I think that's right, although is there an edge case where the query is pending and we submit another query (with a different key), then both queries process in the same epoch? 

Something like
---- epoch n
*query with key k1
*upgrade
---- epoch n+1
*query with key k2
*submit query response for k1
*submit query response for k2

sampocs:
> W.r.t migrating the store - I think that's right, although is there an edge case where the query is pending and we submit another query (with a different key), then both queries process in the same epoch?

Good question. I think in that scenario, k1 will be timed out and the response will be ignored

asalzmann:
Ah yeah good point!

PR #384, opened 2022-11-22
v4/upgrade handler

hieuvubk:
Closes: #XXX

As part of the [v3](https://github.com/Stride-Labs/stride/releases/tag/v3.0.0) upgrade, we added [authz](https://github.com/Stride-Labs/stride/pull/352). However, we forgot to set the store key as part of the upgrade. As a consequence, after the upgrade, validators that attempted to prune had apphash errors. More details from @chillyvee [here](https://github.com/cosmos/cosmos-sdk/issues/13933#issuecomment-1326791953).

## Brief Changelog

- Add upgrade_tests to the v3 / v4 upgrade
- Initialize the params store for the claim and authz modules
- Add v4 upgrade handler and set the `authzkeeper.StoreKey`

## Author's Checklist

I have...

## Reviewers Checklist

I have...

      - [ ] not applicable
      - [ ] jira ticket `XXX` 
      - [ ] README.md 
      - [ ] not documented


faddat:
think it can work?

vuong177:
@hieuvubk we also need the authz module in `initParamsKeeper`

hieuvubk:
> @hieuvubk we also need the authz module in `initParamsKeeper`
thankss

hieuvubk:
I think we should keep upgrade v3 handler. Also, add `upgrade_test` to make sure no check upgrade handler

app/app.go
```diff
@@ -1002,6 +1002,8 @@ func initParamsKeeper(appCodec codec.BinaryCodec, legacyAmino *codec.LegacyAmino
 	paramsKeeper.Subspace(icacallbacksmoduletypes.ModuleName)
 	// this line is used by starport scaffolding # stargate/app/paramSubspace
 
+	paramsKeeper.Subspace(claimtypes.ModuleName)
+	paramsKeeper.Subspace(authz.ModuleName)
```
asalzmann, line 1006: Why is this required?

app/upgrades.go
```diff
@@ -39,7 +40,7 @@ func (app *StrideApp) setupUpgradeHandlers() {
 	// no store upgrades
 	case "v3":
 		storeUpgrades = &storetypes.StoreUpgrades{
-			Added: []string{claimtypes.StoreKey},
+			Added: []string{claimtypes.StoreKey, authzkeeper.StoreKey},
```
asalzmann, line 43: Is this a required step whenever a new module is added as part of an upgrade? Thanks for adding this, still learning how upgrades work

app/upgrades/v3/upgrades_test.go
```diff
@@ -0,0 +1,82 @@
...
+				plan, exists := suite.App.UpgradeKeeper.GetUpgradePlan(suite.Ctx())
+				suite.Require().True(exists)
+
+				suite.Context = suite.Ctx().WithBlockHeight(dummyUpgradeHeight)
+				suite.Require().NotPanics(func() {
+					beginBlockRequest := abci.RequestBeginBlock{}
+					suite.App.BeginBlocker(suite.Context, beginBlockRequest)
+				})
+
+				// make sure claim record was set
```
asalzmann, line 62: Is there any way to verify authz was properly initialized here?

asalzmann (changes requested):
We already ran the upgrade, so this will need to be a v4 upgrade. Thanks for doing this!

chillyvee:
> think it can work?
> 
> worried it already has a store key-- but frankly, it may work. Not sure.

@faddat - I traced the code path and I think this code SHOULD have worked if it was in the previous upgrade.

But like you, because the store key already has a height, I'm not sure if it will be willing to overwrite.

An interesting note is that I don't see any protection from adding a store during an upgrade THAT ALREADY EXISS

https://github.com/cosmos/cosmos-sdk/blob/main/store/rootmulti/store.go#L246
```
		// If it has been added, set the initial version
		if upgrades.IsAdded(key.Name()) {
			storeParams.initialVersion = uint64(ver) + 1
		}
```


That means as long as we're only adding much higher heights for that store name, there shouldn't be any key/height conflict and the new store should just go in.

Then on 

https://github.com/cosmos/cosmos-sdk/blob/main/store/rootmulti/store.go#L289

rs.stores = newStores

So isn't it reasonable to assume that any old heights from the same name store would be forgotten?

Even if the app tried to load the same store, the latest version would be the one just added during the upgrade.

app/upgrades.go
hieuvubk: this step is for new modules added to the upgrade, add module store to app multistore

app/app.go
hieuvubk: Need it to initialize params store for modules

hieuvubk: Unlike other data, module params are stored differently. The prefixed subspace of the parameter store is manage by `params` module. Some modules need their params private for another, we can modifier it

app/upgrades/v3/upgrades_test.go
hieuvubk: I tried to export module genesis then compare it with defaultGenState. Believe we do not modify the init state

hieuvubk:
> > think it can work?
> > worried it already has a store key-- but frankly, it may work. Not sure.
> 
> @faddat - Traced the code path and this code SHOULD have worked if it was in the previous upgrade.
> 
> But like you, because the store key already has a height, not sure if it will be willing to overwrite.
> 
> Keying in on `Added: []string{claimtypes.StoreKey, authzkeeper.StoreKey},`
> 
> An interesting note is that THERE ISN'T ANY PROTECTION from adding a store during an upgrade THAT ALREADY EXISTS (the one concern that we have)
> 
> https://github.com/cosmos/cosmos-sdk/blob/main/store/rootmulti/store.go#L246
> 
> ```
> 		// If it has been added, set the initial version
> 		if upgrades.IsAdded(key.Name()) {
> 			storeParams.initialVersion = uint64(ver) + 1
> 		}
> ```
> 
> Where ver is from https://github.com/cosmos/cosmos-sdk/blob/main/store/rootmulti/store.go#L183-L185
> 
> ```
> func (rs *Store) LoadLatestVersionAndUpgrade(upgrades *types.StoreUpgrades) error {
>         ver := GetLatestVersion(rs.db) <-- assuming this works
>         return rs.loadVersion(ver, upgrades)
> }               
> ```
> 
> And it should because that's a single unified key for the entire store https://github.com/cosmos/cosmos-sdk/blob/main/store/rootmulti/store.go#L1021-L1022
> 
> ```
> func GetLatestVersion(db dbm.DB) int64 {
> 	bz, err := db.Get([]byte(latestVersionKey))
> ```
> 
> That means as long as we're only adding much higher heights for that store name, there shouldn't be any key/height conflict and the new store should just go in. The invalid heights we see now are relatively much lower than the actual height, so there doesn't seem to be any way to create a height conflict.
> 
> Then on
> 
> https://github.com/cosmos/cosmos-sdk/blob/main/store/rootmulti/store.go#L289
> 
> ```
> rs.stores = newStores
> ```
> 
> So isn't it reasonable to assume that any old heights from the same name store would be forgotten?
> 
> Even if the app tried to load the same store, the latest version would be the one just added during the upgrade.
> 
> And YES, increment the version name to "v4"(?) for the upgrade so that it targets "the next upgrade"

Yeah it should be v4

> > think it can work?
> > worried it already has a store key-- but frankly, it may work. Not sure.
> 
> @faddat - Traced the code path and this code SHOULD have worked if it was in the previous upgrade.
> 
> But like you, because the store key already has a height, not sure if it will be willing to overwrite.
> 
> Keying in on `Added: []string{claimtypes.StoreKey, authzkeeper.StoreKey},`
> 
> An interesting note is that THERE ISN'T ANY PROTECTION from adding a store during an upgrade THAT ALREADY EXISTS (the one concern that we have)
> 
> https://github.com/cosmos/cosmos-sdk/blob/main/store/rootmulti/store.go#L246
> 
> ```
> 		// If it has been added, set the initial version
> 		if upgrades.IsAdded(key.Name()) {
> 			storeParams.initialVersion = uint64(ver) + 1
> 		}
> ```
> 
> Where ver is from https://github.com/cosmos/cosmos-sdk/blob/main/store/rootmulti/store.go#L183-L185
> 
> ```
> func (rs *Store) LoadLatestVersionAndUpgrade(upgrades *types.StoreUpgrades) error {
>         ver := GetLatestVersion(rs.db) <-- assuming this works
>         return rs.loadVersion(ver, upgrades)
> }               
> ```
> 
> And it should because that's a single unified key for the entire store https://github.com/cosmos/cosmos-sdk/blob/main/store/rootmulti/store.go#L1021-L1022
> 
> ```
> func GetLatestVersion(db dbm.DB) int64 {
> 	bz, err := db.Get([]byte(latestVersionKey))
> ```
> 
> That means as long as we're only adding much higher heights for that store name, there shouldn't be any key/height conflict and the new store should just go in. The invalid heights we see now are relatively much lower than the actual height, so there doesn't seem to be any way to create a height conflict.
> 
> Then on
> 
> https://github.com/cosmos/cosmos-sdk/blob/main/store/rootmulti/store.go#L289
> 
> ```
> rs.stores = newStores
> ```
> 
> So isn't it reasonable to assume that any old heights from the same name store would be forgotten?
> 
> Even if the app tried to load the same store, the latest version would be the one just added during the upgrade.
> 
> And YES, increment the version name to "v4"(?) for the upgrade so that it targets "the next upgrade"

yeah it should be v4

app/upgrades/v3/upgrades_test.go
faddat: that is what my new computer is for....  I'm going to export-upgrade -- and offer that to every chain that Notional validates before upgrades via ci.

faddat:
really happy to see everyone coming together on this in this manner.  

Thanks Stride, Hieu, Chilly

<3

app/app.go
asalzmann: Got it, thanks for the explanation!

app/upgrades.go
asalzmann: Thanks! Will keep this in mind in the future

app/upgrades/v3/upgrades_test.go
asalzmann: Notional consistently proving your value

app/upgrades/v3/upgrades_test.go
```diff
@@ -0,0 +1,82 @@
...
+			"Test that upgrade does not panic",
+			func() {
+				// Create pool 1
+				suite.Setup()
+			},
+			func() {
+				// run upgrade
+				// TODO: Refactor this all into a helper fn
+				
+				suite.Context = suite.Context.WithBlockHeight(dummyUpgradeHeight - 1)
```
sampocs, line 49: I don't think you need to set `Context` as an attribute on TestHelper here

```go
contextBeforeUpgrade := suite.Ctx().WithBlockHeight(dummyUpgradeHeight - 1)
contextAtUpgrade := suite.Ctx().WithBlockHeight(dummyUpgradeHeight)

plan := upgradetypes.Plan{Name: "v3", Height: dummyUpgradeHeight}
err := suite.App.UpgradeKeeper.ScheduleUpgrade(contextBeforeUpgrade, plan)
suite.Require().NoError(err)

plan, exists := suite.App.UpgradeKeeper.GetUpgradePlan(contextBeforeUpgrade)
suite.Require().True(exists)

suite.Require().NotPanics(func() {
	beginBlockRequest := abci.RequestBeginBlock{}
	suite.App.BeginBlocker(contextAtUpgrade, beginBlockRequest)
})
```

We can also add helper functions to iterate blocks

app/upgrades/v3/upgrades_test.go
```diff
@@ -0,0 +1,82 @@
...
+}
+
+func TestKeeperTestSuite(t *testing.T) {
+	suite.Run(t, new(UpgradeTestSuite))
+}
+
+func (suite *UpgradeTestSuite) TestUpgrade() {
+	testCases := []struct {
+		msg         string
+		pre_update  func()
```
sampocs, line 34: nit: camel case

app/upgrades/v4/upgrades_test.go
```diff
@@ -0,0 +1,83 @@
...
+}
+
+func TestKeeperTestSuite(t *testing.T) {
+	suite.Run(t, new(UpgradeTestSuite))
+}
+
+func (suite *UpgradeTestSuite) TestUpgrade() {
+	testCases := []struct {
+		msg         string
+		pre_update  func()
```
sampocs, line 35: nit: camel case

app/upgrades/v4/upgrades_test.go
```diff
@@ -0,0 +1,83 @@
...
+			"Test that upgrade does not panic",
+			func() {
+				// Create pool 1
+				suite.Setup()
+			},
+			func() {
+				// run upgrade
+				// TODO: Refactor this all into a helper fn
+				
+				suite.Context = suite.Context.WithBlockHeight(dummyUpgradeHeight - 1)
```
sampocs, line 50: same comment about Context as above

app/upgrades/v3/upgrades_test.go
```diff
@@ -0,0 +1,82 @@
...
+	}{
+		{
+			"Test that upgrade does not panic",
+			func() {
+				// Create pool 1
+				suite.Setup()
+			},
+			func() {
+				// run upgrade
+				// TODO: Refactor this all into a helper fn
```
sampocs, line 47: May as well move this to bottom of `test_helpers.go` now instead of adding it as a TODO

```go
func (s *AppTestHelper) ConfirmUpgradeSucceededs(upgradeName string, upgradeHeight int) {
   contextBeforeUpgrade := s.Ctx().WithBlockHeight(upgradeHeight - 1)
   contextAtUpgrade := s.Ctx().WithBlockHeight(upgradeHeight)

   plan := upgradetypes.Plan{Name: upgradeName, Height: upgradeHeight}
   err := s.App.UpgradeKeeper.ScheduleUpgrade(contextBeforeUpgrade, plan)
   s.Require().NoError(err)

   plan, exists := s.App.UpgradeKeeper.GetUpgradePlan(contextBeforeUpgrade)
   s.Require().True(exists)

   s.Require().NotPanics(func() {
	beginBlockRequest := abci.RequestBeginBlock{}
	s.App.BeginBlocker(contextAtUpgrade, beginBlockRequest)
   })
}
```

and then you can add it to the v3 and v4 test
```go
suite.ConfirmUpgradeSucceededs("v3", 5)
suite.ConfirmUpgradeSucceededs("v4", 5)
```

sampocs (commented):
Lgtm! Couple small comments about the unit test setup

app/upgrades/v3/upgrades_test.go
hieuvubk: Thank Sam this look good

asalzmann:
Addressing here https://github.com/Stride-Labs/stride/pull/393

PR #404, opened 2022-11-28
[DO NOT MERGE] Upgrade v4 Debug

sampocs:
* Setup dockernet by running the following:
```
make build-docker build=gr
```
* Then see the upgrade [README](https://github.com/Stride-Labs/stride/tree/main/scripts/upgrades#run-instructions) for instructions
* View stride logs in `scripts/logs/stride.log`

### DIFF
--- a/app/upgrades.go
+++ b/app/upgrades.go
@@ -6,11 +6,12 @@ import (
 	storetypes "github.com/cosmos/cosmos-sdk/store/types"
 	upgradetypes "github.com/cosmos/cosmos-sdk/x/upgrade/types"
 
+	authzkeeper "github.com/cosmos/cosmos-sdk/x/authz/keeper"
+
 	v2 "github.com/Stride-Labs/stride/v3/app/upgrades/v2"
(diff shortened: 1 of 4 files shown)


sampocs:
@faddat Not sure if this helps at all, but I tried a few combinations of adding/deleting the key at different heights. The heights were passed into `UpgradeStoreLoader`. Results were as follows:

| Deleted Key (Height) | Added Key (Height)      | Result    |
|----------------------|-------------------------|-----------|
| yes (upgradeHeight)  | no                      | SUCCEEDED |
| no                   | yes (upgradeHeight)     | FAILED    |
| yes (upgradeHeight)  | yes (upgradeHeight + 5) | SUCCEEDED |
| no                   | yes (upgradeHeight + 5) | SUCCEEDED |

However, I suspect passing a non-upgrade height was simply ignored since the handler is only executed at the upgrade height (supported by the last example in the table). Which would mean the successes were actually silent failures and these tests results are essentially meaningless... but I figured I'd share the results nonetheless!

PR #406, opened 2022-11-29
Add ctx app test helper

hieuvubk:
Closes: #XXX

App Context is not saved when tests are running. So I created a `Ctx` field in `AppTestHelper` to  store app context.


app/apptesting/test_helpers.go
```diff
@@ -149,6 +142,7 @@ func (s *AppTestHelper) CreateTransferChannel(hostChainID string) {
 	s.App = s.StrideChain.App.(*app.StrideApp)
 	s.HostApp = s.HostChain.GetSimApp()
 
+	s.Ctx = s.StrideChain.GetContext()
```
sampocs, line 145: ah yes, this a great way to do it! (nit comment: move it up a line)

```go
s.App = s.StrideChain.App.(*app.StrideApp)
s.HostApp = s.HostChain.GetSimApp()
s.Ctx = s.StrideChain.GetContext()

// Finally confirm the channel was setup properly
```

sampocs (commented):
This is so much cleaner!! Thanks for doing this! This actually opens the door for two additional cleanups:
1. I think we can now remove the `IbcEnabled` attribute from the `AppTestHelper` struct. And then in `CreateTransferChannel`, instead of checking `if !s.IbcEnabled`, you can just check`if s.HostChain == nil`
2. You can remove the `HostCtx` function. It's not actually used anywhere. **If** we ever find that we need the host zone context, then we can add that as a separate attribute to the struc, but let's hold off for now cause I can't think of reason why we would ever need that.

DongLieu:
@sampocs Hi I try to using `s.HostChain == nil` but seem not as we expected. I'll dig into it later

sampocs:
> @sampocs Hi I try to using s.HostChain == nil but seem not as we expected. I'll dig into it later
Ah sorry, actually I think I just remembered that it doesn't work because it's a pointer and the tests are running in parallel. Keeping IbcEnabled in there is fine

sampocs (approved):

PR #446, opened 2022-12-03
IF-STRIDE-STAKEIBC-REDEEM_STAKE

asalzmann:
Users may not be able to redeem stake

# Involved artifacts

- [x/stakeibc/keeper/msg_server_redeem_stake.go](https://github.com/Stride-Labs/stride/blob/v2.0.3/x/stakeibc/keeper/msg_server_redeem_stake.go#L51-L53)

# Description

Users can always redeem from Stride. When they select "redeeem" on the Stride website, Stride will initiate unbonding on the host zone. Once the unbonding period elapses, the users will receive native tokens in their wallets.

However, users cannot unstake an amount greater than staked balance on the host zone.

```go
if msg.Amount > hostZone.StakedBal {
		return nil, sdkerrors.Wrapf(types.ErrInvalidAmount, "cannot unstake an amount g.t. staked balance on host zone: %d", msg.Amount)
	}
```

The problem with the code above is in comparing two different units; `msg.Amount` represents the input **stTokens** amount, while on the other hand `hostZone.StakedBal` describes host zone's total amount of **native** tokens.

# Problem Scenarios

An `ErrInvalidAmount` error could be returned even if the user's redeeming amount doesn't exceed total host zone's staked amount. That's because stTokens are not 1-1 alligned with native tokens. There is `redemptionRate` defined as `nativeToken amount / stToken amount` which can fluctuate between 0.9 and 1.5 for current stride implementation, hard-coded in [stakeibc/types/params.go](https://github.com/Stride-Labs/stride/blob/v2.0.3/x/stakeibc/types/params.go#L25-L26).

# Recommendation

Before comparing with `hostZone.StakedBal`, `msg.Amount` should be exchanged to the native tokens:

```go
nativeAmount := sdk.NewDec(msg.Amount).Mul(hostZone.RedemptionRate).RoundInt()
```
After that, validation check for amount would be correct: 

```go
if nativeAmount > hostZone.StakedBal {
		return nil, sdkerrors.Wrapf(types.ErrInvalidAmount, "cannot unstake an amount g.t. staked balance on host zone: %d", msg.Amount)
	}
```


x/stakeibc/keeper/msg_server_redeem_stake.go
```diff
@@ -48,15 +48,22 @@ func (k msgServer) RedeemStake(goCtx context.Context, msg *types.MsgRedeemStake)
...
-		return nil, sdkerrors.Wrapf(types.ErrInvalidAmount, "cannot unstake an amount g.t. staked balance on host zone: %d", msg.Amount)
-	}
-
+	// construct desired unstaking amount from host zone
 	amt, err := cast.ToInt64E(msg.Amount)
 	if err != nil {
 		k.Logger(ctx).Error(fmt.Sprintf("error casting RedeemStake msg.Amount to int64, err: %s", err.Error()))
 		return nil, sdkerrors.Wrapf(types.ErrInvalidAmount, fmt.Sprintf("invalid amount: %s", err.Error()))
 	}
+	coinDenom := types.StAssetDenomFromHostZoneDenom(hostZone.HostDenom)
```
sampocs, line 57: nit: should we rename this to `stDenom` ?

sampocs (approved):
lgtm!

x/stakeibc/keeper/msg_server_redeem_stake.go
asalzmann: Done

PR #447, opened 2022-12-03
IF-STRIDE-STAKEIBC-UPDATEDELEGATIONBALANCES fix

asalzmann:
Underflow possible when decreasing staked balance for host zone

# Involved artifacts

- [x/stakeibc/keeper/icacallbacks_undelegate.go](https://github.com/Stride-Labs/stride/blob/v2.0.3/x/stakeibc/keeper/icacallbacks_undelegate.go#L128)

# Description

When undelegating from each validator, host zone's staked balance is being decreased according to undelegation amount of particular validator:

```go
for _, undelegation := range undelegateCallback.SplitDelegations {
		undelegateAmt, err := cast.ToInt64E(undelegation.Amount)
		k.Logger(ctx).Info(fmt.Sprintf("UndelegateCallback, Undelegation: %d, validator: %s", undelegateAmt, undelegation.Validator))
		if err != nil {
			errMsg := fmt.Sprintf("Could not convert undelegate amount to int64 in undelegation callback | %s", err.Error())
			k.Logger(ctx).Error(errMsg)
			return sdkerrors.Wrapf(types.ErrIntCast, errMsg)
		}
		undelegateVal := undelegation.Validator
		success := k.AddDelegationToValidator(ctx, zone, undelegateVal, -undelegateAmt)
		if !success {
			return sdkerrors.Wrapf(types.ErrValidatorDelegationChg, "Failed to remove delegation to validator")
		}
		zone.StakedBal -= undelegation.Amount
	}
```

However, `zone.StakedBal` is defined as `uint64`; therefore, it might underflow if `undelegation.Amount` > `zone.StakedBal`. Undelegation amount for each undelegation is calculated in [GetTargetValAmtsForHostZone](https://github.com/Stride-Labs/stride/blob/v2.0.3/x/stakeibc/keeper/unbonding_records.go#L74) and have been additionaly modified during [overflow](https://github.com/Stride-Labs/stride/blob/v2.0.3/x/stakeibc/keeper/unbonding_records.go#L103-L139) handling.

# Problem Scenarios

If underflow happens, staked balance of particular zone will unexpectedly and quietly become tremendously large having adverse consequences during both, `UndelegateCallback` and `DelegateCallback`.

# Recommendation

Add validation check before decreasing the staked balance amount: 

```go
  if undelegation.Amount > zone.StakedBal {
  // handle incoming underflow
  } else {
    zone.StakedBal -= undelegation.Amount
  }
```


sampocs (commented):
Lgtm! but looks like a unit test is broken

sampocs (approved):

PR #450, opened 2022-12-03
V4 changelog

asalzmann:
changelog


sampocs (approved):
I'll let this one slide but in the future we need to add test cases

Issue #419, opened 2022-11-30
AUDIT: Different conditions for max number of validators to rebalance


asalzmann:
_Surfaced from @informalsystems audit of [Stride](https://github.com/Stride-Labs/stride) at commit f9736fe0_

---
finding:
  id: IF-STRIDE-STAKEIBC-REBALANCEVALIDATORS
  title: Different conditions for max number of validators to rebalance
  severity: Informational
  impact: None
  exploitability: None
  type: Implementation
  issue: 
  status: Unresolved
---

# Involved artifacts

- [x/stakeibc/keeper/msg_server_rebalance_validators.go](https://github.com/Stride-Labs/stride/blob/v2.0.3/x/stakeibc/keeper/msg_server_rebalance_validators.go#L48-L55)
- [x/stakeibc/types/message_rebalance_validators.go](https://github.com/Stride-Labs/stride/blob/v2.0.3/x/stakeibc/types/message_rebalance_validators.go#L53-L55)

# Description

There are two different places in the code where conditions for maximum number of validators to rebalance are checked and they don't really match. First one is in the function [ValidateBasic()](https://github.com/Stride-Labs/stride/blob/v2.0.3/x/stakeibc/types/message_rebalance_validators.go#L53-L55):
```go
if (msg.NumRebalance < 1) || (msg.NumRebalance > 10) {
		return sdkerrors.Wrapf(sdkerrors.ErrInvalidRequest, fmt.Sprintf("invalid number of validators to rebalance (%d)", msg.NumRebalance))
	}
```
 and the other one is part of [RebalanceValidators()](https://github.com/Stride-Labs/stride/blob/v2.0.3/x/stakeibc/keeper/msg_server_rebalance_validators.go#L48-L55) implementation.

```go
if maxNumRebalance < 1 {
		k.Logger(ctx).Error(fmt.Sprintf("Invalid number of validators to rebalance %d", maxNumRebalance))
		return nil, types.ErrInvalidNumValidator
	}
	if maxNumRebalance > 4 {
		k.Logger(ctx).Error(fmt.Sprintf("Invalid number of validators to rebalance %d", maxNumRebalance))
		return nil, types.ErrInvalidNumValidator
	}
```
"Magic numbers" are also used inside if conditions.

# Problem Scenarios
`MsgRebalanceValidators` will be processed successfully by  `ValidateBasic()` for any NumRebalance value in {5..10}, but it will fail later on msg execution. 

# Recommendation

[...]

PR #455, opened 2022-12-04
Hoang/check num of val to rebalance

hoangdv2429:
Closes: #419

I also added test for code written in validateBasic of stakeibc, however in order to get it to pass. You will have to change bech32prefix from cosmos to stride (I wrote some comments describe the detail in the code)

## Brief Changelog

Move num val checking from msg_server to validateBasic and add test for it.

## Author's Checklist

I have...

- [x] Run and PASSED locally all Stride integration tests
- [x] If the change is contentful, I either:
    - [x] Added test cases to existing unit tests

I also added test for code written in validateBasic of stakeibc, however in order to get it to pass. You will have to change bech32prefix from cosmos to stride (I wrote some comments describe the detail in the code)

## Reviewers Checklist

I have...

      - [ ] not applicable
      - [ ] jira ticket `XXX` 
      - [ ] README.md 
      - [ ] not documented


hieuvubk:
@hoangdv2429 hey could u run `make test-unit` and check, seem some test are failing

hoangdv2429:
> @hoangdv2429 hey could u run `make test-unit` and check, seem some test are failing

I updated it with app params. all test should pass now.

x/stakeibc/types/message_rebalance_validators_test.go
```diff
@@ -27,6 +32,27 @@ func TestMsgRebalanceValidators_ValidateBasic(t *testing.T) {
 				Creator: sample.AccAddress(),
 			},
 			err: sdkerrors.ErrInvalidAddress,
+		}, {
```
riley-stride, line 35: @sampocs do you think we should move these tests back to `x/stakeibc/keeper/msg_server_rebalance_validators_test.go` so that we have consistency in our testing approach across the codebase?

x/stakeibc/types/message_rebalance_validators.go
```diff
@@ -10,6 +10,8 @@ import (
 )
 
 const TypeMsgRebalanceValidators = "rebalance_validators"
+const MIN_NUM_REBALANCE = 1
+const MAX_NUM_REBALANCE = 10
```
riley-stride, line 14: If you set these constants here, shouldn't you reference them on line 55, rather than leaving the hardcoded values?

riley-stride (changes requested):
Thanks for opening this PR. A few questions and requests. 

1) Defining MIN and MAX as constants makes sense, though I think making them params would give the protocol a bit more flexibility. 

2) Would like to get @sampocs's review on this and his opinion on the testing framework choice.

3) Seems like you are still using the hardcoded values in the ValidateBasic check in `message_rebalance_validators.go`. 

4) Why are you adding the `app/params` directory?

sampocs:
@hoangdv2429 @riley-stride 
Considering the check is now in Validate basic, I think moving it to the testing in types makes sense since our framework is meant for keeper functions. That said, I'm not sure all those files are necessary to just to get the address prefix to work. 

I handled something similar to this in a separate branch that hasn't been merged yet. I'll pull it out into a new PR and you can pull that into this.

sampocs:
For what it's worth, my philosophy on testing (since it's come up across a couple of PRs): 

The "array-of-test-cases" approach is good for simple go tests like those in types and ValidateBasic, _where the input to each test case is small._

However, when it comes to our keeper functions, there is often a lot of state that needs to be mocked out and each test can have various other conditions and branches (for instance, you can imagine an ICA unit test where you might want to close a channel in the middle of a test). In those scenarios, I think the "array-of-test-cases" approach quickly becomes very messy as it's difficult to generalize logic across each test case tests. Plus, the test function itself becomes bloated, which makes it hard to follow what's going on. So in those scenarios, I think the "function-per-test-case" approach makes sense as you see with most of our keeper functions.

All that said, I support the decision to move these tests over in this PR! Curious to hear your guys' thoughts as well!

hieuvubk:
@riley-stride @sampocs Thank u for leaving comment!
Yes I totally agree that we should move MIN & MAX to params.
@sampocs Thank u for helper func its help me so much

hieuvubk:
Also if we use MIN & MAX as params. We have to move the check from `ValidateBasic` to keeper func (of course we can't get module params in `types`)
Pls lmk what u think @riley-stride @sampocs

hoangdv2429:
@sampocs @riley-stride  can you review this PR ? If there is anything wrong with it we'll try to change it asap. Thank you!

sampocs:
Sorry for the delay on this! Would you mind merging in main and resolving the conflicts and then I'll give it another review!

hoangdv2429:
@sampocs sorry I didn't see your message earlier, I merged and resolved all conflicts. Can you have a look

proto/stride/stakeibc/params.proto
```diff
@@ -33,5 +33,7 @@ message Params {
   uint64 safety_max_redemption_rate_threshold = 15;
   uint64 ibc_transfer_timeout_nanos = 16;
   uint64 safety_num_validators = 17;
-  uint64 safety_max_slash_percent = 18;
+  uint64 min_num_rebalance = 18;
+  uint64 max_num_rebalance = 19;
+  uint64 safety_max_slash_percent = 20;
```
sampocs, line 38: Lets keep the slash percent as the same index to make the migration cleaner
```suggestion
  uint64 safety_max_slash_percent = 18;
  uint64 min_num_rebalance = 19;
  uint64 max_num_rebalance = 20;
```

x/stakeibc/keeper/msg_server_rebalance_validators.go
```diff
@@ -26,15 +26,14 @@ func (k msgServer) RebalanceValidators(goCtx context.Context, msg *types.MsgReba
...
-		k.Logger(ctx).Error(fmt.Sprintf("Invalid number of validators to rebalance %d", maxNumRebalance))
-		return nil, types.ErrInvalidNumValidator
-	}
-	if maxNumRebalance > 4 {
-		k.Logger(ctx).Error(fmt.Sprintf("Invalid number of validators to rebalance %d", maxNumRebalance))
-		return nil, types.ErrInvalidNumValidator
+
+	params := k.GetParams(ctx)
+	if (msg.NumRebalance < params.MinNumRebalance) || (msg.NumRebalance > params.MaxNumRebalance) {
+		return nil, fmt.Errorf("invalid number of validators to rebalance (%d)", msg.NumRebalance)
```
sampocs, line 32:
```suggestion
		return nil, errorsmod.Wrapf(types.ErrInvalidNumValidator, "invalid number of validators to rebalance (%d)", msg.NumRebalance)
```

x/stakeibc/keeper/msg_server_rebalance_validators.go
```diff
@@ -26,15 +26,14 @@ func (k msgServer) RebalanceValidators(goCtx context.Context, msg *types.MsgReba
...
 	}
+	
+	maxNumRebalance := cast.ToInt(msg.NumRebalance)
```
sampocs, line 35: You can remove this line and change the forloop below (where maxNumRebalance is used) to 

```
for i := uint64(1); i <= maxNumRebalance; i++ {
```

x/stakeibc/types/params.go
```diff
@@ -27,6 +27,8 @@ var (
 	DefaultMaxStakeICACallsPerEpoch         uint64 = 100
 	DefaultIBCTransferTimeoutNanos          uint64 = 1800000000000 // 30 minutes
 	DefaultSafetyNumValidators              uint64 = 35
+	DefaultMinNumRebalance					uint64 = 1
```
sampocs, line 30: can you double check the linting in this file? the spacing seems off

x/stakeibc/types/message_rebalance_validators.go
```diff
@@ -52,7 +50,7 @@ func (msg *MsgRebalanceValidators) ValidateBasic() error {
 		return err
 	}
 	if (msg.NumRebalance < 1) || (msg.NumRebalance > 10) {
```
sampocs, line 52: This should use the same parameters that were just added

sampocs (commented):
Good stuff! Some nits comments below. One large point:

The main issue for this ticket was that checks in validate basic and the msg_server_rebalance_vals should be consistent, but if the min/max are stored as params in the store, there's no way to grab them from validate basic. 

I think the two options are:
1. Remove the check from validate basic 
2. Use a true golang constant instead of a module param (as recommended in the audit)

I think my vote is to go with option 2 as it's little simpler and then we also don't need a migration for this param. Curious to here @riley-stride's thoughts

x/stakeibc/types/params.go
hoangdv2429: idk why but it's look fine on my side though 
image

hoangdv2429:
@sampocs I made the changes as suggested. Can you have a look ?

sampocs:
Closing in favor of #630 so the min/max variables are represented as constants instead of params.
The audit report suggested just checking in one place (validate basic) and validate basic does not have access to the store, so it makes more sense to just use constants.
This also prevents the need for a store migration

Issue #422, opened 2022-11-30
AUDIT: Coding style recommendations


asalzmann:
_Surfaced from @informalsystems audit of [Stride](https://github.com/Stride-Labs/stride) at commit f9736fe0_

---
finding:
  id: IF-STRIDE-CODINGSTYLE
  title: Coding style recommendations
  severity: Informational
  impact: None
  exploitability: None
  type: Implementation
  issue: 
  status: Unresolved
---

# Involved artifacts

- [x/stakeibc/keeper/msg_server_redeem_stake.go line 129-135](https://github.com/Stride-Labs/stride/blob/v2.0.3/x/stakeibc/keeper/msg_server_redeem_stake.go#L129-L135)
- [x/stakeibc/keeper/host_zone.go](https://github.com/Stride-Labs/stride/blob/v2.0.3/x/stakeibc/keeper/host_zone.go#L101-L121)
- [x/stakeibc/keeper/hooks.go line58](https://github.com/Stride-Labs/stride/blob/v2.0.3/x/stakeibc/keeper/hooks.go#L58)
- [x/stakeibc/keeper/hooks.go line 167](https://github.com/Stride-Labs/stride/blob/v2.0.3/x/stakeibc/keeper/hooks.go#L167)
- [x/stakeibc/keeper/msg_server_redeem_stake.go line 23-26](https://github.com/Stride-Labs/stride/blob/v2.0.3/x/stakeibc/keeper/msg_server_redeem_stake.go#L23-L26)
- [x/stakeibc/keeper/icacallbacks_reinvest.go](https://github.com/Stride-Labs/stride/blob/v2.0.3/x/stakeibc/keeper/icacallbacks_reinvest.go#L61-L62)

# Description

- Redundant casting:
```go
    stTokenAmount, err := cast.ToUint64E(msg.Amount)
	if err != nil {
		errMsg := fmt.Sprintf("Could not convert redemption amount to int64 in redeem stake | %s", err.Error())
		k.Logger(ctx).Error(errMsg)
		return nil, sdkerrors.Wrapf(types.ErrIntCast, errMsg)
	}
	hostZoneUnbonding.StTokenAmount += stTokenAmount
```
Both `msg.Amount` and `hostZoneUnbonding.StTokenAmount` are already `uint64`.

- Redundant `else` keyword:
```go
    if amt >= 0 {
        amt, err := cast.ToUint64E(amt)
        if err != nil {
[...]

PR #461, opened 2022-12-05
remove uneeded casting, refractor transfer from mint

sontrinh16:
Closes: #422

Fix some audited coding style 

## Author's Checklist

I have...

- [x] remove some unnecessary casting to uint64
- [x] remove some unnecessary else statement
- [ ] remove err check in RedeemStake
- [ ] Change type name for ReinvestAmount
- [x] Move send stCoins logic to LiquidStake instead of MintAsset


sontrinh16:
i dont think we can remove the address err check in redeemstake and liquidstake without fixing the test cases right now

sampocs (commented):
Thanks for opening this! I agree on moving the code out of the else statement. 

Can you elaborate on the motivation behind moving the transfer out of the mint function? If it's because the name is unclear, you could just rename it to `MintAndTransferStAsset`

sontrinh16:
i think that function should only have 1 responsibility that is minting as it describe, the transfer should be moving out of the funtion. Hmmm but yeah rename it would be clearer too

asalzmann (approved):
lgtm

sampocs (approved):

sontrinh16:
Base on 2 approves, i will merge the PR

PR #474, opened 2022-12-06
Cleaned Up Middleware Stack Code

sampocs:
Closes: #XXX

Now that the middleware stacks have grown since inception, their definition in `app.go` has become slightly confusing for someone not familiar with middleware. This PR does a simple cleanup to try and rename/ organize in a way that's slightly easier to follow.

**Question for reviewer**: I'm not super familiar with the purpose of the IBC routing. Why do we not add a route for `recordstypes.ModuleName -> recordsStack`? I assumed that's because packets never flow directly to records, but if that was the case, I wouldn't expect a route for stakeibc either.

## Brief Changelog
* Variable renaming / re-ordering / commenting to follow 3 stacks: ICAHost, ICAController, and Records

## Author's Checklist

I have...

## Reviewers Checklist

I have...

      - [ ] not applicable
      - [ ] jira ticket `XXX` 
      - [ ] README.md 
      - [ ] not documented

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -559,7 +559,8 @@ func NewStrideApp(
 	)
 	icaModule := ica.NewAppModule(&app.ICAControllerKeeper, &app.ICAHostKeeper)
 
-	// Define the IBC stacks
+	
+// Define the IBC middleware stacks
 	// Stack one (ICAHost Stack) contains:
 	// - IBC
 	// - ICAHost
--- a/app/app.go
+++ b/app/app.go
@@ -5,8 +5,6 @@ import (
 	"os"
 	"path/filepath"
 
-	porttypes "github.com/cosmos/ibc-go/v3/modules/core/05-port/types"
-
 	"github.com/Stride-Labs/stride/v4/utils"
 
 	"github.com/cosmos/cosmos-sdk/baseapp"
@@ -561,35 +559,43 @@ func NewStrideApp(
 	)
 	icaModule := ica.NewAppModule(&app.ICAControllerKeeper, &app.ICAHostKeeper)
 
-	// Create the middleware stacks
-
-	// Stack one contains
+	// Define the IBC stacks
+	// Stack one (ICAHost Stack) contains:
 	// - IBC
-	// - ICA
-	// - icacallbacks
-	// - stakeibc
+	// - ICAHost
 	// - base app
-	var icamiddlewareStack porttypes.IBCModule
-	icamiddlewareStack = icacallbacksmodule.NewIBCModule(app.IcacallbacksKeeper, stakeibcIBCModule)
-	icamiddlewareStack = icacontroller.NewIBCModule(app.ICAControllerKeeper, icamiddlewareStack)
 	icaHostIBCModule := icahost.NewIBCModule(app.ICAHostKeeper)
 
-	// Stack two contains
+	// Stack two (ICAController Stack) contains
 	// - IBC
-	// - records
+	// - ICAController (middleware)
+	// - icacallbacks (middleware)
+	// - stakeibc
+	// - base app
+	icaCallbacksStack := icacallbacksmodule.NewIBCModule(app.IcacallbacksKeeper, stakeibcIBCModule)
+	icaControllerStack := icacontroller.NewIBCModule(app.ICAControllerKeeper, icaCallbacksStack)
+
+	// Stack three (Records Stack) contains
+	// - IBC
+	// - records (middleware)
 	// - transfer
 	// - base app
 	recordsStack := recordsmodule.NewIBCModule(app.RecordsKeeper, transferIBCModule)
 
 	// Create static IBC router, add transfer route, then set and seal it
+	// Note, authentication module packets are routed to the top level of the middleware stack
 	ibcRouter := ibcporttypes.NewRouter()
 	ibcRouter.
-		AddRoute(ibctransfertypes.ModuleName, recordsStack).
-		AddRoute(icacontrollertypes.SubModuleName, icamiddlewareStack).
+		// ICAHost Stack
 		AddRoute(icahosttypes.SubModuleName, icaHostIBCModule).
-		// Note, authentication module packets are routed to the top level of the middleware stack
-		AddRoute(stakeibcmoduletypes.ModuleName, icamiddlewareStack).
-		AddRoute(icacallbacksmoduletypes.ModuleName, icamiddlewareStack)
+		// ICAController Stack
+		AddRoute(icacontrollertypes.SubModuleName, icaControllerStack).
+		AddRoute(icacallbacksmoduletypes.ModuleName, icaControllerStack).
+		AddRoute(stakeibcmoduletypes.ModuleName, icaControllerStack).
+		// Records Stack
+		AddRoute(ibctransfertypes.ModuleName, recordsStack)
+		// QUESTION: Why don't we have a router for recordstypes.ModuleName -> recordsStack ?
+
 	// this line is used by starport scaffolding # ibc/app/router
 	app.IBCKeeper.SetRouter(ibcRouter)
 


app/app.go
```diff
@@ -561,35 +559,43 @@ func NewStrideApp(
 	)
 	icaModule := ica.NewAppModule(&app.ICAControllerKeeper, &app.ICAHostKeeper)
 
-	// Create the middleware stacks
+	// Define the IBC stacks
```
asalzmann, line 562:
```suggestion
	// Define the IBC middleware stacks
```

app/app.go
```diff
@@ -561,35 +559,43 @@ func NewStrideApp(
...
+	// - ICAHost
+	// - base app
+	icaHostIBCModule := icahost.NewIBCModule(app.ICAHostKeeper)
 
-	// Stack one contains
+	// Stack two (ICAController Stack) contains
 	// - IBC
-	// - ICA
-	// - icacallbacks
+	// - ICAController (middleware)
```
asalzmann, line 571: ICAController, icacallbacks and stakeibc are all middlewares

app/app.go
```diff
@@ -561,35 +559,43 @@ func NewStrideApp(
...
 	// - base app
-	var icamiddlewareStack porttypes.IBCModule
-	icamiddlewareStack = icacallbacksmodule.NewIBCModule(app.IcacallbacksKeeper, stakeibcIBCModule)
-	icamiddlewareStack = icacontroller.NewIBCModule(app.ICAControllerKeeper, icamiddlewareStack)
-	icaHostIBCModule := icahost.NewIBCModule(app.ICAHostKeeper)
+	icaCallbacksStack := icacallbacksmodule.NewIBCModule(app.IcacallbacksKeeper, stakeibcIBCModule)
+	icaControllerStack := icacontroller.NewIBCModule(app.ICAControllerKeeper, icaCallbacksStack)
 
-	// Stack two contains
+	// Stack three (Records Stack) contains
```
asalzmann, line 578: We should probably call this the transfer stack, for clarity

app/app.go
```diff
@@ -561,35 +559,43 @@ func NewStrideApp(
...
+	// - ICAController (middleware)
+	// - icacallbacks (middleware)
 	// - stakeibc
 	// - base app
-	var icamiddlewareStack porttypes.IBCModule
-	icamiddlewareStack = icacallbacksmodule.NewIBCModule(app.IcacallbacksKeeper, stakeibcIBCModule)
-	icamiddlewareStack = icacontroller.NewIBCModule(app.ICAControllerKeeper, icamiddlewareStack)
-	icaHostIBCModule := icahost.NewIBCModule(app.ICAHostKeeper)
+	icaCallbacksStack := icacallbacksmodule.NewIBCModule(app.IcacallbacksKeeper, stakeibcIBCModule)
+	icaControllerStack := icacontroller.NewIBCModule(app.ICAControllerKeeper, icaCallbacksStack)
```
asalzmann, line 576: nit: is there a nice way to format this as a single line of code? something like
```
icamiddlewarestack := icacontroller.NewIBCModule(app.ICAControllerKeeper,
                                    icacallbacksmodule.NewIBCModule(app.IcacallbacksKeeper,
                                    stakeibcIBCModule)
```
this orders the stack in a visually intuitive way (not sure if the formatting works)

app/app.go
```diff
@@ -561,35 +559,43 @@ func NewStrideApp(
...
-	// - records
+	// - records (middleware)
 	// - transfer
 	// - base app
 	recordsStack := recordsmodule.NewIBCModule(app.RecordsKeeper, transferIBCModule)
 
 	// Create static IBC router, add transfer route, then set and seal it
+	// Note, authentication module packets are routed to the top level of the middleware stack
 	ibcRouter := ibcporttypes.NewRouter()
 	ibcRouter.
```
asalzmann, line 588: We should double check that changing the route order here won't cause unexpected changes in how packets are routed

app/app.go
```diff
@@ -561,35 +559,43 @@ func NewStrideApp(
...
-		AddRoute(ibctransfertypes.ModuleName, recordsStack).
-		AddRoute(icacontrollertypes.SubModuleName, icamiddlewareStack).
+		// ICAHost Stack
 		AddRoute(icahosttypes.SubModuleName, icaHostIBCModule).
-		// Note, authentication module packets are routed to the top level of the middleware stack
```
asalzmann, line 590: nit: leave this note in / update with a clarifying comment (I think it answers the question for the reviewer you had)

asalzmann (changes requested):
Couple of minor comments. This code is a bit finicky and not well-tested - should we add tests as part of this PR?

> Question for reviewer: I'm not super familiar with the purpose of the IBC routing. Why do we not add a route for recordstypes.ModuleName -> recordsStack? I assumed that's because packets never flow directly to records, but if that was the case, I wouldn't expect a route for stakeibc either.

The reason is that the middleware stacks for ICA modules in ibc-go v3 are designed in a confusing way. Typically, IBC middlewares sit on _top_ of other modules. For example, when we route incoming transfer packets, they are passed to the `recordsStack`, which contains the records module, which calls the transfer module.

However, ICAController authentication modules (like `stakeibc`) wrap ICAController. So, incoming ICA packets are routed to the stakeibc module, then to the ICAController module. This was flagged as confusing and addressed in ibc-go v5 (it's backwards compatible afaiu, so nothing in our current design should break).

app/app.go
sampocs: talked offline, ordering shouldn't matter since its just adding items to a map

app/app.go
sampocs: I just moved it up to the line above `ibcRouter := ibcporttypes.NewRouter()`

app/app.go
sampocs: my thinking was we should label each stack from either the top or bottom but be consistent about it. Which would mean we either do ICAControllerStack and RecordsStack, or StakeibcStack and TransferStack.

I thought using the top of the stack made the code slightly cleaner since you define them bottom up. What do you think?

app/app.go
sampocs: I'm not sure there's a way to do that cleanly. Ideally, we'd flop the order of arguments if the NewIBCModule functions so that each one shows `NewIBCModule(sameStack, newKeeper)` but we can't change the ICA ones.

The main change here I was trying to get at was changing the naming of `icamiddlewarestack` since that doesn't differentiate between ICAHost or ICAController.

Depending on how we want to name the stack (see earlier comment about ICAController vs Stakeibc), we can do something similar to how you had it before with
```go
var icaControllerStack ibcporttypes.IBCModule = stakeibcIBCModule
icaControllerStack = icacallbacksmodule.NewIBCModule(app.IcacallbacksKeeper, icaControllerStack)
icaControllerStack = icacontroller.NewIBCModule(app.ICAControllerKeeper, icaControllerStack)
```

sampocs:
@asalzmann, I'll wait until we have the app-router and ratelimit merged in, but I think we should go with the following format:

```go
var stakeibcStack ibcporttypes.IBCModule = stakeibcIBCModule
stakeibcStack = icacallbacksmodule.NewIBCModule(app.IcacallbacksKeeper, stakeibcStack)
stakeibcStack = icacontroller.NewIBCModule(app.ICAControllerKeeper, stakeibcStack)


var transferStack ibcporttypes.IBCModule = transferIBCModule
transferStack = recordsmodule.NewIBCModule(app.RecordsKeeper, transferStack)
transferStack = ratelimitmodule.NewIBCModule(app.RateLimitKeeper, transferStack)
transferStack = routermodule.NewIBCModule(app.AppRouterKeeper, transferStack)
```

sampocs:
Open

sampocs:
closing in favor of #618

Issue #437, opened 2022-12-02
RATE-LIMITER: Add Path type and helper functions


sampocs:
Subticket of #414 

## Suggested Design
* Spec for Path datastructure (persisted in store)
```
Attributes
    Id string (f'{Denom}-{ChannelId}')
    Denom string (native denom)
    ChannelId string

Functions
    AddPath(denom, channel) → string (id)
        Adds to store using the ID as a prefix, returns an ID string
    RemovePath(id)
       Deletes from store
    GetPath(id) → path 
    GetAllPaths() → List[path]
    IsNative() → bool 
       Returns true if denom is ustrd, false otherwise
    GetIbcDenomHash()
    GetChannelValue() [LEAVE UNIMPLEMENTED]
```

## Acceptance Criteria
* Unit tests for all functions

PR #479, opened 2022-12-08
[RATE-LIMIT] Path helper functions

sampocs:
Closes: #437 

## Brief Changelog
* Added path keeper functions and unit tests


x/ratelimit/keeper/path.go
```diff
@@ -0,0 +1,78 @@
...
+	"github.com/Stride-Labs/stride/v4/x/ratelimit/types"
+)
+
+func FormatPathId(denom string, channelId string) string {
+	return fmt.Sprintf("%s-%s", denom, channelId)
+}
+
+func (k Keeper) AddPath(ctx sdk.Context, path types.Path) string {
+	store := ctx.KVStore(k.storeKey)
+	prefixStore := prefix.NewStore(store, types.KeyPrefix(types.PathKey))
```
antstalepresh: It might make more sense to change `PathKey` to `PathKeyPrefix`

x/ratelimit/keeper/path.go
```diff
@@ -0,0 +1,78 @@
...
+
+func (k Keeper) GetPath(ctx sdk.Context, pathId string) (path types.Path, found bool) {
+	store := ctx.KVStore(k.storeKey)
+	prefixStore := prefix.NewStore(store, types.KeyPrefix(types.PathKey))
+
+	pathKey := types.KeyPrefix(pathId)
+	pathBytes := prefixStore.Get(pathKey)
+
+	path = types.Path{}
+	err := proto.Unmarshal(pathBytes, &path)
```
antstalepresh: What about using k.cdc.Unmarshal() instead of proto.Unmarshal()?

x/ratelimit/keeper/path.go
```diff
@@ -0,0 +1,78 @@
...
+	prefixStore := prefix.NewStore(store, types.KeyPrefix(types.PathKey))
+
+	iterator := prefixStore.Iterator(nil, nil)
+	defer iterator.Close()
+
+	allPaths := []types.Path{}
+	for ; iterator.Valid(); iterator.Next() {
+
+		path := types.Path{}
+		err := proto.Unmarshal(iterator.Value(), &path)
```
antstalepresh: What about using k.cdc.Unmarshal() instead of proto.Unmarshal()?

x/ratelimit/types/path.go
```diff
@@ -1,9 +1,10 @@
 package types
 
-func (p *Path) IsNative() bool {
-	return p.Denom == "ustrd" // TODO: Replace with param
-}
+import "strings"
 
-func (p *Path) GetIBCDenomHash() {
+// QUESTION FOR REVIEWER: Should I place this here or in keeper/path.go?
```
antstalepresh: I think types/path.go is the right place.

x/ratelimit/keeper/path_test.go
```diff
@@ -1,99 +1,50 @@
...
 
-func (s *KeeperTestSuite) TestAddPath() {
-	// for _, path := range paths {
-	// 	s.App.RatelimitKeeper.AddPath(s.Ctx, path.TraceDenom, path.ChannelId)
-	// }
-	// for _, path := range paths {
-	// 	s.App.RatelimitKeeper.GetPath(s.Ctx, path.)
-	// }
+	actualPath, found := s.App.RatelimitKeeper.GetPath(s.Ctx, expectedPath.Id)
+	s.Require().True(found, "element found")
```
antstalepresh, line 32: should change `element found` to `element not found`

antstalepresh (changes requested):
perfect! I left a few comments.

x/ratelimit/keeper/path.go
sampocs: agreed, updated

x/ratelimit/keeper/path.go
sampocs: hmm, I think this is slightly outdated. I'm looking at the most recent commit and I used k.cdc.Unmarshal

x/ratelimit/keeper/path.go
sampocs: same thing here - I think I had changed it in a later commit

x/ratelimit/keeper/path_test.go
sampocs: So my think was this field is usually shown more as a _description_ than an _error_. Meaning, when it fails, rather than saying: `error: element found`, it shows showing like:

```
element found:
   expected: True
   actual: False
```

And I've found that sometimes putting in an error message for the description can make it a little confusing, ex:
```
element not found:
   expected: True
   actual: False
```

What do you think? Is that what you've seen in your experience or do you think it makes more sense to use it like an error message?

antstalepresh: Afaik, the second param indicates the error message.
Here is the example log when the first param is false.
```
quota_test.go:30: 
                Error Trace:    /Volumes/Data/stride/x/ratelimit/keeper/quota_test.go:30
                Error:          Should be true
                Test:           TestKeeperTestSuite/TestGetQuota
                Messages:       element not found
```
This means that the second param indicates the error message rather than a description.

sampocs: I see. I still find it slightly confusing with the boolean value. How bout I just be more explicit with something like:

"element should have been found, but was not"

antstalepresh (approved):

PR #483, opened 2022-12-08
add  msg types and routes, add creator field to tx msgs

antstalepresh:
Missing msg routes and types are blocking chain initialization and this PR is for those fixes.

## Brief Changelog
  - *Add creator fields to all tx msgs for GetSigners function*
  - *Add tx msg types*
  - *Add tx msg handlers*

### DIFF
--- a/x/ratelimit/types/msgs.go
+++ b/x/ratelimit/types/msgs.go
@@ -1,0 +1,195 @@
+package types
+
+import (
+	sdk "github.com/cosmos/cosmos-sdk/types"
+)
+
+// Msg type for MsgAddQuota
+const TypeMsgAddQuota = "add_quota"
+
+var _ sdk.Msg = &MsgAddQuota{}
+
+func NewMsgAddQuota(name string, maxPercentSend uint64, maxPercentRecv uint64, durationMinutes uint64) *MsgAddQuota {
+	return &MsgAddQuota{
+		Name:            name,
+		MaxPercentSend:  maxPercentSend,
+		MaxPercentRecv:  maxPercentRecv,
+		DurationMinutes: durationMinutes,
+	}
+}
+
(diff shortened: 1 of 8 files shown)


x/ratelimit/types/msgs.go
```diff
@@ -0,0 +1,195 @@
...
+)
+
+// Msg type for MsgAddQuota
+const TypeMsgAddQuota = "add_quota"
+
+var _ sdk.Msg = &MsgAddQuota{}
+
+func NewMsgAddQuota(name string, maxPercentSend uint64, maxPercentRecv uint64, durationMinutes uint64) *MsgAddQuota {
+	return &MsgAddQuota{
+		Name:            name,
```
sampocs, line 14: Should the `Creator` field be included in these `NewMsg` functions?

x/ratelimit/types/msgs.go
```diff
@@ -0,0 +1,195 @@
...
+func (msg *MsgRemoveQuota) Type() string {
+	return TypeMsgRemoveQuota
+}
+
+func (msg *MsgRemoveQuota) GetSigners() []sdk.AccAddress {
+	allocator, err := sdk.AccAddressFromBech32(msg.Creator)
+	if err != nil {
+		panic(err)
+	}
+	return []sdk.AccAddress{allocator}
```
sampocs, line 71:
```suggestion
	creator, err := sdk.AccAddressFromBech32(msg.Creator)
	if err != nil {
		panic(err)
	}
	return []sdk.AccAddress{creator}
```

sampocs (commented):
Looks great! Only one comment and one question - repeated for each message type:

x/ratelimit/types/msgs.go
antstalepresh: Yep, added it!

PR #488, opened 2022-12-08
implement quota functions

antstalepresh:
## Brief Changelog

Implemented quota keeper functions
 
  - AddQuota
  - RemoveQuota
  - GetQuota
  - GetAllQuotas
  - IsExpired
  - CheckExceedsQuota


x/ratelimit/types/msgs.go
```diff
@@ -41,7 +42,14 @@ func (msg *MsgAddQuota) GetSignBytes() []byte {
...
 func (msg *MsgAddQuota) ValidateBasic() error {
-	// TODO:
+	_, err := sdk.AccAddressFromBech32(msg.Creator)
+	if err != nil {
+		return sdkerrors.Wrapf(sdkerrors.ErrInvalidAddress, "invalid creator address (%s)", err)
+	}
+
+	if msg.Name == "" {
+		return sdkerrors.Wrapf(sdkerrors.ErrInvalidRequest, "quota name not set")
+	}
```
sampocs, line 52: should we validate the other fields here as well?

x/ratelimit/types/msgs.go
```diff
@@ -41,7 +42,14 @@ func (msg *MsgAddQuota) GetSignBytes() []byte {
 }
 
 func (msg *MsgAddQuota) ValidateBasic() error {
```
sampocs, line 44: can you add unit tests for these `ValidateBasic` functions?

x/ratelimit/keeper/quota.go
```diff
@@ -0,0 +1,65 @@
...
+
+import (
+	"github.com/cosmos/cosmos-sdk/store/prefix"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+
+	"github.com/Stride-Labs/stride/v4/x/ratelimit/types"
+)
+
+// Add new quota to the store
+func (k Keeper) AddQuota(ctx sdk.Context, quota types.Quota) {
```
sampocs, line 11: My apologies, I had this wrong in the spec, this should probably be `SetQuota` instead of `AddQuota`

x/ratelimit/keeper/quota.go
```diff
@@ -0,0 +1,65 @@
+package keeper
+
+import (
+	"github.com/cosmos/cosmos-sdk/store/prefix"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+
+	"github.com/Stride-Labs/stride/v4/x/ratelimit/types"
+)
+
```
sampocs, line 9: can you add unit tests for these?

x/ratelimit/keeper/quota.go
```diff
@@ -0,0 +1,65 @@
...
+}
+
+// Remove quota from the store
+func (k Keeper) RemoveQuota(ctx sdk.Context, name string) {
+	store := prefix.NewStore(ctx.KVStore(k.storeKey), types.KeyPrefix(types.QuotaKeyPrefix))
+	store.Delete([]byte(name))
+}
+
+// Get quota from the store
+func (k Keeper) GetQuota(ctx sdk.Context, name string) (quota types.Quota) {
```
sampocs, line 25: Might make sense to have to this also return a bool that indicates if it was found

x/ratelimit/keeper/quota.go
```diff
@@ -0,0 +1,65 @@
...
+	}
+
+	k.cdc.MustUnmarshal(b, &quota)
+	return quota
+}
+
+// Get all quotas from the store
+func (k Keeper) GetAllQuotas(ctx sdk.Context) (list []types.Quota) {
+	store := prefix.NewStore(ctx.KVStore(k.storeKey), types.KeyPrefix(types.QuotaKeyPrefix))
+	iterator := sdk.KVStorePrefixIterator(store, []byte{})
```
sampocs, line 40: just curious - is this the same as doing `store.Iterator(nil, nil)`? (that's what I've seen before)

x/ratelimit/keeper/quota.go
```diff
@@ -0,0 +1,65 @@
...
+	return ctx.BlockTime().Unix() > int64(quota.PeriodEnd)
+}
+
+// CheckExceedsQuota checks if new in/out flow is going to reach the max in/out or not
+func (k Keeper) CheckExceedsQuota(direction types.PacketDirection, amount uint64, totalValue uint64, quota types.Quota) bool {
+	if direction == types.PACKET_RECV {
+		return amount > totalValue*quota.MaxPercentRecv/100
+	}
+
+	return amount > totalValue*quota.MaxPercentSend/100
```
sampocs, line 64: does it make more sense to put this in `quota.types` extending the quota object? 
So then you can call it directly from the quota object

sampocs (changes requested):
Looks great! Couple comments below

x/ratelimit/types/msgs.go
antstalepresh: That might be needed but not clear for now.
I am gonna put TODO comment there for now and we can discuss later.

x/ratelimit/types/msgs.go
antstalepresh: ok

x/ratelimit/keeper/quota.go
antstalepresh: ok.

x/ratelimit/keeper/quota.go
antstalepresh: ok.

x/ratelimit/keeper/quota.go
antstalepresh: yes, its just same thing.

x/ratelimit/keeper/quota.go
antstalepresh: ok. I like that.

x/ratelimit/types/msgs.go
antstalepresh: ok

proto/stride/ratelimit/tx.proto
```diff
@@ -5,7 +5,7 @@ option go_package = "github.com/Stride-Labs/stride/v4/x/ratelimit/types";
 
 // Msg defines the Msg service.
 service Msg {
-  rpc AddQuota(MsgAddQuota) returns (MsgAddQuotaResponse);
+  rpc SetQuota(MsgSetQuota) returns (MsgSetQuotaResponse);
```
sampocs, line 8: so sorry for misleading you. The Message should stay as `MsgAddQuota` - I just mean the keeper function can be changed to `SetQuota`

So in the message implementation, it would do something like:

```
func MsgAddQutoa(quota) {
    _, found := k.GetQuota(quota.Name)
    if found {
       error already exists
    }
    k.SetQuota(quota)
}

```

x/ratelimit/types/msgs.go
```diff
@@ -50,6 +50,15 @@ func (msg *MsgAddQuota) ValidateBasic() error {
 	if msg.Name == "" {
 		return sdkerrors.Wrapf(sdkerrors.ErrInvalidRequest, "quota name not set")
 	}
+
+	if msg.MaxPercentRecv == 0 && msg.MaxPercentSend == 0 {
```
sampocs, line 54: We should also validate that they're less than or equal to 100

x/ratelimit/types/msgs_test.go
```diff
@@ -0,0 +1,140 @@
...
+
+	"github.com/Stride-Labs/stride/v4/app/apptesting"
+	cmdcfg "github.com/Stride-Labs/stride/v4/cmd/strided/config"
+	"github.com/Stride-Labs/stride/v4/x/ratelimit/types"
+)
+
+func TestMsgSetQuota(t *testing.T) {
+	cmdcfg.SetupConfig()
+	validAddr, invalidAddr := apptesting.GenerateTestAddrs()
+	tests := []struct {
```
sampocs, line 16: great tests!

x/ratelimit/keeper/quota.go
```diff
@@ -22,16 +22,16 @@ func (k Keeper) RemoveQuota(ctx sdk.Context, name string) {
 }
 
 // Get a quota from the store using quota name
-func (k Keeper) GetQuota(ctx sdk.Context, name string) (quota types.Quota) {
+func (k Keeper) GetQuota(ctx sdk.Context, name string) (quota types.Quota, found bool) {
 	store := prefix.NewStore(ctx.KVStore(k.storeKey), types.KeyPrefix(types.QuotaKeyPrefix))
 
 	b := store.Get([]byte(name))
 	if b == nil {
```
sampocs, line 29: do we also want to check for an empty byte array?

```
if b == nil || len(b) == 0
```
```

File: x/ratelimit/keeper/quota_test.go [hunk 12]
```diff
@@ -0,0 +1,48 @@
...
+	s.Require().Equal(expectedQuota, actualQuota)
+}
+
+func (s *KeeperTestSuite) TestRemoveQuota() {
+	quotas := s.createQuotas()
+	idToRemove := quotas[0].Name
+
+	s.App.RatelimitKeeper.RemoveQuota(s.Ctx, idToRemove)
+	_, found := s.App.RatelimitKeeper.GetQuota(s.Ctx, idToRemove)
+	s.Require().False(found, "removed element found")
```
sampocs, line 40: as discussed, lets make these error messages a little more specific

sampocs (changes requested):
Fantastic unit tests on the messages! My apologies if I wasn't clear about the `AddQuota` → `SetQuota` (more explained below)

x/ratelimit/types/quota.go
```diff
@@ -0,0 +1,10 @@
+package types
+
+// CheckExceedsQuota checks if new in/out flow is going to reach the max in/out or not
+func (q *Quota) CheckExceedsQuota(direction PacketDirection, amount uint64, totalValue uint64) bool {
```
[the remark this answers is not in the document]
antstalepresh: ok

x/ratelimit/types/message_update_rate_limit_test.go
```diff
@@ -0,0 +1,145 @@
...
+				DurationHours:  validDurationHours,
+			},
+			err: "invalid denom",
+		},
+		{
+			name: "invalid channel-id",
+			msg: types.MsgUpdateRateLimit{
+				Creator:        validAddr,
+				Denom:          validDenom,
+				ChannelId:      "chan-0",
```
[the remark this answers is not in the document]
antstalepresh: ok

x/ratelimit/types/quota_test.go
```diff
@@ -0,0 +1,61 @@
...
+		name       string
+		direction  types.PacketDirection
+		amount     uint64
+		totalValue uint64
+		exp        bool
+	}{
+		{
+			name:       "inflow exceed",
+			direction:  types.PACKET_RECV,
+			amount:     200*quota.MaxPercentRecv/100 + 1,
```
sampocs, line 28: Excellent tests! Small comment, might make sense to change to amount to a hard coded number so it's easier on the understand what's happening:

I'd recommend setting the total value to 100 for each of these, and then you can set the amounts to 5 and 15, for "not exceeded", and "exceeded" respectively

You could even define variables right under the quota object like so:
```
totalValue := 100
amountUnderThreshold := 5
amountOverThreshold := 15
```
and then just use those variables in each test case

antstalepresh: ok

sampocs (approved):

Issue #441, opened 2022-12-03
RATE-LIMITER: Add RateLimit type and helper functions 


sampocs:
Subticket of #414 
Blocked by #440, #437, #439 

## Suggested Design
* Spec for RateLimit data structure (persisted in store)
```
Attributes
	Path path
	Flow flow
	Quota quota

Functions
	AddRateLimit(path, flow, quota) 
            Adds a rate limit object to the store, using the pathId as the key

        RemoveRateLimit(pathId)
            Removes from the store 

	GetRateLimit(pathId)
            Grabs a RateLimit object from the store using the path id as the key and returns it

	GetAllRateLimits()
           Iterates through each rate limit object and returns a list
```

## Acceptance Criteria
* RateLimit proto definition
* RateLimit proto-generated go file
* Helper functions implemented
* Unit tests for all functions

Issue #443, opened 2022-12-03
RATE-LIMITER: Add Queries


sampocs:
Subticket of #414 

## Background
* Add the following queries
  * Paths: Returns all path objects
  * Path/{Id}: Returns a path given the path ID ({Denom-Channel})
  * RateLimits: Returns all rate limit objects
  * RateLimit/{PathId}: Returns a rate limit object given the path ID

## Acceptance Criteria
* Query keeper functions
* CLI commands
* Unit tests
* Tested with dockernet

PR #490, opened 2022-12-08
[RATE-LIMIT]: RateLimit Helper Functions and Queries

sampocs:
Addresses: #443 
Closes: #441 

## Brief Changelog
* Added RateLimit keeper functions and queries

### DIFF
--- a/x/ratelimit/keeper/rate_limit.go
+++ b/x/ratelimit/keeper/rate_limit.go
@@ -15,7 +15,7 @@ func CheckRateLimit(direction types.PacketDirection, packet exported.PacketI) er
 
 // Stores/Updates a rate limit object in the store
 func (k Keeper) SetRateLimit(ctx sdk.Context, rateLimit types.RateLimit) {
-	store := prefix.NewStore(ctx.KVStore(k.storeKey), types.KeyPrefix(types.RateLimitKey))
+	store := prefix.NewStore(ctx.KVStore(k.storeKey), types.RateLimitKeyPrefix)
 
 	rateLimitKey := types.KeyPrefix(rateLimit.Path.Id)
 	rateLimitValue := k.cdc.MustMarshal(&rateLimit)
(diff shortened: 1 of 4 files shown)


x/ratelimit/types/keys.go
```diff
@@ -12,3 +12,11 @@ const (
 	// QuerierRoute defines the module's query routing key
 	QuerierRoute = ModuleName
 )
+
+func KeyPrefix(prefix string) []byte {
+	return []byte(prefix)
+}
+
+const (
+	RateLimitKey = "rate-limit"
```
antstalepresh, line 21: It makes more sense to change `RateLimitKey` to `RateLimitKeyPrefix`

x/ratelimit/keeper/rate_limit_test.go
```diff
@@ -0,0 +1,45 @@
...
+	}
+	return rateLimits
+}
+
+func (s *KeeperTestSuite) TestGetRateLimit() {
+	rateLimits := s.createRateLimits()
+	expectedRateLimit := rateLimits[0]
+
+	actualRateLimit, found := s.App.RatelimitKeeper.GetRateLimit(s.Ctx, expectedRateLimit.Path.Id)
+	s.Require().True(found, "element found")
```
antstalepresh, line 27: second param for True() should be "element not found" as that param is to show error message.

x/ratelimit/keeper/rate_limit.go
```diff
@@ -1,13 +1,64 @@
...
+}
+
+// Grabs and returns a rate limit object from the store using the PathId
+func (k Keeper) GetRateLimit(ctx sdk.Context, pathId string) (rateLimit types.RateLimit, found bool) {
+	store := prefix.NewStore(ctx.KVStore(k.storeKey), types.KeyPrefix(types.RateLimitKey))
+
+	rateLimitKey := types.KeyPrefix(pathId)
+	rateLimitValue := store.Get(rateLimitKey)
+
+	if rateLimitValue == nil || len(rateLimitValue) == 0 {
```
antstalepresh, line 40: This checking causes a golang lint error.
`rateLimitValue == nil` is same as `len(rateLimitValue) == 0` so its ok to check only nil value.

antstalepresh (changes requested):
Great work! A few comments here

x/ratelimit/keeper/rate_limit_test.go
sampocs: discussed in the other PR - once we agree on the best approach, I'll update

antstalepresh (approved):

PR #502, opened 2022-12-12
Move Quota Attributes to RateLimit object

sampocs:
Closes: #XXX

## Author's Checklist

I have...

## Reviewers Checklist

I have...

      - [ ] not applicable
      - [ ] jira ticket `XXX` 
      - [ ] README.md 
      - [ ] not documented


x/ratelimit/client/cli/query.go
```diff
@@ -22,8 +24,72 @@ func GetQueryCmd() *cobra.Command {
...
 	)
 	return cmd
 }
+
+// GetCmdQueryRateLimit implements a command to query a specific rate limit
+func GetCmdQueryRateLimit() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "rate-limit [denom] [channel-id]",
+		Short: "Query a specific rate limit",
+		Args:  cobra.ExactArgs(1),
```
antstalepresh, line 38: I think the number of arguments is set wrongly.
There are same issues for some other query/tx CLIs.
Please check.

x/ratelimit/keeper/rate_limit.go
```diff
@@ -8,6 +8,11 @@ import (
 	"github.com/Stride-Labs/stride/v4/x/ratelimit/types"
 )
 
+// QUESTION FOR REVIEWER: Is this the right way to do a composite key?
```
antstalepresh, line 11: Correct!

x/ratelimit/types/message_add_rate_limit.go
```diff
@@ -0,0 +1,74 @@
...
+	_, err := sdk.AccAddressFromBech32(msg.Creator)
+	if err != nil {
+		return sdkerrors.Wrapf(sdkerrors.ErrInvalidAddress, "invalid creator address (%s)", err)
+	}
+
+	if msg.Denom == "" {
+		return sdkerrors.Wrapf(sdkerrors.ErrInvalidRequest, "invalid denom")
+	}
+
+	if !strings.HasPrefix(msg.ChannelId, "channel-") {
```
antstalepresh, line 57: I think we should check one more case that channelId is exactly _channel-_.
What do you think about checking length of channel id too? 
For example, we can accept channel ids that length is bigger than 8 so that it should always look like this `channel-xxx`

proto/stride/ratelimit/ratelimit.proto
```diff
@@ -20,46 +20,25 @@ enum FlowDirection {
...
-  string id = 1;
-
-  // path base_denom - represents the underlying denom (ignoring ibc traces)
-  string base_denom = 2;
-
-  // path channelid
-  string channel_id = 3;
+  string denom = 1;
+  string channel_id = 2;
+  string chain_id = 3;
```
antstalepresh, line 25: Any specific reason to store chain_id?

antstalepresh (changes requested):
Overall great! Some feedbacks here

x/ratelimit/client/cli/query.go
sampocs: great catch! fixing now

x/ratelimit/types/message_add_rate_limit.go
sampocs: Agreed, I'll add a regex check

proto/stride/ratelimit/ratelimit.proto
antstalepresh: Good point and agree with that!

antstalepresh (approved):
Approved!

PR #507, opened 2022-12-13
[RATE-LIMIT] Added Epochs to ratelimit module

sampocs:
Closes: #XXX

## Brief Changelog
* Added epochs module
* Added beginEpoch function
* Added HOUR_EPOCH type


x/ratelimit/keeper/rate_limit.go
```diff
@@ -42,7 +47,7 @@ func (k Keeper) GetRateLimit(ctx sdk.Context, denom string, channelId string) (r
 	rateLimitKey := GetRateLimitItemKey(denom, channelId)
 	rateLimitValue := store.Get(rateLimitKey)
 
-	if rateLimitValue == nil || len(rateLimitValue) == 0 {
```
antstalepresh, line 45: Any specific reason for why not checking nil value?

antstalepresh (changes requested):
Great! left some comments

x/ratelimit/keeper/rate_limit.go
antstalepresh: Ok.

antstalepresh (approved):

PR #515, opened 2022-12-15
[RATE-LIMIT] Implement CheckRateLimit, SendPacket, OnRecvPacket

sampocs:
Closes: #XXX

* Moved ResetRateLimiter to keeper function
* Implemented CheckRateLimit, SendPacket, and OnRecvPacket 
* Added helper functions: ParseDenomFromSendPacket and ParseDenomFromRecvPacket
* Added ratelimit keeper to ICS4Wrapper and removed ICS4Wrapper from rate limit module

### DIFF
--- a/x/ratelimit/keeper/rate_limit.go
+++ b/x/ratelimit/keeper/rate_limit.go
@@ -21,8 +21,6 @@ func (k Keeper) GetChannelValue(ctx sdk.Context, denom string) sdk.Int {
 	return k.bankKeeper.GetSupply(ctx, denom).Amount
 }
 
-// QUESTION FOR REVIEWER: Where should these two functions live?
-
 // Parse the denom from the Send Packet that will be used by the rate limit module
 // The denom that the rate limiter will use for a SEND packet depends on whether
 //    it was a NATIVE token (e.g. ustrd, stuatom, etc.) or NON-NATIVE token (e.g. ibc/...)...
@@ -48,7 +46,7 @@ func ParseDenomFromSendPacket(packet transfertypes.FungibleTokenPacketData) (den
 	return denom
 }
 
-// Parse the denom from the Send Packet that will be used by the rate limit module
+// Parse the denom from the Recv Packet that will be used by the rate limit module
 // The denom that the rate limiter will use for a RECIEVE packet depends on whether it was a source or sink
 // 		Source: The packet's is being recieved by a chain it was just sent from (i.e. the token has gone back and forth)
 //              (e.g. strd is sent -> to osmosis -> and then back to stride)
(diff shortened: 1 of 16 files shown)


x/ratelimit/keeper/rate_limit.go
```diff
@@ -19,13 +21,126 @@ func (k Keeper) GetChannelValue(ctx sdk.Context, denom string) sdk.Int {
...
+		denom = packet.Denom
+	} else {
+		// Non-native assets should be hashed
+		denom = denomTrace.IBCDenom()
+	}
+
+	return denom
+}
+
+// Parse the denom from the Send Packet that will be used by the rate limit module
```
antstalepresh, line 51: Parse the denom from the Send Packet that will be used by the rate limit module
```suggestion
// Parse the denom from the Recv Packet that will be used by the rate limit module
```

antstalepresh (changes requested):
Awesome! just one comment issue.

x/ratelimit/keeper/rate_limit.go
```diff
@@ -19,13 +21,126 @@ func (k Keeper) GetChannelValue(ctx sdk.Context, denom string) sdk.Int {
 	return k.bankKeeper.GetSupply(ctx, denom).Amount
 }
 
-func CheckRateLimit(direction types.PacketDirection, packet exported.PacketI) error {
-	// TODO
-	return nil
+// QUESTION FOR REVIEWER: Where should these two functions live?
```
antstalepresh, line 24: I would prefer to move these two functions into utils/utils.go.
What do you think?

sampocs: I thought about this a bit, and I actually think it might make sense to move ParseDenomFromSendPacket, ParseDenomFromRecvPacket, SendPacket and OnRecvPacket to a separate file called `rate_limit.go` that's at the same level as module.go (since they're not actually keeper functions). I'll take a quick pass at this and push up another commit and we can always revert it if you think it doesn't make sense!

x/ratelimit/keeper/rate_limit.go
sampocs: done

x/ratelimit/keeper/rate_limit.go
sampocs: @antstalepresh  Just pushed it up, see the latest commits, and let me know what you think!

antstalepresh (approved):

x/ratelimit/keeper/rate_limit.go
antstalepresh: That makes sense!

x/ratelimit/keeper/rate_limit.go
antstalepresh: Great!

PR #517, opened 2022-12-15
fix some cli config issues, add tests for clis

antstalepresh:
## Brief Changelog

  - *Fix some cli config issues.*
  - *Add tests for rate limiter clis*

### DIFF
--- a/x/ratelimit/client/cli/cli_test.go
+++ b/x/ratelimit/client/cli/cli_test.go
@@ -1,0 +1,318 @@
+package cli_test
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/cosmos/cosmos-sdk/client/flags"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+
+	clitestutil "github.com/cosmos/cosmos-sdk/testutil/cli"
+
+	strideclitestutil "github.com/Stride-Labs/stride/v4/testutil/cli"
+
+	"github.com/Stride-Labs/stride/v4/testutil/network"
+
+	"github.com/stretchr/testify/suite"
+
+	tmcli "github.com/tendermint/tendermint/libs/cli"
+
+	"github.com/Stride-Labs/stride/v4/x/ratelimit/client/cli"
+
+	"github.com/Stride-Labs/stride/v4/app"
+	"github.com/Stride-Labs/stride/v4/x/ratelimit/types"
+)
+
+type IntegrationTestSuite struct {
+	suite.Suite
+
+	cfg     network.Config
+	network *network.Network
+}
+
+func (s *IntegrationTestSuite) SetupSuite() {
+	s.T().Log("setting up integration test suite")
+
+	s.cfg = network.DefaultConfig()
+
+	genState := app.ModuleBasics.DefaultGenesis(s.cfg.Codec)
+	ratelimitGenState := types.DefaultGenesis()
+	ratelimitGenStateBz := s.cfg.Codec.MustMarshalJSON(ratelimitGenState)
+	genState[types.ModuleName] = ratelimitGenStateBz
+
+	s.cfg.GenesisState = genState
+	s.network = network.New(s.T(), s.cfg)
+
+	_, err := s.network.WaitForHeight(1)
+	s.Require().NoError(err)
+}
+
+func (s *IntegrationTestSuite) TestCmdTxAddRateLimit() {
+	val := s.network.Validators[0]
+	quota := types.Quota{
+		MaxPercentSend: 10,
+		MaxPercentRecv: 20,
+		DurationHours:  1,
+	}
+	path := types.Path{
+		Denom:     sdk.DefaultBondDenom,
+		ChannelId: "channel-0",
+	}
+
+	testCases := []struct {
+		name     string
+		args     []string
+		expQuota types.Quota
+	}{
+		{
+			"add-rate-limit tx",
+			[]string{
+				sdk.DefaultBondDenom,                    // denom
+				path.ChannelId,                          // channelId
+				fmt.Sprintf("%d", quota.MaxPercentSend), // maxPercentSend
+				fmt.Sprintf("%d", quota.MaxPercentRecv), // maxPercentRecv
+				fmt.Sprintf("%d", quota.DurationHours),  // durationHours
+				fmt.Sprintf("--%s=json", tmcli.OutputFlag),
+				fmt.Sprintf("--%s=%s", flags.FlagFrom, val.Address.String()),
+				// common args
+				fmt.Sprintf("--%s=true", flags.FlagSkipConfirmation),
+				fmt.Sprintf("--%s=%s", flags.FlagBroadcastMode, flags.BroadcastBlock),
+				strideclitestutil.DefaultFeeString(s.cfg),
+			},
+			quota,
+		},
+	}
+
+	for _, tc := range testCases {
+		tc := tc
+
+		s.Run(tc.name, func() {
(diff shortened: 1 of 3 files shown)


x/ratelimit/client/cli/cli_test.go
```diff
@@ -0,0 +1,318 @@
...
+				fmt.Sprintf("%d", quota.DurationHours),  // durationHours
+				fmt.Sprintf("--%s=json", tmcli.OutputFlag),
+				fmt.Sprintf("--%s=%s", flags.FlagFrom, val.Address.String()),
+				// common args
+				fmt.Sprintf("--%s=true", flags.FlagSkipConfirmation),
+				fmt.Sprintf("--%s=%s", flags.FlagBroadcastMode, flags.BroadcastBlock),
+				strideclitestutil.DefaultFeeString(s.cfg),
+			})
+			s.Require().NoError(err)
+
```
sampocs, line 244: is there a way to send a transfer message here so the flow can go:
1. Add rate limit
2. Transfer 
3. Check that flow matches transfer
4. Reset
5. Check that flow is 0

?

x/ratelimit/client/cli/cli_test.go
```diff
@@ -0,0 +1,318 @@
...
+			},
+			rateLimit,
+		},
+	}
+
+	for _, tc := range testCases {
+		tc := tc
+
+		s.Run(tc.name, func() {
+			cmd := cli.CmdRemoveRateLimit()
```
sampocs, line 298: I think we should add a rate limit first before testing the remove, something like:
1. Attempt to remove a rate limit (should fail)
2. Add a rate limit
3. Remove that rate limit (should succeed)
4. Check that rate limit was actually removed

Might also make sense to bundle the add rate limit into its own function that you can call from this and ResetRateLimit

x/ratelimit/client/cli/cli_test.go
```diff
@@ -0,0 +1,318 @@
...
+				fmt.Sprintf("--%s=json", tmcli.OutputFlag),
+			})
+			s.Require().NoError(err)
+
+			var result types.RateLimit
+			s.Require().NoError(clientCtx.Codec.UnmarshalJSON(out.Bytes(), &result))
+			s.Require().Equal(tc.expQuota, *result.Quota)
+		})
+	}
+}
```
sampocs, line 110: For all these tests, since there's only one element in the array, can you remove the array and remove the s.Run?

```suggestion
func (s *IntegrationTestSuite) TestCmdTxAddRateLimit() {
	val := s.network.Validators[0]
	expectedQuota := types.Quota{
		MaxPercentSend: 10,
		MaxPercentRecv: 20,
		DurationHours:  1,
	}
	path := types.Path{
		Denom:     sdk.DefaultBondDenom,
		ChannelId: "channel-0",
	}

	args := []string{
		sdk.DefaultBondDenom,                    // denom
		path.ChannelId,                          // channelId
		fmt.Sprintf("%d", quota.MaxPercentSend), // maxPercentSend
		fmt.Sprintf("%d", quota.MaxPercentRecv), // maxPercentRecv
		fmt.Sprintf("%d", quota.DurationHours),  // durationHours
		fmt.Sprintf("--%s=json", tmcli.OutputFlag),
		fmt.Sprintf("--%s=%s", flags.FlagFrom, val.Address.String()),
		// common args
		fmt.Sprintf("--%s=true", flags.FlagSkipConfirmation),
		fmt.Sprintf("--%s=%s", flags.FlagBroadcastMode, flags.BroadcastBlock),
		strideclitestutil.DefaultFeeString(s.cfg),
	}

	cmd := cli.CmdAddRateLimit()
	clientCtx := val.ClientCtx

	_, err := clitestutil.ExecTestCLICmd(clientCtx, cmd, args)
	s.Require().NoError(err)

	// Check if ratelimit was set properly
	cmd = cli.GetCmdQueryRateLimit()
	out, err := clitestutil.ExecTestCLICmd(clientCtx, cmd, []string{
		path.Denom,
		path.ChannelId,
		fmt.Sprintf("--%s=json", tmcli.OutputFlag),
	})
	s.Require().NoError(err)

	var result types.RateLimit
	s.Require().NoError(clientCtx.Codec.UnmarshalJSON(out.Bytes(), &result))
	s.Require().Equal(expectedQuota, *result.Quota)
}
```

x/ratelimit/client/cli/cli_test.go
```diff
@@ -0,0 +1,318 @@
...
+	genState[types.ModuleName] = ratelimitGenStateBz
+
+	s.cfg.GenesisState = genState
+	s.network = network.New(s.T(), s.cfg)
+
+	_, err := s.network.WaitForHeight(1)
+	s.Require().NoError(err)
+}
+
+func (s *IntegrationTestSuite) TestCmdTxAddRateLimit() {
```
sampocs, line 50: Shouldn't these tests fail since the channel value is zero at the time that they're created?

x/ratelimit/client/cli/cli_test.go
```diff
@@ -0,0 +1,318 @@
...
+)
+
+type IntegrationTestSuite struct {
+	suite.Suite
+
+	cfg     network.Config
+	network *network.Network
+}
+
+func (s *IntegrationTestSuite) SetupSuite() {
```
sampocs, line 33: Is there a way to leverage the `app/test_helper` setup for these tests instead? Similar to what we do for the keeper tests

sampocs (changes requested):
Thanks for laying the framework with these tests! Some higher level comments/questions are:

1. Can you remove the array's from these tests since there isn't actually any iteration
2. Do you think it would be possible to leverage the same setup we use for our keeper functions (`apptesting.AppTestHelper`)? I think ideally we'd be able to share the same framework, but I'm not sure if it has all the necessary functionality (but maybe it's easy to add whatever's missing).

x/ratelimit/client/cli/cli_test.go
antstalepresh: Ok, actually it doesn't fail as there are already rate limits added while running previous tests but I will try to add a new one for running remove cli.
And, we already have a very similar test scenario(1-4) on our simapp tests so I don't think to repeat that again for cli tests.

x/ratelimit/client/cli/cli_test.go
antstalepresh: No, when the testnet for cli tests starts running, some bond_denom tokens are minted initially.
Please check `testutil/network/network.go` => `DefaultConfig` function.

x/ratelimit/client/cli/cli_test.go
sampocs: Great point - totally agree!

x/ratelimit/client/cli/cli_test.go
sampocs: > Ok, actually it doesn't fail as there are already rate limits added while running previous tests but I will try to add a new one for running remove cli.

Ah my mistake, I think I misunderstood how these CLI tests run

> And, we already have a very similar test scenario(1-4) on our simapp tests so I don't think to repeat that again for cli tests.

Totally agree on not duplicating that work again

x/ratelimit/client/cli/cli_test.go
sampocs: got it, thanks for the clarification!

x/ratelimit/client/cli/cli_test.go
sampocs: I see, that makes sense. I'll reach out to you offline to talk this one out

sampocs (approved):

PR #527, opened 2022-12-19
add genesis handler and tests, update some keeper tests

antstalepresh:
## Brief Changelog
  - *Updated genesis handler functions*
  - *Added tests for genesis functions*
  - *updated some keeper functions as there were some failures on keeper tests*


x/ratelimit/genesis_test.go
```diff
@@ -0,0 +1,45 @@
...
+}
+
+func TestGenesis(t *testing.T) {
+	genesisState := types.GenesisState{
+		Params:     types.Params{},
+		RateLimits: createRateLimits(),
+	}
+
+	k, ctx := keepertest.RatelimitKeeper(t)
+	ratelimit.InitGenesis(ctx, *k, genesisState)
```
sampocs, line 37: I think you can do something like:

```suggestion
// import keepertest "github.com/Stride-Labs/stride/v4/x/ratelimit/keeper_test" (not sure if this is the right way to )import
(s keepertest.KeeperTestSuite) func TestGenesis() {
        s.Setup()
	genesisState := types.GenesisState{
		Params:     types.Params{},
		RateLimits: createRateLimits(),
	}

	ratelimit.InitGenesis(s.Ctx, s.App.RateLimitKeeper, genesisState)
```

x/ratelimit/keeper/msg_server_test.go
```diff
@@ -41,6 +41,10 @@ func (s *KeeperTestSuite) TestMsgServer_AddRateLimit() {
 
 	denom := addRateLimitMsg.Denom
 	channelId := addRateLimitMsg.ChannelId
+	channelValue := int64(100)
+
+	// Mint tokens for generating channel value
+	s.App.BankKeeper.MintCoins(s.Ctx, minttypes.ModuleName, sdk.NewCoins(sdk.NewInt64Coin(addRateLimitMsg.Denom, channelValue)))
```
sampocs, line 47: can you grab and check the error that's returned from all these?

sampocs (changes requested):
Great work on this! I think we might be able to leverage the AppTestHelper - let me know if you don't think it's possible though

x/ratelimit/genesis_test.go
antstalepresh: That makes sense.
But genesis tests is running at module level, not keeper level.
So I am not sure if its right to use KeeperTestSuite for module level tests.
I have checked other modules in stride repo and module level tests didn't use AppTestingHelper at all.
Please check and let me know what you think.

x/ratelimit/keeper/msg_server_test.go
antstalepresh: Sure. I will.

testutil/keeper/ratelimit.go
```diff
@@ -0,0 +1,20 @@
+package keeper
```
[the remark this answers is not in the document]
antstalepresh: This is related to my answer for bellow comment.

x/ratelimit/genesis_test.go
sampocs: Yeah that's a good point, and on second thought, I don't think it makes sense to use the KeeperTestSuite, but I do still think it makes sense to leverage the AppTestHelper. 

The reason the other module don't use the AppTestHelper is because they were created automatically when Stride was scaffolded with ignite. Our long-term hope was to unify the tests into the same framework (if possible), because having multiple different test frameworks can be a little confusing for someone coming into the repo for the first time if they don't have cosmos experience. 

The functions in `testutil/keeper` run almost the exact same logic that's in AppTestHelper, so I think we can work to remove the folder entirely (and use the rate limit module as the first example with the new structure).

I have a plan for how to do this that will only take a second. I'll open up a PR in a second (with this as the base) and if you think it makes sense, we can add it

antstalepresh: Ok, waiting for your PR.

PR #541, opened 2022-12-22
[RATE-LIMIT] Bash Integration Tests

sampocs:
## Brief Changelog
* Tests in `dockernet/scripts/ratelimit/`
* Run with `bash dockernet/scripts/ratelimit/run_all_tests.sh`
* Each testcase will print either a checkmark or an X - if there are no X's, the tests were successful


dockernet/scripts/ratelimit/test_denoms.sh
```diff
@@ -0,0 +1,142 @@
...
+test_denom_ustrd() {
+    print_header "TESTING DENOMS - USTRD"
+    wait_until_epoch_end
+
+    __test_denom_send_packet_native_ustrd
+    __test_denom_receive_packet_native_ustrd
+}
+
+#############################################
+# ujuno from Juno to Stride, the back to Juno
```
antstalepresh, line 45:
```suggestion
# ujuno from Juno to Stride, then back to Juno
```

antstalepresh (changes requested):
Great work! left two comments

antstalepresh (approved):

PR #554, opened 2023-01-04
Callback Logs Cleanup Pt. 3 (ICQCallbacks)

sampocs:
Closes: #XXX

Continuing the callback logs cleanup with ICQCallbacks.

## Brief Changelog
* Cleaned up logs for the withdrawal balance, delegator shares, and validator exchange rate ICQ callbacks
* Cleaned up interchainquery logs for MsgSubmitQueryResponse
* Removed double logging of errors 
* Campfire rule - misc cleanup/standardization across functions (e.g. variable naming)
* **Message to Reviewer**: I'd recommend reviewing this in VSCode as the github diffs are misleading

## Author's Checklist

I have...

## Reviewers Checklist

I have...

      - [ ] not applicable
      - [ ] jira ticket `XXX` 
      - [ ] README.md 
      - [ ] not documented


x/stakeibc/keeper/hooks.go
```diff
@@ -208,8 +208,8 @@ func (k Keeper) ReinvestRewards(ctx sdk.Context) {
 
 	for _, hostZone := range k.GetAllHostZone(ctx) {
 		// only process host zones once withdrawal accounts are registered
-		withdrawalIca := hostZone.WithdrawalAccount
-		if withdrawalIca == nil {
+		withdrawalAccount := hostZone.WithdrawalAccount
+		if withdrawalAccount == nil || withdrawalAccount.Address == "" {
```
riley-stride, line 212: What's the case where address is an empty string?

x/stakeibc/keeper/msg_server_submit_tx.go
```diff
@@ -42,8 +42,8 @@ func (k Keeper) DelegateOnHost(ctx sdk.Context, hostZone types.HostZone, amt sdk
 	}
 
 	// Fetch the relevant ICA
-	delegationIca := hostZone.DelegationAccount
-	if delegationIca == nil || delegationIca.Address == "" {
+	delegationAccount := hostZone.DelegationAccount
+	if delegationAccount == nil || delegationAccount.Address == "" {
```
riley-stride, line 46: Same question about the empty string case

x/stakeibc/keeper/msg_server_submit_tx.go
```diff
@@ -171,13 +175,13 @@ func (k Keeper) UpdateWithdrawalBalance(ctx sdk.Context, hostZone types.HostZone
 		// use "bank" store to access acct balances which live in the bank module
 		// use "key" suffix to retrieve a proof alongside the query result
 		icqtypes.BANK_STORE_QUERY_WITH_PROOF,
-		append(data, []byte(hostZone.HostDenom)...),
-		ttl, // ttl
-	)
-	if err != nil {
+		queryData,
```
riley-stride, line 178: thanks for cleaning this up, it was much needed

x/stakeibc/keeper/icqcallbacks_validator_exchange_rate_test.go
```diff
@@ -43,8 +46,10 @@ func (s *KeeperTestSuite) SetupValidatorICQCallback() ValidatorICQCallbackTestCa
 	// We don't actually need a transfer channel for this test, but we do need to have IBC support for timeouts
 	s.CreateTransferChannel(HostChainId)
 
-	valAddress := "valoper1"
-	valIndexQueried := 0 // index in the validators array
+	// These must be valid addresses, otherwise the bech decoding will fail
```
riley-stride, line 49: Was the test previously failing?

x/stakeibc/keeper/icqcallbacks_validator_exchange_rate_test.go
```diff
@@ -184,11 +194,21 @@ func (s *KeeperTestSuite) TestValidatorExchangeRateCallback_ValidatorNotFound()
 	tc := s.SetupValidatorICQCallback()
 
 	// Update the callback args to contain a validator address that doesn't exist
-	badCallbackArgs := s.CreateValidatorQueryResponse("fake_val", 0, 0)
+	badCallbackArgs := s.CreateValidatorQueryResponse("fake_val", 1, 1)
 	err := stakeibckeeper.ValidatorExchangeRateCallback(s.App.StakeibcKeeper, s.Ctx, badCallbackArgs, tc.validArgs.query)
 	s.Require().EqualError(err, "no registered validator for address (fake_val): validator not found")
 }
 
+func (s *KeeperTestSuite) TestValidatorExchangeRateCallback_InvalidValidatorAddress() {
```
riley-stride, line 202: Nice! Good thinking to add this

x/stakeibc/keeper/icqcallbacks_withdrawal_balance.go
```diff
@@ -8,119 +8,118 @@ import (
...
-		withdrawalBalanceCoin.Amount, withdrawalBalanceCoin.Denom))
-
-	withdrawalAccount := hostZone.GetWithdrawalAccount()
-	if withdrawalAccount == nil {
-		errMsg := fmt.Sprintf("WithdrawalBalanceCallback: no withdrawal account found for zone: %s", hostZone.ChainId)
-		k.Logger(ctx).Error(errMsg)
-		return sdkerrors.Wrapf(types.ErrICAAccountNotFound, errMsg)
+	// Get the host zone's ICA accounts
+	withdrawalAccount := hostZone.WithdrawalAccount
+	if withdrawalAccount == nil || withdrawalAccount.Address == "" {
```
riley-stride, line 57: Same Q about empty string case

x/stakeibc/keeper/icqcallbacks_withdrawal_balance.go
```diff
@@ -8,119 +8,118 @@ import (
...
 	icqtypes "github.com/Stride-Labs/stride/v4/x/interchainquery/types"
 	"github.com/Stride-Labs/stride/v4/x/stakeibc/types"
 )
 
 // WithdrawalBalanceCallback is a callback handler for WithdrawalBalance queries.
+// The query response will return the withdrawal account balance
+// If the balance is non-zero, ICA MsgSends are submitted to transfer from the withdrawal account
+//  to the delegation account (for reinvestment) and fee account (for commission)
 // Note: for now, to get proofs in your ICQs, you need to query the entire store on the host zone! e.g. "store/bank/key"
 func WithdrawalBalanceCallback(k Keeper, ctx sdk.Context, args []byte, query icqtypes.Query) error {
```
riley-stride, line 21: after refactoring this, any thoughts on whether we can/should modularize it more? it's already a bit cleaner with sdk.Ints, but still pretty hefty

x/stakeibc/keeper/icqcallbacks_validator_exchange_rate.go
```diff
@@ -84,15 +83,13 @@ func ValidatorExchangeRateCallback(k Keeper, ctx sdk.Context, args []byte, query
...
+		validator.InternalExchangeRate.InternalTokensToSharesRate))
 
-	// armed with the exch rate, we can now query the (val,del) delegation
-	err = k.QueryDelegationsIcq(ctx, hostZone, queriedValidator.OperatorAddress)
-	if err != nil {
-		errMsg := fmt.Sprintf("ValidatorCallback: failed to query delegation, zone %s, err: %s", hostZone.ChainId, err.Error())
-		k.Logger(ctx).Error(errMsg)
-		return sdkerrors.Wrapf(types.ErrICQFailed, errMsg)
+	// Armed with the exch rate, we can now query the (validator,delegator) delegation
+	if err := k.QueryDelegationsIcq(ctx, hostZone, queriedValidator.OperatorAddress); err != nil {
```
riley-stride, line 90: what is this syntax? isn't `err != nil` redundant?

```
if err := fn(); err != nil {
   // log or return error
} 
```

x/stakeibc/keeper/icqcallbacks_delegator_shares.go
```diff
@@ -26,108 +25,99 @@ import (
...
-		errMsg := fmt.Sprintf("unable to determine if ICQ callback is inside buffer window, err: %s", err.Error())
-		k.Logger(ctx).Error(errMsg)
-		return sdkerrors.Wrapf(types.ErrOutsideIcqWindow, errMsg)
-	} else if !isWithinWindow {
-		k.Logger(ctx).Error("delegator shares callback is outside ICQ window")
-		return nil
+		return sdkerrors.Wrapf(sdkerrors.ErrInvalidRequest, "unable to determine if ICQ callback is inside buffer window, err: %s", err.Error())
+	}
+	if !withinBufferWindow {
+		// QUESTION: Should this return an error?
```
riley-stride, line 53: Returning a failure here makes the caller aware they're outside the buffer window, and tells them how far outside it they are. It's not necessary, just convenient.

x/stakeibc/keeper/icqcallbacks_delegator_shares.go
```diff
@@ -26,108 +25,99 @@ import (
...
+		return sdkerrors.Wrapf(sdkerrors.ErrInvalidRequest,
+			"validator (%s) internal exchange rate has not been updated this epoch (epoch #%d)", validator.Address, strideEpochTracker.EpochNumber)
 	}
 
-	// TODO: make sure conversion math precision matches the sdk's staking module's version (we did it slightly differently)
-	// note: truncateInt per https://github.com/cosmos/cosmos-sdk/blob/cb31043d35bad90c4daa923bb109f38fd092feda/x/staking/types/validator.go#L431
-	validatorTokens := queriedDelgation.Shares.Mul(validator.InternalExchangeRate.InternalTokensToSharesRate).TruncateInt()
-	k.Logger(ctx).Info(fmt.Sprintf("DelegationCallback: HostZone: %s, Validator: %s, Previous NumTokens: %v, Current NumTokens: %v",
-		hostZone.ChainId, validator.Address, validator.DelegationAmt, validatorTokens))
+	// Calculate the number of tokens delegated (using the internal exchange rate)
```
riley-stride, line 73: Slight preference to include the link / reference to the SDK function we're replicating, since `tokensToShares` is a bit of a confusing concept

x/stakeibc/keeper/icqcallbacks_delegator_shares.go
```diff
@@ -26,108 +25,99 @@ import (
...
+	slashAmount := validator.DelegationAmt.Sub(delegatedTokens)
 	slashPct := sdk.NewDecFromInt(slashAmount).Quo(sdk.NewDecFromInt(validator.DelegationAmt))
-	k.Logger(ctx).Info(fmt.Sprintf("ICQ'd Delegation Amount Mismatch, HostZone: %s, Validator: %s, Delegator: %s, Records Tokens: %v, Tokens from ICQ %v, Slash Amount: %v, Slash Pct: %v!",
-		hostZone.ChainId, validator.Address, queriedDelgation.DelegatorAddress, validator.DelegationAmt, validatorTokens, slashAmount, slashPct))
+	k.Logger(ctx).Info(utils.LogICQCallbackWithHostZone(chainId, ICQCallbackID_Delegation,
+		"Validator was slashed! Validator: %s, Delegator: %s, Delegation in State: %v, Delegation from ICQ %v, Slash Amount: %v, Slash Pct: %v",
+		validator.Address, queriedDelgation.DelegatorAddress, validator.DelegationAmt, delegatedTokens, slashAmount, slashPct))
 
 	// Abort if the slash was greater than 10%
 	tenPercent := sdk.NewDec(10).Quo(sdk.NewDec(100))
```
riley-stride, line 99: While we're here, should we make this a param?

x/stakeibc/keeper/icqcallbacks_delegator_shares.go
```diff
@@ -26,108 +25,99 @@ import (
...
 	if slashPct.GT(tenPercent) {
-		errMsg := fmt.Sprintf("DelegationCallback: Validator (%s) slashed but ABORTING update, slash is greater than 0.10 (%d)", validator.Address, slashPct)
-		k.Logger(ctx).Error(errMsg)
-		return sdkerrors.Wrapf(types.ErrSlashGtTenPct, errMsg)
+		return sdkerrors.Wrapf(types.ErrSlashGtTenPct, "Validator slashed but ABORTING update, slash is greater than 0.10 (%v)", slashPct)
+	}
+
+	// Update the validator weight and delegation reflect to reflect the slash
```
riley-stride, line 104: Why is this block now moved down below the slashPct calc?

x/interchainquery/keeper/msg_submit_query_response_test.go
```diff
@@ -89,29 +89,22 @@ func (s *KeeperTestSuite) TestMsgSubmitQueryResponse_UnknownId() {
...
-func (s *KeeperTestSuite) TestMsgSubmitQueryResponse_ExceededTtl() {
-	tc := s.SetupMsgSubmitQueryResponse()
-
-	tc.query.Ttl = uint64(1) // set ttl to be expired
-	s.App.InterchainqueryKeeper.SetQuery(s.Ctx, tc.query)
-	exceeded, err := s.App.InterchainqueryKeeper.HasQueryExceededTtl(s.Ctx, &tc.validMsg, tc.query)
-	s.Require().NoError(err)
-	s.Require().True(exceeded)
-}
+// TODO: Enable test if query proof is mocked out successfully
```
riley-stride, line 92: Why has this been commented out?

x/interchainquery/keeper/msg_server.go
```diff
@@ -30,174 +30,153 @@ func NewMsgServerImpl(keeper Keeper) types.MsgServer {
...
 	// check if the response has an associated query stored on stride
-	q, found := k.GetQuery(ctx, msg.QueryId)
+	query, found := k.GetQuery(ctx, msg.QueryId)
 	if !found {
-		k.Logger(ctx).Info("[ICQ Resp] ignoring non-existent query response (note: duplicate responses are nonexistent)")
+		k.Logger(ctx).Info("ICQ RESPONSE  | Ignoring non-existent query response (note: duplicate responses are nonexistent)")
 		return &types.MsgSubmitQueryResponseResponse{}, nil // technically this is an error, but will cause the entire tx to fail if we have one 'bad' message, so we can just no-op here.
 	}
 
 	defer ctx.EventManager().EmitEvents(sdk.Events{
```
riley-stride, line 131: Let's make sure we test this thoroughly with the ibc-go relayer, since ibc-go's ICQ relaying assumes on a specific Stride/x/interchainquery event emission schema to detect relayed ICQs. Did you check this against Andrew's schema or test it with the relayer?

riley-stride (changes requested):
Great stuff! Left some questions/comment throughout. 

I reviewed everything except the ICQ proof changes. I'd like to sync with you live on Monday before reviewing those (given the sensitivity).

x/stakeibc/keeper/hooks.go
sampocs: Shouldn't be possible in practice! But an empty/un-initialized `WithdrawalAccount` struct has an empty string for the `Address` field, so I think we should include it in the case where we check for a nil Account to be safe

```go
w := icatypes.WithdrawalAccount{} 
// w.Address is empty string 
```

x/stakeibc/keeper/msg_server_submit_tx.go
sampocs: answered above

x/stakeibc/keeper/icqcallbacks_validator_exchange_rate_test.go
sampocs: Previously we weren't catching the error case of if you couldn't decode the address. Once I changed the original function to catch that error case, this test failed (since "valoper1" is not a valid address)

x/stakeibc/keeper/icqcallbacks_withdrawal_balance.go
sampocs: answered above

x/stakeibc/keeper/icqcallbacks_withdrawal_balance.go
sampocs: Good question! I could probably go either way. My two cents:

I think functions like this where the bulk of the function just checks error cases (e.g. make sure host exists, make sure ICAs are initialzed, etc.) are lower priority refactors. 
That said, I think we should refactor all functions that submit ICAs in such a way that there's a function that first returns just the ICA message itself, and then a separate function that actually executes it. That way we can write a test that checks the contents of the message

x/stakeibc/keeper/icqcallbacks_validator_exchange_rate.go
sampocs:
```go
// this:
if err := fn(); err != nil {
   // log or return error
} 

// is just a shorthand way of doing this:

err := fn()
if err != nil {
    // log or return error
}
```
https://www.calhoun.io/one-liner-if-statements-with-errors/

Is that what you were asking or did I misinterpret your question?

x/stakeibc/keeper/icqcallbacks_delegator_shares.go
sampocs: Ah good call, did not mean to remove that, just meant to remove the TODO.

Adding back in

x/stakeibc/keeper/icqcallbacks_delegator_shares.go
sampocs: Cool, I just noticed some of the ICQ functions returned an error when we were outside the buffer window and other's did not (including this one). So wanted to make sure there wasn't a specific reason for not returning an error in these

x/stakeibc/keeper/icqcallbacks_delegator_shares.go
sampocs: Yeah! Do you remember all the steps for adding a param? We need to add it in our upgrade handler right?

x/stakeibc/keeper/icqcallbacks_delegator_shares.go
sampocs: It just seemed to make more logical sense there from a readability perspective. 

The flow was
```go
// get slash amount
// cast weight 
// do something with slash amount
// do another thing with slash amount
// do something with the casted weight
```

seemed like it made more sense to order it/ group as:
```go
// get slash amount
// do something with slash amount
// do another thing with slash amount

// cast weight 
// do something with the casted weight
```

x/interchainquery/keeper/msg_submit_query_response_test.go
sampocs: I'm not sure how to mock out the state needed to verify a proof, and it seems like whoever wrote these tests ran into the same issue (see all the code commented out at the bottom of this file)

So I figured that if we ever resolve that, then we can re-enable this test

x/interchainquery/keeper/msg_server.go
sampocs: Was there a change that you thought would be incompatible? The change you've highlighted here is just in the logs - it doesn't modify the events at all

Also we can use the integration tests to confirm the ICQs still work with the relayers

sampocs: Also fwiw, I did test the slash queries still work as expected

x/stakeibc/keeper/icqcallbacks_validator_exchange_rate_test.go
riley-stride: Nice catch!

x/interchainquery/keeper/msg_submit_query_response_test.go
riley-stride: I understand why the proof verification check is hard. I think I wrote these tests and couldn't recreate it.

But we should test the case where we exceed Tttl, right? That's the `TestMsgSubmitQueryResponse_ExceededTtl` function that seems to be commented out here

x/stakeibc/keeper/icqcallbacks_delegator_shares.go
riley-stride: Thanks!

x/stakeibc/keeper/icqcallbacks_validator_exchange_rate.go
riley-stride: Makes sense, thanks. For some reason I hadn't used this syntax before.

x/interchainquery/keeper/msg_server.go
riley-stride: Discussed this offline. It works!

riley-stride (changes requested):
Really solid work. ICQ msg_server looks good, thanks for the walkthrough.

Requesting two changes
- add the slashPct param
- bring back the ttl exceeded unit test (possible I'm misunderstanding your message)

sampocs:
@riley-stride 
> bring back the ttl exceeded unit test (possible I'm misunderstanding your message)

I think the issue is that VerifyKeyProof is called (and fails) before the TTL is checked. So there's no way to cover that code unless we mock out the proof

sampocs:
Looks like I can actually just change the query type though to bypass that function

riley-stride (approved):
🚢 🚢 🚢 great stuff! param addition and upgrade handler look good to me

Issue #544, opened 2022-12-22
Move MsgServer to attribute in KeeperTestSuite (remove GetMsgServer)


sampocs:
## Background
`x/stakeibc/keeper/keeper_test.go` defines a `KeeperTestSuite` with a `GetMsgServer` function. I can't remember why a Getter was used for this and I don't think it's necessary.

## Suggested Design
In `x/stakeibc/keeper/keeper_test.go`, change this:
```go
type KeeperTestSuite struct {
	apptesting.AppTestHelper
}

func (s *KeeperTestSuite) SetupTest() {
	s.Setup()
}

func (s *KeeperTestSuite) GetMsgServer() types.MsgServer {
	return keeper.NewMsgServerImpl(s.App.StakeibcKeeper)
}
```
To this:
```go
type KeeperTestSuite struct {
	apptesting.AppTestHelper
        MsgServer types.MsgServer
}

func (s *KeeperTestSuite) SetupTest() {
	s.Setup()
        s.MsgServer = keeper.NewMsgServerImpl(s.App.StakeibcKeeper)
}
```
And update unit tests appropriately
```go
_, err := s.GetMsgServer().LiquidStake(sdk.WrapSDKContext(s.Ctx), &msg)
// changed to
_, err := s.MsgServer.LiquidStake(sdk.WrapSDKContext(s.Ctx), &msg)
```

## Acceptance Criteria
* Unit tests pass

hoangdv2429:
Can I handle this ?

sampocs:
Go for it! Thanks for checking first :)

hoangdv2429:
@sampocs Can you have a look ? Thank you!

sampocs:
After discussion [here](https://github.com/Stride-Labs/stride/pull/555#discussion_r1061667970), decision was made to leave this as is

PR #555, opened 2023-01-04
Chore/remove get msg server

hoangdv2429:
Closes: #544

This will remove GetMsgServer from x/stakeibc/keeper/keeper_test.go .

## Brief Changelog

Changed using of GetMsgServer from a few test files to MsgServer field of KeeperTestSuite instead.

## Author's Checklist

I have...

- [x] Run and PASSED locally all Stride integration tests

## Reviewers Checklist

I have...

      - [ ] not applicable
      - [ ] jira ticket `XXX` 
      - [ ] README.md 
      - [ ] not documented

### DIFF
--- a/x/stakeibc/keeper/msg_server_rebalance_validators_test.go
+++ b/x/stakeibc/keeper/msg_server_rebalance_validators_test.go
@@ -151,7 +151,7 @@ func (s *KeeperTestSuite) TestRebalanceValidators_Successful() {
 	s.Require().NoError(err, "unmarshalling callback args error for callback key (%s)", callbackKey)
 	s.Require().Equal("GAIA", callbackArgs.HostZoneId, "callback host zone id should be GAIA")
 
-		// verify callback rebalance is what we want
+	// verify callback rebalance is what we want
 	s.Require().Equal(2, len(callbackArgs.Rebalancings), "callback should have 2 rebalancing")
 	firstRebal := callbackArgs.Rebalancings[0]
 	s.Require().Equal(sdk.NewInt(104), firstRebal.Amt, "first rebalance should rebalance 104 ATOM")
--- a/x/stakeibc/keeper/msg_server_rebalance_validators_test.go
+++ b/x/stakeibc/keeper/msg_server_rebalance_validators_test.go
@@ -9,6 +9,7 @@ import (
 
 	epochtypes "github.com/Stride-Labs/stride/v4/x/epochs/types"
 	icacallbackstypes "github.com/Stride-Labs/stride/v4/x/icacallbacks/types"
+	stakeibckeeper "github.com/Stride-Labs/stride/v4/x/stakeibc/keeper"
 	stakeibctypes "github.com/Stride-Labs/stride/v4/x/stakeibc/types"
 )
 
@@ -20,10 +21,10 @@ type RebalanceValidatorsTestCase struct {
 }
 
 func (s *KeeperTestSuite) SetupRebalanceValidators() RebalanceValidatorsTestCase {
-	s.SetupTest()
 	// Setup IBC
 	delegationIcaOwner := "GAIA.DELEGATION"
 	delegationChannelId := s.CreateICAChannel(delegationIcaOwner)
+	s.MsgServer = stakeibckeeper.NewMsgServerImpl(s.App.StakeibcKeeper)
 	delegationAddr := s.IcaAddresses[delegationIcaOwner]
 
 	// setup epochs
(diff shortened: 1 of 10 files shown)


x/stakeibc/keeper/msg_server_rebalance_validators_test.go
```diff
@@ -23,6 +24,7 @@ func (s *KeeperTestSuite) SetupRebalanceValidators() RebalanceValidatorsTestCase
 	// Setup IBC
 	delegationIcaOwner := "GAIA.DELEGATION"
 	delegationChannelId := s.CreateICAChannel(delegationIcaOwner)
+	s.MsgServer = stakeibckeeper.NewMsgServerImpl(s.App.StakeibcKeeper)
```
sampocs, line 27: is this line necessary?

x/stakeibc/keeper/msg_server_redeem_stake_test.go
```diff
@@ -25,6 +25,7 @@ type RedeemStakeTestCase struct {
 }
 
 func (s *KeeperTestSuite) SetupRedeemStake() RedeemStakeTestCase {
+	// s.SetupTest()
```
sampocs, line 28: nit: remove

x/stakeibc/keeper/msg_server_restore_interchain_account_test.go
```diff
@@ -32,6 +33,7 @@ type RestoreInterchainAccountTestCase struct {
 
 func (s *KeeperTestSuite) SetupRestoreInterchainAccount() RestoreInterchainAccountTestCase {
 	s.CreateTransferChannel(HostChainId)
+	s.MsgServer = stakeibckeeper.NewMsgServerImpl(s.App.StakeibcKeeper)
```
sampocs, line 36: is this line necessary?

sampocs (commented):
Looks great, thanks for doing this! Couple small comments below

x/stakeibc/keeper/msg_server_rebalance_validators_test.go
hoangdv2429: Yes, it does actually. because the Setup() run in SetupTest() only setup for tests that need to use Stride only. If we move to tests that use ICA or IBC (need interact between 2 chains) then we'll have to have this line, otherwise the context that get pass from function to function will be different, either we have this line or we'll have to change code in CreateICAChannel which I choose to go with the 1st approach.

x/stakeibc/keeper/msg_server_restore_interchain_account_test.go
hoangdv2429: It is the same as above

x/stakeibc/keeper/msg_server_rebalance_validators_test.go
sampocs: Ah shoot. That's sounding familiar - that might be why we originally set it up as a function GetMsgServer (so that it gets executed after the setup)

I think its a confusing UX for someone to need to know that they have to reinitialize MsgServer every time they setup ICA/IBC tests. 

Can you think of a clean way we could update it in CreateICAChannel (or preferably in CreateTransferChannel, similar to how we update the Ctx)? I'm not sure there's an easy way since the msg server is at the module keeper level and app tester is at the app level. 

If there's no clean way, I'd unfortunately probably vote to leave it as a function GetMsgServer

hoangdv2429: Eventhough I can manage to add an extra field to AppTestSuite with new param to get the module's type of keeper to make it work. I think it doesn't make sense to do it, I also checked other modules to see if the GetMsgServer get remove but none of it does. So maybe we should close this as complete and keep GetMsgServer. @sampocs

sampocs: Yeah I agree with all that. Sorry about sending you down this path :( 

Do you mind just adding a comment to each `GetMsgServer` function that describes this so we don't accidentally go down this road again in the future? You can do it in a separate PR

Something like
```
// Dynamically gets the MsgServer for this module's keeper
// A function must be used so that the MsgServer is always created with the most updated App context
//  which can change depending on the type of test 
//  (e.g. tests with only one Stride chain vs tests with multiple chains and IBC support) 
```

hoangdv2429: Got it! I'll add comment on that

sampocs: Thank you!

sampocs:
As discussed above, it will be left as a function in order to support both single chain and multi-chain functionality. @hoangdv2429  to add clarifying comment in new PR

Issue #414, opened 2022-11-30
Epic: IBC rate limiting module


asalzmann:
_heavily inspired by the Osmosis implementation: https://github.com/osmosis-labs/osmosis/tree/main/x/ibc-rate-limit_

# IBC Rate Limit

The IBC Rate Limit module is responsible for adding a governance-configurable rate limit to IBC transfers.
This is a safety control, intended to protect assets on Stride in event of:

* a bug/hack on Stride
* a bug/hack on the counter-party chain
* a bug/hack in IBC itself

This is done in exchange for a potential (one-way) bridge liveness tradeoff, in periods of high deposits or withdrawals.

The architecture of this package is a minimal go package which implements an [IBC Middleware](https://github.com/cosmos/ibc-go/blob/f57170b1d4dd202a3c6c1c61dcf302b6a9546405/docs/ibc/middleware/develop.md) that wraps the [ICS20 transfer](https://ibc.cosmos.network/main/apps/transfer/overview.html) app, and calls a new module, `ibc-rate-limit`.

The `ibc-rate-limit` module then has all of the actual IBC rate limiting logic.

The status of the module is being in a state suitable for some initial governance settable rate limits for high value bridged assets.
Its not in its long term / end state for all channels by any means, but does act as a strong protection we
can instantiate today for high value IBC connections.

## Motivation

The motivation of IBC-rate-limit comes from the empirical observations of blockchain bridge hacks that a rate limit would have massively reduced the stolen amount of assets in:

- [Polynetwork Bridge Hack ($611 million)](https://rekt.news/polynetwork-rekt/)
- [BNB Bridge Hack ($586 million)](https://rekt.news/bnb-bridge-rekt/)
- [Wormhole Bridge Hack ($326 million)](https://rekt.news/wormhole-rekt/)
- [Nomad Bridge Hack ($190 million)](https://rekt.news/nomad-rekt/)
- [Harmony Bridge Hack ($100 million)](https://rekt.news/harmony-rekt/) - (Would require rate limit + monitoring)
- [Dragonberry IBC bug](https://forum.cosmos.network/t/ibc-security-advisory-dragonberry/7702) (can't yet disclose amount at risk, but was saved due to being found first by altruistic Osmosis core developers)

In the presence of a software bug on Stride, IBC itself, or on a counterparty chain, we would like to prevent the bridge from being fully depegged.
This stems from the idea that a 30% asset depeg is ~infinitely better than a 100% depeg.
[...]

PR #556, opened 2023-01-04
IBC Rate Limiting

sampocs:
Co-Author: @antstalepresh 
Closes: #414 

The `ratelimit` module is meant as a safety control in the event of a bug or attack, and prevents massive inflows or outflows of IBC tokens to/from Stride in a short time frame. See module docs [here](https://github.com/Stride-Labs/stride/tree/rate-limiter/x/ratelimit) for an overview of the purpose and tech spec.

## Brief Changelog
* Added `ratelimit` module as IBC middleware around transfer module
* Added governance transactions `AddRateLimit`, `UpdateRateLimit`, `ResetRateLimit` and `RemoveRateLimit`
* Added `rate-limit`, `list-rate-limits` and `rate-limits-by-chain` queries
* Added unit tests and bash integration tests
* Added Juno <> Osmo relayer config for testing transfers between non-stride chains
* Added TODO to add store key when upgrade handler is added
* Added Hour epoch

## Upgrade
The upgrade handler was excluded from the PR as the target upgrade has not been determined yet.
**IN THE ASSOCIATED UPGRADE, WE MUST REMEMBER TO ADD THE STORE KEY AND HOUR EPOCH**
**WE MUST ALSO ADD RATE LIMITS FOR EXISTING DENOMS!**

## Testing
* There's a full suite of integration tests that were meant to be a comprehensive sanity check before launching this module. **They are NOT meant to run in CI and are intentionally pedantic (they take ~30 minutes to run). They're also redundant with the unit tests.**
* Run tests with 
```bash
make start-docker-all && bash dockernet/scripts/ratelimit/run_all_tests.sh
```
* Once the PR has been reviewed and approved, I'll run through the normal stride integration tests and the rate limit integration tests one last time before merging:
- [x] CI Integration Tests for all host zones
- [x] RateLimit Integration Tests

## Updates needed for sdk 46 and ibc-go v5
- [x] Change sdk.Int to cosmossdk/math
- [x] Remove rest client and rest route registered in module.go
- [x] IBC go callback args
- [x] Refactor ICS4Wrapper
- [x] Implement GetAppVesrion
 
## Before Merging
- [x] Do we need to bind PortID? I don't think so since the middleware starts from the transfer port
- [x] Revert dockernet config settings to default
- [x] Add blacklist
- [x] Add evens

## Author's Checklist

I have...

- [X] If the change is contentful, I either:

## Reviewers Checklist

I have...

  - [X] Does this pull request introduce a new feature or user-facing behavior changes? 
      - [ ] not applicable
      - [ ] jira ticket `XXX` 
      - [X] README.md 
      - [ ] not documented


x/ratelimit/README.md
```diff
@@ -0,0 +1,189 @@
...
+#### Source vs Sink
+* **Source**: The packet is being received by a chain it was just sent from (i.e. the token has gone back and forth)
+    * Ex1: `strd` is sent from Stride to Osmosis, and then back to Stride 
+    * Ex2: `ujuno` is sent to Stride, then to Osmosis, then back to Stride 
+    * Stride is the source in the examples above because the token went back and forth from Stride -> Osmosis -> Stride
+* **Sink**: The packet is being received by a chain that either created it or previous received it from somewhere else
+    * Ex1: `uatom` is sent from Cosmoshub to Stride (`uatom` was created on Cosmoshub)
+    * Ex2: `uatom` is sent from Cosmoshub to Osmosis then to Stride. Here the receiving chain (Stride) is not the same as the previous hop (Cosmoshub), so Stride is not acting as a source.
+#### Determining the denom in the rate limit
+* If the chain is acting as a **Sink**: Add on the Stride port and channel and hash it
```
asalzmann, line 72: Nice example!

x/ratelimit/ibc_module.go
```diff
@@ -0,0 +1,158 @@
...
+	return im.app.OnChanCloseConfirm(ctx, portID, channelID)
+}
+
+// OnRecvPacket implements the IBCModule interface
+func (im IBCModule) OnRecvPacket(
+	ctx sdk.Context,
+	packet channeltypes.Packet,
+	relayer sdk.AccAddress,
+) exported.Acknowledgement {
+	if err := ReceiveRateLimitedPacket(ctx, im.keeper, packet); err != nil {
```
asalzmann, line 112: Does a refund occur if this errors?

x/ratelimit/ics4_wrapper.go
```diff
@@ -0,0 +1,38 @@
...
+}
+
+func NewICS4Middleware(channel porttypes.ICS4Wrapper, ratelimitKeeper ratelimitkeeper.Keeper) ICS4Wrapper {
+	return ICS4Wrapper{
+		channel:         channel,
+		rateLimitKeeper: ratelimitKeeper,
+	}
+}
+
+func (i *ICS4Wrapper) SendPacket(ctx sdk.Context, chanCap *capabilitytypes.Capability, packet exported.PacketI) error {
```
asalzmann, line 29: Why is this `SendPacket` used, rather than the `SendPacket` in `ibc_module.go`?

x/ratelimit/rate_limit.go
```diff
@@ -0,0 +1,155 @@
...
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// Middleware implementation for RecvPacket with rate limiting
+func ReceiveRateLimitedPacket(ctx sdk.Context, keeper ratelimitkeeper.Keeper, packet channeltypes.Packet) error {
+	// For a receive packet, the channel on stride is the "Destination" channel
```
asalzmann, line 130: Can you add more detail to this sentence? I think the implication here is that the Stride channel is used to key `RateLimit` objects

x/ratelimit/types/keys.go
```diff
@@ -0,0 +1,23 @@
+package types
+
+const (
+	ModuleName = "ratelimit" // IBC at the end to avoid conflicts with the ibc prefix
```
asalzmann, line 4: What does "IBC at the end to avoid conflicts with the ibc prefix" mean here?

x/ratelimit/keeper/rate_limit.go
```diff
@@ -0,0 +1,121 @@
...
+)
+
+// Get the rate limit byte key built from the denom and channelId
+func GetRateLimitItemKey(denom string, channelId string) []byte {
+	return append(types.KeyPrefix(denom), types.KeyPrefix(channelId)...)
+}
+
+// The total value on a given path (aka, the denominator in the percentage calculation)
+// is the total supply of the given denom
+func (k Keeper) GetChannelValue(ctx sdk.Context, denom string) sdk.Int {
```
asalzmann, line 18: I recall Osmosis summing up across multiple channels - why were they doing that and why don't we have to account for that here?

x/ratelimit/rate_limit.go
```diff
@@ -0,0 +1,155 @@
...
+		denom = packet.Denom
+	} else {
+		// Non-native assets should be hashed
+		denom = denomTrace.IBCDenom()
+	}
+
+	return denom
+}
+
+// Parse the denom from the Recv Packet that will be used by the rate limit module
```
asalzmann, line 42: Given these are in the README, not sure it makes sense to repeat here. Maybe something like

A->B->C->A->C->B->A
A->B: A is acting as a source, B as a sink
B->C: B is a source, C is a sink
etc.

would be more helpful

x/ratelimit/rate_limit.go
```diff
@@ -0,0 +1,155 @@
...
+//    and if it is NON-NATIVE, it will have a prefix (e.g. transfer/channel-2/uosmo)
+//
+// For NATIVE denoms, return as is (e.g. ustrd)
+// For NON-NATIVE denoms, take the ibc hash (e.g. hash "transfer/channel-2/usoms" into "ibc/...")
+func ParseDenomFromSendPacket(packet transfertypes.FungibleTokenPacketData) (denom string) {
+	// Determine the denom by looking at the denom trace path
+	denomTrace := transfertypes.ParseDenomTrace(packet.Denom)
+
+	// Native assets will have an empty trace path and can be returned as is
+	if denomTrace.Path == "" {
```
asalzmann, line 32: We should double-check this assumption - does this if-else block map onto the logic we think it does?
Path == "" => natively minted on chain
Path != "" => IBC token

x/ratelimit/rate_limit.go
```diff
@@ -0,0 +1,155 @@
...
+//
+//			Ex2: juno was sent to Stride, then to Osmosis, then back to Stride
+//              Packet Denom:      transfer/channel-X/transfer/channel-Z/ujuno
+//               -> Remove Prefix: transfer/channel-Z/ujuno
+//               -> Hash:          ibc/...
+func ParseDenomFromRecvPacket(packet channeltypes.Packet, packetData transfertypes.FungibleTokenPacketData) (denom string) {
+	// To determine the denom, first check whether Stride is acting as source
+	if transfertypes.ReceiverChainIsSource(packet.GetSourcePort(), packet.GetSourceChannel(), packetData.Denom) {
+		// Remove the source prefix (e.g. transfer/channel-X/transfer/channel-Z/ujuno -> transfer/channel-Z/ujuno)
+		sourcePrefix := transfertypes.GetDenomPrefix(packet.GetSourcePort(), packet.GetSourceChannel())
```
asalzmann, line 76: Is this logic equivalent to what ICS-20 does when it unwraps vouchers? If so, let's make sure we pull the code from there and reference it

app/app.go
```diff
@@ -576,15 +611,17 @@ func NewStrideApp(
 
 	// Stack two contains
 	// - IBC
+	// - ratelimit
```
asalzmann, line 614: What's the thinking around the ordering here? I'm not sure it matters, just curious. Conceptually, I'd probably move this down closer to transfer (because it's lower level), but not sure there's actually a reason to do so!

asalzmann (commented):
Great stuff! Left a bunch of comments, but conceptually everything makes sense. Thanks for the thorough documentation and testing!

app/app.go
asalzmann: Also, if we change this, we probably need to update this LOC https://github.com/notional-labs/stride/blob/acd8c57188652ce42c53a09c82a2d5bbe49b83af/x/icacallbacks/keeper/keeper.go#L115

README.md
```diff
@@ -155,6 +155,7 @@ We use the following modules from [Osmosis](https://github.com/osmosis-labs/osmo
 ```
 x/epochs
 x/mint
+x/ratelimit
```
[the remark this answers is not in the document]
sampocs: maybe lets just remove it from here - we give credit to the design in the module itself

dockernet/config.sh
```diff
@@ -21,7 +21,7 @@ KEYS_LOGS=$SCRIPT_DIR/logs/keys.log
 HOST_CHAINS=() 
 
 if [[ "${ALL_HOST_CHAINS:-false}" == "true" ]]; then 
-  HOST_CHAINS=(GAIA JUNO OSMO STARS)
+  HOST_CHAINS=(GAIA JUNO OSMO)
```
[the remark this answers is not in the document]
sampocs: Just figured it was overkill, we had also removed it from main

x/README.md
```diff
@@ -13,6 +13,7 @@ category: 62c5c5ff03a5bf069004def2
 `interchainquery` - Issues queries between IBC chains, verifies state proof and executes callbacks.
 `epochs` - Makes on-chain timers which other modules can execute code during.
 `mint` - Controls token supply emissions, and what modules they are directed to.
+`ratelimit` - IBC middleware wrapping the transfer module, thottles large IBC transfers in the event of a bug/hack.
```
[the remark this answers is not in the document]
sampocs: done

app/upgrades.go
```diff
@@ -48,6 +48,10 @@ func (app *StrideApp) setupUpgradeHandlers() {
 			Added: []string{claimtypes.StoreKey},
 		}
 	}
+	// TODO: add ratelimit store key when module is added
```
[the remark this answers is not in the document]
sampocs: Yeah that's not a bad idea

proto/stride/ratelimit/ratelimit.proto
```diff
@@ -0,0 +1,51 @@
...
+  PACKET_RECV = 1;
+}
+
+message Path {
+  string denom = 1;
+  string channel_id = 2;
+}
+
+message Quota {
+  string max_percent_send = 1 [
```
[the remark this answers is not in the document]
sampocs: I don't see a scenario where we would

proto/stride/ratelimit/ratelimit.proto
```diff
@@ -0,0 +1,51 @@
...
+    (gogoproto.customtype) = "github.com/cosmos/cosmos-sdk/types.Int",
+    (gogoproto.nullable) = false
+  ];
+  string max_percent_recv = 2 [
+    (gogoproto.customtype) = "github.com/cosmos/cosmos-sdk/types.Int",
+    (gogoproto.nullable) = false
+  ];
+  uint64 duration_hours = 3;
```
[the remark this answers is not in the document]
sampocs: The osmo spec mentioned they don't want the window to be too short - it should be long enough for a human to react. They even explicitly said they would never make a rate limit that was only 1 hour long.

I agree with their reasoning but curious to hear your thoughts as well!

x/ratelimit/README.md
```diff
@@ -0,0 +1,189 @@
+---
+title: "RateLimit"
+excerpt: ""
+category: 6392913957c533007128548e
+---
+# RateLimit Module
+## Overview
+This `ratelimit` module is a native golang implementation of Osmosis's CosmWasm [`ibc-rate-limit`](https://github.com/osmosis-labs/osmosis/tree/main/x/ibc-rate-limit) module. The module is meant as a safety control in the event of a bug or attack, and prevents massive inflows or outflows of IBC tokens to/from Stride in a short time frame. See [here](https://github.com/osmosis-labs/osmosis/tree/main/x/ibc-rate-limit#motivation) for an excellent summary by the Osmosis team on the motivation for rate limiting.
```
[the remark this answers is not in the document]
sampocs: great points, adjusted

x/ratelimit/README.md
```diff
@@ -0,0 +1,189 @@
...
+    * For `Receive` packets: 
+    $$\text{Exceeds Quota if:} \left(\frac{\text{Inflow} - \text{Outflow} + \text{Packet Amount}}{\text{ChannelValue}}\right) > \text{MaxPercentRecv}$$
+
+## Example Walk-Through
+Using the example above, let's say we created a 24 hour rate limit on `ibc/D24B4564BCD51D3D02D9987D92571EAC5915676A9BD6D9B0C1D0254CB8A5EA34` ("`ibc/uosmo`"), `channel-5`, with a 10% send and receive threshold. 
+1. At the start of the window, the supply will be queried, to determine the channel value. Let's say the total supply was 100
+2. If someone transferred `8uosmo` from `Osmosis -> Stride`, the `Inflow` would increment by 8
+3. If someone tried to transfer another `8uosmo` from `Osmosis -> Stride`, it would exceed the quota since `(8+8)/100 = 16%` (which is greater than 10%) and thus, the transfer would be rejected.
+4. If someone tried to transfer `12ibc/uosmo` from Stride -> Osmosis, the `Outflow` would increment by 12. Notice, even though 12 is greater than 10% the total channel value, the *net* outflow is only `4uatom` (since it's offset by the `8uatom` `Inflow`). As a result, this transaction would succeed.
```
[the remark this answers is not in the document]
sampocs: The idea is if inflows or outflows are handled separately, it opens the door for a DOS attack where someone sends assets back and forth and causes the channel to get throttled.

The osmo spec mentioned it [here](https://github.com/osmosis-labs/osmosis/tree/main/x/ibc-rate-limit#rate-limit-types). I can include it in ours as well

x/epochs/types/genesis.go
```diff
@@ -10,16 +10,20 @@ func NewGenesisState(epochs []EpochInfo) *GenesisState {
 }
 
 var (
-	STRIDE_EPOCH = "stride_epoch"
+	HOUR_EPOCH   = "hour"
```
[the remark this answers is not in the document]
sampocs: yeah, there's a todo in the upgrade handler

x/ratelimit/README.md
```diff
@@ -0,0 +1,189 @@
...
+We always want to refer to the channel ID and denom as they appear on Stride. For instance, in the example above, we would store the rate limit with denom `ibc/D24B4564BCD51D3D02D9987D92571EAC5915676A9BD6D9B0C1D0254CB8A5EA34` and `channel-5`, instead of `uosmo` and `channel-326` (the ChannelID on Osmosis).
+
+However, since the ratelimit module acts as middleware to the transfer module, the respective denoms need to be interpreted using the denom trace associated with each packet. There are a few scenarios at play here...
+
+### Send Packets
+The denom that the rate limiter will use for a send packet depends on whether it was a native token (e.g. ustrd, stuatom, etc.) or non-native token (e.g. ibc/...)...
+#### Native vs Non-Native
+* We can identify if the token is native or not by parsing the denom trace from the packet
+    * If the token is **native**, it **will not** have a prefix (e.g. `ustrd`)
+    * If the token is **non-native**, it **will** have a prefix (e.g. `transfer/channel-X/uosmo`)
```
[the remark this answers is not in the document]
sampocs: Can you elaborate? Are you asking if someone could create a denom with a portId, channelId and slashes?

x/ratelimit/README.md
```diff
@@ -0,0 +1,189 @@
...
+#### Determining the denom in the rate limit
+* For **native** tokens, return as is (e.g. `ustrd`)
+* For **non-native** tokens, take the ibc hash (e.g. hash `transfer/channel-X/uosmo` into `ibc/...`)
+
+### Receive Packets
+The denom that the rate limiter will use for a receive packet depends on whether it was a source or sink
+#### Source vs Sink
+* **Source**: The packet is being received by a chain it was just sent from (i.e. the token has gone back and forth)
```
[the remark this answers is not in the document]
sampocs: I found the spec particularly confusing. That was my (failed) attempt to try and clarify it. 

How about this- kind of a hybrid of what I had and what you sent...

---

### Receive Packets
The denom that the rate limiter will use for a receive packet depends on whether it was a source or sink.

#### Source vs Sink
As a token travels across IBC chains, its path is recorded in the denom trace. 

* **Sink**: If the token moves **forward**, to a chain different than its previous hop, the destination chain acts as a **sink zone**, and the new port and channel are **appended** to the denom trace.
    * Ex1: `uatom` is sent from Cosmoshub to Stride 
      * Stride is the first destination for `uatom`, and acts as a sink zone
      * The IBC denom becomes the hash of: `/{stride-port)/{stride-channel}/uatom`
    * Ex2: `uatom` is sent from Cosmoshub to Osmosis then to Stride
      * Here the receiving chain (Stride) is not the same as the previous hop (Cosmoshub), so Stride, once again, is acting as a sink zone
      *  The IBC denom becomes the hash of: `/{stride-port)/{stride-channel}/{osmosis-port}/{osmosis-channel}/uatom`
   
* **Source**: If the token moves **backwards** (i.e. revisits the last chain it was sent from), the destination chain is acting as a **source zone**, and the port and channel are **removed** from the denom trace - undoing the last hop. Should a token reverse its course completely and head back along the same path to its native chain, the denom trace will unwind and reduce back down to the original base denom.
    * Ex1: `ustrd` is sent from Stride to Osmosis, and then back to Stride 
      * Here the trace reduces from `/{osmosis-port}/{osmosis-channel}/ustrd` to simply `ustrd`
    * Ex2: `ujuno` is sent to Stride, then to Osmosis, then back to Stride 
      * Here the trace reduces from `/{osmosis-port}/{osmosis-channel}/{stride-port}/{stride-channel}/ujuno` to just `/{stride-port}/{stride-channel}/ujuno` (the Osmosis hop is removed)
    * Stride is the source in the examples above because the token went back and forth from Stride -> Osmosis -> Stride

For a more detailed explanation, see the[ ICS-20 ADR](https://github.com/cosmos/ibc-go/blob/main/docs/architecture/adr-001-coin-source-tracing.md#example) and [spec](https://github.com/cosmos/ibc/tree/main/spec/app/ics-020-fungible-token-transfer).

x/ratelimit/ics4_wrapper.go
sampocs: This was by far the most confusing aspect of the ICS Middleware setup. It has to do with the ordering of the execution of each stack: sends are bottom-up, receives are top-down.

**Receive Packet**
1. IBC Channel Module
2. OnRecvPacket in ratelimit module
3. OnRecvPacket in records module
4. OnRecvPacket in transfer module
5. Tokens are minted

**Send Packet**
1. SendPacket from transfer module
2. Tokens are escrowed
3. SendPacket from ICS4 Wrapper (rate limit module)
4. SendPacket from Channel Module

---

In the receive scenario, the execution flow aligns with the chaining defined when setting up and wrapping each middleware. So the **rate limit** module has access to the **records** module, which has access to the **transfer** module.
```go
stack := ratelimit.NewIBCModule(rateLimitKeeper, records.NewIBCModule(recordsKeeper, transferModule))
```

But in the send scenario, the transfer module doesn't have access to that full middleware chaining (since it's at the bottom of the stack) - it only has access to its ICSMiddleware.
```go
transferKeeper := NewKeeper(..., icsMiddleware, ...)
```

That's my understanding of it at least - I'm still not convinced I 100% understand it haha

x/ratelimit/rate_limit.go
sampocs: yeah that's exactly it.

Added some clarifying comments

app/app.go
sampocs: No specific reasoning. I'm fine with moving it closer to transfer

x/ratelimit/rate_limit.go
sampocs: Yeah I used the ibc-go code as a reference
https://github.com/cosmos/ibc-go/blob/main/modules/apps/transfer/keeper/relay.go#L200

x/ratelimit/rate_limit.go
sampocs: ibc-go makes the same assumption
https://github.com/cosmos/ibc-go/blob/main/modules/apps/transfer/keeper/relay.go#L210-L215

x/ratelimit/rate_limit.go
sampocs: I'm not sure I see the argument for having less documentation in the code itself. It seems arguably more important to have this in the code than the README. Is your thinking just that the comment was overwhelming?

I clarified some of the description and added the A,B,C notation in the latest commit, let me know if that alleviates some of the concern

x/ratelimit/keeper/rate_limit.go
sampocs: Osmosis did not sum across multiple channels - they just used the total supply. They mentioned that they _wanted_ to  sum across multiple channels but the functionality wasn't present in the SDK.

Their rationale was that they wanted to keep quota thresholds for the same denom relatively similar. For instance, there aren't a lot of people sending their `uosmo` to the hub, but there might be a lot that are sending it to Stride. So the quota thresholds on those two channels would have to be quite different (since the denominators are of different magnitudes), even though the denom is the same.

And they also felt that using the total escrow balance across each channel was more accurate than using the total supply since most native tokens were kept on Osmosis (as they had little utility elsewhere). 
Meaning: `Total Supply >>> Total Escrow `

In our case, most of our native tokens (stTokens) will be moved off Stride right away (since they have little utility on Stride), so the total supply is likely much closer to the total escrow balance and thus is a more accurate metric for us.
Meaning: `Total Supply ~= Total Escrow`

x/ratelimit/README.md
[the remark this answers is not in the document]
asalzmann: Ah, that's clever!

x/ratelimit/rate_limit.go
asalzmann: sounds good - might be worth adding a comment linking to that code

x/ratelimit/README.md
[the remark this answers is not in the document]
asalzmann: Yeah - my guess is that this is not possible (or shouldn't be).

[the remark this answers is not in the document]
sampocs: I actually think it is possible! But I think in those cases, the `transfer/channel-X/thistokenwasminted` is the `BaseDenom` - I don't think it gets conflated with the prefix.

These types of denoms were actually mentioned in the ibc-go [migration docs](https://ibc.cosmos.network/main/migrations/support-denoms-with-slashes.html)

x/ratelimit/ibc_module.go
sampocs: I believe it does cause the integration tests checked that balances were unchanged after failed txs. 

That said, let's re-evaluate once that other PR is in

app/app.go
sampocs: On second thought, I actually think the most logical ordering of everything is probably:

IBC
ratelimit
transfer
records

At the end of the day, it's pretty arbitrary since traffic floats in both directions. But records feels closest to the core application and rate limit/transfer feel more like related to the actually transport/packets.

asalzmann: That framework makes sense. transfer seems closer to IBC than ratelimit (you can have transfer w/o ratelimit, but not vice-versa)

sampocs: I'm not sure I follow the with/without argument? 

I think of transfer as closer to the application (e.g. it's what mints/burns tokens to users) and ratelimit more as middleware that is meant to sit in between the relaying of packets (IBC) and the handling of those packets (Base App)

sampocs: Another way we might want to think about it, is I think depending on the ordering, we might have to revert the action of the previous module.

---

So if we have 
IBC
transfer
ratelimit

Then for a receive packet (travels down the stack), we have to burn/re-escrow any tokens that were minted/un-escrowed by transfer.

---

Similarly if we have

IBC
ratelimit
transfer

Then for a send packet (travels up the stack), we have to mint/un-escrow any tokens that were burned/escrowed by transfer

asalzmann: > the relaying of packets (IBC) and the handling of those packets

I was holding we'd only ever rate limit transfer packets, so it would only make sense to include ratelimit in a middleware stack that also includes transfer. I guess it's possible we rate-limit other types of packets as well! Putting it next to IBC makes sense

> Another way we might want to think about it, is I think depending on the ordering, we might have to revert the action of the previous module.

Would we want the entire middleware stack to atomically succeed/fail? Looks like right now `SendRateLimitedPacket` throws an error (which should revert all other calls in the middleware stack), but `ReceiveRateLimitedPacket` returns a `channeltypes.NewErrorAcknowledgement(err)` (which I don't think will revert logic executed by other middlewares, but not 100% sure)

sampocs: yeah I think right now there's a bug - I'm refactoring the middleware wiring as we speak.

I think we need to revert one of the cases, and it depends on the ordering of the stack (as described in my last example), but right now we're reverting neither.

I'm also wondering if it makes sense to just have separate stacks?

asalzmann: Does [this error being thrown](https://github.com/Stride-Labs/stride/pull/556/files#diff-62a2ad25069547e2e6afd9aaa01547c698b11f7725c70ef4ce4c5d2c4080e83dR30) not revert sends? I would have expected the error to propagate through the middleware stack

> I'm also wondering if it makes sense to just have separate stacks?

Can you elaborate on this?

sampocs: Hmm actually yeah I think you're right with the current way we set it up. It think the SendPacket you linked above is called in ibc-go [here](https://github.com/cosmos/ibc-go/blob/main/modules/apps/transfer/keeper/relay.go#L136)

But that's just cause we added the rateLimitKeeper as the ics4Wrapper in the transfer keeper - but it's not clear to me yet if that's the right way to set it up

sampocs: Revisited this - 

I think transfer has to sit at the bottom of the stack because the transfer `IBCModule` does not have the `app` attribute used for middleware (e.g. [this](https://github.com/cosmos/ibc-go/blob/v5.1.0/modules/apps/transfer/ibc_module.go#L21-L23) vs [this](https://github.com/cosmos/ibc-go/blob/v5.1.0/modules/apps/29-fee/ibc_middleware.go#L21-L23)). Meaning, if you had transfer at the top of the stack, then there's no way for the Transfer `OnRecvPacket` callback to pass the packet down to a different middleware.

---

That then brings us full circle back to the question of whether the stack should be
```
records
ratelimit
transfer

OR

ratelimit
records
transfer
```

And between those two, I think I'd side with your OG suggestion that keeping ratelimit closest to transfer probably makes the most sense. 

Sorry for the runaround!

sampocs: And then back to the error question…

(once again for context)
* SendPacket flows `transfer` → `ratelimit` → `records`
* RecvPacket flow `records` → `ratelimit` → `transfer`

<br>
There are the following two error cases in the rate limit module:
* SendPacket - As implemented already and discussed above, I think throwing an error from `ratelimit` is adequate because it will propagate back to the transfer module and should atomically fail & discard anything the transfer module had done already
* OnRecvPacket - As implemented, I think returning an AckError is adequate here because nothing has been done yet on the Stride side of things that would need to get undone (contrary to something like OnAckPacket). And when we return the ack error, it should propagate back to `records` and then get emitted in the core IBC keeper

womensrights:
I was just passing by and saw this comment, you might be interested in [this middleware](https://github.com/celestiaorg/celestia-app/pull/1219) celestia have implemented

riley-stride:
> It would be nice if this PR supports blacklisting for specific token IBC transfer (receive/send) and export interface for registration of those tokens in keeper level so that it can be used on other modules.

Sam's suggestion to set the rate limit to 0 for the particular IBC denom in this case sounds like the cleanest solution to this question

app/upgrades.go
```diff
@@ -74,6 +74,14 @@ func (app *StrideApp) setupUpgradeHandlers() {
...
 		}
 	}
+	// TODO: RATE LIMIT UPGRADE
+	//  1. Add ratelimit store key when module is added
+	//     storeUpgrades = &storetypes.StoreUpgrades{
+	// 	     Added: []string{ratelimittypes.StoreKey},
+	//     }
+	//
+	// 2. Add hour epoch to store
+	// 3. Add rate limits for existing denoms
```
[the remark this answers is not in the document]
sampocs: It'd require adding the whole upgrade handler, so I think I'd lean toward doing it in a separate PR. Wdyt?

x/ratelimit/types/quota.go
```diff
@@ -0,0 +1,22 @@
...
+
+import (
+	sdkmath "cosmossdk.io/math"
+)
+
+// CheckExceedsQuota checks if new in/out flow is going to reach the max in/out or not
+func (q *Quota) CheckExceedsQuota(direction PacketDirection, amount sdkmath.Int, totalValue sdkmath.Int) bool {
+	// If there's no channel value (this should be almost impossible), it means there is no
+	// supply of the asset, so we shoudn't prevent inflows/outflows
+	if totalValue.IsZero() {
```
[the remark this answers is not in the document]
sampocs: Good question! That totalValue is only ever set from the total supply of the denom, so I think to get around this they would have to set the supply to 0 (which should be impossible)

app/upgrades.go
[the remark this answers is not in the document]
riley-stride: Agreed, makes sense to put in a separate PR. We should just make sure that goes in the next upgrade.

x/ratelimit/types/quota.go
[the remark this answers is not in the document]
riley-stride: Makes sense!

riley-stride (approved):
Unit testing coverage is really strong. I have no comments and I think this is ready to merge if you feel good about it @sampocs 🎉

x/ratelimit/README.md
```diff
@@ -0,0 +1,206 @@
...
+|   2  |      8usomo Osmosis → Stride     |    Successful   |    8   |    0    |     8%     |             |      100      |
+|   3  |      8usomo Osmosis → Stride     |     Rejected    |   16   |    0    | 16% (>10%) |             |      100      |
+|   3  | State reverted after rejected Tx |                 |    8   |    0    |     8%     |             |      100      |
+|   4  |   12ibc/uosmo Stride → Osmosis   |    Successful   |    8   |    12   |            |      4%     |      100      |
+|   5  |      8usomo Osmosis → Stride     |    Successful   |   16   |    12   |     4%     |             |      100      |
+|   6  |            Quota Reset           |                 |    0   |    0    |            |             |      104      |
+
+## Denom Blacklist
+The module also consists of a blacklist to completely halt all IBC transfers for a given denom. There are keeper functions to add or remove denoms from the blacklist; however, these functions are not exposed externally through transactions or governance, and they should only be leveraged internally from the protocol in extreme scenarios.
```
riley-stride, line 48:
```suggestion
The module also contains a blacklist to completely halt all IBC transfers for a given denom. There are keeper functions to add or remove denoms from the blacklist; however, these functions are not exposed externally through transactions or governance, and they should only be leveraged internally from the protocol in extreme scenarios.
```

x/ratelimit/keeper/rate_limit.go
```diff
@@ -0,0 +1,198 @@
...
+}
+
+// The total value on a given path (aka, the denominator in the percentage calculation)
+// is the total supply of the given denom
+func (k Keeper) GetChannelValue(ctx sdk.Context, denom string) sdkmath.Int {
+	return k.bankKeeper.GetSupply(ctx, denom).Amount
+}
+
+// If the rate limit is exceeded or the denom is blacklisted, we emit an event
+func EmitRateLimitExceededEvent(ctx sdk.Context, denom, channelId string, direction types.PacketDirection, amount sdkmath.Int, err error) {
```
riley-stride, line 28: Nit: might be nice to have separate events for attempts to transfer blacklisted assets vs attempts to transfer assets whose rate limits were exceeded. The former seems important to be able to identify at a moment's notice in potential emergency scenarios

riley-stride (approved):
LGTM!

riley-stride:
Was wondering @sampocs : is there any reason we should prevent from-and-to-Stride transfers of blacklisted assets? (eg bank send from stride1...A to stride1....B. I can't think of any, but curious if you can!

x/ratelimit/keeper/rate_limit.go
sampocs: I went back and forth on that a little. My thinking was you could easily determine whether it was blacklisted or throttled based on the error message and I thought it would be nice to have a common event type to listen to all blocked txs.

Another option is to change the event type to TransferBlocked and then add an attribute that's either RateLimitExceeded or BlacklistedDenom? 
Also fine with just using different event types

Wdyt?

sampocs:
> Was wondering @sampocs : is there any reason we should prevent from-and-to-Stride transfers of blacklisted assets? (eg bank send from stride1...A to stride1....B. I can't think of any, but curious if you can!

Good question! Nothing comes to mind

x/ratelimit/keeper/rate_limit.go
riley-stride: > change the event type to TransferBlocked and then add an attribute that's either RateLimitExceeded or BlacklistedDenom?

I think this would be nice n clean. I know this is nit, but it could make a big difference in the moment if there's ever a blacklist event

PR #577, opened 2023-01-15
fix `GET_VAL_ADDR` param

asalzmann:
**Summary**
`GET_VAL_ADDR` requires a `CHAIN_NAME`, not a `HOST_CHAIN_ID`. In many cases, these are the same (so this error went unnoticed) but Evmos has a different chain name / chain id, which causes a failure.

**Test plan**
Ran integration tests on Gaia, Evmos successfully


sampocs (approved):
Great catch!

PR #589, opened 2023-01-19
Fix Stakeibc Init/Export Genesis

sampocs:
Closes: #XXX

The stakeibc init and export genesis functions had some missing fields, and some unnecessary fields - this PR fixes both functions.

## Brief Changelog
* Added epoch tracker list to stakeibc init and export genesis 
* Removed unnecessary ICAAccount keeper functions
* Removed unnecessary ICAAccount queries
* Removed unnecessary MinValidatorRequirement keeper functions
* Moved epoch tracker key to stakeibc keys
* Removed pagination from epoch tracker query

## Author's Checklist

I have...

## Reviewers Checklist

I have...

      - [ ] not applicable
      - [ ] jira ticket `XXX` 
      - [ ] README.md 
      - [ ] not documented


proto/stride/stakeibc/genesis.proto
```diff
@@ -15,13 +14,10 @@ message GenesisState {
   Params params = 1 [ (gogoproto.nullable) = false ];
   string port_id = 2;
   // list of zones that are registered by the protocol
-  ICAAccount ica_account = 4;
   repeated HostZone host_zone_list = 5 [ (gogoproto.nullable) = false ];
   uint64 host_zone_count = 6;
-  // stores a map from hostZone base denom to hostZone
-  map<string, string> denom_to_host_zone = 9;
```
shellvish, line 22: thanks for removing this + ica_account!

x/stakeibc/genesis.go
```diff
@@ -10,48 +10,23 @@ import (
...
 	}
-	// Set all the hostZone
-	for _, elem := range genState.HostZoneList {
-		k.SetHostZone(ctx, elem)
+	for _, epochTracker := range genState.EpochTrackerList {
+		k.SetEpochTracker(ctx, epochTracker)
 	}
 
-	// Set hostZone count
-	k.SetHostZoneCount(ctx, genState.HostZoneCount)
```
shellvish, line 24: If we're removing HostZoneCount here, I think we should also remove it from the protos. I don't think HostZoneCount is used anywhere in the repo, so this should be safe! 

Also reasonable if we want to split this into another PR given it's not directly related to this issue, but feels like a small change that could be bundled in

x/stakeibc/genesis.go
```diff
@@ -10,48 +10,23 @@ import (
...
-	// genesis.PortId = k.GetPort(ctx)
-	// Get all iCAAccount
-	iCAAccount, found := k.GetICAAccount(ctx)
-	if found {
-		genesis.IcaAccount = &iCAAccount
-	}
+	genesis.Params = k.GetParams(ctx)
+	genesis.HostZoneList = k.GetAllHostZone(ctx)
 	genesis.EpochTrackerList = k.GetAllEpochTracker(ctx)
-	// this line is used by starport scaffolding # genesis/module/export
```
shellvish, line 54: Should we remove the scaffolding? I think this implies we can't use `ignite cli` but I don't think we're using it atm regardless

shellvish (commented):
Hm, I'm very surprised this doesn't add HostZone to the state when we export, and that the EpochTrackers still aren't being added. I'm going to try testing locally to see if we can export these without an upgrade  

Also really appreciate the standardization of GetEpochTracker on LCD vs CLI endpoints!

x/stakeibc/genesis.go
sampocs: I agree! Removing it now

x/stakeibc/genesis.go
sampocs: I think we already removed ignite as a dependency so I don't think it's possible to scaffold anymore

But I also think we understand the setup well enough at this point that it's much cleaner to not scaffold!

x/stakeibc/keeper/min_validator_requirements.go
```diff
@@ -1,34 +0,0 @@
-package keeper
```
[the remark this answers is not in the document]
sampocs: yes! good catch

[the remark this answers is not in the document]
sampocs: there's nothing I love more than scrubbing out unused code lol

x/stakeibc/genesis.go
shellvish: Agreed!

x/stakeibc/keeper/min_validator_requirements.go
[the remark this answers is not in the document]
shellvish: the most addictive drug known to man

shellvish (approved):
Seems like the issue was in state export on v4.0.3 

Sam and I tested the upgrade flow and local-stride; all seems good!

PR #590, opened 2023-01-20
Added Next Packet Sequence Query

sampocs:
Closes: #XXX

When monitoring ICA transactions, there's no clear cut way to check if an ICA was sent. To provide more transparency, this PR introduces a new query that returns the next packet sequence on the channel.

**Questions for Reviewer**:
1. Thoughts on the query name? Should it be more specific (i.e. that it's a send packet)
2. Is this all the info we need from a monitoring perspective? My thinking is we have the following scenarios:
  * _No ICA sent_ 
    *  Identified by incremented send sequence
  * _ICA sent but failed_ 
    *  Identified by incremented packet sequence + records unchanged
  * _ICA sent and pending_ 
    *  Identified by incremented packet sequence + record in state IN_PROGRESS
    * Also potentially identified by querying pending packets/acks
  * _ICA sent and success_ 
    *  Identified by incremented packet sequence + record modified/removed
 
## Brief Changelog
* Added `NextPacketSequence` query, CLI command and unit test

## Author's Checklist

I have...

- [x] OR this change is a trivial rework / code cleanup without any test coverage

## Reviewers Checklist

I have...

      - [ ] not applicable
      - [ ] jira ticket `XXX` 
      - [ ] README.md 
      - [ ] not documented


x/stakeibc/README.md
```diff
@@ -91,24 +91,16 @@ Governance
 
 ## Queries
 
-- `QueryInterchainAccountFromAddressRequest`
```
shellvish, line 94: why delete these queries? are they deprecated?

proto/stride/stakeibc/query.proto
```diff
@@ -65,7 +65,11 @@ service Query {
     option (google.api.http).get = "/Stride-Labs/stride/stakeibc/epoch_tracker";
   }
 
-  // this line is used by starport scaffolding # 2
+  // Queries the next packet sequence for one for a given channel
+  rpc NextPacketSequence(QueryGetNextPacketSequenceRequest)
+      returns (QueryGetNextPacketSequenceResponse) {
+    option (google.api.http).get = "/Stride-Labs/stride/stakeibc/next_packet_sequence/{channel_id}/{port_id}";
```
shellvish, line 71: do you think we should add this query to `stakeibc` or `icacallbacks`?

shellvish (approved):
Minor comments but not blocking, this looks great!

x/stakeibc/README.md
sampocs: The heading is queries so I thought it made more sense to just document the query type, rather than doing QueryTypeRequest and QueryTypeResponse for each

proto/stride/stakeibc/query.proto
sampocs: good question, I was wondering this as well. Maybe icacallbacks does make more sense? What do you think?

sampocs: I think icacallbacks might be slightly misleading cause this applies to all packets, not just ica's. But from a usage perspective, I think it aligns more with callbacks

sampocs: Stakeibc is just kind of a collect-all of misc stuff haha

sampocs: Chatted offline. It does seem more closely aligned with callbacks, but:
  * The purpose of the query is for stakeibc monitoring
  * Icacallbacks is treated more as public infra and if another project imported it, they would not expect that query (i.e. icacallbacks is more well scoped)

So the final consensus is to leave it in stakeibc for now

PR #594, opened 2023-01-26
Initial Instant Redeem Stake implementation (only pulls from Deposit Records)

khdegraaf:

Initial implementation for Instant redemptions, only pulling tokens from Deposit Records in current epoch

**Design for instant redemptions**
To process instant redemptions, the basic idea is simple - we have some ATOM that is about to be staked, which can be used to process redemptions. The reason this works is because 

Total_staked_n+1 = new_staked_n - new_unstaked_n

So we can short-circuit unbondings by decreasing both unstaked and staked amounts. Unstaked tokens are tracked on `HostZoneUnbonding` records (on `EpochUnbondingRecord` records). Staked tokens that can be matched with unbondings can be

1. Newly staked tokens (tracked on `DepositRecord` records)
2. Interest that is about to be reinvested

Newly staked tokens are simple but unpredictable, there may be many epochs where Total_staked_n+1 < 0 (therefore, we can’t process instant unbondings). Interest that is about to be reinvested is slightly more complex but predictable - it accrues daily and is proportional to TVL.

For newly staked tokens, we can just match redemptions to the `DepositRecord` from the current epoch. If the redeemed amount is greater than the amount on the deposit record, get the amount to unbond (using the redemption rate), burn the corresponding stATOM, and send the ATOM to the redeemer, decrementing the deposit record.

For accrued interest, the interest is on an ICA. Therefore we have to IBC transfer tokens back to Stride and allow them to be redeemed for one epoch, before sending the remainder back to the host zone. We will likely need a new data type (`InterestRecord`) that tracks an epoch, amount, and zone. We will also need a new module account to store tokens in and epochly logic to handle IBC token transfers.

**x/stakeibc**
- InstantRedeemStake(creator, amt, zone): checks the sources of tokens that can be unbonded (`DepositRecord`, `InterestRecord`) and processes an instant-redeem-stake if it’s valid  

## Brief Changelog

  - InstantRedeemStake implementation, including cli command and unit testing

## Testing

**Setup**
- Start dockernet, run an integration test to verify everything is funded and running.
```
make start-docker guild=sg
make test-integration-docker
```
- Review everything is OK, and current status of account
```
strided q records list-deposit-record
strided q bank balances stride1uk4ze0x4nvh4fk0xm4jdud58eqn4yxhrt52vv7
 ```
- Make sure there is a deposit record by staking some uosmo
````
strided tx stakeibc liquid-stake 100 uosmo --from val1 --keyring-backend=test --chain-id STRIDE
````
- Try to instant redeem stake a smaller valid amount for success
````
strided tx stakeibc instant-redeem-stake 10 OSMO --from val1 --keyring-backend=test --chain-id STRIDE
````
- Try to instant redeem a larger amount, or wait for epoch to end for failure
````
strided tx stakeibc instant-redeem-stake 200 OSMO --from val1 --keyring-backend=test --chain-id STRIDE
````
- Verify transaction completed by querying the txns, and checking the balances
````
strided q records list-deposit-record
strided q bank balances stride1uk4ze0x4nvh4fk0xm4jdud58eqn4yxhrt52vv7
````

## Author's Checklist

I have...

- [x] If the change is contentful, I either:

## Reviewers Checklist

I have...

      - [ ] not applicable
      - [ ] jira ticket `XXX` 
      - [ ] README.md 
      - [ ] not documented

### CODE REVIEW

File: x/stakeibc/keeper/msg_server_fast_unbond.go [hunk 1]
```diff
@@ -0,0 +1,117 @@
...
+	}
+
+	// Find and subtract this amount from a deposit record if it is big enough
+	epochTracker, found := k.GetEpochTracker(ctx, epochtypes.STRIDE_EPOCH)
+	if !found {
+		k.Logger(ctx).Error("failed to find stride epoch")
+		return nil, sdkerrors.Wrapf(sdkerrors.ErrNotFound, "no epoch number for epoch (%s)", epochtypes.STRIDE_EPOCH)
+	}
+	// Does this use too much gas?
+	depositRecord, found := k.RecordsKeeper.GetDepositRecordByEpochAndChain(ctx, epochTracker.EpochNumber, hostZone.ChainId)
```
asalzmann, line 80: I think technically we can sum up all depositRecords that haven't been transferred (there could be multiple if IBC transfers fail) and process redemptions. Curious what you think

x/stakeibc/keeper/msg_server_fast_unbond.go
```diff
@@ -0,0 +1,117 @@
...
+	hostZone, found := k.GetHostZone(ctx, msg.HostZone)
+	if !found {
+		return nil, sdkerrors.Wrapf(types.ErrInvalidHostZone, "host zone is invalid: %s", msg.HostZone)
+	}
+
+	// get the coins to return, they need to be in the format {amount}{denom}
+	// is safe. The converse is not true.
+	ibcDenom := hostZone.GetIbcDenom()
+	coinString := msg.Amount.String() + ibcDenom
+	outCoin, err := sdk.ParseCoinNormalized(coinString)
```
asalzmann, line 34: I'm finding this naming schema confusing with
outCoin
nativeAmount
inCoin

How can we rename this to make this more clear?

x/stakeibc/keeper/msg_server_fast_unbond.go
```diff
@@ -0,0 +1,117 @@
...
+	if err != nil {
+		k.Logger(ctx).Error(fmt.Sprintf("failed to parse coin (%s)", coinString))
+		return nil, sdkerrors.Wrapf(err, "failed to parse coin (%s)", coinString)
+	}
+	// construct desired unstaking amount from host zone
+	stDenom := types.StAssetDenomFromHostZoneDenom(hostZone.HostDenom)
+	nativeAmount := sdk.NewDecFromInt(msg.Amount).Mul(hostZone.RedemptionRate).RoundInt()
```
asalzmann, line 41: If the `Amount` input represents native tokens, and this value represents stTokens, we should divide here

```
stATOM = ATOM / ATOM_RR
```

asalzmann (commented):
Overall, starting to look good! I found the naming of the coin variables a bit confusing - what does `Amount` on `MsgFastUnbond` represent?

x/stakeibc/keeper/msg_server_fast_unbond.go
```diff
@@ -0,0 +1,117 @@
...
+		return nil, sdkerrors.Wrapf(sdkerrors.ErrInvalidCoins, "balance is lower than redemption amount. redemption amount: %v, balance %v: ", msg.Amount, balance.Amount)
+	}
+
+	// Find and subtract this amount from a deposit record if it is big enough
+	epochTracker, found := k.GetEpochTracker(ctx, epochtypes.STRIDE_EPOCH)
+	if !found {
+		k.Logger(ctx).Error("failed to find stride epoch")
+		return nil, sdkerrors.Wrapf(sdkerrors.ErrNotFound, "no epoch number for epoch (%s)", epochtypes.STRIDE_EPOCH)
+	}
+	// Does this use too much gas?
```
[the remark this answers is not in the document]
khdegraaf: Sure.  This is copy/pasta from similar code with the same comment.  Should I clean it up in other places?

x/stakeibc/keeper/msg_server_fast_unbond.go
```diff
@@ -0,0 +1,117 @@
...
+	}
+
+	// safety check: redemption rate must be within safety bounds
+	rateIsSafe, err := k.IsRedemptionRateWithinSafetyBounds(ctx, hostZone)
+	if !rateIsSafe || (err != nil) {
+		errMsg := fmt.Sprintf("IsRedemptionRateWithinSafetyBounds check failed. hostZone: %s, err: %s", hostZone.String(), err.Error())
+		return nil, sdkerrors.Wrapf(types.ErrRedemptionRateOutsideSafetyBounds, errMsg)
+	}
+
+	// TODO(TEST-112) bigint safety
```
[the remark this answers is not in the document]
khdegraaf: This is copy/pasta from other code.  Should I remove all instances of this?

x/stakeibc/keeper/msg_server_fast_unbond.go
```diff
@@ -0,0 +1,117 @@
...
+	}
+
+	// then make sure host zone is valid
+	hostZone, found := k.GetHostZone(ctx, msg.HostZone)
+	if !found {
+		return nil, sdkerrors.Wrapf(types.ErrInvalidHostZone, "host zone is invalid: %s", msg.HostZone)
+	}
+
+	// get the coins to return, they need to be in the format {amount}{denom}
+	// is safe. The converse is not true.
```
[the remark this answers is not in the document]
khdegraaf: Haha, great question.  This snippet is copied from https://github.com/Stride-Labs/stride/blob/main/x/stakeibc/keeper/msg_server_liquid_stake.go#L31-L32, and it appears that the meaning got a bit garbled by Vishay's commit here: https://github.com/Stride-Labs/stride/commit/72e5b3d9b9cf5ac34a6acc8293555f3b350940bd.  I had planned to review and fix up a few of these issues before requesting a PR, but hadn't done so yet.  The second line should be deleted most likely in both places where this exists.

x/stakeibc/keeper/msg_server_fast_unbond.go
```diff
@@ -0,0 +1,117 @@
...
+	// construct desired unstaking amount from host zone
+	stDenom := types.StAssetDenomFromHostZoneDenom(hostZone.HostDenom)
+	nativeAmount := sdk.NewDecFromInt(msg.Amount).Mul(hostZone.RedemptionRate).RoundInt()
+
+	if nativeAmount.GT(hostZone.StakedBal) {
+		return nil, sdkerrors.Wrapf(types.ErrInvalidAmount, "cannot unstake an amount g.t. staked balance on host zone: %v", msg.Amount)
+	}
+
+	// safety check: redemption rate must be within safety bounds
+	rateIsSafe, err := k.IsRedemptionRateWithinSafetyBounds(ctx, hostZone)
```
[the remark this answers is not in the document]
khdegraaf: Both LiquidStake and RedeemStake have this same check in similar places.  From a purely functional perspective, it seems mostly immaterial to me.  What would be the rationale for doing this?

x/stakeibc/keeper/msg_server_fast_unbond.go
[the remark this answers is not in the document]
khdegraaf: fixed.

proto/stride/stakeibc/tx.proto
```diff
@@ -150,4 +162,4 @@ message MsgUpdateValidatorSharesExchRate {
 
 message MsgUpdateValidatorSharesExchRateResponse {}
 
-// this line is used by starport scaffolding # proto/tx/message
\ No newline at end of file
+// this line is used by starport scaffolding # proto/tx/message
```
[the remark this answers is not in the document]
khdegraaf: fixed

x/stakeibc/keeper/msg_server_fast_unbond.go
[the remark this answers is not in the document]
khdegraaf: fixed

x/stakeibc/keeper/msg_server_fast_unbond.go
khdegraaf: Fair point.  I wasn't aware of that detail when I wrote this.  I'll modify to address this.

x/stakeibc/keeper/msg_server_fast_unbond.go
khdegraaf: The naming makes this a bit confusing, but msg.Amount is the amount of stTokens to redeem (just like RedeemStake function), so I am computing

`ATOM = stATOM * ATOM_RR`

This code is identical to https://github.com/Stride-Labs/stride/blob/main/x/stakeibc/keeper/msg_server_redeem_stake.go#L50-L51 in RedeemStake function.

x/stakeibc/keeper/msg_server_fast_unbond.go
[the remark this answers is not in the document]
asalzmann: Agreed, functionally immaterial. I was thinking it would be nice break up the logical flow of the function to
1) safety checks
2) setting up the coin variables
3) executing the coin-related logic

Open to alternatives though! What do you think?

x/stakeibc/keeper/msg_server_fast_unbond.go
asalzmann: Yes, we try to process transfers each epoch, but a transfer could fail. Any deposit record that has status `TRANSFER_QUEUE` should be eligible. I believe we have existing filtering functions in hooks.go

x/stakeibc/keeper/msg_server_fast_unbond.go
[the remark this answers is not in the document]
khdegraaf: That makes sense.  I have edited the code to reflect that.  Let me know what you think.

x/stakeibc/keeper/msg_server_fast_unbond.go
khdegraaf: OK, I think I have fixed this.  I found the previous schema confusing as well.  Let me know what you think.

x/stakeibc/keeper/msg_server_fast_unbond.go
khdegraaf: I think this is a lot clearer after I fixed the naming and reorganized the code.

x/stakeibc/keeper/msg_server_fast_unbond.go
khdegraaf: OK, I have code to handle this.

x/stakeibc/keeper/msg_server_instant_redeem_stake.go
```diff
@@ -0,0 +1,112 @@
...
+	rateIsSafe, err := k.IsRedemptionRateWithinSafetyBounds(ctx, hostZone)
+	if !rateIsSafe || (err != nil) {
+		errMsg := fmt.Sprintf("IsRedemptionRateWithinSafetyBounds check failed. hostZone: %s, err: %s", hostZone.String(), err.Error())
+		return nil, sdkerrors.Wrapf(types.ErrRedemptionRateOutsideSafetyBounds, errMsg)
+	}
+
+	// construct desired unstaking amount from host zone
+	nativeAmount := sdk.NewDecFromInt(msg.Amount).Mul(hostZone.RedemptionRate).RoundInt()
+	// Redemption amount must not be greater than staked balance in that zone.
+	if nativeAmount.GT(hostZone.StakedBal) {
```
riley-stride, line 42: Could this check ever trigger? (in cases that do not involve a hack/bug). If not, I wonder if we even need this check in `msg_server_redeem_stake`.

Might be simpler to use a stricter check below, after we sum the DRs: `nativeAmount.GT(moduleAccount balance)`.

x/stakeibc/keeper/msg_server_instant_redeem_stake.go
```diff
@@ -0,0 +1,112 @@
...
+	}
+
+	// Find and subtract this amount from a deposit record if it is big enough
+	depositRecords := k.RecordsKeeper.GetAllDepositRecord(ctx)
+	pendingDepositRecords := utils.FilterDepositRecords(depositRecords, func(record recordstypes.DepositRecord) (condition bool) {
+		return record.Status == recordstypes.DepositRecord_TRANSFER_QUEUE && record.HostZoneId == hostZone.ChainId
+	})
+	totalPendingDeposits := utils.SumDepositRecords(pendingDepositRecords)
+	fmt.Printf("totalPendingDeposits = %v\n", totalPendingDeposits)
+	if nativeAmount.GT(totalPendingDeposits) {
```
riley-stride, line 70: Open product question: if a user invokes `instant-redeem-stake` and the module account does not contain sufficient balance to fulfill the instant redemption, should the tx fail, or should it initiate a regular `redeem-stake`? I lean towards failing the tx (as you implemented here) to avoid confusion.

x/stakeibc/keeper/msg_server_instant_redeem_stake.go
```diff
@@ -0,0 +1,112 @@
...
+	if balance.Amount.LT(msg.Amount) {
+		return nil, sdkerrors.Wrapf(sdkerrors.ErrInvalidCoins, "balance is lower than redemption amount. redemption amount: %v, balance %v: ", msg.Amount, balance.Amount)
+	}
+
+	// Find and subtract this amount from a deposit record if it is big enough
+	depositRecords := k.RecordsKeeper.GetAllDepositRecord(ctx)
+	pendingDepositRecords := utils.FilterDepositRecords(depositRecords, func(record recordstypes.DepositRecord) (condition bool) {
+		return record.Status == recordstypes.DepositRecord_TRANSFER_QUEUE && record.HostZoneId == hostZone.ChainId
+	})
+	totalPendingDeposits := utils.SumDepositRecords(pendingDepositRecords)
```
riley-stride, line 68: I believe we should only ever have 1 DepositRecord with status TRANSFER_QUEUE per HostZone at a time, so might not need to sum here, and might be able to replace the loop on ln 75 with a simple subtraction. @asalzmann can you check me on this?

x/stakeibc/keeper/msg_server_instant_redeem_stake.go
```diff
@@ -0,0 +1,112 @@
...
+	fmt.Printf("totalPendingDeposits = %v\n", totalPendingDeposits)
+	if nativeAmount.GT(totalPendingDeposits) {
+		return nil, sdkerrors.Wrapf(types.ErrInvalidAmount, "cannot fast unbond an amount %v g.t. pending deposit balance on host zone: %v", nativeAmount, msg.Amount)
+	}
+	// Subtract all of nativeAmount from one or more pending deposit records
+	x := nativeAmount
```
riley-stride, line 74: Might be able to replace this loop with a simple subtraction; see comment on line 68

x/stakeibc/keeper/msg_server_instant_redeem_stake.go
```diff
@@ -0,0 +1,112 @@
...
+	for _, depositRecord := range pendingDepositRecords {
```
riley-stride, line 75: Might not need this loop if we only have one depositRecord matching the conditions, per comment on line 68

x/stakeibc/keeper/msg_server_instant_redeem_stake_test.go
```diff
@@ -0,0 +1,192 @@
...
+		Address:        zoneAddress.String(),
+		StakedBal:      unbondAmount,
+	}
+
+	epochTracker := stakeibctypes.EpochTracker{
+		EpochIdentifier: epochtypes.STRIDE_EPOCH,
+		EpochNumber:     1,
+	}
+
+	oldDepositRecord := recordtypes.DepositRecord{
```
riley-stride, line 61: Do these need Status TRANSFER_QUEUE?

x/stakeibc/keeper/msg_server_instant_redeem_stake_test.go
```diff
@@ -0,0 +1,192 @@
...
+		Id:                 0,
+		DepositEpochNumber: 0,
+		HostZoneId:         "GAIA",
+		Amount:             depositAmount,
+	}
+
+	initialDepositRecord := recordtypes.DepositRecord{
```
riley-stride, line 68: Makes sense to test the case with multiple DRs, but only one should have Status TRANSFER_QUEUE per host zone I believe

x/stakeibc/keeper/msg_server_instant_redeem_stake_test.go
```diff
@@ -0,0 +1,192 @@
...
+		validMsg: stakeibctypes.MsgInstantRedeemStake{
+			Creator:  user.acc.String(),
+			HostZone: HostChainId,
+			Amount:   unbondAmount,
+		},
+	}
+}
+
+// TODO: Need to add tests for at least multiple deposit records, non 1.0 Redemption Rates, and probably some other basic scenarios.
+func (s *KeeperTestSuite) TestInstantRedeemStake_Successful() {
```
riley-stride, line 97: Could we add tests for
- no new funds were staked in this epoch (q: in this case, will the DR exist with value 0 or will it not exist?)
- instant redeem staking the entire amount of the DR, then trying to instant redeem stake again after that (second should fail I think, but I believe the DR should remain, with value 0)

riley-stride (changes requested):
This is starting to come together. 

Requesting some changes, main point: I think we can simplify some of the logic in the msg_server since only one DR can match your criteria.

x/stakeibc/keeper/msg_server_instant_redeem_stake.go
khdegraaf: It shouldn't be possible other than hack/bug.  It does seem to have limited utility, I put it in here since you had it in `msg_server_redeem_stake` .

x/stakeibc/keeper/msg_server_instant_redeem_stake.go
khdegraaf: Yes, I think this behavior is a bit cleaner.

x/stakeibc/keeper/msg_server_instant_redeem_stake.go
khdegraaf: I did it that way initially, and changed it to this upon @asalzmann's request and comment above.  I reviewed the code, and it does look like there can be multiple entries if they timeout.

x/stakeibc/keeper/msg_server_instant_redeem_stake_test.go
khdegraaf: Previous failed transfers when they time out will be set back to TRANSFER_QUEUE, so we will try again in the future.  So the case where we have multiple entries is when they are all this.

x/stakeibc/keeper/msg_server_instant_redeem_stake_test.go
khdegraaf: Each epoch creates a record, with value of 0.  So if there are no new funds, it is there with a value of 0.  I'll add both cases when I do more scenarios.

x/stakeibc/keeper/msg_server_instant_redeem_stake.go
shellvish: +1 I'd bias towards replacing this with the more restrictive check `nativeAmount.GT(moduleAccount balance)`

x/stakeibc/keeper/msg_server_instant_redeem_stake.go
```diff
@@ -0,0 +1,110 @@
...
+	}
+
+	// safety check: redemption rate must be within safety bounds
+	rateIsSafe, err := k.IsRedemptionRateWithinSafetyBounds(ctx, hostZone)
+	if !rateIsSafe || (err != nil) {
+		errMsg := fmt.Sprintf("IsRedemptionRateWithinSafetyBounds check failed. hostZone: %s, err: %s", hostZone.String(), err.Error())
+		return nil, sdkerrors.Wrapf(types.ErrRedemptionRateOutsideSafetyBounds, errMsg)
+	}
+
+	// construct desired unstaking amount from host zone
```
shellvish, line 39: I think we should add some logic here to charge a fee on `InstantRedeemStake` (even though initial fee will be zero), just to make it easier to publicly signal that this won't be a free feature forever.

I think this should just be defining some `feeAmount` here, then creating a new `RedeemAmount` that is `msg.Amount - feeAmount`, and then transferring the fee to the community pool at the end of the function.

shellvish (commented):
This is looking quite good so far! 

One observation: this design will require us to be able to initiate an IBC transfer from the HostZone. This is a perfectly reasonable assumption, and we should be able to guarantee it for all HostZones going forward. Stargaze, however, doesn't have this in their ICA AllowMessages; we should put up a proposal to change that before this feature goes live.

x/stakeibc/keeper/msg_server_instant_redeem_stake.go
khdegraaf: I thought about this some more and carefully reviewed the code.  Given that the StakedBal actually represents the amount staked in the host zone after the funds have been transferred there, and therefore no longer in the Deposit Account, this check is actually wrong and doesn't make sense for instant-redeem-stake (although usually not an issue as amount in host zone will be much larger).  Strictly speaking, and absent any bugs, the amount in the Deposit Account should always be the amount in the deposit records (which I check that it doesn't exceed immediately after) and that's it.  If there is any discrepancy, not the SendCoins call should fail if the amount to send is greater than the amount in the account.  It sounds like you are fine with simply removing this check?  Or did you want to see an explicit semi-redundant check that this amount doesn't exceed the module account balance, maybe for a better error message?

x/stakeibc/keeper/msg_server_instant_redeem_stake.go
khdegraaf: See comment above.

x/stakeibc/keeper/msg_server_instant_redeem_stake.go
khdegraaf: See comment above.

x/stakeibc/keeper/msg_server_instant_redeem_stake.go
khdegraaf: So an absolute amount, not a percentage then?  Also, are you thinking this is just a constant in code to signal intention, or a module parameter?

x/stakeibc/keeper/msg_server_instant_redeem_stake_test.go
```diff
@@ -0,0 +1,201 @@
...
+	hostZone := stakeibctypes.HostZone{
+		ChainId:        HostChainId,
+		HostDenom:      Atom,
+		IbcDenom:       IbcAtom,
+		RedemptionRate: sdk.NewDec(1.0),
+		Address:        zoneAddress.String(),
+		StakedBal:      unbondAmount,
+	}
+
+	epochTracker := stakeibctypes.EpochTracker{
```
sampocs, line 58: I don't think you need the epoch tracker for these tests

x/stakeibc/keeper/msg_server_instant_redeem_stake_test.go
```diff
@@ -0,0 +1,201 @@
...
+		DepositEpochNumber: 1,
+		HostZoneId:         "GAIA",
+		Amount:             depositAmount,
+		Status:             recordtypes.DepositRecord_TRANSFER_QUEUE,
+	}
+
+	s.App.StakeibcKeeper.SetHostZone(s.Ctx, hostZone)
+	s.App.StakeibcKeeper.SetEpochTracker(s.Ctx, epochTracker)
+	s.App.RecordsKeeper.SetDepositRecord(s.Ctx, oldDepositRecord)
+	s.App.RecordsKeeper.SetDepositRecord(s.Ctx, initialDepositRecord)
```
sampocs, line 82:
```suggestion
	s.App.RecordsKeeper.SetDepositRecord(s.Ctx, depositRecord1)
	s.App.RecordsKeeper.SetDepositRecord(s.Ctx, depositRecord2)
```

proto/stride/stakeibc/tx.proto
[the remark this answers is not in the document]
khdegraaf: fixed.

x/stakeibc/keeper/msg_server_instant_redeem_stake_test.go
```diff
@@ -0,0 +1,201 @@
...
+
+	// User STUATOM balance should have DECREASED by the amount unbonded
+	expectedUserStAtomBalance := user.stAtomBalance.SubAmount(msg.Amount)
+	actualUserStAtomBalance := s.App.BankKeeper.GetBalance(s.Ctx, user.acc, StAtom)
+	s.CompareCoins(expectedUserStAtomBalance, actualUserStAtomBalance, "user stuatom balance")
+	// User IBC/UATOM balance should have INCREASED by the amount unbonded
+	expectedUserAtomBalance := user.atomBalance.AddAmount(msg.Amount)
+	actualUserAtomBalance := s.App.BankKeeper.GetBalance(s.Ctx, user.acc, IbcAtom)
+	s.CompareCoins(expectedUserAtomBalance, actualUserAtomBalance, "user ibc/uatom balance")
+	// zoneAccount IBC/UATOM balance should have DECREASED by the size of the stake
```
sampocs, line 121:
```suggestion
	// zoneAccount IBC/UATOM balance should have DECREASED by the size of the redemption
```

x/stakeibc/keeper/msg_server_instant_redeem_stake_test.go
```diff
@@ -0,0 +1,201 @@
...
+	expectedzoneAccountAtomBalance := zoneAccount.atomBalance.SubAmount(msg.Amount)
+	actualzoneAccountAtomBalance := s.App.BankKeeper.GetBalance(s.Ctx, zoneAccount.acc, IbcAtom)
+	s.CompareCoins(expectedzoneAccountAtomBalance, actualzoneAccountAtomBalance, "zoneAccount ibc/uatom balance")
+	// Bank supply of STUATOM should have DECREASED by the size of the stake
```
sampocs, line 125:
```suggestion
	// Bank supply of STUATOM should have DECREASED by the size of the redemption
```

x/stakeibc/keeper/msg_server_instant_redeem_stake_test.go
```diff
@@ -0,0 +1,201 @@
...
+	expectedBankSupply := initialStAtomSupply.SubAmount(msg.Amount)
+	actualBankSupply := s.App.BankKeeper.GetSupply(s.Ctx, StAtom)
+	s.CompareCoins(expectedBankSupply, actualBankSupply, "bank stuatom supply")
+
+	// Validate Instant Redeem Stake, subsequent call will not have funds.
+	_, err = s.GetMsgServer().InstantRedeemStake(sdk.WrapSDKContext(s.Ctx), &msg)
+	s.Require().EqualError(err, fmt.Sprintf("balance is lower than redemption amount. redemption amount: %v, balance %v: : invalid coins", msg.Amount, actualUserStAtomBalance.Amount))
+}
+
+func (s *KeeperTestSuite) TestInstantRedeemStake_InvalidCreatorAddress() {
```
sampocs, line 135: For future reference, these tests are mostly covered by the ValidateBasic unit test, so you really only one of these test cases to cover the error path in MsgInstantRedeem.

That said, no need to remove these cause there's nothing wrong with more tests!

x/stakeibc/keeper/msg_server_instant_redeem_stake_test.go
```diff
@@ -0,0 +1,201 @@
...
+	_, err := s.GetMsgServer().InstantRedeemStake(sdk.WrapSDKContext(s.Ctx), &invalidMsg)
+
+	s.Require().EqualError(err, fmt.Sprintf("balance is lower than redemption amount. redemption amount: %v, balance %v: : invalid coins", invalidMsg.Amount, tc.initialState.initialAmount))
+}
+
+func (s *KeeperTestSuite) TestInstantRedeemStake_InvalidHostAddress() {
+	tc := s.SetupInstantRedeemStake()
+
+	// Update hostzone with invalid address
+	badHostZone, _ := s.App.StakeibcKeeper.GetHostZone(s.Ctx, tc.validMsg.HostZone)
```
sampocs, line 195:
```suggestion
	badHostZone := tc.validMsg.HostZone
```

x/stakeibc/keeper/msg_server_instant_redeem_stake_test.go
```diff
@@ -0,0 +1,201 @@
...
+	stakeibctypes "github.com/Stride-Labs/stride/v5/x/stakeibc/types"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	_ "github.com/stretchr/testify/suite"
+)
+
+type InstantRedeemStakeState struct {
+	epochNumber         uint64
+	depositRecordAmount sdkmath.Int
+	initialAmount       sdkmath.Int
+	hostZone            stakeibctypes.HostZone
```
sampocs, line 17: I dont think you need `epochNumber` or `initialAmount` here since they're not leveraged directly by any of the tests

x/stakeibc/keeper/msg_server_instant_redeem_stake_test.go
sampocs: +1 on testing the DR goes unchanged. Would be good to add a case that redeems 750_000 - it should deplete one DR completely and then deplete the second by 50%.
Also agree with your comment - I'd definitely add a test with a non-1.0 redemption rate!

x/stakeibc/keeper/msg_server_fast_unbond.go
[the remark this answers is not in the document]
sampocs: yeah if you notice stale comments like this in other places, please feel free to remove em! 

A lot of these basic message server functions were written 6+months ago and could probably be cleaned up a bit (e.g. the variable naming). Definitely don't feel like your functions have to look identical - if you notice something could be more clear, go for it!

x/stakeibc/keeper/msg_server_instant_redeem_stake.go
sampocs: My vote is to just remove. I think it makes sense in normal redeem stake (I believe it was in the audit report), but here it is irrelevant since only not-yet-staked funds are of concern

x/stakeibc/keeper/msg_server_instant_redeem_stake.go
sampocs: +1 on just failing immediately

sampocs (commented):
Great work! No high level comments on top of the other reviews - just some nit comments below:

x/stakeibc/keeper/msg_server_instant_redeem_stake_test.go
khdegraaf: FYI, I fixed this.

x/stakeibc/keeper/msg_server_instant_redeem_stake_test.go
khdegraaf: Yeah, I did when I wrote it initially, but I don't anymore

x/stakeibc/keeper/msg_server_instant_redeem_stake_test.go
khdegraaf: Yeah, currently true for epochNumber, but initialAmount is used in a test.

x/stakeibc/keeper/msg_server_instant_redeem_stake.go
riley-stride: Agreed with removing it from here for the reasons Kevin listed.

IMO we should also remove it from `msg_server_redeem_stake`. @sampocs why does it makes sense to keep there, even if it was listed in the audit report?

x/stakeibc/keeper/msg_server_instant_redeem_stake.go
riley-stride: Agreed, resolving

x/stakeibc/keeper/msg_server_instant_redeem_stake_test.go
riley-stride: Thanks, resolving

x/stakeibc/keeper/msg_server_instant_redeem_stake_test.go
riley-stride: Great point. Resolving

x/stakeibc/keeper/msg_server_instant_redeem_stake_test.go
riley-stride: Leaving this open until we add these tests

x/stakeibc/keeper/msg_server_instant_redeem_stake.go
shellvish: Oh apologies, `feeAmount` above would be a calculated value, I think the fee should be a percentage. 

I think we should make it a module parameter, to make it a bit more explicit (and this _feels_ like a variable that governance should be allowed to change). 

Let me know what you think though!

x/stakeibc/client/cli/tx_instant_redeem_stake.go
```diff
@@ -0,0 +1,50 @@
...
+	"github.com/cosmos/cosmos-sdk/client/flags"
+	"github.com/cosmos/cosmos-sdk/client/tx"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
+	"github.com/spf13/cobra"
+
+	"github.com/Stride-Labs/stride/v5/x/stakeibc/types"
+)
+
+var _ = strconv.Itoa(0)
```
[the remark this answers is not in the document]
khdegraaf: done

x/stakeibc/keeper/msg_server_instant_redeem_stake.go
```diff
@@ -0,0 +1,106 @@
...
+	depositRecords := k.RecordsKeeper.GetAllDepositRecord(ctx)
+	pendingDepositRecords := utils.FilterDepositRecords(depositRecords, func(record recordstypes.DepositRecord) (condition bool) {
+		return record.Status == recordstypes.DepositRecord_TRANSFER_QUEUE && record.HostZoneId == hostZone.ChainId
+	})
+	totalPendingDeposits := utils.SumDepositRecords(pendingDepositRecords)
+	if nativeAmount.GT(totalPendingDeposits) {
+		return nil, sdkerrors.Wrapf(types.ErrInvalidAmount, "cannot fast unbond an amount %v g.t. pending deposit balance on host zone: %v", nativeAmount, msg.Amount)
+	}
+	// Subtract all of nativeAmount from one or more pending deposit records
+	x := nativeAmount
```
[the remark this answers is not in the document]
khdegraaf: I guess I am trying to signal that this value is purely temporary and has no long lasting meaning.  You can look at the code overall, and nativeAmount is a constant once it is initialized.  Given my background in functional programming, I have a bit of a bias in that direction.  Strictly speaking, we don't use nativeAmount after this point, but we do use nativeCoin.Amount, so we could've.  Leaving this value alone I think reduces the chances for confusion.   Naming the temporary keeping track of amount left as "x" means you can't infer anything from the name (it might not even go to 0), which I think is more accurate.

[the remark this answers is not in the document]
sampocs: Chatted offline - agreed on functional paradigm and indicating the variable is temporary / not modifying `nativeAmount`
Proposed solution is just to continue using a separate variable but use one that is more descriptive than simply `x` (such as `remainingNativeAmount`)

x/stakeibc/keeper/msg_server_instant_redeem_stake.go
riley-stride: this is now removed

x/stakeibc/keeper/msg_server_instant_redeem_stake_test.go
riley-stride: Most of these tests have been added, just missing one case

x/stakeibc/keeper/msg_server_instant_redeem_stake.go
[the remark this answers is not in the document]
riley-stride: This has been changed, ready to be resolved @sampocs

x/stakeibc/keeper/msg_server_instant_redeem_stake_test.go
riley-stride: @sampocs this is addressed. ready to resolve

x/stakeibc/keeper/msg_server_instant_redeem_stake_test.go
riley-stride: @sampocs this is addressed. ready to resolve

utils/utils.go
```diff
@@ -29,6 +30,15 @@ func FilterDepositRecords(arr []recordstypes.DepositRecord, condition func(recor
 	return ret
 }
 
+func SumDepositRecords(arr []recordstypes.DepositRecord) sdkmath.Int {
```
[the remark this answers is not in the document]
khdegraaf: done.

x/stakeibc/keeper/msg_server_instant_redeem_stake_test.go
khdegraaf: All cases are in there now.

x/stakeibc/keeper/msg_server_instant_redeem_stake.go
khdegraaf: This has been implemented, although not tested specifically for non-zero values.  Let me know if you still have any issues here.

PR #596, opened 2023-01-27
Planetary Governance

AustinoBombino:
This PR introduces the x/liquidgov module responsible for facilitating token lockups, host zone proposal querying, and host zone liquid voting.

## Author's Checklist

I have...

## Reviewers Checklist

I have...

  - [X] Does this pull request introduce a new feature or user-facing behavior changes? 
      - [ ] not applicable
      - [ ] jira ticket `XXX` 
      - [ ] README.md 
      - [ ] not documented


app/app.go
```diff
@@ -284,6 +290,8 @@ type StrideApp struct {
 	ScopedIcacallbacksKeeper capabilitykeeper.ScopedKeeper
 	IcacallbacksKeeper       icacallbacksmodulekeeper.Keeper
 	ClaimKeeper              claimkeeper.Keeper
+
```
shellvish, line 293: nit: we can probably remove this line

x/liquidgov/client/cli/query.go
```diff
@@ -0,0 +1,32 @@
+package cli
+
+import (
+	"fmt"
+	// "strings"
+
+	"github.com/spf13/cobra"
+
+	"github.com/cosmos/cosmos-sdk/client"
+	// "github.com/cosmos/cosmos-sdk/client/flags"
```
shellvish, line 10: just curious, why comment these out?

x/liquidgov/keeper/hooks.go
```diff
@@ -0,0 +1,131 @@
...
+// epochs hooks
+func (h Hooks) BeforeEpochStart(ctx sdk.Context, epochInfo epochstypes.EpochInfo) {
+	h.k.BeforeEpochStart(ctx, epochInfo)
+}
+
+func (h Hooks) AfterEpochEnd(ctx sdk.Context, epochInfo epochstypes.EpochInfo) {
+	h.k.AfterEpochEnd(ctx, epochInfo)
+}
+
+// // Update the epoch information in the stakeibc epoch tracker
```
shellvish, line 48: We can likely remove this I think, or use the EpochTracker from StakeIBC

Minor question, do you think it's better to use new epoch hooks in `x/liquidgov`, or call `x/liquidgov` functions from `x/stakeibc`? I don't have a strong view either way, curious what you think. I generally think it's slightly cleaner to have it in its own module (like you have here), but I think this is the first time we have epochly logic outside of `x/stakeibc`

x/liquidgov/keeper/hooks.go
```diff
@@ -0,0 +1,131 @@
...
+// 				// tally votes
+// 				vote := k.TallyVotes(ctx, proposal)
+
+// 				// cast vote on host
+// 				k.CastVoteOnHost(ctx, hostZone, vote)	// submits vote ICA on host
+// 			}
+// 		}
+// 	}
+// }
+// function may also want to prune old proposals
```
shellvish, line 107: I have a slight bias to move pruning to a separate function, just to make things a bit easier to follow

x/liquidgov/keeper/icqcallbacks_mirror_proposals.go
```diff
@@ -0,0 +1,50 @@
...
+
+	// Confirm host exists
+	chainId := query.ChainId
+	_, found := k.stakeibcKeeper.GetHostZone(ctx, chainId)
+	if !found {
+		return sdkerrors.Wrapf(stakeibctypes.ErrHostZoneNotFound, "no registered zone for queried chain ID (%s)", chainId)
+	}
+
+	// get bytes of last entry in range query
+	lastPropBytes := sdk.InclusiveEndBytes(args)
```
shellvish: If we don't just pull the last entry, does this return a series of gov proposals?

x/liquidgov/keeper/msg_server_submit_tx.go
```diff
@@ -0,0 +1,50 @@
...
+
+	govtypes "github.com/cosmos/cosmos-sdk/x/gov/types"
+
+	"github.com/Stride-Labs/stride/v5/x/liquidgov/types"
+)
+
+func (k Keeper) MirrorProposals(ctx sdk.Context, hostZone stakeibctypes.HostZone) error {
+	k.Logger(ctx).Info(utils.LogWithHostZone(hostZone.ChainId, "Submitting ICQ for proposals to %s", hostZone.ChainId))
+
+	endTime := ctx.BlockTime().AddDate(1, 0, 0) // TODO veryify range query works, then change logic to compare to latest proposal on stride
```
shellvish: Naive question, what does endTime do? Is it querying all proposals that end before endTime, that end after endTime, or something else?

shellvish (commented):
Left some light comments, but this is coming along really nicely! I think after the ICQ is done, lockups should be relatively straightforward. ICA might give a bit of an issue, but definitely manageable overall. 

Really excited to see how this develops, and thanks a lot for the extremely readable code!

x/liquidgov/keeper/msg_server_submit_tx.go
AustinoBombino: This is outdated. I was attempting to pull a range based on the endTime but what it actually does is get an entry in the ActiveProposalQueue which only stores proposal id's. The proposal iterator uses it to get the end of the queue based on an endTime.

x/liquidgov/keeper/icqcallbacks_mirror_proposals.go
AustinoBombino: This is also outdated. It wouldn't pull a range, the actual query that was being sent out would just return a proposal id in the ActiveProposalQueue.

x/liquidgov/client/cli/query.go
AustinoBombino: I think ignite scaffolding generated them commented.

x/liquidgov/keeper/proposal.go
```diff
@@ -60,7 +60,7 @@ func (keeper Keeper) DeleteProposal(ctx sdk.Context, proposalID uint64, chain_id
 	store.Delete(types.ProposalKey(proposalID, chain_id))
 }
 
-// GetProposalID gets the highest proposal ID
+// GetProposalID gets the highest ICQ'd proposal ID
```
asalzmann, line 63: I think there could be multiple concurrent proposals - might want to rename to `GetHighestProposalID`

x/liquidgov/keeper/icqcallbacks_mirror_proposals.go
```diff
@@ -27,27 +25,29 @@ func MirrorProposalsCallback(k Keeper, ctx sdk.Context, args []byte, query icqty
 
 	highestID, _ := k.GetProposalID(ctx, chainId)
 
-	var proposals govtypesv1.Proposals
-	codec.NewLegacyAmino().UnmarshalJSON(args, &proposals)
-
-	// if err != nil {
-	// 	return sdkerrors.Wrapf(stakeibctypes.ErrMarshalFailure, "unable to unmarshal query response into Proposal type, err: %s", err.Error())
-	// }
+	queriedProposal := govtypesv1beta1.Proposal{}
```
asalzmann, line 28: We may have to branch this by SDK version on the host chain (if the gov module was update between SDK 45/46)

x/liquidgov/keeper/msg_server_submit_tx.go
```diff
@@ -2,33 +2,30 @@ package keeper
...
 
 func (k Keeper) MirrorProposals(ctx sdk.Context, hostZone stakeibctypes.HostZone) error {
 	k.Logger(ctx).Info(utils.LogWithHostZone(hostZone.ChainId, "Submitting ICQ for proposals to %s", hostZone.ChainId))
 
-	// query custom proposals path
-	path := strings.Join([]string{"custom", govtypes.QuerierRoute, v1beta1.QueryProposals}, "/")
-	// query data 1 page, default limit, voting period, no owner, no depositor
-	queryData := codec.NewLegacyAmino().MustMarshalJSON(v1.NewQueryProposalsParams(1, 0, v1.StatusVotingPeriod, nil, nil))
+	// get highest proposal ID on stride
+	highestID, _ := k.GetProposalID(ctx, hostZone.ChainId)
```
asalzmann, line 25: What would happen if multiple proposals were created within a single epoch?

asalzmann (commented):
Took a look at `b5c94905dc5e2d688dae424b5bdea10b32c6c5b4` - looking good! One design decision to consider (shouldn't refactor now IMO but it is worth noting the tradeoffs), should we mirror proposals in hooks or should we expose a user-facing function that can issue an ICQ, mirroring a proposal? I could see some pros/cons for each approach.

Unrelated question, do these ICQs take into account deposits when querying proposals? If not, should they?

x/liquidgov/keeper/icqcallbacks_mirror_proposals.go
jstr1121: Not just host chain version, how to handle custom proposals of custom module on the zone? (Could we decode this and show this? Or should we just fetch proposal ID and the content is shown on frontend side?)

riley-stride: Are there many custom proposal types across cosmos? Imo we should start with the basic code gov proposal types, then we can add the others in a future ugrade

riley-stride:
open

PR #611, opened 2023-02-08
Auction module

hackartist:
Types refactored for multiple types of auction with different settings and logic.  All three auction types can be run and tested now with the default pools created in genesis.

A SealedBidAuction has two phases: first, bidders submit bids which are hashed with a secret salt only they know.  No one can tell what the bid actually is. Second, bidders reveal their original bid and the salt so the protocol can hash and confirm this was the hash received in the first phase.

How to test:
Locally create bid JSON as an env var.  Orders object has list of Price, Volume pairs
`bid_data='{"Orders":[{"Price":102000,"Volume":4000000000},{"Price":93000,"Volume":3000000000]}'`
Create a salt value which can be anything you want, keep it secret until the reveal phase of the auction
`salt_data='seasalt'`
Create a sha256 of the bid_data and salt_data appended.  The -n is important to avoid unwanted new lines
`hash_data=$(echo -n "$bid_data$salt_data" | sha256sum | cut -c 1-64)`

With Stride, Gaia, and relayers running, in another shell window run
`tail -f ./dockernet/logs/stride.log | grep --line-buffered "\[auction\]"`
This will stream the stride log and filter for only events coming from the auction

Start an auction with
`strided tx auction start-auction cosmoshub-4 0 --from stride1uk4ze...`
During the first phase, submit a sealed bid
`strided tx auction submit-sealed-bid cosmoshub-4 0 $hash_data --from stride1uk4ze...`
When you see the state change to REVEAL in the logs, reveal the bid with
`strided tx auction reveal-bid cosmoshub-4 0 "$bid_data" $salt_data --from stride1uk4ze...`
Shortly after the state should change to Payout and the auction will resolve printing out who would be paid how much.  You can do this whole process in several different shells with different data and from addresses to see a more complicated resolution.

image

In general the funds taken in and send out are in different tokens like stATOM in and ATOM out.
You can see the underlying data structure which runs the auction and properties by running
`strided q auction list-auction-pool`

Also for the purposes of testing, there are two other auction pools with ID 1 to run AscendingAuction and ID 2 to run DescendingAuction 

Start an ascending auction with
`strided tx auction start-auction cosmoshub-4 1 --from stride1uk4ze...`
While running, submit a bid
`strided tx auction submit-open-bid cosmoshub-4 1 "$bid_data" --from stride1uk4ze...`
When the auction reaches its time limit with no bids coming in recently the auction will end.  If you submit bids near to the end of the auction you will see the ending time get extended with this info printed in the logs.

Start a descending auction with
`strided tx auction start-auction cosmoshub-4 2 --from stride1uk4ze...`
When the auction starts, the logs will print out the current bid which will move down every few blocks.  This bid is the price in the bid object, when you send in bids all that matters here is the volume. Submit a bid
`strided tx auction submit-open-bid cosmoshub-4 2 "$bid_data" --from stride1uk4ze...`
You should see in the logs instant "payout" the volume requested.  This will either end the auction if your requested volume covered the supply or it will continue if there is supply remaining.  The auction ends when supply is gone, or the bid gets below the minimum bid setting.

Next steps: 
- Include important parameters like which zoneId ==> which coins this auction is for, should be the first prefix in the KVStore
- Pass in parameters to whole module via stakeibc for example instead of fixed hard coded values
- Ability for reserve auctions to run by having more than one pool with different parameter settings 
- Add collateral to each bid and refund if they reveal correctly to avoid many price probing bids


proto/stride/auction/auction_pool.proto
```diff
@@ -0,0 +1,104 @@
...
+// even if the supply in the address of the pool changes during that time
+// Depending on the algorithm chosen there are different specific params
+
+// English style auction specific parameters
+// Starts low with increasing bids, auction ends if no new bids occur 
+// for endDuration blocks in a row and goes to current winner
+// When a higher bid comes in, the lastBlock is set now + endDuration
+message AscendingAuction {
+  AuctionState status = 1;
+  AscendingAuctionProperties auctionProperties = 2;
```
antstalepresh, line 56: The common practice for defining message fields is to use snake_case, not camelCase.
Please convert all camelCase fields to snake_case.

proto/stride/auction/auction_pool.proto
```diff
@@ -0,0 +1,104 @@
...
+  optional DescendingAuction descendingAuction = 3;
+  optional SealedBidAuction sealedBidAuction = 4;
+}
+
+// Auction Pool has an address with some assets and restrictions on the types
+// of auctions which can be run for this pool.
+message AuctionPool {
+  uint64 id = 1;
+  AuctionPoolProperties poolProperties = 2;
+  Auction latestAuction = 8;
```
antstalepresh, line 103:
```suggestion
  Auction latestAuction = 3;
```

proto/stride/auction/bid.proto
```diff
@@ -0,0 +1,45 @@
...
+/* ****************************************************************************
+An Order is a pair of price and volume
+A Bid is one or more orders
+
+A SealedBid is who sent it and the hash of the bid + secretSalt
+An OpenBid is who sent it with the Bid data structure
+**************************************************************************** */
+
+message Order {
+    uint64 price = 1 [
```
antstalepresh, line 20: The type of price field must be always integer? I think it should be decimal type more generally.
What do you think?

proto/stride/auction/properties.proto
```diff
@@ -0,0 +1,69 @@
...
+    ASCENDING = 1;
+    DESCENDING = 2;
+    SEALEDBID = 3;
+}
+
+message AscendingAuctionProperties {
+    uint64 supply = 1;
+    uint64 minAllowedBid = 2;
+    uint64 maxAllowedBid = 3;
+    float redemptionRate = 4;
```
antstalepresh, line 34: Please don't use float for proto messages since float is not safety for managing decimals.
Instead, please use sdk decimal type.
example,
string redemption_rate = 4 [
    (gogoproto.nullable) = false
]

app/app.go
```diff
@@ -212,6 +217,7 @@ var (
 		claimtypes.ModuleName:           nil,
 		interchainquerytypes.ModuleName: nil,
 		icatypes.ModuleName:             nil,
+		auctionmoduletypes.ModuleName:   {authtypes.Minter, authtypes.Burner, authtypes.Staking},
```
antstalepresh, line 220: Why do we need mint, burn, staking perms for auction module account?

hackartist: We don't, good catch. This was an automatic update which the tool "ignite" made when I said I wanted the auction keeper to contain a bank keeper but it wasn't anything I was trying to add on purpose. We only need the ability to send tokens and receive tokens

proto/stride/auction/bid.proto
hackartist: This was something I was unsure about because different denom have different subunits.  I asked about this and was told most are 10^6 so 1ATOM is represented as 1000000 but some might be 10^12 or 10^18 subunits just depending on the denom.  If you think the sdk DEC is a better way to represent this or if that is what we do in other parts of the code base I'm happy to make this change.

proto/stride/auction/auction_pool.proto
hackartist: So won't this make the type private to the module?  Or will the proto generation step still generate it with different capitalization so that it can be used across different packages?

proto/stride/auction/query.proto
```diff
@@ -0,0 +1,54 @@
...
+// Query defines the gRPC querier service.
+service Query {
+
+  // Parameters queries the parameters of the module.
+  rpc Params(QueryParamsRequest) returns (QueryParamsResponse) {
+    option (google.api.http).get = "/stride/v5/auction/params";
+  }
+
+  // Queries a list of AuctionPool items.
+  rpc AuctionPool(QueryGetAuctionPoolRequest)
```
antstalepresh, line 22:
```suggestion
  rpc AuctionPool(QueryAuctionPoolRequest)
```
Common practice for defining query req & resp msgs are Query[rpc name]Request/Response

proto/stride/auction/params.proto
```diff
@@ -0,0 +1,17 @@
...
+
+package stride.auction;
+
+import "gogoproto/gogo.proto";
+
+option go_package = "github.com/Stride-Labs/stride/v5/x/auction/types";
+
+// Params defines the defaults for the whole auction module.
+message Params {
+    uint64 defaultSealedAuctionDuration = 1;
```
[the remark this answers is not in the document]
hackartist: So it would be good for a gov proposal to be able to change this.  An actual auction only lasts a few minutes when kicked off but once started its duration is fixed, this value is a module wide parameter for what the length of the next auction will be.  It would be fine if a gov proposal voted auctions to last shorter or longer going forward and thus it would be fine for this value to get modified while the chain is running.  I'll continue to think about it and research keys.go

testutil/keeper/auction.go
```diff
@@ -0,0 +1,53 @@
...
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	typesparams "github.com/cosmos/cosmos-sdk/x/params/types"
+	"github.com/stretchr/testify/require"
+	"github.com/tendermint/tendermint/libs/log"
+	tmproto "github.com/tendermint/tendermint/proto/tendermint/types"
+	tmdb "github.com/tendermint/tm-db"
+)
+
+func AuctionKeeper(t testing.TB) (*keeper.Keeper, sdk.Context) {
+	storeKey := sdk.NewKVStoreKey(types.StoreKey)
```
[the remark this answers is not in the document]
hackartist: Yes, good catch, this is another one of the changes ignite had made automatically but I will simplify to match the style of those other two testutil keeper methods

proto/stride/auction/params.proto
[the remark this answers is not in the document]
antstalepresh: Gotcha. If these are going to changed by gov proposal, my only suggestion is to remove `default` prefix from these fields.

proto/stride/auction/bid.proto
antstalepresh: I see. Lets keep it as it is now.

x/auction/keeper/msg_server_reveal_bid.go
```diff
@@ -0,0 +1,77 @@
...
+// If their sealed bid hash matches the hash we generate then nothing has changed and their revealed bid is entered
+
+func (k msgServer) RevealBid(goCtx context.Context, msg *types.MsgRevealBid) (*types.MsgRevealBidResponse, error) {
+	ctx := sdk.UnwrapSDKContext(goCtx)
+
+	ctx.Logger().Info(fmt.Sprintf("[auction]"))
+	ctx.Logger().Info(fmt.Sprintf("[auction] Reveal request for %+v", msg))
+
+	auctionPools := k.Keeper.GetAllAuctionPool(ctx)
+	for _, pool := range auctionPools {
```
antstalepresh, line 25: We should remove this for loop and use GetAuctionPool function instead to get auction pool by id.

x/auction/keeper/msg_server_start_auction.go
```diff
@@ -0,0 +1,35 @@
...
+// Get the name of the target pool, if the end-block is already before the current blockchain height, then we can start a new auction
+
+func (k msgServer) StartAuction(goCtx context.Context, msg *types.MsgStartAuction) (*types.MsgStartAuctionResponse, error) {
+	ctx := sdk.UnwrapSDKContext(goCtx)
+
+	now := cast.ToUint64(ctx.BlockHeight())
+	ctx.Logger().Info(fmt.Sprintf("[auction] Request to start auction for pool %d at block %d", msg.GetPoolID(), now))
+
+	auctionPools := k.Keeper.GetAllAuctionPool(ctx)
+	for _, pool := range auctionPools {
```
antstalepresh, line 22: We should remove this for loop and use GetAuctionPool function instead to get auction pool by id.

x/auction/keeper/msg_server_submit_open_bid.go
```diff
@@ -0,0 +1,77 @@
...
+func (k msgServer) SubmitOpenBid(goCtx context.Context, msg *types.MsgSubmitOpenBid) (*types.MsgSubmitOpenBidResponse, error) {
+	ctx := sdk.UnwrapSDKContext(goCtx)
+
+	now := cast.ToUint64(ctx.BlockHeight())
+
+	ctx.Logger().Info(fmt.Sprintf("[auction]"))
+	ctx.Logger().Info(fmt.Sprintf("[auction] Open bid incoming %s at block %d", msg.GetBid(), now))
+
+	auctionPools := k.Keeper.GetAllAuctionPool(ctx)
+	for _, pool := range auctionPools {
```
antstalepresh, line 26: We should remove this for loop and use GetAuctionPool function instead to get auction pool by id.

x/auction/keeper/msg_server_submit_sealed_bid.go
```diff
@@ -0,0 +1,92 @@
...
+func (k msgServer) SubmitSealedBid(goCtx context.Context, msg *types.MsgSubmitSealedBid) (*types.MsgSubmitSealedBidResponse, error) {
+	ctx := sdk.UnwrapSDKContext(goCtx)
+
+	now := cast.ToUint64(ctx.BlockHeight())
+
+	ctx.Logger().Info(fmt.Sprintf("[auction]"))
+	ctx.Logger().Info(fmt.Sprintf("[auction] Sealed bid incoming %s at block %d", msg.GetHashedBid(), now))
+
+	auctionPools := k.Keeper.GetAllAuctionPool(ctx)
+	for _, pool := range auctionPools {
```
antstalepresh, line 27: We should remove this for loop and use GetAuctionPool function instead to get auction pool by id.

x/auction/keeper/msg_server_reveal_bid.go
```diff
@@ -0,0 +1,77 @@
...
+				if sealedAuction != nil && sealedAuction.GetStatus() == types.AuctionState_REVEAL {
+
+					creator := msg.GetCreator()
+					sealedBids := sealedAuction.GetSealedBids()
+					ctx.Logger().Info(fmt.Sprintf("[auction] Here are the sealedAuction %+v", pool.GetLatestAuction().GetSealedBidAuction()))
+					ctx.Logger().Info(fmt.Sprintf("[auction] Here are the sealedbids map %+v", sealedBids))
+
+					inSealedBids := false
+					sealedValue := ""
+					for _, sealedBid := range sealedBids {
```
antstalepresh, line 39: We need to avoid using for loop in development as much as possible.
I suggest to add id field in auction type and save a map info into kv store.
The format of the map key can be bytes(pool id+auction id+bidder address) and the value can be auction type string.

x/auction/keeper/msg_server_reveal_bid.go
```diff
@@ -0,0 +1,77 @@
...
+
+					if inSealedBids {
+						//ctx.Logger().Info(fmt.Sprintf("[auction] Before the hash Bid is %s", msg.GetBid()+msg.GetSalt()))
+						hash := hex.EncodeToString(crypto.Sha256([]byte(msg.GetBid() + msg.GetSalt())))
+
+						if sealedValue == hash {
+							ctx.Logger().Info(fmt.Sprintf("[auction] Hash Match! On hash %s", hash))
+
+							var revealedBid types.Bid
+							json.Unmarshal([]byte(msg.GetBid()), &revealedBid)
```
antstalepresh, line 54: Please fix all lint issues.
I recommend you to install golangci-lint package on your machine and check lint issues before pushing updates.
```golangci-lint run ./x/auction```

File: x/auction/keeper/msg_server_reveal_bid.go [hunk 15]
```diff
@@ -0,0 +1,77 @@
...
+							inSealedBids = true
+							sealedValue = sealedBid.GetHashedBid()
+						}
+					}
+
+					if inSealedBids {
+						//ctx.Logger().Info(fmt.Sprintf("[auction] Before the hash Bid is %s", msg.GetBid()+msg.GetSalt()))
+						hash := hex.EncodeToString(crypto.Sha256([]byte(msg.GetBid() + msg.GetSalt())))
+
+						if sealedValue == hash {
```
antstalepresh, line 50: In sdk development, its not good practice to use if statement in if statement repeatedly.
Recommended change;
```
if !inSealedBids {
   return;
}

if sealedValue != hash {
   return;
}

...
```

x/auction/types/message_start_auction.go
```diff
@@ -0,0 +1,47 @@
...
+	}
+	return []sdk.AccAddress{creator}
+}
+
+func (msg *MsgStartAuction) GetSignBytes() []byte {
+	bz := ModuleCdc.MustMarshalJSON(msg)
+	return sdk.MustSortJSON(bz)
+}
+
+func (msg *MsgStartAuction) ValidateBasic() error {
```
antstalepresh, line 41: We should also check if Zone is valid.
Please check ValidateBasic functions for all messages and see if there are any other missing validation checks.

x/auction/types/genesis_test.go
```diff
@@ -0,0 +1,75 @@
...
+            desc:     "default is valid",
+            genState: types.DefaultGenesis(),
+            valid:    true,
+        },
+        {
+            desc:     "valid genesis state",
+            genState: &types.GenesisState{
+            	
+                AuctionPoolList: []types.AuctionPool{
+	{
```
antstalepresh, line 26: This looks a bit messy. Can you please clean up this file?
Not only this one but please also check all other files you have written again and clean this kind of styling issues.

antstalepresh:
Overall looking great!
I left some comments for performance improvement.
A few more things I want to suggest are,
1. Try to leave comments as many as possible to make codebases more readable.
2. Put empty lines in appropriate places for better looking.
3. Use go linter package to fix lint issues before pushing updates.
4. Try to split up codes into several PRs in the future. (large PRs make us difficult to review in detail)

x/auction/keeper/msg_server_reveal_bid.go
shellvish: +1, the main reason for this is efficiency. By introducing large loops like this, if we ever had lots of bids submitting for an auction, the gas cost to reveal a bid would become quite high. A Stride "block" only has so much gas, so this also introduces a maximum to the number of bids that could be submitted. A lookup up table is nice and constant time!

x/auction/keeper/msg_server_reveal_bid.go
shellvish: +1, would be much easier to read as well

x/auction/keeper/msg_server_reveal_bid.go
shellvish: VSCode should do this automatically, let me know if you're having trouble setting it up Ethan!

x/auction/keeper/msg_server_submit_sealed_bid.go
```diff
@@ -0,0 +1,92 @@
...
+						poolAddress, _ := sdk.AccAddressFromBech32(pool.GetPoolProperties().PoolAddress)
+
+						if k.bankKeeper.BlockedAddr(poolAddress) {
+							ctx.Logger().Info(fmt.Sprintf("[auction] pool address is blocked!"))
+						}
+						if k.bankKeeper.BlockedAddr(bidderAddress) {
+							ctx.Logger().Info(fmt.Sprintf("[auction] bidder address is blocked!"))
+						}
+
+						ctx.Logger().Info(fmt.Sprintf("[auction] About to send collateral %+v %+v %+v", coin, bidderAddress, poolAddress))
```
shellvish, line 68: nit: "Sending collateral %+v..."

x/auction/keeper/msg_server_submit_sealed_bid.go
```diff
@@ -0,0 +1,92 @@
...
+						collateralErr = k.bankKeeper.SendCoins(ctx, bidderAddress, poolAddress, coins)
```
shellvish, line 69: I would change this structure to (pseudocode) ```
if collateralErr, print Err and return```
e.g. a general flow of 
"Assume the success case, handle all failure cases immediately, then assume the next success case, etc."

shellvish (commented):
This is looking really great! I left some comments, would love to take a second look after you address these. 

On the loops, to give a bit more intuition, you build more security into the chain when we know operations will take a deterministic amount of resources. Otherwise, the chain could get into a state where the looped operation gets too expensive to run feasibly (sometimes it might exceed the gas in a single block, which would make the operation impossible). Storing maps with pointers ends up becoming much more predictable. 

After this round of comments, I think most of what we're missing is just testing. This is getting close!

asalzmann:
let's keep this open!

Issue #434, opened 2022-12-02
Redemption rate safety bound invariants by host zone


riley-stride:
## Background

We currently upper and lower bound all redemption rates by fixed, hardcoded constants [0.9,1.5]. We should instead vary these bounds by host zones. We also should put these bounds up to governance as parameters.

In this way
- zones with higher inflation rates have wider bounds, while those with lower inflation have tighter bounds 
- a simple param change gov vote can update the bounds, rather than requiring a software upgrade

## Suggested Design

1. Add two new maps in stakeibc params: 
- maxrrs = {chainidA: 1.5, chainidB: 2.0,…}
- minrrs = {chainidA: 0.6, chainidB: 0.8,…}
2. Use the proper hostzone's bounds (indexed from these maps) everywhere the current bounds are used

## Acceptance Criteria
- all existing and new tests should pass
- write two new unit tests, they should pass:
      - the redemption rate on a host zone is more than the upper bound and halts the chain
      - the redemption rate on a host zone is less than the upper bound and halts the chain
- manually verify we can change the params with a gov vote
- make sure we initialize these params in the software upgrade

PR #614, opened 2023-02-10
Safety bound per hostzone

jstr1121:
Closes: #434

- Add two new maps in stakeibc params
- Use the proper hostzone's bounds (indexed from these maps) everywhere the current bounds are used

## Author's Checklist

I have...

## Reviewers Checklist

I have...

      - [ ] not applicable
      - [ ] jira ticket `XXX` 
      - [ ] README.md 
      - [ ] not documented


x/stakeibc/types/params.go
```diff
@@ -242,8 +255,103 @@ func isCommission(i interface{}) error {
...
+		if !rate.IsPositive() {
+			return fmt.Errorf("redemption rate should be positive")
+		}
+	}
+	return nil
+}
+
 // Validate validates the set of params
 func (p Params) Validate() error {
+	if err := isPositive(p.DepositInterval); err != nil {
```
riley-stride, line 297: Great catch, thank you for adding these in @jstr1121!

x/stakeibc/keeper/keeper_test.go
```diff
@@ -44,3 +45,28 @@ func (s *KeeperTestSuite) GetMsgServer() types.MsgServer {
...
+	rrSafe, err := s.App.StakeibcKeeper.IsRedemptionRateWithinSafetyBounds(s.Ctx, zone)
+	s.Require().NoError(err)
+	s.Require().True(rrSafe)
+
+	zone.RedemptionRate = sdk.NewDecWithPrec(16, 1) // 1.6
+	rrSafe, err = s.App.StakeibcKeeper.IsRedemptionRateWithinSafetyBounds(s.Ctx, zone)
+	s.Require().NoError(err)
+	s.Require().True(rrSafe)
+
+	zone.RedemptionRate = sdk.NewDecWithPrec(21, 1) // 2.1
```
riley-stride, line 68: Could you add a test for a zone that does NOT have a custom bound, to verify that in that case it uses the default bounds?

riley-stride (changes requested):
This looks good @jstr1121! Just requesting a few additional test cases.

x/stakeibc/keeper/keeper_test.go
sampocs: Can you also add another host zone to the params and add some tests across both hosts? 
E.g. Osmo 0.3-2, Gaia 1.5-2.5
Test cases:
 0.1 - out of both
0.3 - in for osmo, out for gaia
1.5 - in for both
2.2 - out for osmo, in for gaia

sampocs (commented):
Nice work Jun! Only comment is on adding some more tests cases

proto/stride/stakeibc/params.proto
```diff
@@ -34,4 +34,6 @@ message Params {
   uint64 ibc_transfer_timeout_nanos = 16;
   uint64 safety_num_validators = 17;
   uint64 safety_max_slash_percent = 18;
+  map<string, string> max_redemption_rates = 19; // maxrrs = {chainidA: "1.5", chainidB: "2.0",…}
```
sampocs, line 37: Could using a map here potentially lead to a non-determinism error?

jstr1121: What sort of non-determinism error? When it's not set on map, it's using default value.

x/stakeibc/keeper/keeper_test.go
jstr1121: Okay got it.

x/stakeibc/types/params.go
jstr1121: I think on other modules these are missing - will need to add those on a new PR - since it's for stakeibc related PR.

x/stakeibc/keeper/keeper_test.go
jstr1121: https://github.com/Stride-Labs/stride/pull/614/commits/85508ca1f81e4d93780cf6a82585e5f0a84d3c1c

proto/stride/stakeibc/params.proto
sampocs: I believe just storing a map can cause a consensus error because maps are non-determinstic in go - meaning different validators could attempt to store the same map, but the serialized representation of it could be slightly differently. 
You can get around this by just creating another prefix store that acts as a map

x/stakeibc/keeper/keeper_test.go
sampocs: beautiful!

proto/stride/stakeibc/params.proto
riley-stride: @jstr1121 following up on this, we should use a prefix store that acts as a map. Here's an example https://github.com/Stride-Labs/stride/pull/556/commits/aa0af4927db96778d7565210a1a12f2946f900d3

riley-stride: We'd like to get this PR into the v7 happening next week. Do you think we can make that happen?

jstr1121: I think we wanted to put this in Params - otherwise, we will need to have a separate proposal to change these values.

sampocs (commented):
Lgtm! Only thing is we might want rename `safety_min_redemption_rate_threshold` and `safety_max_redemption_rate_threshold` to `default_min_redemption_rate_threshold` and `default_max_redemption_rate_threshold`

riley-stride:
> Lgtm! Only thing is we might want rename `safety_min_redemption_rate_threshold` and `safety_max_redemption_rate_threshold` to `default_min_redemption_rate_threshold` and `default_max_redemption_rate_threshold`

Agreed, that'd be much more intuitive.

Otherwise LGTM as well!

jstr1121:
> Lgtm! Only thing is we might want rename `safety_min_redemption_rate_threshold` and `safety_max_redemption_rate_threshold` to `default_min_redemption_rate_threshold` and `default_max_redemption_rate_threshold`

Made change here
https://github.com/Stride-Labs/stride/pull/614/commits/1e216da16b3a72696f9bf3b2a6d50e75798915f7

riley-stride (approved):
lgtm

Issue #622, opened 2023-02-21
Lack of check exist of acctAddress in mint module


ThanhNhann:
## Background

`SetupNewModuleAccount` function doesn't have a check if the `acctAddress` is exist
https://github.com/Stride-Labs/stride/blob/94e3a91dc582218a477771d9a694966bf901fdf5/x/mint/keeper/keeper.go#L233-L242
so in some cases, we can face some problems with this, I have run simulation for this and face it
## Suggested Design

- Add a check if the `acctAddress` is exist

## Acceptance Criteria

- Check for existing of `acctAddress` was added
- All test pass

riley-stride:
Can you elaborate on 
```
so in some cases, we can face some problems with this, I have run simulation for this and face it
```
I don't understand what will break in this case?

Also, if you are running simulation tests, could you share your code for that? We'd like to start using them.

ThanhNhann:
Sure @riley-stride, I coding simulation test for Strid, and will update information of fail soon in the pr of this issue

PR #623, opened 2023-02-21
check module account is exist

ThanhNhann:
Closes: #622

## Author's Checklist

I have...

## Reviewers Checklist

I have...

      - [ ] not applicable
      - [ ] jira ticket `XXX` 
      - [ ] README.md 
      - [ ] not documented

### DIFF
--- a/x/mint/keeper/keeper.go
+++ b/x/mint/keeper/keeper.go
@@ -233,8 +233,8 @@ func (k Keeper) DistributeMintedCoin(ctx sdk.Context, mintedCoin sdk.Coin) error
 func (k Keeper) SetupNewModuleAccount(ctx sdk.Context, submoduleName string, submoduleNamespace string) {
 	// create and save the module account to the account keeper
 	acctAddress := k.GetSubmoduleAddress(submoduleName, submoduleNamespace)
-	acctAddressIsExist := k.accountKeeper.GetAccount(ctx, acctAddress) 
-	if acctAddressIsExist != nil {
+	checkAcctAddress := k.accountKeeper.GetAccount(ctx, acctAddress) 
+	if checkAcctAddress != nil {
 		return
 	}
 	acc := k.accountKeeper.NewAccount(
--- a/x/mint/keeper/keeper.go
+++ b/x/mint/keeper/keeper.go
@@ -233,6 +233,10 @@ func (k Keeper) DistributeMintedCoin(ctx sdk.Context, mintedCoin sdk.Coin) error
 func (k Keeper) SetupNewModuleAccount(ctx sdk.Context, submoduleName string, submoduleNamespace string) {
 	// create and save the module account to the account keeper
 	acctAddress := k.GetSubmoduleAddress(submoduleName, submoduleNamespace)
+	acctAddressIsExist := k.accountKeeper.GetAccount(ctx, acctAddress) 
+	if acctAddressIsExist != nil {
+		return
+	}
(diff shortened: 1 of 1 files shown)


sampocs:
Can you give some more details on the error that you saw in the simulation tests? From briefly eyeballing it, it looks like this is only called when the chain is started up for the first time, and I think the address' are also deterministic so setting an account that already exists would just override it?

Curious to get @riley-stride's opinion as he's most familiar with this code

ThanhNhann:
Yeah, when I code for simulation test in the case of `TestAppImportExport`, it will always fail because you have to start a new app, and I see that, the error is because we had checked the existing module account. Code for simulation test is below:
https://github.com/notional-labs/stride/blob/716a2d449374f56c7544174e923b75a5c390a836/app/sim_test.go#L83

riley-stride:
> when I code for simulation test in the case of `TestAppImportExport`, it will always fail because you have to start a new app, and I see that, the error is because we don't have a check the existing module account so that it fails for the test after. This test simulates an upgrade chain work with export genesis from the old app and import to the new app. The code for simulation test with update checking is below: https://github.com/notional-labs/stride/blob/716a2d449374f56c7544174e923b75a5c390a836/app/sim_test.go#L83

Can you point us to the line that breaks in the sim tests? When do we create a module account twice?

ThanhNhann:
Hi @riley-stride, full code for this fail test is in this function https://github.com/notional-labs/stride/blob/cd4d171f8492d42956a933c2e679eaa1bea1e195/app/sim_test.go#L84
Steps to run:
  - `make test-sim-import-export` and will show the fail like image below 
  Screenshot 2023-02-22 at 12 36 57
  - Run the line below failed notification in the image, in this case is: `go test ./app -run TestAppImportExport -Enabled=true -NumBlocks=10 -Genesis= -Verbose=true -Commit=true -Seed=7 -Period=2 -ExportParamsPath /var/folders/5y/7n2nyf8x5799th2r_zmw58x00000gn/T/sim-logs-1372263546/sim_params-7.json -ExportStatePath /var/folders/5y/7n2nyf8x5799th2r_zmw58x00000gn/T/sim-logs-1372263546/sim_state-7.json -v -timeout 24h`  and it's will show the error like image below 
  
Screenshot 2023-02-22 at 12 41 07

Insight: 
- The image above show that their are some key/value pairs different with `app` and  `newapp`, 
- I debuged and record all the moduleAccount address in https://github.com/notional-labs/stride/blob/nhan%2Ffail_simualtion_test/app/test.json for the app and https://github.com/notional-labs/stride/blob/nhan%2Ffail_simualtion_test/app/test2.json for the newapp, and see that three accounts seted up by https://github.com/notional-labs/stride/blob/cd4d171f8492d42956a933c2e679eaa1bea1e195/x/mint/genesis.go#L11 will be make a fail. 
- In the file `test.json`, we can check that there are total `925` accounts were created with the max id is `924` and the id for three accounts above are `922,923,924`, after that I also check for file `test2.json` and see that, there are also total `925` accounts were created but with max id is `927` and the id for three accounts above is `925,926,927` and don't have id `922,923,924`, and you know what, the address of accounts `925,926,927` in `test2.json` are the same with accounts `922,923,924` in `test.json`, in this case, I think it mean when we run new app with genesis exported from old app, the information of three accounts above will be added more, not update like another accounts because we don't have a check exist for them. So I added it and run ok now, if you have any questions, feel free to ask, thank!

PR #630, opened 2023-02-24
Removed duplicate check for min/max validators in rebalance

sampocs:
Closes: #419

The min/max number of validators during rebalancing was checked twice - once in validate basic, and again in the actual message. There was also a different number used in each. 
Per the audit report, the fix was to remove the duplicate check (just keep it in validate basic) and add constants for the min/max values.

## Brief Changelog
* Added constants for the min/max number of validators to rebalance
* Removed duplicate check from msg_server
* Adjusted unit tests
* Added helper function to get admin address for validate basic unit tests

## Author's Checklist

I have...

    - [x] Added test cases to existing unit tests

## Reviewers Checklist

I have...

      - [ ] not applicable
      - [ ] jira ticket `XXX` 
      - [ ] README.md 
      - [ ] not documented


app/apptesting/test_helpers.go
```diff
@@ -356,6 +357,14 @@ func GenerateTestAddrs() (string, string) {
 	return validAddr, invalidAddr
 }
 
+// Grabs an admin address to test validate basic on admin txs
+func GetAdminAddress() (address string, ok bool) {
+	for address := range utils.Admins {
+		return address, true
```
riley-stride, line 363: nit: is this just going to always return the first admin address? kinda confusing to do that in a loop; possible it's a golang idiom

riley-stride (approved):
lgtm

app/apptesting/test_helpers.go
sampocs: the admins list is a map haha (not sure why tbh) but this is the only way to grab the "first item" from a map as far as I know

sampocs: actually I guess is does make sense as a map cause we're generally starting from a supplied address and checking if it's an admin

PR #638, opened 2023-03-02
v7 upgrade handler

riley-stride:
Closes: #XXX

## Context and Purpose
* Added upgrade handler and unit tests for v7

## Brief Changelog
* Added rate limit and autopilot store keys
* Prop 8 (reward allocation)
* Prop 153 (incentive diversification)
* Added hour epoch (for rate limits)
* Increased stride inflation
* Added stride messages as ICA Host allow messages 
* Added default safety param attributes to host zones (min/max redemption rate, halted)
* Changed juno unbonding frequency to 5
* Created reward collector module account


app/upgrades/v7/upgrades.go
```diff
@@ -0,0 +1,183 @@
...
+	}
+
+	k.SetEpochInfo(ctx, hourEpoch)
+}
+
+// Increase stride inflation from 285 STRD to 1078 STRD
+func IncreaseStrideInflation(ctx sdk.Context, k mintkeeper.Keeper) {
+	ctx.Logger().Info("Increasing STRD inflation")
+
+	epochProvisions := sdk.NewDec(1_078_767_123_000_000)
```
shellvish, line 111: I think this should be 1_078_767_123, but please double check me on that!

sampocs: Good catch haha, you’re right!

app/upgrades/v7/upgrades.go
```diff
@@ -0,0 +1,201 @@
...
+	accountKeeper authkeeper.AccountKeeper,
+	bankKeeper bankkeeper.Keeper,
+	epochsKeeper epochskeeper.Keeper,
+	icahostKeeper icahostkeeper.Keeper,
+	mintKeeper mintkeeper.Keeper,
+	stakeibcKeeper stakeibckeeper.Keeper,
+) upgradetypes.UpgradeHandler {
+	return func(ctx sdk.Context, _ upgradetypes.Plan, vm module.VersionMap) (module.VersionMap, error) {
+		ctx.Logger().Info("Starting upgrade v7...")
+
```
asalzmann, line 60: TODO: we should disable autopilot (set `DefaultActive` to false in params)

asalzmann: resolved

shellvish (approved):
Looks good, we should also make sure to double-check all of these in localstride

I'm pretty confident this inflation number is correct, although would be good to get one more independent check that this is the right value 

Thanks for doing this, lots of nit details in this upgrade that are handled very cleanly!

app/upgrades/v7/README.md
```diff
@@ -0,0 +1,4 @@
+# Upgrade v7 Changelog
+1. XXX
```
asalzmann, line 2: @shellvish to do in a separate PR

app/upgrades/v7/upgrades.go
```diff
@@ -0,0 +1,199 @@
...
+	stakeibckeeper "github.com/Stride-Labs/stride/v6/x/stakeibc/keeper"
+	"github.com/Stride-Labs/stride/v6/x/stakeibc/types"
+	stakeibctypes "github.com/Stride-Labs/stride/v6/x/stakeibc/types"
+)
+
+// Note: ensure these values are properly set before running upgrade
+var (
+	UpgradeName             = "v7"
+	IncentiveProgramAddress = "stride1tlxk4as9sgpqkh42cfaxqja0mdj6qculqshy0gg3glazmrnx3y8s8gsvqk"
+	StrideFoundationAddress = "stride1yz3mp7c2m739nftfrv5r3h6j64aqp95f3degpf"
```
asalzmann, line 38: nit: please label this F4

app/upgrades/v7/README.md
asalzmann, line 2: @shellvish to do in a separate PR

app/upgrades/v7/upgrades.go
```diff
@@ -0,0 +1,199 @@
...
+	k.SetParams(ctx, params)
+}
+
+// Add the min/max redemption rate to each host zone as safety bounds, using the default for each
+// Also set the Halted boolean to false
+func AddRedemptionRateSafetyChecks(ctx sdk.Context, k stakeibckeeper.Keeper) {
+	ctx.Logger().Info("Setting min/max redemption rate safety bounds on each host zone")
+
+	// Get default min/max redemption rate
+	defaultMinRedemptionRate := sdk.NewDecWithPrec(int64(k.GetParam(ctx, types.KeyDefaultMinRedemptionRateThreshold)), 2)
```
asalzmann, line 145: I think we need to set this param before getting it - should this happen before running `AddRedemptionRateSafetyChecks`?

app/upgrades/v7/upgrades.go
```diff
@@ -0,0 +1,199 @@
...
+	defaultMaxRedemptionRate := sdk.NewDecWithPrec(int64(k.GetParam(ctx, types.KeyDefaultMaxRedemptionRateThreshold)), 2)
+
+	for _, hostZone := range k.GetAllHostZone(ctx) {
+
+		hostZone.MinRedemptionRate = defaultMinRedemptionRate
```
asalzmann, line 150: What's the thinking around using defaults vs. tuning to zones?

app/upgrades/v7/upgrades.go
```diff
@@ -0,0 +1,199 @@
...
+	if !found {
+		return stakeibctypes.ErrHostZoneNotFound
+	}
+	junoHostZone.UnbondingFrequency = unbondingFrequency
+	k.SetHostZone(ctx, junoHostZone)
+
+	return nil
+}
+
+// Incentive diversification (Prop #153) - Send 3M STRD from Incentive Program to Stride Foundation
```
asalzmann, line 175:
```suggestion
// Incentive diversification (Prop #153) - Send 3M STRD from Incentive Program to Stride Foundation
// so that it can be swapped for USDC and deployed as incentives 
```

app/upgrades/v7/upgrades.go
```diff
@@ -0,0 +1,199 @@
...
+
+	incentiveProgramAddress, err := sdk.AccAddressFromBech32(IncentiveProgramAddress)
+	if err != nil {
+		return err
+	}
+	strideFoundationAddress, err := sdk.AccAddressFromBech32(StrideFoundationAddress)
+	if err != nil {
+		return err
+	}
+	amount := sdk.NewCoin("ustrd", sdk.NewInt(3_000_000_000_000))
```
asalzmann, line 187: nit: thoughts on using a `constants.go` file for things like this? 

The benefit I see is that it's easier to review the constants, and especially for constants like this where 3m STRD are being sent, it should reduce human error in the review process

asalzmann (commented):
PR needs a description - LMK if I can help with that! Otherwise lgtm pending the comments

app/upgrades/v7/upgrades.go
asalzmann: Done

app/upgrades/v7/upgrades.go
asalzmann: Done

app/upgrades/v7/upgrades.go
asalzmann: Done

app/upgrades/v7/upgrades.go
asalzmann: Done

app/upgrades/v7/upgrades.go
sampocs: Good call! Wanna move the addresses and "ustrd" there as well?

app/upgrades/v7/upgrades.go
sampocs: great catch! Thanks for resolving

app/upgrades/v7/upgrades.go
sampocs: done

PR #646, opened 2023-03-02
Lower Max Validator Threshold

shellvish:


sampocs (approved):
Lgtm, thanks for doing this!

PR #651, opened 2023-03-08
Code Cleanup: MsgLiquidStake

sampocs:
Closes: #XXX

Cleaned up MsgLiquidStake

## Brief Changelog
* Cleaned up variable naming, commenting, logging, and errors
* Moved validation checks as upstream as possible so that they occur before any state changes (relevant to protocol directed liquid stakes)
* Added filter to GetDepositRecordsByEpochAndChain to only grab transfer records
* Added event emission after LS
* Added check to error if the LS would result in 0 stTokens

## Author's Checklist

I have...

## Reviewers Checklist

I have...

      - [ ] not applicable
      - [ ] jira ticket `XXX` 
      - [ ] README.md 
      - [ ] not documented


asalzmann (changes requested):
Would prefer to push this to v8 since it touches our most sensitive code. Any reason we have to get this in for v7?

x/stakeibc/keeper/msg_server_liquid_stake.go
```diff
@@ -17,114 +15,91 @@ import (
 func (k msgServer) LiquidStake(goCtx context.Context, msg *types.MsgLiquidStake) (*types.MsgLiquidStakeResponse, error) {
```
asalzmann, line 15: Can you add a comment specifying that this function has to function even if run non-atomically?

x/stakeibc/keeper/msg_server_liquid_stake.go
```diff
@@ -17,114 +15,91 @@ import (
...
+		return nil, errorsmod.Wrapf(types.ErrInvalidToken, "no host zone found for denom (%s)", msg.HostDenom)
 	}
 
-	// Creator owns at least "amount" of inCoin
-	balance := k.bankKeeper.GetBalance(ctx, sender, ibcDenom)
-	if balance.IsLT(inCoin) {
-		k.Logger(ctx).Error(fmt.Sprintf("balance is lower than staking amount. staking amount: %v, balance: %v", msg.Amount, balance.Amount))
-		return nil, errorsmod.Wrapf(sdkerrors.ErrInsufficientFunds, "balance is lower than staking amount. staking amount: %v, balance: %v", msg.Amount, balance.Amount)
+	// Get user and module account addresses
+	userAddress, err := sdk.AccAddressFromBech32(msg.Creator)
```
asalzmann, line 25: nit: sender is clearer to me, but I don't have a strong view! other addresses (e.g. modules) could stake, whereas `userAddress` sounds like an individual

x/stakeibc/keeper/msg_server_liquid_stake.go
```diff
@@ -17,114 +15,91 @@ import (
...
-	// transfer those coins to the user
-	err = k.bankKeeper.SendCoinsFromModuleToAccount(ctx, types.ModuleName, sender, stCoins)
-	if err != nil {
-		k.Logger(ctx).Error("Failed to send coins from module to account")
-		return errorsmod.Wrapf(sdkerrors.ErrInvalidRequest, "Failed to send %s from module to account", stCoins.GetDenomByIndex(0))
+	// Mint the stTokens and transfer them to the user
+	stDenom := types.StAssetDenomFromHostZoneDenom(msg.HostDenom)
+	stCoin := sdk.NewCoin(stDenom, stAmount)
+	if err := k.bankKeeper.MintCoins(ctx, types.ModuleName, sdk.NewCoins(stCoin)); err != nil {
+		return nil, errorsmod.Wrapf(err, "Failed to mint coins")
```
asalzmann, line 79: Given this needs to work non-atomically - should we refund the coins sent in here if an error is thrown? I'd err on the side of no, but it will require a software upgrade to fix this if something goes wrong

asalzmann (commented):
LGTM - although we should get multiple reviews before merging

x/stakeibc/keeper/msg_server_liquid_stake.go
sampocs: Agreed `user` isn't great, but I think sender is a bit too generic considering there are multiple sends going on in this function. Can you think of a good third option?

asalzmann: `msgSender` perhaps?

x/stakeibc/keeper/msg_server_liquid_stake.go
sampocs: that's a good idea! Lemme read through the function again and get back to ya
Feels like there's other stuff we'd have to revert too (e.g. burn coins if the send from module account fails)

x/stakeibc/keeper/msg_server_liquid_stake.go
asalzmann: `caller`, `liquidStakingAddress` could also work

sampocs: I like those last two! How bout `liquidStakerAddress`?

asalzmann: sgtm!

x/stakeibc/keeper/msg_server_liquid_stake.go
```diff
@@ -17,114 +15,91 @@ import (
...
-	if !isIbcToken {
-		k.Logger(ctx).Error("invalid token denom - denom is not an IBC token (%s)", ibcDenom)
-		return nil, errorsmod.Wrapf(types.ErrInvalidToken, "denom is not an IBC token (%s)", ibcDenom)
+	hostZoneAddress, err := sdk.AccAddressFromBech32(hostZone.Address)
+	if err != nil {
+		return nil, errorsmod.Wrapf(err, "host zone address is invalid")
 	}
 
-	// safety check: redemption rate must be above safety threshold
+	// Safety check: redemption rate must be above safety threshold
```
shellvish, line 34: nit: must be within safety bounds

x/stakeibc/keeper/msg_server_liquid_stake.go
```diff
@@ -17,114 +15,91 @@ import (
...
+	stDenom := types.StAssetDenomFromHostZoneDenom(msg.HostDenom)
+	stCoin := sdk.NewCoin(stDenom, stAmount)
+	if err := k.bankKeeper.MintCoins(ctx, types.ModuleName, sdk.NewCoins(stCoin)); err != nil {
+		return nil, errorsmod.Wrapf(err, "Failed to mint coins")
+	}
+	if err := k.bankKeeper.SendCoinsFromModuleToAccount(ctx, types.ModuleName, userAddress, sdk.NewCoins(stCoin)); err != nil {
+		return nil, errorsmod.Wrapf(err, "Failed to send %s from module to account", stCoin.String())
 	}
 
-	k.Logger(ctx).Info(fmt.Sprintf("[MINT ST ASSET] success on %s.", hz.GetChainId()))
```
shellvish, line 128: nit: should we add a successful log after liquid staking?

x/stakeibc/keeper/msg_server_liquid_stake.go
```diff
@@ -17,114 +15,91 @@ import (
...
 
-	k.hooks.AfterLiquidStake(ctx, sender)
-	return &types.MsgLiquidStakeResponse{}, nil
-}
-
-func (k msgServer) MintStAssetAndTransfer(ctx sdk.Context, sender sdk.AccAddress, amount sdkmath.Int, denom string) error {
-	stAssetDenom := types.StAssetDenomFromHostZoneDenom(denom)
+	// The tokens that are sent to the protocol are denominated in the ibc hash of the native token on stride (e.g. ibc/xxx)
+	nativeDenom := hostZone.IbcDenom
+	nativeCoin := sdk.NewCoin(nativeDenom, msg.Amount)
```
shellvish, line 52: piconit, should line 52 be below the if-statement?

x/stakeibc/keeper/msg_server_liquid_stake.go
```diff
@@ -17,114 +15,91 @@ import (
...
+	if balance.IsLT(nativeCoin) {
+		return nil, errorsmod.Wrapf(sdkerrors.ErrInsufficientFunds, "balance is lower than staking amount. staking amount: %v, balance: %v", msg.Amount, balance.Amount)
 	}
 
-	// Mints coins to the module account, will error if the module account does not exist or is unauthorized.
+	// Determine the amount of stTokens to mint using the redemption rate
+	stAmount := (sdk.NewDecFromInt(msg.Amount).Quo(hostZone.RedemptionRate)).TruncateInt()
+	if stAmount.IsZero() {
+		return nil, errorsmod.Wrapf(types.ErrInsufficientLiquidStake,
+			"Liquid stake of %s%s would return 0 stTokens", msg.Amount.String(), hostZone.HostDenom)
```
shellvish, line 67: good catch!

x/stakeibc/keeper/msg_server_liquid_stake.go
```diff
@@ -17,114 +15,91 @@ import (
...
 	}
 
-	// transfer those coins to the user
-	err = k.bankKeeper.SendCoinsFromModuleToAccount(ctx, types.ModuleName, sender, stCoins)
-	if err != nil {
-		k.Logger(ctx).Error("Failed to send coins from module to account")
-		return errorsmod.Wrapf(sdkerrors.ErrInvalidRequest, "Failed to send %s from module to account", stCoins.GetDenomByIndex(0))
+	// Mint the stTokens and transfer them to the user
+	stDenom := types.StAssetDenomFromHostZoneDenom(msg.HostDenom)
+	stCoin := sdk.NewCoin(stDenom, stAmount)
```
shellvish, line 77: much better than `ParseCoinsNormalized` above

shellvish (approved):
Looks good to me! 

I think we should revert changes though if `MintCoins` or `SencdCoinsFromModuleToAccount` fails, even though it'll be a bit tedious in the code. Happy to help out here!

x/stakeibc/keeper/msg_server_liquid_stake.go
sampocs: done

x/stakeibc/keeper/msg_server_liquid_stake.go
sampocs: done

x/stakeibc/keeper/msg_server_liquid_stake.go
sampocs: chatted offline, I think we're in favor of leaving the function as is instead of creating a non-atomic version considering the low that the function fails with impartial state, and the higher probabillity of introducing a new bug by trying to add in reversal logic

x/stakeibc/keeper/msg_server_liquid_stake.go
```diff
@@ -17,114 +15,91 @@ import (
...
-	// create a deposit record of these tokens (pending transfer)
+	// Grab the deposit record that will be used for record keeping
 	strideEpochTracker, found := k.GetEpochTracker(ctx, epochtypes.STRIDE_EPOCH)
 	if !found {
-		k.Logger(ctx).Error("failed to find stride epoch")
 		return nil, errorsmod.Wrapf(sdkerrors.ErrNotFound, "no epoch number for epoch (%s)", epochtypes.STRIDE_EPOCH)
 	}
-	// Does this use too much gas?
-	depositRecord, found := k.RecordsKeeper.GetDepositRecordByEpochAndChain(ctx, strideEpochTracker.EpochNumber, hostZone.ChainId)
+	depositRecord, found := k.RecordsKeeper.GetTransferDepositRecordByEpochAndChain(ctx, strideEpochTracker.EpochNumber, hostZone.ChainId)
```
riley-stride, line 45: Great change

x/stakeibc/keeper/msg_server_liquid_stake_test.go
```diff
@@ -156,65 +157,50 @@ func (s *KeeperTestSuite) TestLiquidStake_DifferentRedemptionRates() {
...
-	s.FundAccount(tc.user.acc, sdk.NewInt64Coin(badDenom, 1000000000))
-	_, err := s.GetMsgServer().LiquidStake(sdk.WrapSDKContext(s.Ctx), &tc.validMsg)
 
-	s.Require().EqualError(err, fmt.Sprintf("denom is not an IBC token (%s): invalid token denom", badHostZone.IbcDenom))
+	_, err := s.GetMsgServer().LiquidStake(sdk.WrapSDKContext(s.Ctx), &tc.validMsg)
+	s.Require().EqualError(err, "host zone address is invalid: decoding bech32 failed: string not all lowercase or all uppercase")
 }
 
-func (s *KeeperTestSuite) TestLiquidStake_InsufficientBalance() {
+func (s *KeeperTestSuite) TestLiquidStake_RateBelowMinThreshold() {
```
riley-stride, line 193: Can we add one for above threshold?

riley-stride: And halted zone please

x/stakeibc/keeper/msg_server_liquid_stake.go
sampocs: done

riley-stride (approved):

x/stakeibc/keeper/msg_server_liquid_stake.go
sampocs: I was thinking it's not really necessary to have logs in transactions cause it's more common to check the status by querying the tx than by looking in the logs. But not opposed to adding it if you think it makes sense!

x/stakeibc/keeper/msg_server_liquid_stake.go
sampocs: I can move it down! It's more just a question of readability (NewCoin will not fail if the denom is not an ibc denom)

x/stakeibc/keeper/msg_server_liquid_stake.go
```diff
@@ -17,114 +15,91 @@ import (
 func (k msgServer) LiquidStake(goCtx context.Context, msg *types.MsgLiquidStake) (*types.MsgLiquidStakeResponse, error) {
 	ctx := sdk.UnwrapSDKContext(goCtx)
 
-	// Init variables
-	// deposit `amount` of `denom` token to the stakeibc module
-	// NOTE: Should we add an additional check here? This is a pretty important line of code
-	// NOTE: If sender doesn't have enough inCoin, this errors (error is hard to interpret)
-	// check that hostZone is registered
-	// strided tx stakeibc liquid-stake 100 uatom
+	// Get the host zone from the base denom in the message (e.g. uatom)
```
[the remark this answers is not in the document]
sampocs: good call! Will merge in main and double check that it gets added

x/stakeibc/keeper/msg_server_liquid_stake_test.go
sampocs: yessir! Great ideas

asalzmann (approved):

PR #652, opened 2023-03-09
Validator Rebalancing Update

shellvish:

Fixes a minor Abs bug (resulting in more inefficient rebalancing) as well as cleans up some redundant computation.


sampocs (approved):
lgtm, nice catch!

PR #658, opened 2023-03-09
Safely Add Module Account

sampocs:
Closes: #XXX

* Borrowed from osmosis - in the event that we are attempting to create a module account with the same address as an active user account, we should error instead of overriding it.

## Brief Changelog
* Added `CreateModuleAccount` function to first check that a module account can be created
* Modified `msg_server_register_host_zone.go` to use the new function when creating module accounts for each host zone
* **Question for reviewer**: Are there any additional accounts to add?

## Author's Checklist

I have...

## Reviewers Checklist

I have...

      - [ ] not applicable
      - [ ] jira ticket `XXX` 
      - [ ] README.md 
      - [ ] not documented


utils/module_account.go
```diff
@@ -0,0 +1,76 @@
+package utils
+
+// Borrowed from Osmosis
+// https://github.com/osmosis-labs/osmosis/blob/5297ea4b641c0c1090cb253a34a510dd0c85b407/osmoutils/module_account.go
+// Allows the safe creation of module accounts in the event that funds have already been sent to the address
```
asalzmann, line 5:
```suggestion
// Allows the safe creation of module accounts in the event that funds have already been sent to the address and the address has sent a tx successfully 
```

asalzmann (approved):
LGTM! How did you test this?

sampocs:
@asalzmann 
> LGTM! How did you test this?

Just ran integration tests (since we create module accounts when registering host zones). Lmk if you can think of an additional way to test it.

Not sure how we can test the case where the creation fails because there's an active base account at that address (i.e. an account with a sequence number greater than 0) since that would involve figuring out the mnemonic that creates the module account address

PR #659, opened 2023-03-10
remove unnecessary errs in hooks

shellvish:

Hooks.go contains an unnecessary Error, and the Error was handled incorrectly (`return` instead of `continue`). Given that we can't make any atomicity assumptions in `hooks.go`, I removed the extra Error + Return statement. 

## Brief Changelog

Modify error handling in `UpdateRedemptionRates`


asalzmann (approved):
lgtm

PR #669, opened 2023-03-14
Add readme for integration failure fix

jstr1121:

Add description to resolve integration test failure on local environment.

## Brief Changelog

Update README on dockernet for integration test failure fix


sampocs (changes requested):
I don't think it's necessary to include this note that ibc-transfer used to fail. I think in your case, your setup was just a little outdated (and it seemed like you had made some changes to the submodules that might've broken things)

I would rephrase this to:
```
### Debugging Integration Test Failures

If running into issues on dockernet, please make sure your local repo is sync'd with the latest changes.

git pull
git submodule sync --recursive
git submodule --init --recursive
```

PR #676, opened 2023-03-22
Add stumee 

EbrahimUmee:


robert-zaremba (commented):
You should use our official docker image, no need to create a custom one.
https://github.com/umee-network/umee/pkgs/container/umeed

PR #677, opened 2023-03-22
Add Multiple Validators in One Transaction

sampocs:
Closes: #XXX

The `add-validator` function was previous limited to adding only one validator at a time. This PR alters the transaction so that a JSON with an arbitrary number of validators can be supplied to improve the dev experience.
**NOTE FOR REVIEWER**: It might be easier to scroll through commits since files were renamed

## Brief Changelog
* Change `add-validator` → `add-validators`
* Modify `add-validators` to take a list of validators as a JSON instead of a single validator
* Removed unused fields from validator struct. **We should confirm with localstride that this doesn't require a migration**. 
  * I did test by running the old binary, and then querying the host zones with this binary. The removed fields did not appear and the query did not error - indicating that this new binary can successfully deserialize the old store (which should mean we do not need a state migration).

### Setup
* Start dockernet
```bash
make start-docker build=sgr
```

* Confirm there's one validator
```bash
```

### Success Case
* Create a json file called `add_validator_admin.json` as follows:
```json
{
    "validators": [
        {"name": "val-from-admin-1", "address": "cosmosvaloper196ax4vc0lwpxndu9dyhvca7jhxp70rmcvrj90c", "weight": 10},
        {"name": "val-from-admin-2", "address": "cosmosvaloper1v5y0tg0jllvxf5c3afml8s3awue0ymju89frut", "weight": 0},
        {"name": "val-from-admin-3", "address": "cosmosvaloper1clpqr4nrk4khgkxj78fcwwh6dl3uw4epsluffn", "weight": 30}
    ]
}
```

* Create json file called `add_validator_gov.json` as follows:
```json
{
    "description": "Proposal to add GAIA validators",
    "hostZone": "GAIA",
    "validators": [
        {"name": "val-from-gov-1", "address": "cosmosvaloper1lzhlnpahvznwfv4jmay2tgaha5kmz5qxerarrl"},
        {"name": "val-from-gov-2", "address": "cosmosvaloper16k579jk6yt2cwmqx9dz5xvq9fug2tekvlu9qdv"},
        {"name": "val-from-gov-3", "address": "cosmosvaloper1g48268mu5vfp4wk7dk89r0wdrakm9p5xk0q50k"}
     ],
    "deposit": "64000000ustrd"
}
```

* Add validators through admin function
```bash
strided tx stakeibc add-validators GAIA add_validator_admin.json --from admin 
```

* Confirm admin validators were added successfully
```bash
```

* Add validators though governance (**you have 30 seconds to get the votes in!**)
```
strided tx gov submit-legacy-proposal add-validators add_validator_gov.json --from val1 --gas auto -y 
```
* After the voting period has passed (30 seconds), confirm the gov validators were added. Note that the weight is equal to 5 (the lowest weight, excluding 0)
```bash
```

### Failure Case - Invalid Host Zone
* Attempt to add validators with a host zone that does not exist, it should fail
```bash
strided tx stakeibc add-validators FAKE add_validator_admin.json --from admin 
```

### Failure Case - Admin - Validator already exists
* For each of the following, inspect the tx hash to identify the failure
* Replace `add_validator_admin.json` with the following
```json
{
    "validators": [
        {"name": "val-from-admin-8", "address": "cosmosvaloper14kn0kk33szpwus9nh8n87fjel8djx0y070ymmj", "weight": 10},
        {"name": "val-from-admin-1", "address": "cosmosvaloper1vvwtk805lxehwle9l4yudmq6mn0g32px9xtkhc", "weight": 10}
    ]
}
```
* Attempt to add the validators - it should fail because the validator **name** already exists
```
strided tx stakeibc add-validators GAIA add_validator_admin.json --from admin 
```

* Now replace it again with the following
```json
{
    "validators": [
        {"name": "val-from-admin-8", "address": "cosmosvaloper14kn0kk33szpwus9nh8n87fjel8djx0y070ymmj", "weight": 10},
        {"name": "val-from-admin-9", "address": "cosmosvaloper196ax4vc0lwpxndu9dyhvca7jhxp70rmcvrj90c", "weight": 10}
    ]
}
```
* Attempt to add the validators - it should fail because the validator **address** already exists
```
strided tx stakeibc add-validators GAIA add_validator_admin.json --from admin 
```

### Failure Case - Gov - Validator already exists
* For each of the following, the tx should fail immediately. **Note**: The error log is very verbose, but you can see the relevant error message if you look close enough
* Replace `add_validator_gov.json` with the following
```json
{
    "description": "Proposal to add GAIA validators",
    "hostZone": "GAIA",
    "validators": [
        {"name": "val-from-admin-8", "address": "cosmosvaloper14kn0kk33szpwus9nh8n87fjel8djx0y070ymmj", "weight": 10},
        {"name": "val-from-admin-1", "address": "cosmosvaloper1vvwtk805lxehwle9l4yudmq6mn0g32px9xtkhc", "weight": 10}
    ],
    "deposit": "64000000ustrd"
}
```
* Attempt to add the validators - it should fail because the validator **name** already exists
```bash
strided tx gov submit-legacy-proposal add-validators add_validator_gov.json --from val1 --gas auto
```

* Now replace it again with the following
```json
{
    "description": "Proposal to add GAIA validators",
    "hostZone": "GAIA",
    "validators": [
        {"name": "val-from-admin-8", "address": "cosmosvaloper14kn0kk33szpwus9nh8n87fjel8djx0y070ymmj", "weight": 10},
        {"name": "val-from-admin-9", "address": "cosmosvaloper196ax4vc0lwpxndu9dyhvca7jhxp70rmcvrj90c", "weight": 10}
    ],
    "deposit": "64000000ustrd"
}

```
* Attempt to add the validators - it should fail because the validator **address** already exists
```
strided tx gov submit-legacy-proposal add-validators add_validator_gov.json --from val1 --gas auto
```

## Author's Checklist

I have...

## Reviewers Checklist

I have...

      - [ ] not applicable
      - [ ] jira ticket `XXX` 
      - [ ] README.md 
      - [ ] not documented


x/stakeibc/types/message_add_validators.go
```diff
@@ -0,0 +1,69 @@
...
+	}
+	if err := utils.ValidateAdminAddress(msg.Creator); err != nil {
+		return err
+	}
+
+	if len(msg.Validators) == 0 {
+		return errorsmod.Wrapf(sdkerrors.ErrInvalidRequest, "at least one validator must be provided")
+	}
+
+	for i, validator := range msg.Validators {
```
riley-stride, line 59: can we validate validator addresses, even though they're on a different chain?

riley-stride: maybe also check weights >= 0 here

x/stakeibc/types/message_add_validators_test.go
```diff
@@ -0,0 +1,93 @@
...
+)
+
+func TestMsgAddValidators_ValidateBasic(t *testing.T) {
+	validNonAdminAddress, invalidAddress := apptesting.GenerateTestAddrs()
+	adminAddress, ok := apptesting.GetAdminAddress()
+	require.True(t, ok)
+
+	valName1 := "val1"
+	valName2 := "val2"
+	valAddress1 := "cosmosvaloper1"
```
riley-stride, line 19: should we test with a valid and an invalid address?

x/stakeibc/simulation/add_validator.go
```diff
@@ -19,7 +19,7 @@ func SimulateMsgAddValidator(
 	return func(r *rand.Rand, app *baseapp.BaseApp, ctx sdk.Context, accs []simtypes.Account, chainID string,
 	) (simtypes.OperationMsg, []simtypes.FutureOperation, error) {
 		simAccount, _ := simtypes.RandomAcc(r, accs)
-		msg := &types.MsgAddValidator{
+		msg := &types.MsgAddValidators{
```
riley-stride, line 22: do we actually run these sim tests? if not let's remove this

x/stakeibc/migrations/v2/convert.go
```diff
@@ -11,8 +11,6 @@ func convertToNewValidator(oldValidator oldstakeibctypes.Validator) stakeibctype
 	return stakeibctypes.Validator{
 		Name:                 oldValidator.Name,
 		Address:              oldValidator.Address,
```
riley-stride, line 13: I realize these fields are unused but why are we changing the v2 upgrade logic?

x/stakeibc/keeper/msg_server_add_validators_test.go
```diff
@@ -0,0 +1,102 @@
...
+	s.App.StakeibcKeeper.SetHostZone(s.Ctx, hostZone)
+
+	return AddValidatorsTestCase{
+		hostZone:           hostZone,
+		validMsg:           validMsg,
+		expectedValidators: expectedValidators,
+	}
+}
+
+func (s *KeeperTestSuite) TestAddValidators_Successful() {
```
riley-stride, line 50: Can we add a negative weight test?

x/stakeibc/migrations/v2/convert.go
riley-stride: I see your PR comment
```
I did test by running the old binary, and then querying the host zones with this binary. The removed fields did not appear and the query did not error - indicating that this new binary can successfully deserialize the old store (which should mean we do not need a state migration).
```
If a new node is catching up and processes an add-validator tx from before this upgrade but after v2, won't it have issues writing the val with the extraneous fields and mismatch with what was written by the current vals (who wrote the extraneous fields to store)?

riley-stride (approved):

x/stakeibc/types/message_add_validators.go
sampocs: I was messing around with the validator addresses but couldn't really figure out how to do it with the functions at our disposal cause we need each chain's valoper prefix. We could add the valoper prefix to our host zone though - I don't think this is the first time we could have benefitted from it - but maybe that's for a separate PR

Weight is a uint so it has to be >= 0

x/stakeibc/types/message_add_validators_test.go
sampocs: the invalid address is in one of the test cases (empty address)

See my comment above about validating the true validity of the address

x/stakeibc/simulation/add_validator.go
sampocs: Should we just wipe the whole folder? If so, I'll do that in a separate PR

x/stakeibc/migrations/v2/convert.go
sampocs: > I realize these fields are unused but why are we changing the v2 upgrade logic?

It's a very good question. The reason I changed these files was to get around a build error - they use the most updated Validator data type which doesn't have the commission status field.

I was a bit worried about breaking the old migration code. And I considered adding the old types here (old as in what's on main, not v2). But decided against it because:
   1) I would have to create a new v3 folder which makes it appear as if there's a migration for this upgrade which there is not, and 
  2) I don't think this code really matters anymore because it's only executed by the v5 upgrade handler which must be run with the v5 binary.

But that said, I'm curious to hear your take!

sampocs: > If a new node is catching up and processes an add-validator tx from before this upgrade but after v2, won't it have issues writing the val with the extraneous fields and mismatch with what was written by the current vals (who wrote the extraneous fields to store)?

I don't think they will because they'll have to be running the v6 or v7 binary which have the old types in there. The only situation where they would have this mismatch is if they were running the v5 upgrade logic from the v8 binary, which I don't think happens

x/stakeibc/keeper/msg_server_add_validators_test.go
sampocs: it's not possible because the weight is a uint64

x/stakeibc/simulation/add_validator.go
sampocs: chatted offline - planning to remove in separate PR

x/stakeibc/types/message_add_validators.go
sampocs: chatted offline - doesn't seem to be a way to validate the validators address without knowing the host zone valoper prefix. We can revisit if we ever start storing the prefix

x/stakeibc/migrations/v2/convert.go
riley-stride: makes sense, resolving

riley-stride (approved):

PR #680, opened 2023-03-23
Brian/instant redeem

antstalepresh:

Final implementation for Instant redemptions, pulling tokens from Deposit Records in current epoch and accrued rewards

**Design for instant redemptions**
To process instant redemptions, the basic idea is simple - we have some ATOM that is about to be staked, which can be used to process redemptions. The reason this works is because 

Total_staked_n+1 = new_staked_n - new_unstaked_n

So we can short-circuit unbondings by decreasing both unstaked and staked amounts. Unstaked tokens are tracked on `HostZoneUnbonding` records (on `EpochUnbondingRecord` records). Staked tokens that can be matched with unbondings can be

1. Newly staked tokens (tracked on `DepositRecord` records)
2. Interest that is about to be reinvested

Newly staked tokens are simple but unpredictable, there may be many epochs where Total_staked_n+1 < 0 (therefore, we can’t process instant unbondings). Interest that is about to be reinvested is slightly more complex but predictable - it accrues daily and is proportional to TVL.

For newly staked tokens, we can just match redemptions to the `DepositRecord` from the current epoch. If the redeemed amount is greater than the amount on the deposit record, get the amount to unbond (using the redemption rate), burn the corresponding stATOM, and send the ATOM to the redeemer, decrementing the deposit record.

For accrued interest, the interest is on an ICA. Therefore we have to IBC transfer tokens back to Stride and allow them to be redeemed for one epoch, before sending the remainder back to the host zone. We will likely need a new data type (`InterestRecord`) that tracks an epoch, amount, and zone. We will also need a new module account to store tokens in and epochly logic to handle IBC token transfers.

**x/stakeibc**
- InstantRedeemStake(creator, amt, zone): checks the sources of tokens that can be unbonded and processes an instant-redeem-stake if it’s valid  

## Brief Changelog

  - InstantRedeemStake implementation, including cli command and unit testing

## Integration tests
Start dockernet, run dockernet integration test
```
make start-docker build=sg
make test-integration-docker
```
Screenshot 2023-03-23 at 9 37 21 AM

Here are steps for testing instant redemption via example CLI commands. 
- Check account balance, host zone and deposit records status.
```
strided q records list-deposit-record
strided q bank balances stride1uk4ze0x4nvh4fk0xm4jdud58eqn4yxhrt52vv7
 ```
- Make sure there is a deposit record by staking some uatom
````
strided tx stakeibc liquid-stake 100 uatom --from val1 --keyring-backend=test --chain-id STRIDE
````
- Try to instant redeem stake a smaller valid amount for success
````
strided tx stakeibc instant-redeem-stake 10 GAIA --from val1 --keyring-backend=test --chain-id STRIDE
````
- Try to instant redeem a larger amount, or wait for epoch to end for failure
````
strided tx stakeibc instant-redeem-stake 200 GAIA --from val1 --keyring-backend=test --chain-id STRIDE
````
- Verify transaction completed by querying the txns, and checking the balances
````
strided q records list-deposit-record
strided q bank balances stride1uk4ze0x4nvh4fk0xm4jdud58eqn4yxhrt52vv7
````

## Author's Checklist


asalzmann:
still need to review

PR #689, opened 2023-03-24
Autopilot JSON Memo

sampocs:
Closes: #XXX

To support increased flexibility with memo-based packet forwarding middleware, autopilot has been refactored to accept a JSON format instead of a string. 

Format:
```json
{ 
    "autopilot": {
          "receiver": "strideXXX", 
          "{module_name}": { "{additiional_field}": "{value}" }
    }
}
```

Example:
```json
{ 
    "autopilot": {
          "receiver": "strideXXX", 
          "stakeibc": {
               "stride_address": "strideXXX",
               "action": "LiquidStake",
          }
    }
}
```

## Brief Changelog
* Refactored parsing logic to accept a JSON instead of a string
* Unified the 1-click liquid staking code for before and after ibc-go v5 (previously each version was handled differently)
* Added Claim memo parsing to support airdrops from non-118 coin types that require an IBC transfer
* Added address mapping from non-118 coin types to claim records

## Author's Checklist

I have...

## Reviewers Checklist

I have...

      - [ ] not applicable
      - [ ] jira ticket `XXX` 
      - [ ] README.md 
      - [ ] not documented


x/autopilot/types/params.go
```diff
@@ -8,12 +8,14 @@ import (
 )
 
 const (
-	// DefaultActive is the default value for the active param (set to true)
-	DefaultActive = false
+	// Default active value for each autopilot supported module
+	DefaultStakeibcActive = false
+	DefaultClaimActive    = true
```
asalzmann, line 13: Wdyt of setting this to false, and overriding to true in the upgrade?

x/autopilot/module_ibc.go
```diff
@@ -123,51 +123,54 @@ func (im IBCModule) OnRecvPacket(
...
-	newPacket := packet
-	newPacket.Data = bz
-
-	// process the transfer receipt
-	// NOTE: this code is pulled from packet-forwarding-middleware
-	ack := im.app.OnRecvPacket(ctx, newPacket, relayer)
-	if ack.Success() {
-		return im.keeper.TryLiquidStaking(ctx, packet, newData, parsedReceiver, ack)
+	// Handle for each module
+	if parsedPacketMetadata.Stakeibc.Enabled {
```
asalzmann, line 148: should we be checking whether StakeibcActive is true here? Feels slightly safer to put the module param check here

x/autopilot/module_ibc.go
```diff
@@ -123,51 +123,54 @@ func (im IBCModule) OnRecvPacket(
...
+		newPacket.Data = bz
+
+		// process the transfer receipt
+		// NOTE: this code is pulled from packet-forwarding-middleware
+		ack := im.app.OnRecvPacket(ctx, newPacket, relayer)
+		if ack.Success() {
+			return im.keeper.TryLiquidStaking(ctx, packet, newData, &parsedPacketMetadata.Stakeibc, ack)
+		}
+		return ack
+	} else if parsedPacketMetadata.Claim.Enabled {
```
asalzmann, line 166: same comment as above, for ClaimActive

x/autopilot/module_ibc.go
```diff
@@ -123,51 +123,54 @@ func (im IBCModule) OnRecvPacket(
...
+		// TODO
```
asalzmann, line 167: cc @shellvish

x/autopilot/types/parser.go
```diff
@@ -1,51 +1,132 @@
...
+		return nil, err
+	}
+	if len(strings.TrimSpace(r.AirdropId)) == 0 {
+		return nil, ErrInvalidClaimAirdropId
+	}
+
+	return &ClaimPacketMetadata{
+		AirdropId:     r.AirdropId,
+		StrideAddress: address,
+		Enabled:       true,
```
asalzmann, line 91: I'm finding the `Enabled` concept a bit confusing. How are you thinking about this?

proto/stride/autopilot/params.proto
```diff
@@ -9,6 +9,7 @@ option go_package = "github.com/Stride-Labs/stride/v7/x/autopilot/types";
 // next id: 1
 message Params {
   option (gogoproto.goproto_stringer) = false;
-  // optionally, turn off this module
-  bool active = 1;
+  // optionally, turn off each module
+  bool stakeibc_active = 1;
```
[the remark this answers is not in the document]
sampocs: in upgrade handler PR

x/autopilot/types/params.go
sampocs: I was wondering the same thing. I lean slightly towards setting them like this to keep our params as close to the defaults as possible (given how clunky it is to change them in the upgrade handler)

x/autopilot/module_ibc.go
sampocs: i was wondering the same thing - agreed here makes a little more sense

sampocs: actually module_ibc doesn't have access to the keeper to grab the params. We can either leave it as is or move this whole block to the keeper. Wdyt?

x/autopilot/types/parser.go
sampocs: Yeah the variable name is pretty bad. The idea is that each packet is either just a normal packet or it's a packet intended for claim/liquid stake forwarding. 

To me, having an abstraction make sense:
* Grab the packet
* Inspect it to see if there's any fowarding specifications
* Store that info in a struct so that later you can easily check it with an if statement as what's done in module_ibc

That said, maybe some better name options would be `RoutingEnabled`, `Forward`, `RoutingTriggered`, etc.

x/autopilot/types/params.go
shellvish: +1 I would also keep them to the defaults

x/autopilot/module_ibc.go
shellvish: I can submit a PR for this after this is merged in! probably cleanest (given how sensitive the change will be)

shellvish (approved):
LGTM although I'm fairly unfamiliar with this code. Might be good to walk through `parser.go` together?

x/autopilot/genesis_test.go
```diff
@@ -12,7 +12,10 @@ import (
 
 func TestGenesis(t *testing.T) {
 	expectedGenesisState := types.GenesisState{
-		Params: types.Params{Active: true},
+		Params: types.Params{
+			StakeibcActive: true,
```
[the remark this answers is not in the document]
sampocs: this is just a unit test - i'm just setting em to an arbitrary value

x/autopilot/types/parser.go
```diff
@@ -1,51 +1,132 @@
...
+	if r == nil {
+		return &ClaimPacketMetadata{Enabled: false}, nil
+	}
+
+	// Validate the stride address and airdrop ID
+	address, err := sdk.AccAddressFromBech32(r.StrideAddress)
+	if err != nil {
+		return nil, err
+	}
+	if len(strings.TrimSpace(r.AirdropId)) == 0 {
```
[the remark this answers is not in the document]
sampocs: this is more of a validate basic test - it's not stateful. But I think that's a good check we should put in when it comes time for the actual implementation

x/autopilot/module_ibc.go
```diff
@@ -123,51 +123,54 @@ func (im IBCModule) OnRecvPacket(
...
+	var metadata string
+	if data.Memo != "" {
+		metadata = data.Memo
+	} else {
+		metadata = data.Receiver
 	}
 
 	// parse out any forwarding info
-	parsedReceiver, err := types.ParseReceiverData(data.Receiver)
+	parsedPacketMetadata, err := types.ParsePacketMetadata(metadata)
```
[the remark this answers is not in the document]
sampocs: it's a very good question. I think it's a bit tough to dynamically check the flag for each module since the parsing is done once. A couple options I see:
* Add `AtLeastOneRouteActive` helper function to params and call this from module_ibc before any parsing. Only downside here is that we need to remember to modify this function every time we add a parameter.
* Check the `Active` params in the place where each module's metadata is validated. Downside here is that its kinda clunky to pass the params into the parser
* Leave as is. One v minor upside of the current setup is it helps us identify the case where we passed a packet with the memo fields filled out properly - yet the module was inactive

x/autopilot/types/params.go
asalzmann: It's just weird to have the defaults as false/true due to external circumstances

x/autopilot/module_ibc.go
```diff
@@ -123,51 +125,83 @@ func (im IBCModule) OnRecvPacket(
...
-			}, ack)
-		}
-		return ack
+	// ibc-go v5 has a Memo field that can store forwarding info
+	// For older version of ibc-go, the data must be stored in the receiver field
+	var metadata string
+	if data.Memo != "" { // ibc-go v5+
+		metadata = data.Memo
+	} else { // before ibc-go v5
+		metadata = data.Receiver
```
asalzmann, line 134: The common case will be where `data.Receiver` is just an address - thoughts on adding an explicit check for that case and short-circuiting the rest of the logic, if it's just an address? A bit redundant, but it protects standard IBC transfers from any of the autopilot logic interfering

sampocs: yeah good call. Something like:
```go
_, err := sdk.AccAddressFromBech32(data.Receiver)
if err == nil && data.Memo == "" {
   return im.app.OnRecPacket(...)
}
```
?

sampocs: done

x/autopilot/README.md
```diff
@@ -17,7 +17,8 @@ Note: This will support more functions that can reduce number of users' operatio
 ## Params
 
 ```
```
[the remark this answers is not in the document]
sampocs: done

x/autopilot/module_ibc.go
sampocs: done

x/autopilot/module_ibc.go
sampocs: done

x/autopilot/module_ibc.go
sampocs: done

x/autopilot/types/parser.go
sampocs: refactored

PR #690, opened 2023-03-24
Add Function to Update Airdrop Addresses

shellvish:

This adds a function to `claim` that allows the keeper to update an airdrop address. This will be used to match airdrops on zones that have different cointypes, e.g. Evmos.

### DIFF
--- a/x/claim/keeper/claim.go
+++ b/x/claim/keeper/claim.go
@@ -356,6 +356,15 @@ func (k Keeper) SetClaimRecord(ctx sdk.Context, claimRecord types.ClaimRecord) e
 	return nil
 }
 
+func (k Keeper) DeleteClaimRecord(ctx sdk.Context, addr sdk.AccAddress, airdropId string) error {
+	store := ctx.KVStore(k.storeKey)
+	prefixStore := prefix.NewStore(store, append([]byte(types.ClaimRecordsStorePrefix), []byte(airdropId)...))
+
+	prefixStore.Delete(addr)
+
+	return nil
+}
+
 // Get airdrop distributor address
 func (k Keeper) GetAirdropDistributor(ctx sdk.Context, airdropIdentifier string) (sdk.AccAddress, error) {
 	airdrop := k.GetAirdropByIdentifier(ctx, airdropIdentifier)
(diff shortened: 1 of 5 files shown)


x/claim/keeper/claim.go
```diff
@@ -803,3 +803,47 @@ func (k Keeper) DeleteAirdropAndEpoch(ctx sdk.Context, identifier string) error
...
+		return false
+	}
+
+	bz := prefixStore.Get(existingAccAddress)
+	claimRecord := types.ClaimRecord{}
+	err = proto.Unmarshal(bz, &claimRecord)
+	if err != nil {
+		k.Logger(ctx).Error(fmt.Sprintf("%s error unmarshalling claim record for address %s on airdrop %s", logPrefix, existingAddress, airdropId))
+		return false
+	}
```
sampocs, line 838: can you just use `k.GetClaimRecord` for this? We should probably also add a `DeleteClaimRecord` keeper function so that this function can just read

```go
oldClaimRecord, err := k.GetClaimRecord(ctx, existingStrideAddress, airdropId)
if err != nil {
   return errorsmod.Wrapf(err, "unable to find claim record for address %s", existingStrideAddress.String())
}

newClaimRecord := oldClaimRecord
newClaimRecord.Address := newStrideAddress
k.SetClaimRecord(ctx, newClaimRecord)

k.DeleteClaimRecord(ctx, oldClaimRecord)
```

sampocs (changes requested):
Thanks for doing this! Will review unit tests in a bit but dropping these nit boys for the time being

x/claim/keeper/claim_test.go
```diff
@@ -667,3 +669,128 @@ func (suite *KeeperTestSuite) TestGetClaimMetadata() {
...
+
+func (suite *KeeperTestSuite) TestUpdateAirdropAddress_AirdropNotFound() {
+	_, _, evmosParsedAddress, strideAddress := suite.SetupUpdateAirdropAddressChangeTests()
+
+	// update the address
+	success := suite.app.ClaimKeeper.UpdateAirdropAddress(suite.ctx, evmosParsedAddress, strideAddress, "stride")
+	suite.Require().False(success, "airdrop address update shouldn't work with incorrect airdrop id")
+}
+
+func (suite *KeeperTestSuite) TestUpdateAirdropAddress_StrideAddressIncorrect() {
```
riley-stride, line 789: Can we add a test for the case where airdrop exists, address is parseable, but address does not have a claim record in the airdrop

x/claim/keeper/claim_test.go
```diff
@@ -667,3 +669,128 @@ func (suite *KeeperTestSuite) TestGetClaimMetadata() {
...
+	coins, err = suite.app.ClaimKeeper.GetUserTotalClaimable(suite.ctx, evmosParsedAccAddress, airdropId, false)
+	suite.Require().NoError(err)
+	suite.Require().Equal(coins, sdk.NewCoins(sdk.NewCoin("stake", sdk.NewIntFromUint64(100_000_000))), "parsed evmos address should be allowed to claim")
+	claims, err = suite.app.ClaimKeeper.GetClaimStatus(suite.ctx, evmosParsedAccAddress)
+	suite.Require().NoError(err)
+	properClaims := []types.ClaimStatus{{AirdropIdentifier: airdropId, Claimed: false}}
+	suite.Require().Equal(claims, properClaims, "evmos address should have 1 claim record before update")
+
+	// update the address
+	success := suite.app.ClaimKeeper.UpdateAirdropAddress(suite.ctx, evmosParsedAddress, strideAddress, airdropId)
```
riley-stride, line 749: nit: would be slightly cleaner if UpdateAirdropAddress returned an error

riley-stride (changes requested):
1) Agree with Sam's requested changes
2) reviewed unit tests, looks good! Suggested we add one small one

x/claim/keeper/claim.go
```diff
@@ -803,3 +803,47 @@ func (k Keeper) DeleteAirdropAndEpoch(ctx sdk.Context, identifier string) error
...
+
+	// verify that the strideAddress is valid
+	_, err := sdk.AccAddressFromBech32(newStrideAddress)
+	if err != nil {
+		k.Logger(ctx).Error(fmt.Sprintf("%s invalid stride address %s", logPrefix, newStrideAddress))
+		return false
+	}
+
+	// note: existingAccAddress will be a STRIDE address with the same coin type as existingAddress
```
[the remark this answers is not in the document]
shellvish: For sure

x/claim/keeper/claim.go
shellvish: Agreed, changed!

x/claim/keeper/claim_test.go
shellvish: Agreed, changed!

x/claim/keeper/claim_test.go
shellvish: definitely, good call

PR #693, opened 2023-03-25
Add airdrop to autopilot

shellvish:
Adds "UpdateAirdropClaim" logic to autopilot


x/autopilot/keeper/airdrop.go
```diff
@@ -0,0 +1,46 @@
...
+	transfertypes "github.com/cosmos/ibc-go/v5/modules/apps/transfer/types"
+	channeltypes "github.com/cosmos/ibc-go/v5/modules/core/04-channel/types"
+
+	"github.com/Stride-Labs/stride/v7/utils"
+	"github.com/Stride-Labs/stride/v7/x/autopilot/types"
+)
+
+func (k Keeper) TryUpdateAirdropClaim(
+	ctx sdk.Context,
+	packet channeltypes.Packet,
```
sampocs, line 20: I don't think this packet arg is used - but I'm wondering if there's anything else we should check in the packet

wdyt?

x/claim/README.md
```diff
@@ -54,6 +54,21 @@ type ClaimRecord struct {
 
 ```
 
+## A Note on Address Mappings
+
+When an airdrop is created, we call `LoadAllocationData` to load the airdrop data from the allocations file.
```
sampocs, line 59: amazing docs, thanks for adding this!!

x/claim/README.md
```diff
@@ -54,6 +54,21 @@ type ClaimRecord struct {
...
+
+When an airdrop is created, we call `LoadAllocationData` to load the airdrop data from the allocations file. 
+This will apply `utils.ConvertAddressToStrideAddress` on each of those addresses, and then store those with the `ClaimRecords`.
+For an airdrop to, say, the Cosmos Hub, this will be the proper Stride address associated with that account. 
+`claim` state will only ever store this Stride address.
+
+However, for zones with a different coin type, _this will be an incorrect Stride address_. This should not cause any issues though,
+as this Stride address will be unusable.
+
+In order to claim that airdrop, the user will have to verify that they own the corresponding Evmos address. When the user tries to verify,
```
sampocs, line 67: Looking back on this - it might make more sense to key our addresses by `[]byte(addressString)` instead of `AccAddressFromBech32(addressString)` (they both return bytes) 

Definitely not necessary for this upgrade as it'd require a migration. But it'd be a nice future simplification to avoid the scenario of having some correct / some incorrect addresses.

x/claim/keeper/claim.go
```diff
@@ -803,3 +812,42 @@ func (k Keeper) DeleteAirdropAndEpoch(ctx sdk.Context, identifier string) error
...
+	// when new airdrops are ingested, we call utils.ConvertAddressToStrideAddress to convert
+	// the host zone (e.g. Evmos) address to a Stride address. The same conversion must be done
+	// if you're attempting to access a claim record for a non-Stride-address.
+	existingAccAddress, err := sdk.AccAddressFromBech32(existingStrideAddress)
+	if err != nil {
+		return errorsmod.Wrapf(types.ErrClaimNotFound, fmt.Sprintf("error getting claim record for address %s on airdrop %s", existingStrideAddress, airdropId))
+	}
+	claimRecord, err := k.GetClaimRecord(ctx, existingAccAddress, airdrop.AirdropIdentifier)
+	if (err != nil) || (claimRecord.Address == "") {
+		fmt.Printf("error getting claim record for address %s on airdrop %s", existingStrideAddress, airdropId)
```
sampocs, line 838: nit: rm

x/claim/keeper/claim.go
```diff
@@ -803,3 +812,42 @@ func (k Keeper) DeleteAirdropAndEpoch(ctx sdk.Context, identifier string) error
...
+		return errorsmod.Wrapf(types.ErrClaimNotFound, fmt.Sprintf("error getting claim record for address %s on airdrop %s", existingStrideAddress, airdropId))
+	}
+	claimRecord.Address = newStrideAddress
+	err = k.SetClaimRecord(ctx, claimRecord) // this does NOT delete the old record, because claims are indexed by address
+	if err != nil {
+		return errorsmod.Wrapf(types.ErrModifyingClaimRecord, fmt.Sprintf("error setting claim record from address %s to address %s on airdrop %s", existingStrideAddress, newStrideAddress, airdropId))
```
sampocs, line 844: nit, break line at `fmt.Sprintf`

sampocs (commented):
Beautiful PR - thanks for doing this!! 

Let's leave this as a separate PR until Aidan review the JSON PR and then we can merge it over and test!

x/autopilot/keeper/airdrop.go
```diff
@@ -0,0 +1,46 @@
...
+
+	// update the airdrop
+	airdropId := packetMetadata.AirdropId
+	k.Logger(ctx).Info(fmt.Sprintf("updating airdrop address %s (orig %s) to %s for airdrop %s",
+		senderStrideAddress, data.Sender, newStrideAddress, airdropId))
+	err := k.claimKeeper.UpdateAirdropAddress(ctx, senderStrideAddress, newStrideAddress, airdropId)
+	if err != nil {
+		k.Logger(ctx).Error(fmt.Sprintf("failed to update airdrop address: %v", err))
+	}
+
```
sampocs, line 44: just noticed the error case is never returned here. you can actually reduce all this code by consolidating it to:
```suggestion
	return k.claimKeeper.UpdateAirdropAddress(ctx, senderStrideAddress, newStrideAddress, airdropId)
```
(if an error is returned here, it will be logged from module_ibc)

x/claim/README.md
shellvish: Agreed!

x/claim/keeper/claim.go
shellvish: Ah good catch, thank you for removing!

PR #697, opened 2023-03-28
clean up gov spam PR

sontrinh16:

Add ante handler to prevent spam gov by adding minimal deposit amount when submit a proposal

## Brief Changelog

Add Gov prevent spam decorator to ante decorator

## Author's Checklist

I have...

- [x]  Run and PASSED locally all integration tests
- [x] Add new testcase for decorator logic

### DIFF
--- a/app/ante/gov_filter.go
+++ b/app/ante/gov_filter.go
@@ -1,0 +1,154 @@
+package ante
+
+import (
+	"github.com/cosmos/cosmos-sdk/codec"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+
+	errorsmod "cosmossdk.io/errors"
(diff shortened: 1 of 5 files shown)


app/ante/gov_filter.go
```diff
@@ -0,0 +1,154 @@
...
+
+	return next(ctx, tx, simulate)
+
+}
+
+func (gpsd GovPreventSpamDecorator) CheckSpamSubmitProposalMsg(ctx sdk.Context, msgs []sdk.Msg) error {
+
+	validMsg := func(m sdk.Msg) error {
+
+		switch msg := m.(type) {
```
vuong177, line 62: we should use if condition instead of switch case.

PR #699, opened 2023-03-29
Add Unit Test to Check Host Zone Min/Max RR

shellvish:

We are missing a unit test that checks the Min/Max RRs are set properly when a host zone is created.


sampocs (approved):
lgtm, thanks for adding these!

PR #706, opened 2023-04-06
[LSM] Scaffolding

sampocs:
Closes: #XXX

## Author's Checklist

I have...

## Reviewers Checklist

I have...

      - [ ] not applicable
      - [ ] jira ticket `XXX` 
      - [ ] README.md 
      - [ ] not documented


proto/stride/stakeibc/lsm.proto
```diff
@@ -0,0 +1,37 @@
...
+
+message LSMTokenDeposit {
+  string denom = 1;
+  string chain_id = 2;
+  string validator_address = 3;
+  string amount = 4 [
+    (gogoproto.customtype) = "github.com/cosmos/cosmos-sdk/types.Int",
+    (gogoproto.nullable) = false
+  ];
+  uint64 record_id = 5;
```
ethan-stride, line 27: If the denom is going to contain the unique ending to the NFT like "cosmosval42" why do we also need this field for the record_id "42".  Won't we always want to look up using the full denom and this would create additional logic to go back and forth from the denom to the record_id alone?

proto/stride/stakeibc/lsm.proto
```diff
@@ -0,0 +1,37 @@
...
+import "cosmos/base/v1beta1/coin.proto";
+import "gogoproto/gogo.proto";
+
+option go_package = "github.com/Stride-Labs/stride/v8/x/stakeibc/types";
+
+enum LSMDepositStatus {
+  option (gogoproto.goproto_enum_prefix) = false;
+
+  TRANSFER_IN_PROGRESS = 0;
+  TRANSFER_FAILED = 1;
```
ethan-stride, line 13: If transfer failed, it could be for temporary reasons so we would try again eventually or once it fails to transfer is it done forever?  If the former, what is the trigger for trying to resend all transfer failed tokens?  If we immediately retry then we don't need this state, if we retry at a predictable set time in the future, do we need to track how many times this token has failed to transfer and give up if the error is persistent?

proto/stride/stakeibc/validator.proto
```diff
@@ -16,7 +16,15 @@ message ValidatorExchangeRate {
...
-  string delegation_amt = 5 [
+  string balanced_delegation = 5 [
+    (gogoproto.customtype) = "github.com/cosmos/cosmos-sdk/types.Int",
+    (gogoproto.nullable) = false
+  ];
+  string unbalanced_delegation = 8 [
+    (gogoproto.customtype) = "github.com/cosmos/cosmos-sdk/types.Int",
+    (gogoproto.nullable) = false
+  ];
+  string cumulative_lsm_intake = 9 [
```
ethan-stride, line 27: This number only goes up.  But why do we need it?  Over time it could become larger than the total amount of ATOM which exists if people were to loop for some unlikely reason.  This is unrelated to any global limits on amount of LSM possible at any given time.

x/stakeibc/keeper/lsm_token_deposit.go
```diff
@@ -0,0 +1,51 @@
...
+
+func (k Keeper) SetLSMTokenDeposit(ctx sdk.Context, deposit types.LSMTokenDeposit) {
+	// TODO [LSM]
+}
+
+func (k Keeper) RemoveLSMTokenDeposit(ctx sdk.Context, denom string) {
+	// TODO [LSM]
+}
+
+func (k Keeper) GetLSMTokenDeposit(ctx sdk.Context, denom string) (deposit types.LSMTokenDeposit, found bool) {
```
ethan-stride, line 17: For Remove and Get don't we also need the validator information to arrive at a single specific deposit?  The prefix store will use chain_id then validator then denom as key prefixes, the CTX can tell us the chain_id but can it also tell us the validator or should this also be an input (or return multiple deposits across all validators which match this denom)?

x/stakeibc/keeper/lsm_token_deposit.go
```diff
@@ -0,0 +1,51 @@
...
+	// TODO [LSM]
+	return
+}
+
+func (k Keeper) GetAllLSMTokenDeposit(ctx sdk.Context) []types.LSMTokenDeposit {
+	// TODO [LSM]
+	return []types.LSMTokenDeposit{}
+}
+
+func (k Keeper) AddLSMTokenDeposit(ctx sdk.Context, deposit types.LSMDepositStatus) {
```
ethan-stride: Debatable if this should be a separate function, why not make this the default behavior for the SetLSMTokenDeposit?  You always want to call Add and never call Set right because a direct call to set in the case when there already is a deposit for this chain+val+denom will overwrite the existing with the second.

x/stakeibc/keeper/lsm_token_deposit.go
```diff
@@ -0,0 +1,51 @@
...
+	// See if a deposit already exists for this denom
+	// If so, increment the amount
+	// otherwise, create a new deposit
+}
+
+func (k Keeper) UpdateLSMTokenDepositStatus(ctx sdk.Context, deposit types.LSMDepositStatus, status types.LSMDepositStatus) {
+	// TODO [LSM]
+}
+
+func (k Keeper) GetLSMDepositsForHostZone(ctx sdk.Context, deposit types.LSMDepositStatus, chainId string) []types.LSMDepositStatus {
```
ethan-stride: Do we want a similar version of this function for validator as well to be used when rebalancing?

x/stakeibc/keeper/lsm_token_deposit.go
```diff
@@ -0,0 +1,51 @@
...
+}
+
+func (k Keeper) AddLSMTokenDeposit(ctx sdk.Context, deposit types.LSMDepositStatus) {
+	// TODO [LSM]
+	// See if a deposit already exists for this denom
+	// If so, increment the amount
+	// otherwise, create a new deposit
+}
+
+func (k Keeper) UpdateLSMTokenDepositStatus(ctx sdk.Context, deposit types.LSMDepositStatus, status types.LSMDepositStatus) {
```
ethan-stride: Wrong type on variable deposit for this and 2-3 other methods here.  IDE probably autocompleted LSMDepositStatus but any of the variables named deposit are supposed to be LSMTokenDeposit types.

ethan-stride (commented):
Overall makes sense, only actual error I saw was a few wrong types on params for the keeper functions.  Left a few questions on why certain fields or functions are separate as they seem to be repeating other existing fields or functions.

proto/stride/stakeibc/lsm.proto
sampocs: Very good question. At one point, there was an ICA we needed to send where we needed to specify the record ID in the transaction. It's possible that it's no longer necessary now that we're not hoarding though - I'll double check the spec!

proto/stride/stakeibc/lsm.proto
sampocs: There's no expected failure scenario that I'm aware of. So a failure would likely indicate that something unpredictable and very wrong happened - at which point we would just resolve it in an upgrade

proto/stride/stakeibc/validator.proto
sampocs: it's only there for the slashing safety mechanism. One other option I considered is we can reset it every time we get a response from the query. I'd be in favor of this approach, but we'd need a good variable name otherwise I could see it causing more confusion

maybe:
`progress_towards_exchange_rate_query`
?

proto/stride/stakeibc/lsm.proto
sampocs: Ah okay - just checked. It was previously used for the `withdrawRewards` transaction (when we were hoarding LSM tokens), but now that we're not, it's not needed.

Good catch! I'll remove

x/stakeibc/keeper/lsm_token_deposit.go
sampocs: done

x/stakeibc/keeper/lsm_token_deposit.go
sampocs: If I recall correctly, we never have to grab the deposits for a specific validator (but I could be wrong).

I know for sure during rebalance, we want to grab them for the whole host zone.

Let's hold off on adding a validator version for now and we can always add it in in the future if we find that a function would use it

x/stakeibc/keeper/lsm_token_deposit.go
sampocs: Good call out! I think we have two options here:

1. Take three parameters, `chainId, validatorAddress, denom`
2. Take an ID field (that represents the concatenation of the fields)

I don't think this function has a clear use case yet since we typically access all the records from a chain. So I don't have a strong opinion. Maybe we start with option 2 and we can adjust later if needed? Wdyt?

ethan-stride: As I have continued to think about it more, the denom contains information about the validator (and implicitly the chain_id) already.  You would never be able to have the same denom present on more than one validator.  Handing in a single field like denom would force and iteration over the prefix store whereas all 3 params would allow immediate lookup but this small efficiency different probably doesn't matter.  It is making me wonder if the validator address should be part of the prefix at all (if the prefix store was just chain_id and then denom, you can lookup a specific deposit and get the validator address from the deposit without it being a key in the prefix store)

proto/stride/stakeibc/validator.proto
ethan-stride: Sure sounds reasonable and makes it clearer where it will be used.

x/stakeibc/keeper/lsm_token_deposit.go
sampocs: that's a good point - I think we can remove the validator from the prefix. The only purpose before was when we had to look up all the deposits associated with a given validator, but that's also no longer relevant now that we're not hoarding

sampocs: chatted offline - agreed that validator field is not necessary, but we still need chainId for efficient lookups

PR #711, opened 2023-04-10
core lsm deposit keeper functions with unit tests

ethan-stride:
This change adds the core getter and setter functions for the lsm_token_deposits and unit tests

### DIFF
--- a/x/stakeibc/keeper/lsm_token_deposit_test.go
+++ b/x/stakeibc/keeper/lsm_token_deposit_test.go
@@ -1,0 +1,64 @@
+package keeper_test
+
+import (
+	"math/rand"
+	"strconv"
+	"testing"
+
+	sdkmath "cosmossdk.io/math"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/stretchr/testify/require"
+
+	keepertest "github.com/Stride-Labs/stride/v8/testutil/keeper"
+	"github.com/Stride-Labs/stride/v8/testutil/nullify"
+	"github.com/Stride-Labs/stride/v8/x/stakeibc/keeper"
+	"github.com/Stride-Labs/stride/v8/x/stakeibc/types"
+)
+
+func createSetNLSMTokenDeposit(keeper *keeper.Keeper, ctx sdk.Context, n int) []types.LSMTokenDeposit {
+	items := make([]types.LSMTokenDeposit, n)
+	for i := range items {
+		validatorAddr := "validatorAddress"
+		tokenRecordId := rand.Int()
+
+		items[i].Denom = validatorAddr + strconv.Itoa(tokenRecordId)
+		items[i].ValidatorAddress = validatorAddr
+		items[i].ChainId = strconv.Itoa(i)
+		items[i].Amount = sdkmath.ZeroInt()
+		items[i].Status = types.TRANSFER_IN_PROGRESS
+		keeper.SetLSMTokenDeposit(ctx, items[i])
+	}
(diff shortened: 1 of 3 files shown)


x/stakeibc/keeper/lsm_token_deposit_test.go
```diff
@@ -0,0 +1,64 @@
...
+		_, found := keeper.GetLSMTokenDeposit(ctx, item.ChainId, item.Denom)
+		require.False(t, found)
+	}
+}
+
+func TestLSMTokenDepositGetAll(t *testing.T) {
+	keeper, ctx := keepertest.StakeibcKeeper(t)
+	items := createSetNLSMTokenDeposit(keeper, ctx, 10)
+	require.ElementsMatch(t,
+		nullify.Fill(items),
```
sampocs, line 61: remove nullify and add a test case for the length
```go
allDeposits := GetAllLSMTokenDeposit(s.Ctx)
s.Require().Len(len(allDeposits), 10)
s.Require().ElementsMatch(items, allDeposits)
```

x/stakeibc/keeper/lsm_token_deposit_test.go
```diff
@@ -0,0 +1,64 @@
...
+		items[i].Amount = sdkmath.ZeroInt()
+		items[i].Status = types.TRANSFER_IN_PROGRESS
+		keeper.SetLSMTokenDeposit(ctx, items[i])
+	}
+	return items
+}
+
+func TestLSMTokenDepositGet(t *testing.T) {
+	keeper, ctx := keepertest.StakeibcKeeper(t)
+	items := createSetNLSMTokenDeposit(keeper, ctx, 10)
```
sampocs, line 36: superrrr nit here and no need to change in this PR since these tests are so straightforward, but the `items/got` naming is some of the generic ignite variable naming.
In future PRs (especially with more complex tests), I would change it to `expected/actual` like:

```go
expectedDeposits := createSetNLSMTokenDeposit(keeper, ctx, 10)
for _, expectedDeposit := range expectedDeposits {
   actualDeposit, found := keeper.GetLSMTokenDeposit(ctx, expectedDeposit.ChainId, expectedDeposit.Denom)
   ...
   s.Require().Equal(expectedDeposit, actualDeposit)
}
```

sampocs (changes requested):
Really great work here Ethan! Code is squeaky clean and looks functionally correct!

Left a few nit comments that's mostly tribal knowledge that I wouldn't have expected you to know (e.g. moving us away from the ignite generated testing framework).I figured it's good to start moving us in more of the direction of standardized best practices instead of how stakeibc was setup!

sampocs (approved):
Changes look great!

Only small comment is that when adding the description in the require statements, there's no need to explicitly print the values that are being compared.

For instance, 
```go
s.Require().Equal(expected, actual, "found deposit %+v did not match expected %+v", actual, expected)
```
Is a bit redundant because `expected` and `actual` will be printed automatically it fails, so there's no need for `"%+v did not match expected %+v"`

Instead the description should just help identify what records are being compared (this is only really necessary when you're looping through a list and it's unclear which index the failure occurred on).

So instead you could do:
```go
s.Require().Equal(expected, actual, "deposit records did not match for %s", expected.Denom)
// or
s.Require().Equal(expected, actual, "deposit records did not match for item %d",  i)
```

(Feel free to try to break one of the tests on your own to see what the error output looks like - which might make this a little more clear)

Once this change is addressed, go ahead and merge it! (no need to request a re-review again)

PR #714, opened 2023-04-12
[LSM] Implement RefundLSMToken

sampocs:
Implemented helper function to refund an LSM token to a user in the event that the slash query failed

### DIFF
--- a/x/stakeibc/keeper/lsm.go
+++ b/x/stakeibc/keeper/lsm.go
@@ -108,3 +108,22 @@ func (k Keeper) ShouldQueryValidatorExchangeRate(ctx sdk.Context, validator type
 	// Submit query if the query interval checkpoint has been breached
 	return oldProgress.Quo(queryInterval).LT(newProgress.Quo(queryInterval))
 }
+
+// Helper function to Refund and LSM Token to the user if both:
+//  a) the LSMLiquidStake transaction is interrupted with a validator exchange rate query, and
+//  b) the handling of the query callback fails
+func (k Keeper) RefundLSMToken(ctx sdk.Context, lsmLiquidStake types.LSMLiquidStake) error {
+	liquidStakerAddress := lsmLiquidStake.Staker
+	hostZoneAddress, err := sdk.AccAddressFromBech32(lsmLiquidStake.HostZone.Address)
+	if err != nil {
+		return errorsmod.Wrapf(err, "host zone address is invalid")
+	}
+	refundToken := lsmLiquidStake.LSMIBCToken
+
+	if err := k.bankKeeper.SendCoins(ctx, hostZoneAddress, liquidStakerAddress, sdk.NewCoins(refundToken)); err != nil {
+		return errorsmod.Wrapf(err,
+			"failed to send tokens from Module Account (%s) to Staker (%s)", hostZoneAddress.String(), liquidStakerAddress.String())
+	}
+
+	return nil
+}
--- a/x/stakeibc/keeper/lsm.go
+++ b/x/stakeibc/keeper/lsm.go
@@ -109,7 +109,7 @@ func (k Keeper) ShouldQueryValidatorExchangeRate(ctx sdk.Context, validator type
 	return oldProgress.Quo(queryInterval).LT(newProgress.Quo(queryInterval))
 }
 
-// Helper function to Refund and LSM Token to the user if both:
+// Helper function to Refund an LSM Token to the user if both:
 //  a) the LSMLiquidStake transaction is interrupted with a validator exchange rate query, and
 //  b) the handling of the query callback fails
 func (k Keeper) RefundLSMToken(ctx sdk.Context, lsmLiquidStake types.LSMLiquidStake) error {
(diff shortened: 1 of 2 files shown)


x/stakeibc/keeper/lsm.go
```diff
@@ -108,3 +108,22 @@ func (k Keeper) ShouldQueryValidatorExchangeRate(ctx sdk.Context, validator type
 	// Submit query if the query interval checkpoint has been breached
 	return oldProgress.Quo(queryInterval).LT(newProgress.Quo(queryInterval))
 }
+
+// Helper function to Refund and LSM Token to the user if both:
```
ethan-stride, line 112: typo: and -> an

x/stakeibc/keeper/lsm_test.go
```diff
@@ -169,3 +169,42 @@ func (s *KeeperTestSuite) TestShouldQueryValidatorExchangeRate() {
...
+
+	stakerBalance := s.App.BankKeeper.GetBalance(s.Ctx, liquidStakerAddress, lsmTokenIBCDenom)
+	s.CompareCoins(lsmToken, stakerBalance, "staker should have received their LSM token back")
+
+	moduleBalance := s.App.BankKeeper.GetBalance(s.Ctx, hostZoneAddress, lsmTokenIBCDenom)
+	s.True(moduleBalance.IsZero(), "module account should no longer have the LSM token")
+
+	// Attempt to refund again, it should fail from an insufficient balance
+	err = s.App.StakeibcKeeper.RefundLSMToken(s.Ctx, liquidStake)
+	s.Require().ErrorContains(err, "insufficient funds")
```
ethan-stride, line 204: Should we do this with different error types instead of string checking?

ethan-stride (commented):
lgtm except for small typo

x/stakeibc/keeper/lsm.go
sampocs: done

x/stakeibc/keeper/lsm_test.go
sampocs: Good question! I've found using the string approach is a little easier to manage cause once you wrap an error, you can no longer check the type

For instance, if you return 
```go
errorsmod.Wrapf(types.SomeErrorType, "some additional context")
```
then you _cannot_ do the following
```go
s.Require().ErrorIs(types.SomeErrorType, err) 
```

And we wrap most of our errors. Also, there's often a whole stack trace for the error, so it's nice to just check for the part that relevant to the test case

x/stakeibc/keeper/lsm.go
```diff
@@ -108,3 +108,22 @@ func (k Keeper) ShouldQueryValidatorExchangeRate(ctx sdk.Context, validator type
...
+//  b) the handling of the query callback fails
+func (k Keeper) RefundLSMToken(ctx sdk.Context, lsmLiquidStake types.LSMLiquidStake) error {
+	liquidStakerAddress := lsmLiquidStake.Staker
+	hostZoneAddress, err := sdk.AccAddressFromBech32(lsmLiquidStake.HostZone.Address)
+	if err != nil {
+		return errorsmod.Wrapf(err, "host zone address is invalid")
+	}
+	refundToken := lsmLiquidStake.LSMIBCToken
+
+	if err := k.bankKeeper.SendCoins(ctx, hostZoneAddress, liquidStakerAddress, sdk.NewCoins(refundToken)); err != nil {
```
riley-stride, line 123: Sanity checking my understanding: given the record on the tokens the user deposited is in state PENDING, they will not be ibc-transfered to the host at the top of the epoch, so we can be sure that they will be sitting in the x/stakeibc module account, right?

riley-stride:
Lgtm

One thing to flag to you to keep in mind as we design this. Might not make sense to design around it now, we could come back later: what about the case where the user is LSMliquidstake'ing using x/autopilot? In that case, if the query fails and a refund needs to be issued, this refund logic would send the tokens to the module account that is executing middleware, but they would not go back to the user on the host chain afaiu. 

We could prob handle that case by ibc-transfering failed-and-refunded tokens from module account during the endBlocker, or we could ibc-transfer in the refund logic here, but that'd require storing down the user's host zone address (maybe augmenting types.LSMLiquidStake)

Just something to keep in mind as we design this

riley-stride (approved):

x/stakeibc/keeper/lsm.go
sampocs: close! The transfer actually happens asap (instead of epochly), but this refund is only called in the case where we have the async-query-interrupted LSM Liquid Stake - and in that case the transfer is not initiated until after the query is returned

sampocs:
Really good point wrt autopillot. How important do you think it is that we consider autopilot when going through this LSM design? 

My gut says, we shouldn't worry about autopilot until after this is deployed and that the incremental adjustments we'd need to enable it wouldn't be that dramatic

x/stakeibc/keeper/lsm.go
ethan-stride: Transfers can fail as well which is why we have that state in the LSMDepositStatus.  You had said before that it is unlikely to transiently fail so once it fails tha particular token would likely fail again forever... would we also want to send the token back to the original holder in the case that we get a transfer fail or will we hold on to it forever?

sampocs: Sorry just seeing this. I don't think we'll want to refund that token since we've already sent the user stTokens at that point

PR #715, opened 2023-04-12
[LSM] ICA DetokenizeShares and Callback

sampocs:
## Context
After every LSM liquid stake, the LSM Token must be "detokenized" on the host zone - meaning it is converted back into native stake. 

This PR implements the function that triggers the ICA (`DetokenizeLSMDeposit`) and the corresponding callback (`DetokenizeSharesCallback`)

When everything is wired together at the end, each detokenization ICA will get "queued" after the LSM Token is transferred to the host. Then the EndBlocker will pick up the token from the queue, and submit the ICA using `DetokenizeLSMDeposit`. The EndBlocker will be implemented in a separate PR.

## Brief Changelog
* Added LSM's `RedeemTokensFromShares` proto since the SDK versions will not be the same between Stride and LSM. This proto type is required for the ICA.
* Updated proto scripts to generate this proto
* Implemented `DetokenizeLSMDeposit` and `DetokenizeSharesCallback`, and added corresponding unit tests
* **Note to reviewer**: I implemented the callback tests using two testing frameworks. Please provide your opinion as to which you think is the easiest to follow (my vote is the array approach for this particular example)


x/stakeibc/keeper/lsm.go
```diff
@@ -1,12 +1,55 @@
...
 	"github.com/Stride-Labs/stride/v8/x/stakeibc/types"
 )
 
-func (k Keeper) TransferLSMTokenToHost(ctx sdk.Context, lsmTokenDeposit types.LSMTokenDeposit) error {
-	// TODO [LSM]
+// Submits an ICA to "Redeem" an LSM Token - meaning converting the token into native stake
+// This function is called in the EndBlocker which means if the ICA submission fails,
+//   any modified state is not reverted
+// The deposit Status is intentionally updated before the ICA is submitted even though it will NOT be reverted
+//   if the ICA fails to send. This is because a failure is likely caused by a closed ICA channel, and the status
```
riley-stride, line 18: From the logic in `DetokenizeCallback`, it seems like the case you're designing around here is timeout: if timeout, leave the status as `DETOKENIZATION_IN_PROGRESS`. 

Can timeouts only occur if the channel closes / is closed, since it's an ordered channel?

In that case, you say `When the ICA channel is restored, the deposit status will get reset, and the ICA will be attempted again.`. Where's the logic to reset the status when the channel re-opens?

scripts/protocgen.sh
```diff
@@ -8,16 +8,23 @@ go get github.com/regen-network/cosmos-proto/protoc-gen-gocosmos 2>/dev/null
 # get cosmos sdk from github
 # go get github.com/cosmos/cosmos-sdk@v0.45.11 2>/dev/null
 
+generate_protos() {
```
riley-stride, line 11: Was regular proto generation not working?

x/stakeibc/keeper/icacallbacks_detokenize_test.go
```diff
@@ -0,0 +1,175 @@
...
+	deposit, found := s.App.StakeibcKeeper.GetLSMTokenDeposit(s.Ctx, HostChainId, LSMTokenBaseDenom)
+	s.Require().True(found, "deposit should not have been removed")
+	s.Require().Equal(types.DETOKENIZATION_FAILED.String(), deposit.Status.String(), "deposit status")
+}
+
+/////// NOTE TO REVIEWER ////////
+// I'm demonstrating two testing approaches here. This bottom test is redundant with the tests above
+// Let me know which test approach you prefer!
+
+func (s *KeeperTestSuite) TestDetokenizeCallback() {
```
riley-stride, line 96: I agree with you, these tests are cleaner to read. They take a minute to get your head into, though. If we're gonna use these for "less complex" tests, would be nice to add a one line comment to each case like is done in the approach in some of the tests above (e.g. "Call the callback with a successful ack, but invalid callback args"), I find that very helpful to parse the tests and make sure there's good coverage.

riley-stride (commented):
One question on the timeout and status `DETOKENIZATION_IN_PROGRESS`

x/stakeibc/keeper/lsm.go
sampocs: > Can timeouts only occur if the channel closes / is closed, since it's an ordered channel?
its more that IF a timeout occurs, then the channel gets closed automatically

So an example is:
* ICA #1 is submitted, channel gets closed
* ICA #2 attempted immediately after, but SubmitTxs fails since the channel is closed

Since this is all triggered from the begin blocker, if we revert the record right away, the begin blocker will try to submit it again in the next block. But I think we actually just want to ignore the record until the channel is restored
Fwiw, there's also a clause in the being blocker that only submits this ICA if the channel is open, so I'm actually not even sure this case is even relevant. But in any event, it felt like it made more sense to update the status rather than infinitely retry

> Where's the logic to reset the status when the channel re-opens?
And there will be a restore-channel transaction that reverts all IN_PROGRESS records to QUEUE

scripts/protocgen.sh
sampocs: I just moved the proto script to it's own function. It was called on just the stride directory previously, but now I had to add the cosmos directory. So rather than doing
```bash
  proto_dirs=$(find ./stride -path -prune -o -name '*.proto' -print0 | xargs -0 -n1 dirname | sort | uniq)
  for dir in $proto_dirs; do
    for file in $(find "${dir}" -maxdepth 1 -name '*.proto'); do
      if grep go_package "$file" &>/dev/null; then
        buf generate --template buf.gen.gogo.yaml "$file"
      fi
    done
  done

  proto_dirs=$(find ./cosmos -path -prune -o -name '*.proto' -print0 | xargs -0 -n1 dirname | sort | uniq)
  for dir in $proto_dirs; do
    for file in $(find "${dir}" -maxdepth 1 -name '*.proto'); do
      if grep go_package "$file" &>/dev/null; then
        buf generate --template buf.gen.gogo.yaml "$file"
      fi
    done
  done
```

I just did 
```bash
generate_protos() {
  package="$1"
  proto_dirs=$(find $package -path -prune -o -name '*.proto' -print0 | xargs -0 -n1 dirname | sort | uniq)
  for dir in $proto_dirs; do
    for file in $(find "${dir}" -maxdepth 1 -name '*.proto'); do
      if grep go_package "$file" &>/dev/null; then
        buf generate --template buf.gen.gogo.yaml "$file"
      fi
    done
  done
}

generate_protos "./stride"
generate_protos "./cosmos"
```

sampocs: this was because I had to add the Detokenize MsgType from LSM

x/stakeibc/keeper/icacallbacks_detokenize_test.go
sampocs: that's a great idea. 

But yeah i think there's definitely a trade off here - gonna message you offline to double click on how you felt it was tougher to get your head into it (definitely something we wanna avoid!)

x/stakeibc/keeper/icacallbacks_detokenize.go
```diff
@@ -21,7 +27,28 @@ func DetokenizeCallback(k Keeper, ctx sdk.Context, packet channeltypes.Packet, a
...
+			icacallbackstypes.AckResponseStatus_TIMEOUT, packet))
+		return nil
+	}
+
+	// If the iCA failed, update the deposit status
+	if ackResponse.Status == icacallbackstypes.AckResponseStatus_FAILURE {
+		k.Logger(ctx).Error(utils.LogICACallbackStatusWithHostZone(chainId, ICACallbackID_Detokenize,
+			icacallbackstypes.AckResponseStatus_FAILURE, packet))
+
+		k.UpdateLSMTokenDepositStatus(ctx, *detokenizeCallback.Deposit, types.DETOKENIZATION_FAILED)
```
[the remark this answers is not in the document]
ethan-stride: I see an example below where the Deposit field for this type wants the dereferenced form, so I think I understand what is going on now although I would have thought that it would require parentheses.  Nevermind about the original comment

x/stakeibc/keeper/lsm.go
ethan-stride: You said a restore-channel transaction would revert all IN_PROGRESS to QUEUE but is that known to be safe? Are there guarantees on how long the detokenization will take so some from previous blocks which really are being detokenized can't still be IN_PROGRESS at the time when a channel gets closed?  It seems like you want the ones which were just moved to IN_PROGRESS only and not all which are that status unless these can be guaranteed to be the same thing.

x/stakeibc/keeper/lsm.go
```diff
@@ -1,12 +1,55 @@
...
+
+	errorsmod "cosmossdk.io/errors"
 	sdk "github.com/cosmos/cosmos-sdk/types"
 
+	"github.com/golang/protobuf/proto" //nolint:staticcheck
+
 	"github.com/Stride-Labs/stride/v8/x/stakeibc/types"
 )
 
-func (k Keeper) TransferLSMTokenToHost(ctx sdk.Context, lsmTokenDeposit types.LSMTokenDeposit) error {
```
ethan-stride, line 9: What happened to this function?  Where is the logic which checks for TRANSFER_IN_PROGRESS or TRANFER_FAILED states?  Is that just part of a different PR and not relevant here?

x/stakeibc/keeper/lsm_test.go
```diff
@@ -0,0 +1,73 @@
...
+	s.Require().NoError(err, "no error expected when unmarshalling callback data")
+
+	s.Require().Equal(initalDeposit, *callbackData.Deposit, "callback data LSM deposit")
+
+	// Remove connection ID and re-submit - should fail
+	hostZone.ConnectionId = ""
+	err = s.App.StakeibcKeeper.DetokenizeLSMDeposit(s.Ctx, hostZone, initalDeposit)
+	s.Require().ErrorContains(err, "unable to submit detokenization ICA")
+
+	// Remove delegation account and re-submit - should also fail
```
ethan-stride, line 69: You should reset the ConnectionId to the real value for this test to be independent.  It could fail for the same reason as the previous test

ethan-stride (commented):
Overall makes sense.  I have a few questions about the callback types and how they get Marshal-ed I can ask you offline.

scripts/protocgen.sh
riley-stride: Cool that makes sense

x/stakeibc/keeper/lsm.go
sampocs: Sorry I think I confused myself a bit between the spec and the ticket definitions (didn't realize this was assigned to you). 

Before I thought we would have separate Mint and Transfer functions but then I realized the mint code is actually pretty straight forward so I thought i made more sense to just bundle those together into one function.

I wound up putting this code in the `FinishLSMLiquidStake` function [[here](https://github.com/Stride-Labs/stride/pull/713/files#diff-e5e494420dbb9b9ea933fbe70b5bb22b5af4c72a53940963a375dc32f8c754bfR195-R228)]. 

You're still tasked with the callback though!

x/stakeibc/keeper/lsm_test.go
sampocs: the tests were intentionally set up this way because the delegation check is upstream of the connection check. I'll add a clarifying comment

sampocs: actually, you're probably right, it's probably a bit safer to reset the connection ID in case the original function gets re-ordered

x/stakeibc/keeper/icacallbacks_detokenize.go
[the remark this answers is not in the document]
ethan-stride: Yes I understood that part of it in terms of the mechanics because it is similar to C, but I wasn't getting at first that the struct is defined with a pointer (in fact this is something I still want to ask about because a normal Marshal would save the pointer instead of the data which I assume is not useful) but I was confused as much about the precedence and was reading it wrong.  The dereference applies to the whole statement, ie. the dot for a subfield takes precedence over the * for dereferencing.  That's why I was thinking it would need parenthesis but it forced me to go read up a bit on go... (I think it also works this way in C, I just hadn't done it in a few years).  But thank you for carefully explaining all the same.

x/stakeibc/keeper/icacallbacks_detokenize_test.go
ethan-stride: I think we are in agreement on this as well as the slack casual poll so I am resolving this conversation

x/stakeibc/keeper/lsm.go
ethan-stride: You are saying a completely new channel needs to be made and any/all in flight requests would need to be resent... how does this work on the hubside?  So let;s say there is a detokenize in process so it was recieved hubside and it is doing it, but the channel closes.  Will the detokenize fail because it is unable to callback?

x/stakeibc/keeper/lsm_test.go
ethan-stride: Yeah I knew that was how it was implemented otherwise when you ran this test it would give you the wrong comment and the unit test would fail, but you are saying you did this on purpose as part of the test design?  It seems pretty independent to me, ie. it can fail due to bad connectionId or due to bad delegation account address but there isn't intended to be interaction effects?

sampocs: there's no interaction effects. both of those checks are essentially independent lines of code and (in this case in particular) they are don't commingle.

I cleaned this up though in one of the later commits

x/stakeibc/keeper/lsm.go
sampocs: when the channel closes, it closes on both ends. The important part here is the ordering. If you send out 5 packets back to back, the last 4 cannot be submitted to the hub until the first packet was either an ack success or ack failure.
And if packet 1 times out, then the channel closes on both the hub and Stride and packets 2-5 can never be submitted to the hub

PR #717, opened 2023-04-13
[LSM] Renamed ICA Accounts in Host Zone

sampocs:
The host zone ICA addresses currently use an unnecessary ICAAccount struct. This PR replaces the struct with just the address.

**I'd recommend going commit by commit when reviewing**


riley-stride (approved):
Great refactor, minus 500 LOC... LFG!

PR #719, opened 2023-04-13
Ethan/lsm deposit queries

ethan-stride:
Closes: third ticket to Implement the LSMDeposits queries

This change adds the queries so either through http GET or CLI interfaces the LSMTokenDeposit state in the keeper can be requested and shown given different filters.

Note in the protos there are optional fields for denom and status but chain_id always must be present.
There are 3 different http GET url formats specified due to the different fields which might be present
Similarly there are 3 different CLI interfaces.

There is only one grpc function implementation because it is able to do field checking as part of its operation.

## Brief Changelog

  - *Created the query and rpc protos which generate the types*
  - *implemented the grpc call in the keeper functions*
  - *Added the CLI interfaces* 

## Author's Checklist

I have...

- Run and PASSED locally all GAIA integration tests
- Verified that everything builds and all previous unit tests still pass

I am currently looking into how to best unit test the grpc function and the CLI commands.
We generally do have unit tests for the grpc commands but only on one or two of the CLI commands yet.
This is ready for review but feel free to require that I write grpc unit tests before this ticket closes.

## Reviewers Checklist

I have...

      - [ ] not applicable
      - [ ] jira ticket `XXX` 
      - [ ] README.md 
      - [ ] not documented


x/stakeibc/client/cli/query_lsm_deposits.go
```diff
@@ -0,0 +1,107 @@
...
+	"errors"
+
+	"github.com/cosmos/cosmos-sdk/client"
+	"github.com/cosmos/cosmos-sdk/client/flags"
+	"github.com/spf13/cobra"
+
+	"github.com/Stride-Labs/stride/v8/x/stakeibc/types"
+)
+
+func CmdLSMDepositsHostZone() *cobra.Command {
```
sampocs, line 14: I think we should just have one command with optional flags. So the command is just `lsm-deposits` which defaults to all deposits. And then there's optional `--host-zone`, `--denom`, and `--status` flags

Here's an example of how to add the flags [[this](https://github.com/Stride-Labs/stride/pull/619/files#diff-0087d5d0ca1a2154aadad78b1582fa52f24d66f8583a0e08ba76c63be6d4adadR81) function, [flag](https://github.com/Stride-Labs/stride/pull/619/files#diff-0087d5d0ca1a2154aadad78b1582fa52f24d66f8583a0e08ba76c63be6d4adadR133) and [flag](https://github.com/Stride-Labs/stride/pull/619/files#diff-0087d5d0ca1a2154aadad78b1582fa52f24d66f8583a0e08ba76c63be6d4adadR95)]

x/stakeibc/keeper/grpc_query_lsm_deposits.go
```diff
@@ -0,0 +1,53 @@
...
+	if req.GetDenom() != "" {
+		deposit, found := k.GetLSMTokenDeposit(ctx, req.GetChainId(), req.GetDenom())
+		if found {
+			allDeposits = append(allDeposits, deposit)
+		}
+	}
+
+	// Case 2: They are asking for all deposits with chainId + status --> Get
+	//   if they included a status and denom, already handled above, denom is missing for this branch
+	if req.GetXStatus() != nil && req.GetDenom() == "" {
```
sampocs, line 35: is the `X` notation how go + proto handles optionals?

proto/stride/stakeibc/query.proto
```diff
@@ -147,3 +162,13 @@ message QueryAddressUnbondingsResponse {
   repeated AddressUnbonding address_unbondings = 1
       [ (gogoproto.nullable) = false ];
 }
+
+message QueryLSMDepositsRequest {
+  string chain_id = 1;
```
sampocs, line 167: why is chain id mandatory?

sampocs (commented):
Love the initiative here going with a new design! Left a couple immediate comments but lets chat on this live

sampocs (commented):
Sorry for the misdirection here, but now that I'm looking the implementation, I see that it is a bit odd to have a denom filter when there's only one record per denom.

I think it might make sense to instead support 2 queries:
1. `LSMDeposit`:  takes a chain ID and denom, and returns a specific record
3. `AllLSMDeposits`: returns all deposits but with chainId, _validator_, and status filters

proto/stride/stakeibc/query.proto
sampocs: I think I'd actually recommend just setting all these to strings. When someone wants to filter by status, it's an easier UX to pass the string than pass the status int.

And then there's no need to make these optional cause you can just check if the string is empty

x/stakeibc/keeper/grpc_query_lsm_deposits.go
ethan-stride: Yes, inside the generated pb file from the protos a field which has optional has X versions of each field, these X versions do a bunch of checking around nulls etc.  The reason I have to use the XStatus != nil check is because Status is an enum, the default value of the enum corresponds to the first state (0 int32 under the hood) which is a valid Status and not what we want.  So if they put in nothing for Status by the time we cast it to the actual LSMDepositStatus type it now looks like the first state (in the proto language manual they actually recommend you make the first state of an enum explicitly a catchall value like STATUS_UNDEFINED because of this https://protobuf.dev/programming-guides/proto3/#enum)

sampocs: ah I'm glad you mentioned that. I think we should adapt that as well so the 0-value of our enum's isn't just the first status!

proto/stride/stakeibc/query.proto
ethan-stride: Yep as we discussed, avoid the optionals by making all strings and allow chain_id to be empty as well if you want to get all deposts.

ethan-stride:
image

ethan-stride:
Just updated this design to follow the new suggested CLI and http formats, take a look and if the design looks good I will then write the unit tests for it.  I was able to test locally both the CLI and the localhost:1317 endpoints

x/stakeibc/keeper/grpc_query_lsm_deposits.go
```diff
@@ -0,0 +1,72 @@
...
+	sdk "github.com/cosmos/cosmos-sdk/types"
+
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+
+	"github.com/Stride-Labs/stride/v8/x/stakeibc/types"
+)
+
+func (k Keeper) LSMDeposit(c context.Context, req *types.QueryLSMDepositRequest) (*types.QueryLSMDepositResponse, error) {
+	if req == nil || req.GetChainId() == "" || req.GetDenom() == "" {
```
sampocs, line 15: Not a big deal at all, but we prefer accessing attributes of the protos directly (`req.ChainId`) instead of using the getters (`req.GetChainId()`) especially since you already check that `req` is not nil. 

We had a mix of both throughout our codebase, and decided we should start standardizing it, and the attribute approach is what we've seen throughout other projects and seems to be more of the standard in go

sampocs (commented):
Changes look great! Couple nit comments, but go ahead and move forward with the unit tests!

x/stakeibc/keeper/grpc_query_lsm_deposits.go
ethan-stride: On this I have to respectfully disagree.  I do agree we should be consistent but the Getter functions work in some cases where you can't do the direct property version.  In this exact situation it would be the same, but in the case of optional fields, the whole Xvalues stuff happening under the hood exposes the same GetValue() style getters.  If we want to be consistent we should do it that way.  Unrealted, it also makes it easier to mock that function in some places where you don't want to setup a ton of fake data for tests but you want to affect everything which gets returned for a certain case...

proto/stride/stakeibc/query.proto
```diff
@@ -147,3 +163,22 @@ message QueryAddressUnbondingsResponse {
   repeated AddressUnbonding address_unbondings = 1
       [ (gogoproto.nullable) = false ];
 }
+
+message QueryLSMDepositRequest {
+  string chain_id = 1;
+  string denom = 3;
```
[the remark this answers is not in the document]
ethan-stride: done

x/stakeibc/client/cli/query_lsm_deposits.go
```diff
@@ -0,0 +1,86 @@
...
+
+	"github.com/cosmos/cosmos-sdk/client"
+	"github.com/cosmos/cosmos-sdk/client/flags"
+	"github.com/spf13/cobra"
+
+	"github.com/Stride-Labs/stride/v8/x/stakeibc/types"
+)
+
+const (
+	FlagChainId          = "host-zone"
```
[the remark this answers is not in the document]
ethan-stride: done

x/stakeibc/client/cli/query_lsm_deposits.go
```diff
@@ -0,0 +1,86 @@
...
+
+	flags.AddQueryFlagsToCmd(cmd)
+
+	return cmd
+}
+
+func CmdLSMDeposits() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "show-lsm-deposits --host-zone=[chain_id] --validator=[validator_address] --status=[status]",
+		Short: "shows all lsm-deposits filtered by optional flags chain-id validate-address and status",
```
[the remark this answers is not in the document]
ethan-stride: done

x/stakeibc/client/cli/query_lsm_deposits.go
```diff
@@ -0,0 +1,86 @@
...
+	}
+
+	flags.AddQueryFlagsToCmd(cmd)
+
+	return cmd
+}
+
+func CmdLSMDeposits() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "show-lsm-deposits --host-zone=[chain_id] --validator=[validator_address] --status=[status]",
```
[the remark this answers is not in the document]
ethan-stride: done

x/stakeibc/client/cli/query_lsm_deposits.go
```diff
@@ -0,0 +1,86 @@
...
+
+const (
+	FlagChainId          = "host-zone"
+	FlagValidatorAddress = "validator"
+	FlagStatus           = "status"
+)
+
+func CmdLSMDeposit() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "show-lsm-deposit [chain-id] [denom]",
```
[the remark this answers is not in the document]
ethan-stride: done

x/stakeibc/client/cli/query_lsm_deposits.go
```diff
@@ -0,0 +1,86 @@
...
+		Short: "shows either 0 or 1 deposits which match given denom on chain-id",
```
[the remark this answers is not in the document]
ethan-stride: done

x/stakeibc/keeper/grpc_query_lsm_deposits.go
```diff
@@ -0,0 +1,72 @@
...
+
+	var foundDeposit types.LSMTokenDeposit
+	ctx := sdk.UnwrapSDKContext(c)
+
+	deposit, found := k.GetLSMTokenDeposit(ctx, req.GetChainId(), req.GetDenom())
+	if found {
+		foundDeposit = deposit
+	}
+
+	return &types.QueryLSMDepositResponse{Deposit: foundDeposit}, nil
```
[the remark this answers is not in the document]
ethan-stride: done... except if it is not found we should return a nil?  Deposit itself will be a nil unless you want the outer type to not be nil and only the inner deposit?

x/stakeibc/keeper/grpc_query_lsm_deposits.go
```diff
@@ -0,0 +1,72 @@
...
+	// Filter for matches by hand if status is given
+	filtered = []types.LSMTokenDeposit{}
+	if req.GetStatus() != "" {
+		for _, deposit := range deposits {
+			if deposit.Status.String() == req.GetStatus() {
+				filtered = append(filtered, deposit)
+			}
+		}
+		deposits = filtered
+	}
```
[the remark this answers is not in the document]
ethan-stride: I like it although it does take a bit more brain power to read...
I am going to remove lines 52-54 though, in general I want as few return points from the function as possible because it makes things much easier to reason about all the paths which could have been taken.  If it returns an error or something that is fine but I prefer to keep it to one expected return at the very end of the function if possible.

x/stakeibc/keeper/grpc_query_lsm_deposits.go
ethan-stride: Also might not affect us in this case but sometimes the value can be a null type whereas the Getters will coerce it to be the expected type -- in fact this is the source of the 0th enum problem, if there is no input a direct call to the property could get you a nil but the Getter will make sure to return a valid value of the expected type and for enums that default is the first legal value.  If we are doing checks expecting a certain value the Getters are slightly safer.

x/stakeibc/keeper/grpc_query_lsm_deposits.go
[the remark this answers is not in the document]
sampocs: ok sorry, my suggestion was wrong - you can just return an empty response
```go
&types.QueryLSMDepositResponse{}, sdkerrors.ErrNotFound
```
In reality, the response doesn't matter if an error is returned

x/stakeibc/keeper/grpc_query_lsm_deposits.go
sampocs: I agree that if we did have optionals, this notation would make more sense, but I don't think we have any optional's in our codebase so that doesn't seem like a good case to optimize for. Also, I think the coercion does more harm than good. 

For instance, let's say we're trying to register an ICA channel and we need the chainId from the host zone struct for the channel name. At this point in our code base, if we have a nil host zone, we've got a big unexpected problem on our hands and the attribute approach would error as we'd prefer. But if we used a getter, this would pass an empty string for the chainID and we would register a channel with the incorrect name. This applies to most cases I think as well - we rarely pass around pointers, but when we do, if we somehow accidentally passed a nil, we should error right away, rather than masking the problem with a zero value.

But beyond that, imo the readability is much better when accessing with attributes, as (in go) Getters imply there's some calculation going on behind the scenes that's different than just accessing a field (contrary to a language like java).

But I'd also reiterate, that the attribute is common across every other codebase I've looked at (osmosis, cosmos-sdk, ibc-go, etc.) so I assume they're following a similar thinking.

x/stakeibc/keeper/grpc_query_lsm_deposits.go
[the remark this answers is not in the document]
sampocs: nvm, you're right we should return `nil, sdkerrors.ErrNotFound`

x/stakeibc/keeper/grpc_query_lsm_deposits_test.go
```diff
@@ -0,0 +1,109 @@
+package keeper_test
+
+import (
+	_ "github.com/stretchr/testify/suite"
+
+	"github.com/Stride-Labs/stride/v8/x/stakeibc/types"
+)
+
+// Setup LSM deposits in keeper
```
sampocs, line 9: nit: remove I think?

x/stakeibc/keeper/grpc_query_lsm_deposits_test.go
```diff
@@ -0,0 +1,109 @@
...
+func (s *KeeperTestSuite) TestLSMDeposit() {
+	// setup expected deposit in stakeibckeeper
+	initToken := types.LSMTokenDeposit{ChainId: "1", Denom: "validator70027"}
+	s.App.StakeibcKeeper.SetLSMTokenDeposit(s.Ctx, initToken)
+
+	// input nil, no chain-id, no denom --> error invalid input
+	invalidInputs := []*types.QueryLSMDepositRequest{}
+	noChainInput := types.QueryLSMDepositRequest{ChainId: "", Denom: "validator93740"}
+	noDenomInput := types.QueryLSMDepositRequest{ChainId: "42", Denom: ""}
+	invalidInputs = append(invalidInputs, nil, &noChainInput, &noDenomInput)
```
sampocs, line 20: I don't think invalidInputs is used - you can remove

sampocs (commented):
can you add descriptions to these require statements to help identify the test case?
Edit: whoops sorry meant to add this as a normal comment

x/stakeibc/keeper/grpc_query_lsm_deposits.go
sampocs: chatted offline, agreed on accessing using the attribute instead of the getters. marking this resolved

x/stakeibc/keeper/grpc_query_lsm_deposits.go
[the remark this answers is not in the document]
sampocs: Resolving this as the cleanup looks good, but in the future I'd recommend using guard clauses when possible to avoid running unnecessary code

PR #721, opened 2023-04-14
[LSM] Removed Unnecessary Overflow Check

sampocs:
This overflow check is a little ridiculous considering it's only incremented each query we send


ethan-stride (approved):
LGTM

PR #725, opened 2023-04-18
[LSM] Update Redemption Rate Calculation - Unit tests

sampocs:
Pt 2

* Added unit tests for redemption rate sub functions
* Added unit tests to cover LSM specific components of redemption rate

### DIFF
--- a/x/stakeibc/keeper/hooks.go
+++ b/x/stakeibc/keeper/hooks.go
@@ -160,9 +160,9 @@ func (k Keeper) UpdateRedemptionRates(ctx sdk.Context, depositRecords []recordst
 			continue
 		}
 
-		depositAccountBalance := k.GetDepositAccountBalance(hostZone, depositRecords)
-		undelegatedBalance := k.GetUndelegatedBalance(hostZone, depositRecords)
-		lsmDelegatedBalance := k.GetLSMDelegations(ctx, hostZone.ChainId)
+		depositAccountBalance := k.GetDepositAccountBalance(hostZone.ChainId, depositRecords)
+		undelegatedBalance := k.GetUndelegatedBalance(hostZone.ChainId, depositRecords)
+		lsmDelegatedBalance := k.GetTokenizedDelegation(ctx, hostZone.ChainId)
 		nativeDelegatedBalance := sdk.NewDecFromInt(hostZone.TotalBalancedDelegations.Add(hostZone.TotalUnbalancedDelegations))
 
 		k.Logger(ctx).Info(utils.LogWithHostZone(hostZone.ChainId, "Redemption Rate Components - "+
@@ -185,12 +185,12 @@ func (k Keeper) UpdateRedemptionRates(ctx sdk.Context, depositRecords []recordst
 }
 
 // Determine the undelegated balance from the deposit records queued for staking
-func (k Keeper) GetUndelegatedBalance(hostZone types.HostZone, depositRecords []recordstypes.DepositRecord) sdk.Dec {
+func (k Keeper) GetUndelegatedBalance(chainId string, depositRecords []recordstypes.DepositRecord) sdk.Dec {
 	// filter to only the deposit records for the host zone with status DELEGATION_QUEUE
 	undelegatedDepositRecords := utils.FilterDepositRecords(depositRecords, func(record recordstypes.DepositRecord) (condition bool) {
 		return ((record.Status == recordstypes.DepositRecord_DELEGATION_QUEUE ||
 			record.Status == recordstypes.DepositRecord_DELEGATION_IN_PROGRESS) &&
-			record.HostZoneId == hostZone.ChainId)
+			record.HostZoneId == chainId)
 	})
 
 	// sum the amounts of the deposit records
@@ -204,12 +204,12 @@ func (k Keeper) GetUndelegatedBalance(hostZone types.HostZone, depositRecords []
 
 // Determine the deposit account balance, representing native tokens that have been deposited
 // from liquid stakes, but have not yet been transferred to the host
-func (k Keeper) GetDepositAccountBalance(hostZone types.HostZone, depositRecords []recordstypes.DepositRecord) sdk.Dec {
+func (k Keeper) GetDepositAccountBalance(chainId string, depositRecords []recordstypes.DepositRecord) sdk.Dec {
 	// filter to only the deposit records for the host zone with status DELEGATION
 	depositAccountRecords := utils.FilterDepositRecords(depositRecords, func(record recordstypes.DepositRecord) (condition bool) {
 		return (record.Status == recordstypes.DepositRecord_TRANSFER_QUEUE ||
 			record.Status == recordstypes.DepositRecord_TRANSFER_IN_PROGRESS) &&
-			record.HostZoneId == hostZone.ChainId
+			record.HostZoneId == chainId
 	})
 
 	// sum the amounts of the deposit records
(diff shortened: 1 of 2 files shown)


x/stakeibc/keeper/hooks.go
```diff
@@ -184,46 +184,42 @@ func (k Keeper) UpdateRedemptionRates(ctx sdk.Context, depositRecords []recordst
...
-func (k Keeper) GetUndelegatedBalance(hostZone types.HostZone, depositRecords []recordstypes.DepositRecord) sdk.Dec {
-	// filter to only the deposit records for the host zone with status DELEGATION_QUEUE
-	undelegatedDepositRecords := utils.FilterDepositRecords(depositRecords, func(record recordstypes.DepositRecord) (condition bool) {
-		return ((record.Status == recordstypes.DepositRecord_DELEGATION_QUEUE ||
-			record.Status == recordstypes.DepositRecord_DELEGATION_IN_PROGRESS) &&
-			record.HostZoneId == hostZone.ChainId)
-	})
-
-	// sum the amounts of the deposit records
+// Determine the deposit account balance, representing native tokens that have been deposited
```
ethan-stride, line 187: "deposited from liquid stakes" sounds like the opposite thing it sounds like they are handing you a liquid stake for redemption.   I would say "native tokens that have been deposited to create new liquid stakes, ..."

x/stakeibc/keeper/hooks.go
```diff
@@ -184,46 +184,42 @@ func (k Keeper) UpdateRedemptionRates(ctx sdk.Context, depositRecords []recordst
...
+func (k Keeper) GetDepositAccountBalance(chainId string, depositRecords []recordstypes.DepositRecord) sdk.Dec {
+	// sum on deposit records with status TRANSFER_QUEUE or TRANSFER_IN_PROGRESS
 	totalAmount := sdkmath.ZeroInt()
-	for _, depositRecord := range undelegatedDepositRecords {
-		totalAmount = totalAmount.Add(depositRecord.Amount)
+	for _, depositRecord := range depositRecords {
+		transferStatus := (depositRecord.Status == recordstypes.DepositRecord_TRANSFER_QUEUE ||
+			depositRecord.Status == recordstypes.DepositRecord_TRANSFER_IN_PROGRESS)
+
+		if depositRecord.HostZoneId == chainId && transferStatus {
```
ethan-stride, line 196: Why do we call it HostZoneId in some places and chainId in most other places? Is this supposed to signify that the deposit is actually on stride but the token is of a type linked to HostZone or is this a naming convention which we have moved away from for new types but it is hard to change the core types?

x/stakeibc/keeper/hooks.go
```diff
@@ -184,46 +184,42 @@ func (k Keeper) UpdateRedemptionRates(ctx sdk.Context, depositRecords []recordst
...
-	// filter to only the deposit records for the host zone with status DELEGATION
-	depositAccountRecords := utils.FilterDepositRecords(depositRecords, func(record recordstypes.DepositRecord) (condition bool) {
-		return (record.Status == recordstypes.DepositRecord_TRANSFER_QUEUE ||
-			record.Status == recordstypes.DepositRecord_TRANSFER_IN_PROGRESS) &&
-			record.HostZoneId == hostZone.ChainId
-	})
-
-	// sum the amounts of the deposit records
+// Determine the undelegated balance from the deposit records queued for staking
+func (k Keeper) GetUndelegatedBalance(chainId string, depositRecords []recordstypes.DepositRecord) sdk.Dec {
```
ethan-stride, line 205: I would call all of these methods which end in Balance --> Total instead.  It is both clearer that the function is about to do a sum and return a single number but also it avoids confustion around the Balanced/Unbalanced delegations terminology

x/stakeibc/keeper/hooks.go
```diff
@@ -184,46 +184,42 @@ func (k Keeper) UpdateRedemptionRates(ctx sdk.Context, depositRecords []recordst
...
+		}
 	}
 
 	return sdk.NewDecFromInt(totalAmount)
 }
 
 // Returns the total delegated balance that's stored in LSM tokens
 // These are identified by any status besides "DEPOSIT_PENDING"
-func (k Keeper) GetLSMDelegations(ctx sdk.Context, chainId string) sdk.Dec {
+func (k Keeper) GetTokenizedDelegation(ctx sdk.Context, chainId string) sdk.Dec {
```
ethan-stride, line 222: GetDelegatedTokenizedTotal
Most other places you use Delegated or Undelegated as an adjective -- just have two adjectives.  This name makes it sound like there is one big delegation and that you are about to return that delegation when really you are about to sum over the many LSM delegations

x/stakeibc/keeper/hooks.go
```diff
@@ -184,46 +184,42 @@ func (k Keeper) UpdateRedemptionRates(ctx sdk.Context, depositRecords []recordst
...
-	// sum the amounts of the deposit records
+// Determine the deposit account balance, representing native tokens that have been deposited
+// from liquid stakes, but have not yet been transferred to the host
+func (k Keeper) GetDepositAccountBalance(chainId string, depositRecords []recordstypes.DepositRecord) sdk.Dec {
+	// sum on deposit records with status TRANSFER_QUEUE or TRANSFER_IN_PROGRESS
 	totalAmount := sdkmath.ZeroInt()
-	for _, depositRecord := range undelegatedDepositRecords {
-		totalAmount = totalAmount.Add(depositRecord.Amount)
+	for _, depositRecord := range depositRecords {
+		transferStatus := (depositRecord.Status == recordstypes.DepositRecord_TRANSFER_QUEUE ||
```
ethan-stride, line 193: What about the opposite direction? When someone wants to redeem and we have to undelegate (21 days) an then send the tokens from the host zone back over to the stride chain... aren't there going to be similar transfer queue states for that case until the tokens are given to the original user? Or do we take the stToken immediately so everything after the unbonding is triggered it isn't an ammount we need to care about in these totals?

x/stakeibc/keeper/hooks.go
sampocs: A liquid stake is a transaction - so this is saying deposits from the transaction.

The user doesn't "create a liquid stake" (noun), they "liquid stake" (verb)

x/stakeibc/keeper/hooks.go
sampocs: it's just an inconsistency across our codebase that I've been slowly trying to remedy.

You'll often see HostZoneId, HostChainId, ChainId, etc. but they're all the same

x/stakeibc/keeper/hooks.go
sampocs: we never transfer the LSM tokens back to Stride. When they are detokenized, the NFT is burned and the record is deleted

We do take the stTokens immediately during unbonding

x/stakeibc/keeper/hooks.go
sampocs: I don't really see how the Balance here (noun - similar to the balance of a bank account) could be confused with the Balanced adjective used when describing a delegation

I think GetUndelegatedTotal is a bit unclear as it doesn't describe what the total is.

Would GetTotalUndelegatedBalance clarify it?

riley-stride (approved):
Looks good to me; solid tests and ty for cleaning up the comments too, this is more understandable now.

x/stakeibc/keeper/hooks.go
ethan-stride: A balance can only be the amount in a single account, in contrast a total is summed across different accounts.  This function is summing across deposit records to find a total but that total sits in several different validators accounts not one single account balance correct?  Or is this total we compute literally all sitting in the single module account "DelegationAddress" on the host chain but controlled by Stride?  If that is the case then I agree Balance makes more sense here

sampocs: Yeah so it's the latter, it's the balance of tokens sitting in the delegation account. So if you did a balance query for the delegation account - you would get the same number

The reason the implementation is a bit deceiving is because we can't simply do a balance query without issuing an ICQ (since it's on the host chain), so instead we turn to our internal record keeping which tracks the single balance across multiple records, divided up across epochs

x/stakeibc/keeper/hooks.go
sampocs: renaming to `GetTotalTokenizedDelegations` to align with host zones

sampocs: going to handle this in the other PR

PR #729, opened 2023-04-19
V9 Upgrade Handler

sampocs:
Closes: #XXX

## Brief Changelog
* Added upgrade handler for v9 upgrade
* Updated go relayer to include a [fix](https://github.com/cosmos/relayer/issues/1163) for an issue where the issues weren't registering properly

## Author's Checklist

I have...

## Reviewers Checklist

I have...

      - [ ] not applicable
      - [ ] jira ticket `XXX` 
      - [ ] README.md 
      - [ ] not documented


riley-stride (approved):
looks good to me, but we need @shellvish and @aidansalzmann on this too

app/upgrades/v9/upgrades.go
```diff
@@ -0,0 +1,64 @@
...
+	claimParams, err := claimKeeper.GetParams(ctx)
+	if err != nil {
+		return errorsmod.Wrapf(err, "unable to get claim parameters")
+	}
+
+	for _, airdrop := range claimParams.Airdrops {
+		// Add the chain ID to each airdrop
+		chainId, ok := AirdropChainIds[airdrop.AirdropIdentifier]
+		if !ok {
+			ctx.Logger().Error(fmt.Sprintf("Chain ID not specified for %s airdrop", chainId))
```
shellvish, line 45: nit: should we assign `airdrop.ChainId = ""` here?

shellvish (approved):
Looks great! We should query all of the airdrop params after Dockernet / Localstride, just to confirm they work as intended

app/upgrades/v9/upgrades.go
sampocs: hmm good question

sampocs: ah actually, that should happen by default

sampocs: I think the questions is more should we leave it as an empty string and skip over or throw an error here.
I think I lean towards empty string so we don't halt the chain

sampocs: just added a test case for the empty chain ID

riley-stride: > I lean towards empty string so we don't halt the chain
Agreed

asalzmann: All of our current airdrops should map onto chains, so I think we should throw an error here (I don't see why it would throw an error, but if we can't find a chainId for an airdrop maybe something has gone wrong)

asalzmann (approved):

app/upgrades/v9/upgrades.go
sampocs: chatted offline, leaving it be considering how low the impact would be if this situation happened

PR #737, opened 2023-04-23
[LSM] Move LSMTokenDeposit to records

sampocs:
Closes: #XXX

We need the `LSMTokenDeposit` to be in the records module so that we can access the relevant keeper functions during the LSM Token IBC Transfer callback. It also seems like the rational place to put the struct as it's a similar object to our DepositRecods/HostZoneUnbonding records, etc.

Random side comment - we should revisit whether it's necessary to have everything in records in a second module. It's been a common point of confusion to all newcomers entering our code base. I think the reason it needs to be a second module is because we needed to separate out our transfer and ICA middleware stacks, but it's possible that that's no longer necessary with newer versions of IBC-go. 

## Brief Changelog
* Moved LSMTokenDeposit keeper functions and queries to records
* Updated locations where LSMTokenDeposit's were accessed and unit tests

## Author's Checklist

I have...

## Reviewers Checklist

I have...

      - [ ] not applicable
      - [ ] jira ticket `XXX` 
      - [ ] README.md 
      - [ ] not documented


riley-stride (approved):
Looks good. To your point about what goes in each module: imo before merging this we should get in tight sync on a heuristic for what goes in stakeibc and what goes in records, and why. Tagging @aidansalzmann who wrote the first version of records.

x/records/keeper/lsm_token_deposit_test.go
```diff
@@ -93,28 +93,28 @@ func (s *KeeperTestSuite) TestAddLSMTokenDeposit() {
...
 
 func (s *KeeperTestSuite) TestUpdateLSMTokenDepositStatus() {
-	statuses := []types.LSMDepositStatus{
-		types.TRANSFER_IN_PROGRESS,
-		types.TRANSFER_FAILED,
-		types.DETOKENIZATION_QUEUE,
-		types.DETOKENIZATION_IN_PROGRESS,
-		types.DETOKENIZATION_FAILED,
+	statuses := []types.LSMTokenDeposit_Status{
+		types.LSMTokenDeposit_TRANSFER_IN_PROGRESS,
```
ethan-stride, line 105: We should add one for the DEPOSIT_QUEUE and update the expected test behavior but maybe it should its own small PR.  That state didn't exist when these tests were written and accordingly I don't think any tests test for the transitions between DEPOSIT_QUEUE and TRANSFER_IN_PROGRESS

x/records/types/keys.go
```diff
@@ -40,4 +45,5 @@ const (
 	EpochUnbondingRecordCountKey = "EpochUnbondingRecord-count-"
 	DepositRecordKey             = "DepositRecord-value-"
 	DepositRecordCountKey        = "DepositRecord-count-"
+	LSMTokenDepositKey           = "LSMTokenDeposit"
```
ethan-stride, line 48: Should we make this naming consistent with LSMTokenDepositRecordKey and should we put a dash on the end as well?

x/stakeibc/keeper/icacallbacks_detokenize_test.go
```diff
@@ -14,12 +15,12 @@ import (
...
 	// Store the LSMDeposit record with status DETOKENIZATION_IN_PROGRESS
-	deposit := types.LSMTokenDeposit{
+	deposit := recordstypes.LSMTokenDeposit{
 		ChainId: HostChainId,
 		Denom:   LSMTokenBaseDenom,
-		Status:  types.DETOKENIZATION_IN_PROGRESS,
+		Status:  recordstypes.LSMTokenDeposit_DETOKENIZATION_IN_PROGRESS,
 	}
-	s.App.StakeibcKeeper.SetLSMTokenDeposit(s.Ctx, deposit)
+	s.App.RecordsKeeper.SetLSMTokenDeposit(s.Ctx, deposit)
```
ethan-stride, line 23: I assume this works when you ran the tests but just wondering why here in the stakeibc module why don't we have to refer to this as s.App.StakeibcKeeper.RecordsKeeper since everywhere else in this module we have to go one level in to get to the records keeper?

proto/stride/records/records.proto
```diff
@@ -0,0 +1,101 @@
...
+message EpochUnbondingRecord {
+  uint64 epoch_number = 1;
+  repeated HostZoneUnbonding host_zone_unbondings = 3;
+  reserved 2;
+}
+
+message LSMTokenDeposit {
+  enum Status {
+    DEPOSIT_PENDING = 0;
+    TRANSFER_IN_PROGRESS = 1;
```
ethan-stride, line 86: Checked that in the generated pb file doing the statuses this way may change the term we refer to in go to LSMTokenDeposit_TRANSFER_IN_PROGRESS but the actual string itself does not change and will still be "TRANSFER_IN_PROGRESS".  This is important for the CLI and http GET interfaces when filtering by status since that is what the user would have to type in.  No issue, should be good, just calling it out.

ethan-stride (commented):
lgtm with one or two questions

x/records/keeper/lsm_token_deposit_test.go
sampocs: assuming you mean DEPOSIT_PENDING, yeah I think it'd be good to update those - wanna take a stab at that today?

x/records/types/keys.go
sampocs: The notation without the dash is more standard (the dash notation comes from ignite).
So it's a question of whether it makes sense to follow the standard or keep things consistent. I could go either way. Wdyt?

Fwiw, it a perfect world I would update the other keys as well but that would require a state migration

x/stakeibc/keeper/icacallbacks_detokenize_test.go
sampocs: The `App` from the unit tests has access to everything and all keepers.
Whereas, when we're in the stakeibc module, we only have access to the keepers that are dependencies of stakeibc [[here](https://github.com/Stride-Labs/stride/blob/main/x/stakeibc/keeper/keeper.go#L34-L50)]

x/records/types/keys.go
ethan-stride: I like the dash because it visually separates the different parts, if the validator has an aphanumeric name it isn't always clear where the boundary of the base and the rest of the key is.

ethan-stride: I was also asking if you think we should put "Record" into the name of the key like the others have Record in the names

x/stakeibc/keeper/icacallbacks_detokenize_test.go
ethan-stride: Good to know for testing

x/records/types/keys.go
sampocs: the actual key is the byte encoding of this so it's not human readable

sampocs: oh yeah good question (re: adding Record suffix). Let's revisit the renaming of all the LSM stuff at the end - and do that in a separate PR, but yeah we'll probably want to call it `TokenizedShareRecord` or something of the like

ethan-stride: ok makes sense.  Then this PR lgtm

PR #741, opened 2023-04-26
remove add lsm deposit, enforce single record per denom

ethan-stride:
Closes:  enforce one record per denom

Although the user could choose to split up their LSM tokens into multiple liquid stake actions, since we use the denom and chain id as the key in the records, we only want to have one being processed at a time.  If there is no problem, these records should be short lived anyway on the order of minutes until the LSM token is converted to native tokens and the NFT is deleted.  This PR enforces that condition that there can only be one record at a time for a given LSM NFT denom.

## Brief Changelog

remove the AddLSMTokenDeposit
add check to the StartLSMLiquidStake message to prevent new stake with pre-existing denom
Convert all AddLSMTokenDeposit to SetLSMTokenDeposit
Update tests

## Author's Checklist

I have...

- Run and PASSED locally all GAIA integration tests

## Reviewers Checklist

I have...

      - [ ] not applicable
      - [ ] jira ticket `XXX` 
      - [ ] README.md 
      - [ ] not documented

### DIFF
--- a/x/stakeibc/keeper/msg_server_lsm_liquid_stake.go
+++ b/x/stakeibc/keeper/msg_server_lsm_liquid_stake.go
@@ -20,20 +20,22 @@ import (
 
 // Exchanges a user's LSM tokenized shares for stTokens using the current redemption rate
 // The LSM tokens must live on Stride as an IBC voucher (whose denomtrace we recognize)
-//   before this function is called
+//
+//	before this function is called
 //
 // The typical flow:
 //   - A staker tokenizes their delegation on the host zone
 //   - The staker IBC transfers their tokenized shares to Stride
 //   - They then call LSMLiquidStake
-//     - The staker's LSM Tokens are sent to the Stride module account
-//     - The staker recieves stTokens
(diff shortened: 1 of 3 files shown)


x/stakeibc/keeper/msg_server_lsm_liquid_stake.go
```diff
@@ -20,20 +20,22 @@ import (
...
+//   - The staker recieves stTokens
 //
 // As a safety measure, at period checkpoints, the validator's exchange rate is queried and the transaction
 // is not settled until the query returns
 // As a result, this transaction has been split up into a (1) Start and (2) Finish function
-//   If no query is needed, (2) is called immediately after (1)
-//   If a query is needed, (2) is called in the query callback
+//
+//	If no query is needed, (2) is called immediately after (1)
+//	If a query is needed, (2) is called in the query callback
```
sampocs, line 38: I think this may have happened accidentally from your linter/formatter?

sampocs (approved):
Lgtm. Only thing was it looks like some of the docstring got misformatted

PR #742, opened 2023-04-26
[LSM] Move token escrow out of StartLSMLiquidStake

sampocs:
Closes: #XXX

Originally, in the event of a slash query interrupting the LSMLiquidStaking flow, the staker's LSM tokens were escrowed while the query was in flight. However, this introduces complexity in the error case as an error during the query callback could prevent tokens from being refunded. As a result, we've opted to remove the escrow step in StartLSMLiquidStake. 

Now in the callback, we confirm the user has sufficient balance and fail the transaction otherwise. 

## Brief Changelog
* Remove `RefundToken function`
* Moved transfer of LSMToken from `StartLSMLiquidStake` to `FinishLSMLiquidStake`
* Added a `ValidateLSMLiquidStake` function so that the checks can be duplicated in both Start and FinishLSMLiquidStake

## Author's Checklist

I have...

## Reviewers Checklist

I have...

      - [ ] not applicable
      - [ ] jira ticket `XXX` 
      - [ ] README.md 
      - [ ] not documented


x/stakeibc/keeper/msg_server_lsm_liquid_stake.go
```diff
@@ -177,32 +130,59 @@ func (k Keeper) SubmitValidatorSlashQuery(ctx sdk.Context, lsmLiquidStake types.
...
+
+	// Get the staker's address and the host zone's deposit account address (which will custody the tokens)
+	liquidStakerAddress := sdk.MustAccAddressFromBech32(lsmTokenDeposit.StakerAddress)
+	hostZoneAddress, err := sdk.AccAddressFromBech32(hostZone.DepositAddress)
+	if err != nil {
+		return errorsmod.Wrapf(err, "host zone address is invalid")
+	}
+
+	// Transfer the LSM token to the deposit account
+	lsmIBCToken := sdk.NewCoin(lsmTokenDeposit.IbcDenom, lsmTokenDeposit.Amount)
```
ethan-stride, line 163: So we don't actually take control of the LSM token from the user and hand them the stTokens until the finish part of the function because we may have had to wait for the query to give us an answer back and now this avoids any need to refund, but in the start part of the function we set the record for the LSMTokenDeposit... shouldn't we also wait until the transaction is actually going to take place to set this record?

Otherwise we need to explicitly delete that record if anything has gone wrong enough to prevent the user from getting their stTokens but I don't see anywhere that happens in the finish part of the function unless by throwing an error in the finish part it is part of the same state updates as in the start part so throwing an error would negate any changes which would have happened to the records?

ethan-stride (commented):
One big question about the start liquid stake, query, finish liquid stake flow, but otherwise looks good to me.

riley-stride (approved):
lgtm

x/stakeibc/keeper/msg_server_lsm_liquid_stake.go
sampocs: sorry somehow just seeing this! Very good point, I think we should delete the LSMTokenDeposit if the finish function fails. We could also keep it for monitoring purposes, but since it's easy for the user to manipulate, I'd vote removing it!

sampocs: Going to handle this in the other PR since it's part of the callback

PR #747, opened 2023-05-01
[LSM] Batching ICA rebalance calls into smaller groups

ethan-stride:
Closes: Rebalance with smaller batchs of ICA messages

The original version sent all ICA messages for rebalancing in a single transaction but we know that this will fail in reality as that many messages will be too large to fit in a single block.  Currently batching these messages 5 at a time but we can update this to be higher if we see these go through consistently.

## Brief Changelog
Sending the msgs in batches and saving the callback params in corresponding batches
Removed the numRebalances parameter from a few of the functions because we no longer need it.

## Author's Checklist

Currently having an unrelated issue testing locally because of a dockernet issue, but this is a very small change.
Ran the unit tests in validator_selection_test.go by hand

## Reviewers Checklist

I have...

      - [ ] not applicable
      - [ ] jira ticket `XXX` 
      - [ ] README.md 
      - [ ] not documented

### DIFF
--- a/x/stakeibc/keeper/validator_selection.go
+++ b/x/stakeibc/keeper/validator_selection.go
@@ -26,15 +26,13 @@ type RebalanceValidatorDelegationChange struct {
 // Note: this cannot be run more than once in a single unbonding period
 func (k Keeper) RebalanceAllHostZones(ctx sdk.Context, dayNumber uint64) {
 	for _, hostZone := range k.GetAllActiveHostZone(ctx) {
-		numRebalance := uint64(len(hostZone.Validators))
-
 		if dayNumber%hostZone.UnbondingPeriod != 0 {
 			k.Logger(ctx).Info(utils.LogWithHostZone(hostZone.ChainId,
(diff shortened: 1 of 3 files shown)


x/stakeibc/keeper/validator_selection.go
```diff
@@ -59,29 +57,40 @@ func (k Keeper) RebalanceDelegationsForHostZone(ctx sdk.Context, chainId string,
...
-	// marshall the callback
-	rebalanceCallback := types.RebalanceCallback{
-		HostZoneId:   hostZone.ChainId,
-		Rebalancings: rebalacings,
-	}
-	rebalanceCallbackBz, err := k.MarshalRebalanceCallbackArgs(ctx, rebalanceCallback)
-	if err != nil {
-		return errorsmod.Wrapf(err, "unable to marshal rebalance callback args")
-	}
+	icaBatchSize := 5
```
sampocs, line 62: Let's pull this out as a constant. Not sure it makes sense to add it as a full blown module parameter (since we have way to many of those right now), but maybe at least just add it to the top of this file
```go
const IcaBatchSize = 5
```

sampocs (commented):
Looks great! One small comment

PR #749, opened 2023-05-02
Ethan/lsm unittest bugfix

ethan-stride:
Closes: unit test breaking bug

The bug was introduced when removing a seemingly vestigial parameter numRebalance.
The current version of the rebalance algorithm uses a "while" style loop and breaks out when a condition is met.
This numRebalance was being computed several function calls higher and passed through a chain of function calls.
It looked like the original version of the algorithm needed this param but many things would simply now by removing it.
The direct unit tests passed when run by hand.

Lesson Learned: ALWAYS run `make test-unit` to run all unit tests before pushing
It is true that this parameter is no longer needed in the current algorithm.
However, many tests at the ICA messaging level explicitly setup messages with this parameter set
Their purpose is not to do a full rebalance but just a partial rebalance and check the intermediate states.
All of these useful tests got broken when this param was removed, go couldn't warn us because all function signatures and calls were fine, just a previously useful field in the msg was now being ignored and not passed into the function calls.
I think these tests (in msg_server_rebalance_validators_test.go) are actually useful still and so this PR is reversing the change the previous PR had made to remove the param.
Added a cautionary message that although this param is locally useless looking, far away tests actually depend on it.

## Brief Changelog

Added the numRebalance field back into the several functions which need it in validator_selection.go

## Author's Checklist

-  Run all the unit tests with `make test-unit` 

## Reviewers Checklist

I have...

      - [ ] not applicable
      - [ ] jira ticket `XXX` 
      - [ ] README.md 
      - [ ] not documented

### DIFF
--- a/x/stakeibc/keeper/msg_server_rebalance_validators.go
+++ b/x/stakeibc/keeper/msg_server_rebalance_validators.go
@@ -13,7 +13,7 @@ func (k msgServer) RebalanceValidators(goCtx context.Context, msg *types.MsgReba
 	ctx := sdk.UnwrapSDKContext(goCtx)
 	k.Logger(ctx).Info(fmt.Sprintf("RebalanceValidators executing %v", msg))
 
-	if err := k.RebalanceDelegationsForHostZone(ctx, msg.HostZone); err != nil {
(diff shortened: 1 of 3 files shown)


sampocs:
closing in favor of #758

PR #751, opened 2023-05-03
[LSM] Unbonding

sampocs:
## Context and Purpose
Now that we accept LSM Liquid Stakes, the distribution of stake can deviate from the ideal distribution as defined by the validator weights. The distribution will only be inconsistent until we rebalance. In the meantime during unbonding, we can unbond in a way that moves us back towards the ideal distribution by targetting validators that have an excess delegation from LSM.

## Redemption Algorithm
1. Determine total unbond amount for the epoch
2. Determine the ideal/balanced delegation for each validator with the total amount of stake that’s left after unbonding (i.e. the delegations for each validator as if we were to unbond and then rebalance). Hereinafter referred to as the "balanced delegation".
3. Determine the unbond capacity for each validator
   1. The capacity is defined as the difference between the validator’s balanced delegation and their total delegation
   2. If the validator’s current delegation is greater than their balanced delegation, we can unbond only up to the balanced delegation (that way we don’t unfairly penalize a validator for having LSM delegations). Their capacity will be equal to this difference.
   3. If the validator’s balanced delegation is greater than the current delegation, that means, they’re “owed” more than they have. However, this is an unbonding, not a rebalancing, and delegations can only be reduced. Thus, the capacity for these validator’s should be 0 and they should have nothing unbonded.
5. Sort the validator’s by the “BalanceRatio” which indicates their proportional degree of imbalance
   1.  The ratio is equal to the balanced delegation divided by the current delegation
       1. Ex1: Ideal Balanced Delegation 80,  Current Delegation 100 (surplus of 20), Ratio: 0.8
        2. Ex2: Ideal Balanced Delegation 480, Current Delegation 500 (surplus of 20), Ratio: 0.96
        3. Notice these two validators had the same net surplus, but the validator from example 2 is proportionally more balanced as the surplus is a smaller percentage of their overall delegation
   2. This ratio determines the priority for unbonding
   3. Validator’s with zero-weight will always have an ideal balanced delegation of 0, a “BalanceRatio” of 0, and will always be prioritized first
6. Iterate through the list of validator’s in order of priority, and unbond as much as possible from each until the total unbond amount has been accounted for

See excalidraw for an example walkthrough

## Changelog
* Added function `GetTotalUnbondAmountAndRecordIds` to sum the total unbond amount from the epoch unbonding records
* Added function `GetValidatorUnbondCapacity` which takes a list of validators along with their "balanced delegations" and returns an array of validator capacities (with zero-capacity validator's excluded)
* Added function `SortUnbondingCapacityByPriority` to sort the list of validators by their balanced ratio
* Added function `GetUnbondingICAMessages` to iterate through the unbonding capacities in order of priority, unbonding as much as possible from each validator
* Added function` UnbondFromHostZone` that acts as the main driving function for submitting the undelegation ICAs. It does the following:
  * Determines the total unbond amount
  * Determines the balanced delegations
  * Determines each validator's unbond capacity
  * Builds the ICA messages and callback data
  * Submits the ICA
* Cleaned up `InitiateAllHostZoneUnbondings`
* Added unit tests for all the above
* Modified test CreateICAChannel test helper to return the new channel's portID
* Added test utils to search for events

**Note**: When working through this ticket, I first cleaned up the unbonding functions while verifying unit tests still passed. I then implemented the new unbonding algorithm. You're welcome to walk the commits if you prefer to have a better sense of the impact of each change, but a lot of the "cleaned up" code was eventually removed when the new algorithm was implemented.


x/stakeibc/keeper/icacallbacks_undelegate_test.go
```diff
@@ -119,7 +119,7 @@ func (s *KeeperTestSuite) SetupUndelegateCallback() UndelegateCallbackTestCase {
 		SplitDelegations:        []*types.SplitDelegation{&val1SplitDelegation, &val2SplitDelegation},
 		EpochUnbondingRecordIds: []uint64{epochNumber},
 	}
-	callbackArgsBz, err := s.App.StakeibcKeeper.MarshalUndelegateCallbackArgs(s.Ctx, callbackArgs)
+	callbackArgsBz, err := proto.Marshal(&callbackArgs)
```
riley-stride, line 122: Could you explain why this can now be simplified and why we can do away with `MarshalUndelegateCallbackArgs`?

x/stakeibc/keeper/msg_server_restore_interchain_account_test.go
```diff
@@ -48,8 +47,7 @@ func (s *KeeperTestSuite) SetupRestoreInterchainAccount(createDelegationICAChann
 	var channelID, portID string
 	if createDelegationICAChannel {
 		owner := "GAIA.DELEGATION"
-		channelID = s.CreateICAChannel(owner)
-		portID = icatypes.PortPrefix + owner
+		channelID, portID = s.CreateICAChannel(owner)
```
riley-stride, line 50: Is the addition of the portID return arg to `CreateICAChannel` for the sole purpose of simplifying this logic? Not against it necessarily, just trying to understand

x/stakeibc/keeper/icacallbacks_undelegate_test.go
sampocs: all `MarshalUndelegateCallbackArgs` and `UnmarshalUndelegateCallbackArgs` are doing is calling proto.Marshal, proto.Unmarshal under the hood!

I think we should do away with all our custom marshal/unmarshal functions cause they add ~20 lines of unnecessary code per callback and add an extra step/ more friction to the process of adding a new callback.

x/stakeibc/keeper/msg_server_restore_interchain_account_test.go
sampocs: yeah exactly. It's called in a lot of setup functions and I've found that every time I go to create a channel in a test, I wind up referencing/copying-pasting those lines haha

I think we should also start adding stakeibc specific test helpers to keeper_test.go (e.g. `CreateDelegationICAChannel`, but that's a ticket for another day!

x/stakeibc/keeper/unbonding_records_get_host_zone_unbondings_msgs_test.go
```diff
@@ -1,226 +1,813 @@
...
+}
+
+// Helper function to check that an undelegation ICA was submitted and that the callback data
+// holds the expected unbondings for each validator
+func (s *KeeperTestSuite) CheckUnbondingMessages(tc UnbondingTestCase, expectedUnbondings []ValidatorUnbonding) {
+	// Trigger unbonding
+	err := s.App.StakeibcKeeper.UnbondFromHostZone(s.Ctx, tc.hostZone)
+	s.Require().NoError(err, "no error expected when calling unbond from host")
+
+	// Check that sequence number incremented from a sent ICA
```
riley-stride, line 108: Why use the sequence # to check whether an ICA was submitted, rather than checking some result from the ICA directly? Haven't seen this approach before.

x/stakeibc/keeper/unbonding_records_get_host_zone_unbondings_msgs_test.go
```diff
@@ -1,226 +1,813 @@
...
+		validators = append(validators, &types.Validator{
+			Address:    address,
+			Delegation: currentDelegation,
+		})
+		balancedDelegations[address] = balancedDelegation
+	}
 
-	// there's rounding in the logic that distributes stake amongst validators, so one or the other of the balances will be correct, depending on the rounding
-	// at least one will be correct, and the other will be off by 1 by rounding, so we check and OR condition
-	s.Require().True(val1Unbonded || val2Unbonded, "unbonding amt should be the correct amount")
```
riley-stride, line 153: this code that checks unbondAmts per val was removed, why? Is it sufficiently covered in ln 491, and if so, is the rounding that previously required us to check "one or the other" in previous lines 134-150 now changed such that we can check _all_ conditions are satisfied?

x/stakeibc/keeper/unbonding_records_get_host_zone_unbondings_msgs_test.go
```diff
@@ -1,226 +1,813 @@
...
 	}
-	// write the host zone with zero-delegation validators back to the store
-	s.App.StakeibcKeeper.SetHostZone(s.Ctx, tc.hostZone)
-
-	// TODO: check epoch unbonding record ids here
-	_, _, _, _, err := s.App.StakeibcKeeper.GetHostZoneUnbondingMsgs(s.Ctx, tc.hostZone)
-	s.Require().EqualError(err, fmt.Sprintf("Could not unbond %v on Host Zone %s, unable to balance the unbond amount across validators: not found", tc.amtToUnbond.Mul(sdkmath.NewInt(int64(len(tc.epochUnbondingRecords)))), tc.hostZone.ChainId))
-}
 
-func (s *KeeperTestSuite) TestGetTargetValAmtsForHostZone_Success() {
```
riley-stride, line 205: why did we remove the TestGetTargetValAmtsForHostZone_Success test?

x/stakeibc/keeper/unbonding_records_initiate_all_unbondings_test.go
```diff
@@ -109,49 +109,51 @@ func (s *KeeperTestSuite) SetupInitiateAllHostZoneUnbondings() InitiateAllHostZo
...
+	// tests that we can successful initiate a host zone unbonding for GAIA and OSMO
 	s.SetupInitiateAllHostZoneUnbondings()
-	success, successful_unbondings, failed_unbondings := s.App.StakeibcKeeper.InitiateAllHostZoneUnbondings(s.Ctx, 12)
-	s.Require().True(success, "initiating unbondings returns true")
-	s.Require().Len(successful_unbondings, 2, "initiating unbondings returns 2 successful unbondings")
-	s.Require().Len(failed_unbondings, 0, "initiating unbondings returns 0 failed unbondings")
+	s.App.StakeibcKeeper.InitiateAllHostZoneUnbondings(s.Ctx, 12)
+
+	// An event should be emitted for each if they were successful
+	s.CheckEventValueEmitted(types.EventTypeUndelegation, types.AttributeKeyHostZone, HostChainId)
```
riley-stride, line 117: lines 117-170: why are we checking event emissions in unit tests rather than checking changes in underlying state?

x/stakeibc/keeper/unbonding_records.go
```diff
@@ -43,171 +65,250 @@ func (k Keeper) CreateEpochUnbondingRecord(ctx sdk.Context, epochNumber uint64)
...
+		return fmt.Errorf("there are no validators on %s with sufficient unbond capacity", hostZone.ChainId)
+	}
+
+	// Sort the unbonding capacity by priority
+	// Priority is determined by checking the how proportionally unbalanced each validator is
+	// Zero weight validators will come first in the list
+	prioritizedUnbondCapacity := SortUnbondingCapacityByPriority(validatorUnbondCapacity)
+
+	// Get the undelegation ICA messages and split delegations for the callback
+	msgs, unbondings, err := k.GetUnbondingICAMessages(hostZone, totalAmountToUnbond, prioritizedUnbondCapacity)
```
riley-stride, line 263: what's the max number of unbonding msgs we can fix in the host zone ICA block size? recall the limit is ~35 for delegation ICA msgs. We should probably enforce a cap on the # of msgs here as well for now, maybe implemented inside GetUnbondingICAMessages based on a param

riley-stride (commented):
Looking really good! Diagrams helped a lot, ty.  Added some nit questions / comments.

x/stakeibc/keeper/unbonding_records_get_host_zone_unbondings_msgs_test.go
sampocs: did you have anything specific in mind? the ICA isn't relayed here so there's not a lot that we can check

x/stakeibc/keeper/unbonding_records_get_host_zone_unbondings_msgs_test.go
sampocs: yeah it's better covered by all the `TestUnbondFromHostZone` tests. And tbh, I never really followed what that rounding error meant - looking at the test itself, there's no reason there should be a rounding error here. I think it might've just been a bug in the test

x/stakeibc/keeper/unbonding_records_get_host_zone_unbondings_msgs_test.go
sampocs: It's better covered by [this](https://github.com/Stride-Labs/stride/blob/lsm/x/stakeibc/keeper/validator_selection_test.go#L180) test that was added during the rebalance PR

x/stakeibc/keeper/unbonding_records.go
sampocs: ah yeah good call! Wdyt about just adding the size as a constant rather than a param? Feels like we've been going kinda crazy with the module params and none of them are actually used

x/stakeibc/keeper/unbonding_records_initiate_all_unbondings_test.go
sampocs: messaged you on slack to explain in detail, but TLDR is we don't need to check the nitty gritty in this test since that's covered by all the `TestUnbondFromHostZone` tests. All we need to do is make sure these didn't fail, but since no error is throw, we can verify that with events

x/stakeibc/keeper/unbonding_records_get_host_zone_unbondings_msgs_test.go
riley-stride: I can't think of another way, this is creative!

x/stakeibc/keeper/unbonding_records.go
riley-stride: on board with that

riley-stride (approved):

sampocs:
@riley-stride just added unbonding batch size in last commit if you wanna peep that real quick

x/stakeibc/keeper/unbonding_records.go
```diff
@@ -43,171 +69,261 @@ func (k Keeper) CreateEpochUnbondingRecord(ctx sdk.Context, epochNumber uint64)
...
+}
 
-	// Determine the desired unbonding amount for each validator based based on our target weights
-	targetUnbondingsByValidator, err := k.GetTargetValAmtsForHostZone(ctx, hostZone, totalAmountToUnbond)
-	if err != nil {
-		errMsg := fmt.Sprintf("Error getting target val amts for host zone %s %v: %s", hostZone.ChainId, totalAmountToUnbond, err)
-		k.Logger(ctx).Error(errMsg)
-		return nil, sdkmath.ZeroInt(), nil, nil, errorsmod.Wrap(types.ErrNoValidatorAmts, errMsg)
+// Determine the unbonding "slack" that each validator has
+// The slack is determined by the difference between their current delegation
```
ethan-stride, line 95: You used the term capacity everywhere else.  Maybe this comment was written at a different time, but we should update it to be consistent

x/stakeibc/keeper/unbonding_records.go
```diff
@@ -43,171 +69,261 @@ func (k Keeper) CreateEpochUnbondingRecord(ctx sdk.Context, epochNumber uint64)
...
+		validatorB := validatorUnbondCapacity[j]
+
+		// Sort by the balance ratio first - in ascending order - so the more unbalanced validators appear first
+		if !validatorA.GetBalanceRatio().Equal(validatorB.GetBalanceRatio()) {
+			return validatorA.GetBalanceRatio().LT(validatorB.GetBalanceRatio())
 		}
-		finalUnbondingsByValidator[validator.Address] = targetUnbondAmount
-	}
 
-	// If there was overflow (i.e. there was at least one validator without sufficient delegations to cover their unbondings)
```
ethan-stride, line 105: I love how much simpler and more understandable the new code is.  Deleting lines is that much less surface area for bugs

x/stakeibc/keeper/unbonding_records.go
```diff
@@ -43,171 +69,261 @@ func (k Keeper) CreateEpochUnbondingRecord(ctx sdk.Context, epochNumber uint64)
...
-		HostZoneId:              hostZone.ChainId,
-		SplitDelegations:        splitDelegations,
-		EpochUnbondingRecordIds: epochUnbondingRecordIds,
+	return msgs, unbondings, nil
+}
+
+// Submits undelegation ICA messages for a given host zone
+//
+// The total unbond amount is determined from the epoch unbonding records
+// The unbond amount is then distributed amongst validators proportional to how unbalanced they are.
```
ethan-stride, line 217: This comment line is not 100% true.  We order by how out of balance they are but then fill up to use the least messages.  As written in english you just described the more fair algorithm where we split the amount up and none send their full capacity but we had decided against this since it is more complicated to compute and it often causes way more messages needing to be sent.

x/stakeibc/keeper/unbonding_records.go
```diff
@@ -43,171 +69,261 @@ func (k Keeper) CreateEpochUnbondingRecord(ctx sdk.Context, epochNumber uint64)
...
+		return nil
 	}
 
-	return msgs, totalAmountToUnbond, marshalledCallbackArgs, epochUnbondingRecordIds, nil
-}
+	// Determine the ideal balanced delegation for each validator after the unbonding
+	//   (as if we were to unbond and then rebalance)
+	// This will serve as the starting point for determining how much to unbond each validator
+	currentTotalDelegation := k.GetTotalValidatorDelegations(hostZone)
+	delegationAfterUnbonding := currentTotalDelegation.Sub(totalAmountToUnbond)
```
ethan-stride, line 246: It is so simple now and makes good reuse of functions originally defined for the rebalance, this might be some of my favorite code in the codebase.  If we make the liquidstake delegate code also dynamically move things towards the more balanced state (maybe doing this same exact algorithm in reverse) instead of just blindly splitting according to the weights I think we actually improved the aethetics and the average performance of the protocol at the same time!

x/stakeibc/keeper/unbonding_records.go
ethan-stride: It shouldn't be a param anyway, I thought the purpose of the param was something that could be voted on to change dynamically -- it should be something which the stride users have some say in because it affects them to be a param, anything which is a technical tuning should just be done internally by us in development.

x/stakeibc/keeper/unbonding_records_get_host_zone_unbondings_msgs_test.go
```diff
@@ -1,226 +1,813 @@
...
+	validators := []*types.Validator{
+		{Address: "valA", Weight: 0, Delegation: sdkmath.NewInt(25)},
+		{Address: "valB", Weight: 0, Delegation: sdkmath.NewInt(50)},
+		{Address: "valC", Weight: 0, Delegation: sdkmath.NewInt(25)},
+	}
+	tc := s.SetupTestUnbondFromHostZone(totalWeight, totalStake, totalUnbondAmount, validators)
+
+	// Call unbond - it should fail
+	err := s.App.StakeibcKeeper.UnbondFromHostZone(s.Ctx, tc.hostZone)
+	s.Require().ErrorContains(err, "No non-zero validators found for host zone")
```
ethan-stride, line 336: It is a weird case, but why would we want to fail?
Imagine we have validators which do have weights, they have a balance but all of it gets unbonded and sent out.
Then these 3 validators get LSM deposits.  In that case we have no weighted validators with any delegation and we have multiple 0 weight validators which do have some... we should still happily continue working in that case?

riley-stride (approved):
batching looks good to me

x/stakeibc/keeper/unbonding_records_get_host_zone_unbondings_msgs_test.go
```diff
@@ -1,226 +1,813 @@
...
+			Address:    validatorCapacity.ValidatorAddress,
+			Delegation: validatorCapacity.CurrentDelegation,
+		})
+		balancedDelegations[validatorCapacity.ValidatorAddress] = validatorCapacity.BalancedDelegation
+	}
 
-	val1Unbonded := strings.Contains(actualUnbondMsg1, val1UnbondAmt)
-	val2Unbonded := strings.Contains(actualUnbondMsg2, val2UnbondAmt)
+	// Add validators with no capacity - none of these should be in the returned list
+	deficits := []int64{0, 10, 50}
```
ethan-stride, line 530: This is the first test that is hard to understand... what is deficits?
So you set up 6 validators, the first three have capacity, but the last 3 are off by these positive deficits meaning they can't actually be affected by an unbonding...
But I don't see how the numbers add up... When you add the last 3 validators that would completely change the actual capacities wouldn't it?  Are we assuming the weights would be such that this all works out so we take it as granted the capacities were already computed correctly?

x/stakeibc/keeper/unbonding_records_get_host_zone_unbondings_msgs_test.go
```diff
@@ -1,226 +1,813 @@
...
+		})
+		balancedDelegations[address] = balancedDelegation
+	}
 
-	// there's rounding in the logic that distributes stake amongst validators, so one or the other of the balances will be correct, depending on the rounding
-	// at least one will be correct, and the other will be off by 1 by rounding, so we check and OR condition
-	s.Require().True(val1Unbonded || val2Unbonded, "unbonding amt should be the correct amount")
+	// Check capacity matches expectations
+	actualUnbondCapacity := s.App.StakeibcKeeper.GetValidatorUnbondCapacity(s.Ctx, validators, balancedDelegations)
+	s.Require().Len(actualUnbondCapacity, len(expectedUnbondCapacity), "number of expected unbondings")
```
ethan-stride, line 549: This is a weird way to do it, why not match the ones below and make it more readable with this:
s.Require().Equal(len(actual...), len(expected...), "message")

ethan-stride (commented):
Wow Sam is a testing machine! Very nice PR, very well explained and commented throughout.
Excellent paragraph explaining it in the PR, the only thing I might add to explain where the motivation is coming from is what would happen in the counter factual that we took more than the capacity from someone.  You mention in one place that we are trying to be fair to LSM and non LSM holders alike, but if we ever took more than capacity that validator would be able to complain that they have less than they would have specifically because they held LSM which is out of their control.  This is important because the only way it would come back up is with a rebalance which is rare in time compared to unbondings -- so they would be justified in saying that on average they are being short changed overtime because of the LSM they have.  As long as we don't go past capacity they should be agnostic to holding LSM or not since since on the next rebalance they were expecting to drop by up to that amount anyway.

Awesome clean code and really well thought through tests.

x/stakeibc/keeper/unbonding_records.go
```diff
@@ -17,6 +20,29 @@ import (
...
+	UndelegateICABatchSize = 30
+)
+
+// QUESTION: Should I move this and the rebalance structs to types?
+// That feels like the proper location, but since it's only really used by this file
+// if feels a bit more hidden in types
+type ValidatorUnbondCapacity struct {
+	ValidatorAddress   string
+	CurrentDelegation  sdkmath.Int
+	BalancedDelegation sdkmath.Int
```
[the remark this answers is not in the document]
sampocs: yeah I've been struggling with the name as well. I've been going back and forth between ideal and balanced

x/stakeibc/keeper/unbonding_records.go
```diff
@@ -17,6 +20,29 @@ import (
...
+	Capacity           sdkmath.Int
+}
+
+// The ratio of ideal balanced delegation to the current delegation
+// This represents how proportionally unbalanced each validator is
+// The smaller number means their current delegation is much larger
+// then their fair portion of the current total stake
+func (c *ValidatorUnbondCapacity) GetBalanceRatio() sdk.Dec {
+	return sdk.NewDecFromInt(c.BalancedDelegation).Quo(sdk.NewDecFromInt(c.CurrentDelegation))
```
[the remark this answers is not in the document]
sampocs: ah that's very clever!! going to leave it as is for readability, but will keep that trick in my back pocket!

x/stakeibc/keeper/unbonding_records.go
sampocs: i was probably feeling a bit redundant in the comments. I'll change to capacity

x/stakeibc/keeper/unbonding_records.go
sampocs: Poor phrasing on my part, but I was just trying to indicate that validators that are more unbalanced, have more of their delegation targeted. And I was also trying to be more succinct since I described it in detail in other places.

Hows this?
```
// First, the total unbond amount is determined from the epoch unbonding records
// Then that unbond amount is distributed amongst validators by first targeting validators
//   who have a proportionally larger delegation than their entitled to (due to the natural 
//   unbalancing from LSM liquid stakes)
// As a result, unbondings lead to a more balanced distribution of stake across validators
```

x/stakeibc/keeper/unbonding_records.go
sampocs: I agree! Really great work coming up with this algorithm!

x/stakeibc/keeper/unbonding_records_get_host_zone_unbondings_msgs_test.go
sampocs: why is it more unreadable? I would have thought the opposite cause the require statement is clearly indicating that it is checking the length

x/stakeibc/keeper/unbonding_records_get_host_zone_unbondings_msgs_test.go
sampocs: This is the test for the subfunction `GetValidatorUnbondCapacity`. So all it checks is the difference between the balanced delegation and the current delegation. It doesn't consider the total stake at all.

any suggestion on how I could make the deficit variable more clear? Or maybe I just extend the comment?

x/stakeibc/keeper/unbonding_records_get_host_zone_unbondings_msgs_test.go
sampocs: that is a good point. This was an error that was thrown from within GetTargetAmts. I think the idea is that if we have all 0 weight validators, something is clearly very wrong.

If we were throwing the error from UnbondFromHostZone, I'd agree we should remove it, but I think in this case, I'd lean towards not touching GetTargetAmts cause it's called in other places

x/stakeibc/keeper/unbonding_records.go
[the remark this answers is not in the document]
ethan-stride: I like target best but I like ideal second best

x/stakeibc/keeper/unbonding_records_get_host_zone_unbondings_msgs_test.go
ethan-stride: This is definitely weird and unreadable, not your fault it is weird the testing library had a function like this.
Len(actual..., len(expected), "message")    <<< so it looks like a call where the input is all 3 things
but it is doing a comparison between the first and second argument, the first argument is a list, the second argument is a number.

x/stakeibc/keeper/unbonding_records_get_host_zone_unbondings_msgs_test.go
ethan-stride: Yeah I was confused because the capacity calculations have to depend on the total and the unbond amount and the weights but that isn't what this test is trying to test at all.  You aren't worried about the weights and we can take as given that the capacities are already correct and set.  I was confused about the seemingly arbitrary numbers in deficits but I was missing the point

PR #752, opened 2023-05-04
[LSM] Increment Validator Slash Query Progress

sampocs:
## Context
There's a progress tracker on the validator that dictates when to issue the slash ICQs. This progress number should be incremented after the LSM Liquid stake has succeeded.

## Changelog
* Added utility function `IncrementValidatorSlashQueryProgress`
* Incremented progress in `FinishLSMLiquidStake`
* Added/updated unit tests


riley-stride (approved):
lgtm

PR #756, opened 2023-05-05
[LSM] Updated Delegation in Detokenize Callback

sampocs:
In the detokenize callback, we have to increment the delegation on the validator and the total delegation amount for the host zone

## Brief changelog
* Cleaned up AddDelegationToValidator and added unit test
* Updated delegation in detokenize callback and added corresponding case in unit test

### DIFF
--- a/x/stakeibc/keeper/host_zone.go
+++ b/x/stakeibc/keeper/host_zone.go
@@ -101,28 +101,34 @@ func (k Keeper) GetAllActiveHostZone(ctx sdk.Context) (list []types.HostZone) {
 	return
 }
 
-func (k Keeper) AddDelegationToValidator(ctx sdk.Context, hostZone types.HostZone, validatorAddress string, amount sdkmath.Int, callbackId string) (success bool) {
+// Updates a validator's delegation
+// Note: This updates modifies the original validator struct
+func (k Keeper) AddDelegationToValidator(
+	ctx sdk.Context,
+	hostZone types.HostZone,
+	validatorAddress string,
+	amount sdkmath.Int,
+	callbackId string,
+) error {
 	for _, validator := range hostZone.Validators {
 		if validator.Address == validatorAddress {
 			k.Logger(ctx).Info(utils.LogICACallbackWithHostZone(hostZone.ChainId, callbackId,
 				"  Validator %s, Current Delegation: %v, Delegation Change: %v", validator.Address, validator.Delegation, amount))
 
-			if amount.GTE(sdkmath.ZeroInt()) {
-				validator.Delegation = validator.Delegation.Add(amount)
-				return true
+			// If the delegation change is negative, make sure it wont cause the delegation to fall below zero
+			if amount.IsNegative() && amount.Abs().GT(validator.Delegation) {
+				return errorsmod.Wrapf(types.ErrValidatorDelegationChg,
+					"Delegation change (%v) is greater than validator (%s) delegation %v",
+					amount.Abs(), validatorAddress, validator.Delegation)
 			}
-			absAmt := amount.Abs()
-			if absAmt.GT(validator.Delegation) {
-				k.Logger(ctx).Error(fmt.Sprintf("Delegation amount %v is greater than validator %s delegation amount %v", absAmt, validatorAddress, validator.Delegation))
-				return false
-			}
-			validator.Delegation = validator.Delegation.Sub(absAmt)
-			return true
+
+			validator.Delegation = validator.Delegation.Add(amount)
+			return nil
 		}
 	}
 
-	k.Logger(ctx).Error(fmt.Sprintf("Could not find validator %s on host zone %s", validatorAddress, hostZone.ChainId))
-	return false
+	return errorsmod.Wrapf(types.ErrValidatorNotFound,
+		"Could not find validator %s on host zone %s", validatorAddress, hostZone.ChainId)
 }
 
 // Increments the validators slash query progress tracker
--- a/x/stakeibc/keeper/host_zone.go
+++ b/x/stakeibc/keeper/host_zone.go
@@ -101,11 +101,12 @@ func (k Keeper) GetAllActiveHostZone(ctx sdk.Context) (list []types.HostZone) {
 	return
 }
 
-// Updates a validator's delegation
-// Note: This updates modifies the original validator struct
+// Updates a validator's individual delegation, and the corresponding total delegation on the host zone
+// Note: This modifies the original host zone struct. The calling function must Set this host zone
+// for changes to persist
 func (k Keeper) AddDelegationToValidator(
 	ctx sdk.Context,
-	hostZone types.HostZone,
+	hostZone *types.HostZone,
 	validatorAddress string,
 	amount sdkmath.Int,
 	callbackId string,
@@ -116,13 +117,22 @@ func (k Keeper) AddDelegationToValidator(
 				"  Validator %s, Current Delegation: %v, Delegation Change: %v", validator.Address, validator.Delegation, amount))
 
 			// If the delegation change is negative, make sure it wont cause the delegation to fall below zero
-			if amount.IsNegative() && amount.Abs().GT(validator.Delegation) {
-				return errorsmod.Wrapf(types.ErrValidatorDelegationChg,
-					"Delegation change (%v) is greater than validator (%s) delegation %v",
-					amount.Abs(), validatorAddress, validator.Delegation)
+			if amount.IsNegative() {
+				if amount.Abs().GT(validator.Delegation) {
+					return errorsmod.Wrapf(types.ErrValidatorDelegationChg,
+						"Delegation change (%v) is greater than validator (%s) delegation %v",
+						amount.Abs(), validatorAddress, validator.Delegation)
+				}
+				if amount.Abs().GT(hostZone.TotalDelegations) {
+					return errorsmod.Wrapf(types.ErrValidatorDelegationChg,
+						"Delegation change (%v) is greater than total delegation amount on host %s (%v)",
+						amount.Abs(), hostZone.ChainId, hostZone.TotalDelegations)
+				}
 			}
 
 			validator.Delegation = validator.Delegation.Add(amount)
+			hostZone.TotalDelegations = hostZone.TotalDelegations.Add(amount)
+
 			return nil
 		}
 	}
--- a/x/stakeibc/keeper/host_zone_test.go
+++ b/x/stakeibc/keeper/host_zone_test.go
@@ -216,3 +216,32 @@ func (s *KeeperTestSuite) TestIncrementValidatorSlashQueryProgress() {
 	err = s.App.StakeibcKeeper.IncrementValidatorSlashQueryProgress(s.Ctx, HostChainId, "fake_val", stakeAmount)
 	s.Require().ErrorContains(err, "validator not found")
 }
+
+func (s *KeeperTestSuite) TestAddDelegationToValidator() {
+	hostZone := types.HostZone{
+		Validators: []*types.Validator{
+			{Address: "other_val", Delegation: sdkmath.NewInt(1000)},
+			{Address: ValAddress, Delegation: sdkmath.NewInt(3000)},
+			{Address: "other_val2", Delegation: sdkmath.NewInt(3000)},
+		},
+	}
+	updatedIndex := 1
+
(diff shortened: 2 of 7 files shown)


x/stakeibc/keeper/icacallbacks_detokenize.go
```diff
@@ -46,12 +46,23 @@ func DetokenizeCallback(k Keeper, ctx sdk.Context, packet channeltypes.Packet, a
...
 	deposit := detokenizeCallback.Deposit
 	k.RecordsKeeper.RemoveLSMTokenDeposit(ctx, deposit.ChainId, deposit.Denom)
 
 	// Update delegation on the host zone and validator
-	// TODO [LSM] - Use helper function
+	err := k.AddDelegationToValidator(ctx, hostZone, deposit.ValidatorAddress, deposit.Amount, ICACallbackID_Detokenize)
+	if err != nil {
+		return err
+	}
+	hostZone.TotalDelegations = hostZone.TotalDelegations.Add(deposit.Amount)
```
riley-stride, line 64: Wdyt about putting this update inside `AddDelegationToValidator`? I think we would always want to update them in tandem, but might be overlooking an edge case

x/stakeibc/keeper/icacallbacks_delegate.go
```diff
@@ -92,9 +92,9 @@ func DelegateCallback(k Keeper, ctx sdk.Context, packet channeltypes.Packet, ack
 	// Update delegations on the host zone
 	for _, splitDelegation := range delegateCallback.SplitDelegations {
 		hostZone.TotalDelegations = hostZone.TotalDelegations.Add(splitDelegation.Amount)
-		success := k.AddDelegationToValidator(ctx, hostZone, splitDelegation.Validator, splitDelegation.Amount, ICACallbackID_Delegate)
-		if !success {
-			return errorsmod.Wrapf(types.ErrValidatorDelegationChg, "Failed to add delegation to validator")
+		err := k.AddDelegationToValidator(ctx, hostZone, splitDelegation.Validator, splitDelegation.Amount, ICACallbackID_Delegate)
```
riley-stride, line 95: Shouldn't we increment TotalDelegations here? If so, seems like a good reason to nest that update inside `AddDelegationToValidator`

riley-stride: If so, we should add a check on TotalDelegations to the `TestAddDelegationToValidator` unit test too

x/stakeibc/keeper/icacallbacks_detokenize_test.go
```diff
@@ -28,22 +54,32 @@ func (s *KeeperTestSuite) SetupTestDetokenizeCallback() []byte {
...
 
 	// Check that the deposit was removed
 	_, found := s.App.RecordsKeeper.GetLSMTokenDeposit(s.Ctx, HostChainId, LSMTokenBaseDenom)
 	s.Require().False(found, "deposit should have been removed")
+
+	// Check that the delegation was updated
+	hostZone, found := s.App.StakeibcKeeper.GetHostZone(s.Ctx, HostChainId)
+	s.Require().True(found, "host zone should have been found")
+	s.Require().Equal(tc.expectedTotalDelegation, hostZone.TotalDelegations.Int64(), "host zone total delegation")
+	s.Require().Equal(tc.expectedValidatorDelegation, hostZone.Validators[0].Delegation.Int64(), "validator delegation")
```
riley-stride, line 82: Nit: could we add another validator to this test setup and here check that its value remains unchanged?

riley-stride (changes requested):
Main questions is whether we should increment `TotalDelegations` within `AddDelegationToValidator`, and whether we need that update in `icacallbacks_delegate.go`

x/stakeibc/keeper/icacallbacks_delegate.go
sampocs: yeah see my other message, it seems a bit trickier, but I agree it would be safer to update both, so I'll give it another look

riley-stride: replied in the other comment

x/stakeibc/keeper/icacallbacks_detokenize_test.go
sampocs: done

x/stakeibc/keeper/icacallbacks_detokenize.go
sampocs: I don't think we need it now. Just putting it on the radar!

x/stakeibc/keeper/icacallbacks_detokenize_test.go
riley-stride: sweet, resolving!

riley-stride:
> @riley-stride pointer approach seems to work well.
> 
> Since this helper is used to both add and subtract an amount from the delegation, should we rename it from `AddDelegationToValidator` to `UpdateValidatorDelegation`? Or do you think that `Add` is fine cause it implies that passing a negative amount will decrement?

I like `Add` because the negative case is implied (as you stated) and `Update` could refer to any generic operation, not necessarily arithmetic.

x/stakeibc/keeper/host_zone_test.go
```diff
@@ -248,4 +251,11 @@ func (s *KeeperTestSuite) TestAddDelegationToValidator() {
 	// Attempt to modify a validator that doesn't exist - it should fail
 	err = s.App.StakeibcKeeper.AddDelegationToValidator(s.Ctx, hostZone, "does_not_exist", sdkmath.NewInt(1000), "")
 	s.Require().ErrorContains(err, "validator not found")
+
+	// Attempt to subtract more than the total delegations on the host - it should fail
```
riley-stride, line 255: Can you explain this test? My understanding is that `AddDelegationToValidator` is only called when new delegations are added to a validator

x/stakeibc/keeper/host_zone_test.go
```diff
@@ -248,4 +251,11 @@ func (s *KeeperTestSuite) TestAddDelegationToValidator() {
...
+	// Here, w set the validator's delegation to be much higher than the TotalDelegation
+	//   (which should not be possible in practice)
+	hostZone.Validators[updatedIndex].Delegation = sdkmath.NewInt(10000)
+	err = s.App.StakeibcKeeper.AddDelegationToValidator(s.Ctx, hostZone, ValAddress, sdkmath.NewInt(-7000), "")
+	s.Require().ErrorContains(err, "Delegation change (7000) is greater than total delegation amount on host")
```
riley-stride, line 260: Wouldn't the `amount.Abs().GT(validator.Delegation)` condition trigger before the `amount.Abs().GT(hostZone.TotalDelegations)` condition in this case?

x/stakeibc/keeper/host_zone_test.go
```diff
@@ -248,4 +251,11 @@ func (s *KeeperTestSuite) TestAddDelegationToValidator() {
 	// Attempt to modify a validator that doesn't exist - it should fail
 	err = s.App.StakeibcKeeper.AddDelegationToValidator(s.Ctx, hostZone, "does_not_exist", sdkmath.NewInt(1000), "")
 	s.Require().ErrorContains(err, "validator not found")
+
+	// Attempt to subtract more than the total delegations on the host - it should fail
+	// Here, w set the validator's delegation to be much higher than the TotalDelegation
+	//   (which should not be possible in practice)
+	hostZone.Validators[updatedIndex].Delegation = sdkmath.NewInt(10000)
```
riley-stride, line 258: Isn't this an impossible state to reach, where the TotalDelegations and sum of Delegations are out of sync

riley-stride (commented):
Main think I don't understand is the TotalDelegations vs Delegation error handling and test case

x/stakeibc/keeper/host_zone.go
```diff
@@ -116,13 +117,22 @@ func (k Keeper) AddDelegationToValidator(
...
-				return errorsmod.Wrapf(types.ErrValidatorDelegationChg,
-					"Delegation change (%v) is greater than validator (%s) delegation %v",
-					amount.Abs(), validatorAddress, validator.Delegation)
+			if amount.IsNegative() {
+				if amount.Abs().GT(validator.Delegation) {
+					return errorsmod.Wrapf(types.ErrValidatorDelegationChg,
+						"Delegation change (%v) is greater than validator (%s) delegation %v",
+						amount.Abs(), validatorAddress, validator.Delegation)
+				}
+				if amount.Abs().GT(hostZone.TotalDelegations) {
```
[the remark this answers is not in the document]
sampocs: yeah it's not really required but it was in our unbonding flow and I think it was mentioned in one of the audits.

I agree in practice it should never be tripped, but probably doesn't hurt

x/stakeibc/keeper/host_zone_test.go
sampocs: yeah see previous comment

riley-stride: fair enough, can't hurt to keep the extra safety check in

x/stakeibc/keeper/host_zone.go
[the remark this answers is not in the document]
riley-stride: fair enough, can't hurt to keep the extra safety check in

riley-stride (approved):

sampocs:
@riley-stride 

> I like Add because the negative case is implied (as you stated) and Update could refer to any generic operation, not necessarily arithmetic.

Well said, I agree!

x/stakeibc/keeper/icacallbacks_detokenize.go
```diff
@@ -46,12 +46,22 @@ func DetokenizeCallback(k Keeper, ctx sdk.Context, packet channeltypes.Packet, a
...
+		return errorsmod.Wrapf(types.ErrHostZoneNotFound, "Host zone not found: %s", chainId)
+	}
+
 	// If the ICA succeeded, remove the token deposit
 	deposit := detokenizeCallback.Deposit
 	k.RecordsKeeper.RemoveLSMTokenDeposit(ctx, deposit.ChainId, deposit.Denom)
 
 	// Update delegation on the host zone and validator
-	// TODO [LSM] - Use helper function
+	err := k.AddDelegationToValidator(ctx, &hostZone, deposit.ValidatorAddress, deposit.Amount, ICACallbackID_Detokenize)
```
ethan-stride, line 60: This is the only place in the codebase I have seen this pattern (antipattern?) of passing the object by reference and updating it as a side effect in the function instead of just returning a changed version of the object.  Is it just because hostZone is big and complex or is there a reason to do it like this?  Why not also setHostZone inside the addDelegationToValidator then?

ethan-stride (commented):
LGTM

x/stakeibc/keeper/icacallbacks_detokenize.go
sampocs: Great question! I personally dislike this pattern, but decided on it for the following reasons:
a) most functions loop and modify multiple validators in a row so it's slightly more efficient to just do one HostZone set at the end (although this probably doesn't matter much)
b) more importantly, since this is often called from epochs (which runs in the EndBlocker), I was hesitant to Set prematurely as the state wouldn't get reverted (however, there is a function that we'll integrate soon that will let us do the reversion)
c) it was set up this way originally so, while it's not perfect right now, it's at least not worse haha

All that said, I think the ideal approach is to do (a) the refactor I mentioned before where we have a separate prefix store for validators and (b) add in the error catching in the EndBlocker to revert state - and then come back and clean this up once those are in place

x/stakeibc/keeper/host_zone.go
```diff
@@ -101,28 +101,44 @@ func (k Keeper) GetAllActiveHostZone(ctx sdk.Context) (list []types.HostZone) {
 	return
 }
 
-func (k Keeper) AddDelegationToValidator(ctx sdk.Context, hostZone types.HostZone, validatorAddress string, amount sdkmath.Int, callbackId string) (success bool) {
+// Updates a validator's individual delegation, and the corresponding total delegation on the host zone
```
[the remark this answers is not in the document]
ethan-stride: We debated this to death, it ultimately doesn't matter that much.  It is unlikely to get out of sync even without helpers to set it in the different places.  As Sam pointed out, having it in the queries so people can see this number externally is useful so might as well leave it alone.

PR #757, opened 2023-05-05
[LSM] Retry Interchain Queries

sampocs:
Now that interchain queries can be UX blocking, it's important that we resubmit them after a timeout. A retry is also net positive so I don't see why we wouldn't want to do it for the other queries. 

## Brief Changelog
* Renamed field `Timeout` to `TimeoutTimestamp` in the query struct (this represents the timeout as a unix timestamp)
* Added field `TimeoutDuration` to query struct which dictates the length of the timeout (i.e. the time between the current block time and the timeout timestamp)
* Updated the endblocker logic in abci to requeue queries if they've been timed out
* Updated MsgSubmitQueryResponse to not delete timeout queries (so they can be re-submitted from the above)
* Modified SubmitICQRequest to set the timeout timestamp based on the current block time and timeout duration. Previously the timeout timestamp was specified when submitting a query - now the duration is specified instead.
* Changed epochly queries to timeout in 1 hour (since they'll get automatically resubmitted)
* Added a helper function to retry the ICQ which resets the timeout timestamp (using the timeout duration) 
* Updated relevant query submissions and unit tests


x/interchainquery/keeper/abci.go
```diff
@@ -40,6 +41,12 @@ func (k Keeper) EndBlocker(ctx sdk.Context) {
 			query.RequestSent = true
 			k.SetQuery(ctx, query)
 		}
+		// Re-queue timed-out queries
+		if query.TimeoutTimestamp < uint64(ctx.BlockTime().UnixNano()) {
+			if err := k.RetryICQRequest(ctx, query); err != nil {
```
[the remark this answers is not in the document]
riley-stride: Nevermind you icrement the `query.TimeoutTimestamp` inside `RetryICQRequest`, looks good to me!

x/interchainquery/keeper/keeper.go
```diff
@@ -67,3 +71,20 @@ func (k *Keeper) SubmitICQRequest(ctx sdk.Context, query types.Query, forceUniqu
 
 	return nil
 }
+
+// Re-submit an ICQ, generally used after a timeout
+func (k *Keeper) RetryICQRequest(ctx sdk.Context, query types.Query) error {
```
riley-stride, line 76: Seems like this code is mostly a copy paste from `SubmitICQRequest`, should we pull it out of both fns into a helper? I personally think it is more clear as written, just a bit redundant

sampocs: I had that same debate as well, but there's actually 1 big difference - SubmitICQRequest generates the query ID, whereas retry uses the same query ID

x/stakeibc/keeper/icacallbacks_reinvest.go
```diff
@@ -110,23 +111,17 @@ func ReinvestCallback(k Keeper, ctx sdk.Context, packet channeltypes.Packet, ack
...
-		CallbackModule: types.ModuleName,
-		CallbackId:     ICQCallbackID_FeeBalance,
-		Timeout:        timeout,
+		ChainId:         chainId,
+		ConnectionId:    hostZone.ConnectionId,
+		QueryType:       icqtypes.BANK_STORE_QUERY_WITH_PROOF,
+		RequestData:     queryData,
+		CallbackModule:  types.ModuleName,
+		CallbackId:      ICQCallbackID_FeeBalance,
+		TimeoutDuration: time.Hour,
```
riley-stride, line 124: Out of curiosity how'd you choose this?

riley-stride (approved):
lgtm

sampocs:
> Is your thinking for why we can retry ICQs at any time during the epoch that they have no interaction with other ICAs or record keeping? the only case where they interact afaik is we begin with N tokens in the withdrawal acct, then a new delegation ICA triggers passive accounting of M accrued rewards into the withdrawal acct. that could occur simultaneous to the ICQ or while the callback is in flight, but that's OK because the ICQ callback will just sweep N tokens, and the M will remain in the withdrawal acct for the next ICQ. So all is safe, no adverse interactions.

Yup that was my thinking! I think the only query that interacts with ICAs is the delegator shares query at this point and we have safety guards on that.

I'm glad you brought up the withdrawal account case, I actually hadn't thought about that - but agreed it should be all good!

x/interchainquery/keeper/keeper.go
riley-stride: ah good point, makes sense to keep as is then

x/stakeibc/keeper/icacallbacks_reinvest.go
riley-stride: yeah retrying 6x/epoch seems like a reasonable cadence to me

x/interchainquery/keeper/msg_server.go
```diff
@@ -150,20 +150,21 @@ func (k msgServer) SubmitQueryResponse(goCtx context.Context, msg *types.MsgSubm
...
+	if query.TimeoutTimestamp < currBlockTime {
 		k.Logger(ctx).Error(utils.LogICQCallbackWithHostZone(query.ChainId, query.CallbackId,
-			"QUERY TIMEOUT - QueryId: %s, TTL: %d, BlockTime: %d", query.Id, query.Timeout, ctx.BlockHeader().Time.UnixNano()))
+			"QUERY TIMEOUT - QueryId: %s, TTL: %d, BlockTime: %d", query.Id, query.TimeoutTimestamp, ctx.BlockHeader().Time.UnixNano()))
 		return &types.MsgSubmitQueryResponseResponse{}, nil
 	}
 
+	// Immediately delete the query so it cannot process again
+	// We don't want to delete timed-out queries because they'll be retried
+	k.DeleteQuery(ctx, query.Id)
```
ethan-stride, line 166: Does this also delete the callback data associated with the icq?

proto/stride/interchainquery/v1/genesis.proto
```diff
@@ -15,7 +16,9 @@ message Query {
   string callback_module = 13;
   string callback_id = 8;
```
ethan-stride, line 17: When we decide to retry the query do we have to change the callback id?  Isn't the normal flow the icq handler follows that it catches the response, looksup the callback and data we prepared before we sent the query, executes whatever function is attached to the listener given the marshalled callback data, then it removes the callback data... so now we want to retry and we want the same callback data to still be there in case the icq works the second time.  Are we sure the callback won't have been removed on the first handling?

ethan-stride (commented):
Tried to leave helpful comments but really I think I need to learn more about the ICQ flow.  Nothing looks blatantly wrong but I don't yet have the full picture of how the listener to the response from ICQ result handles callback data

x/interchainquery/keeper/abci.go
```diff
@@ -40,6 +41,12 @@ func (k Keeper) EndBlocker(ctx sdk.Context) {
 			query.RequestSent = true
```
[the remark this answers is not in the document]
sampocs: this is the only place that sets a query with RequestSent = true

x/interchainquery/keeper/abci.go
```diff
@@ -18,6 +18,7 @@ func (k Keeper) EndBlocker(ctx sdk.Context) {
 	events := sdk.Events{}
 	// emit events for periodic queries
 	k.IterateQueries(ctx, func(_ int64, query types.Query) (stop bool) {
+		// Submit new queries
 		if !query.RequestSent {
 			k.Logger(ctx).Info(fmt.Sprintf("Interchainquery event emitted %s", query.Id))
 			// QUESTION: Do we need to emit this event twice?
```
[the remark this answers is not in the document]
sampocs: the block time is static through the EndBlocker, but you're on to something - I think we should return from the RequestSent branch

x/interchainquery/keeper/msg_server.go
sampocs: yup!

proto/stride/interchainquery/v1/genesis.proto
sampocs: man i see how this is confusing 😞 

ICQ callbacks are handled differently than ICA callbacks. For ICA callbacks you're right, but for ICQ callbacks, the callback data is just stored on the query object and the query response submission passes a query ID that does the lookup

x/interchainquery/keeper/abci.go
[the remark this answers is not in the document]
ethan-stride: Yeah but that isn't the issue, the real question is could there be anywhere else that we but the query back into the keeper with a future timeout, ie. we intend to retry the query again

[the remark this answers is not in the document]
sampocs: no, there shouldn't be. If retrying the query, the [helper](https://github.com/Stride-Labs/stride/pull/757/files#diff-92e4e0bd0be2857299d6eaf8265f75c848758e93c0e2f05d40aa9be5ac45ad55R76) function will be used

PR #758, opened 2023-05-05
[LSM] Added unit test for RebalanceDelegationsForHostZone

sampocs:
## Brief Changelog
* Added unit tests for RebalanceDelegationsForHostZone
* Removed unit tests in msg_server_rebalance_validators as they're redundant
* Moved setup function that was in `msg_server_rebalance_validators` and needed by the rebalance callback into the callback itself


x/stakeibc/keeper/validator_selection_test.go
```diff
@@ -1,18 +1,159 @@
...
+	"fmt"
 	"math/rand"
 
 	sdkmath "cosmossdk.io/math"
+	"github.com/gogo/protobuf/proto" //nolint:staticcheck
 
 	sdk "github.com/cosmos/cosmos-sdk/types"
+	ibctesting "github.com/cosmos/ibc-go/v5/testing"
+
 	// TODO [LSM]: Revert type
```
ethan-stride, line 13: What is this comment about?

x/stakeibc/keeper/validator_selection_test.go
```diff
@@ -1,18 +1,159 @@
...
+	strideEpochTracker := types.EpochTracker{
+		EpochIdentifier:    epochtypes.STRIDE_EPOCH,
+		EpochNumber:        1,
+		NextEpochStartTime: uint64(s.Coordinator.CurrentTime.UnixNano() + 30_000_000_000), // dictates timeouts
+	}
+
+	s.App.StakeibcKeeper.SetEpochTracker(s.Ctx, strideEpochTracker)
+
+	// Build expected redelegation messages
+	expectedRebalancings := []types.Rebalancing{
```
ethan-stride, line 62: Here we only have two rebalancing messages but in reality we batch these when there are many,  Do you want to make a test that verifies the batched sends and their callbacks don't get messed up?

ethan-stride (approved):
LGTM

x/stakeibc/keeper/validator_selection_test.go
sampocs: its cause we had to change to proto types to the ones from LSM

x/stakeibc/keeper/validator_selection_test.go
sampocs: yeah that's a good call

sampocs: done

riley-stride (approved):
LGTM

PR #759, opened 2023-05-07
[LSM] Detokenization ICA/ICQ Overlap Check

sampocs:
Closes: #XXX

See #745 for full context. When checking if there was overlap between a slash ICQ and an ICA that changes the delegation on a validator, we need to also check for detokenization ICAs. Delegation and undelegation ICAs were already checked in #745.

It's worth mentioning that new LSM liquid stakes are halted while a slash query is in progress, so there is no potential for an infinite retry loop with this approach.

## Brief Changelog
* Added check for detokenization in progress records when determining if there was ICA/ICQ overlap
* Added unit test

## Author's Checklist

I have...

## Reviewers Checklist

I have...

      - [ ] not applicable
      - [ ] jira ticket `XXX` 
      - [ ] README.md 
      - [ ] not documented

### DIFF
--- a/x/stakeibc/keeper/icqcallbacks_delegator_shares.go
+++ b/x/stakeibc/keeper/icqcallbacks_delegator_shares.go
@@ -112,7 +112,8 @@ func DelegatorSharesCallback(k Keeper, ctx sdk.Context, args []byte, query icqty
 // The number of tokens returned from the query must be consistent with the tokens
 //   stored in our internal record keeping during this callback, otherwise the comparision
 //   between the two is invalidated
-// As a result, we must avoid a race condition between the ICQ and an delegate or undelegate ICA
+// As a result, we must avoid a race condition between the ICQ and a delegate, undelegate,
+//   or detokenization ICA
 //
 // More specifically, we must avoid the following cases:
 //  Case 1)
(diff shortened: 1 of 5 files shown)


x/stakeibc/keeper/icqcallbacks_delegator_shares.go
```diff
@@ -154,6 +163,15 @@ func (k Keeper) CheckDelegationChangedDuringQuery(
 		}
 	}
 
+	// Check that there are no LSMTokenDeposits in state IN_PROGRESS - indictive of an Detokenization ICA
+	for _, lsmTokenDeposit := range k.RecordsKeeper.GetLSMDepositsForHostZone(ctx, chainId) {
```
ethan-stride, line 167: Use GetLSMDepositsForHostZoneWithStatus because then if there are really none in progress, you loop 0 times

x/stakeibc/keeper/icqcallbacks_delegator_shares.go
```diff
@@ -62,28 +62,35 @@ func DelegatorSharesCallback(k Keeper, ctx sdk.Context, args []byte, query icqty
...
-		validator.SlashQueryInProgress = false
-		hostZone.Validators[valIndex] = &validator
-		k.SetHostZone(ctx, hostZone)
 	}
 
+	// If there was no ICA/ICQ overlap, update the validator to indicate that the query
+	//  is no longer in progress (which will unblock LSM liquid stakes to that validator)
+	validator.SlashQueryInProgress = false
+	hostZone.Validators[valIndex] = &validator
+	k.SetHostZone(ctx, hostZone)
```
ethan-stride, line 92: Not for this PR, but this is evidence we should do the refactor to change the way validators are stored in a prefix store. Just to update a property on the validator takes 3 lines of code is asking for mistakes to be made.

ethan-stride (commented):
LGTM

x/stakeibc/keeper/icqcallbacks_delegator_shares.go
sampocs: ah good call, forgot about that one!

x/stakeibc/keeper/icqcallbacks_delegator_shares.go
sampocs: amen to that

riley-stride (approved):
lgtm

PR #760, opened 2023-05-07
[LSM] Slash Query Interval Check Relative to TVL

sampocs:
Closes: #XXX

At fixed intervals, we must query validator slashes to make sure we are not minting stTokens at an incorrect rate. Previously, this checkpoint was specified as a fixed number of native tokens (e.g. 100k, meaning every 100k ATOM that's LSM Liquid staked to a validator, a query would be issued). However, this solution does not scale well to multiple host zones as the same number of tokens between chains does not imply the same dollar value, and the price of a token is constantly moving (which could potentially require an updated checkpoint value if the token price changed dramatically).

A more elegant solution is to define this checkpoint as a percentage of TVL (e.g. 1% implies every time 1% of the total TVL has been staked, the query is issued).


x/stakeibc/keeper/lsm.go
```diff
@@ -160,24 +160,32 @@ func (k Keeper) GetValidatorFromLSMTokenDenom(denom string, validators []*types.
...
+// (e.g. 1% means every LS that causes the progress to breach 1% of TVL triggers the query)
+func (k Keeper) ShouldCheckIfValidatorWasSlashed(
+	ctx sdk.Context,
+	validator types.Validator,
+	totalHostZoneStake sdkmath.Int,
+	transactionStakeAmount sdkmath.Int,
+) bool {
 	params := k.GetParams(ctx)
-	queryInterval := sdk.NewIntFromUint64(params.ValidatorSlashQueryInterval)
+	queryInterval := sdk.NewDecWithPrec(int64(params.ValidatorSlashQueryInterval), 2) // percentage
```
ethan-stride, line 172: It wasn't clear until I read through the whole PR how this number is supposed to be, even the comment afterwards saying this is a percentage.  If I am understanding right, this is going to be something like 0.01 to indicate every 1% of the total locked will trigger a new check.

Why doing precision 2?  What if we want to change from 1% to 1.5% so we need 0.015 but then won't prec of 2 give us either 0.01 or 0.02?

x/stakeibc/keeper/lsm.go
```diff
@@ -160,24 +160,32 @@ func (k Keeper) GetValidatorFromLSMTokenDenom(denom string, validators []*types.
...
 
 	oldProgress := validator.SlashQueryProgressTracker
-	newProgress := validator.SlashQueryProgressTracker.Add(stakeAmount)
+	newProgress := validator.SlashQueryProgressTracker.Add(transactionStakeAmount)
 
 	// Submit query if the query interval checkpoint has been breached
 	// Ex: Query Interval: 1000, Old Progress: 900, New Progress: 1100
 	//     => OldProgress/Interval: 0, NewProgress/Interval: 1
-	return oldProgress.Quo(queryInterval).LT(newProgress.Quo(queryInterval))
+	return oldProgress.Quo(checkpoint).LT(newProgress.Quo(checkpoint))
```
ethan-stride, line 188: I don't think this algorithm will work exactly the same way as before, because originally queryInterval was a fixed number but now checkpoint changes with each time we call this function.  It is possible that an amount which would have pushed us over the limit has also increased the checkpoint step size enough that a check doesn't happen.

ethan-stride: Just re-read the way this function gets called, so the total amount on the host zone doesn't get updated until the callback, so this function doing the check at the start isn't going to have the size of "checkpoint" directly affected by the current deposit happening now.  It is still a bit hard to think about edge cases since checkpoint is still changing in size with each call which was not the case before.  My intuition is that with a queryInterval that is small like 1-2% this still works as intended but with a large enough query interval we could have unexpected behavior.  I'll work out some numerical examples

x/stakeibc/types/params.go
```diff
@@ -27,7 +27,7 @@ var (
 	DefaultIBCTransferTimeoutNanos     uint64 = 1800000000000 // 30 minutes
 	DefaultSafetyNumValidators         uint64 = 35
 	DefaultSafetyMaxSlashPercent       uint64 = 10
-	DefaultValidatorSlashQueryInterval uint64 = 100_000_000_000 // denominated in number of native tokens
+	DefaultValidatorSlashQueryInterval uint64 = 1 // denominated in percentage of TVL (1 => 5%)
```
ethan-stride, line 30: (1 => 5%)???
Is this a typo and it was supposed to be (5 => 5%)

ethan-stride (commented):
Still thinking about if the algorithm is still sound given that the checkpoint is changing with every deposit... if it is then this looks good.

x/stakeibc/keeper/lsm.go
```diff
@@ -160,24 +160,32 @@ func (k Keeper) GetValidatorFromLSMTokenDenom(denom string, validators []*types.
...
+func (k Keeper) ShouldCheckIfValidatorWasSlashed(
+	ctx sdk.Context,
+	validator types.Validator,
+	totalHostZoneStake sdkmath.Int,
+	transactionStakeAmount sdkmath.Int,
+) bool {
 	params := k.GetParams(ctx)
-	queryInterval := sdk.NewIntFromUint64(params.ValidatorSlashQueryInterval)
+	queryInterval := sdk.NewDecWithPrec(int64(params.ValidatorSlashQueryInterval), 2) // percentage
+	checkpoint := queryInterval.Mul(sdk.NewDecFromInt(totalHostZoneStake)).TruncateInt()
```
ethan-stride, line 173: One other issue, the totalHostZoneStake can go up but also can go down if people choose to unbond. So imagine oldProgress and newProgress which are both quantities that only increase from transactions adding stake would have been on different sides of the line like the example in the comment below, but a change in the totalHostZoneStake going down now makes them both on the same side...

ex:  previous totalHostZoneStake 100000 --> previous checkpoint 1000
oldprogress 900 new progress 1100
but before this staking transaction someone unbonded 10000 making 
current totalHostZoneStake 90000 --> current checkpoint 900
then oldprgress/checkpoint = 1 and newprogress/checkpoint = 1.222 => 1 on int truncate
So this example should have triggered but because of the unbond it would not!!!

x/stakeibc/keeper/lsm.go
sampocs: we unfortunately are limited to storing our params as integers right now. Not sure how it got that way, but it was likely something from ignite. It's on by list of future TODOs to allow decimal types

That said, I don't think we need any additional precision and all the other params follow a similar standard for percents so I think it would be confusing if we instead had a param set to "10" which really mean 0.010

x/stakeibc/keeper/lsm.go
sampocs: yeah we were just discussing the same thing in a thread on slack (I'll tag you).
TLDR, is that if the interval is small (which it will be), the checkpoint does not move much and is technically more "correct" as the TVL has changed

x/stakeibc/types/params.go
sampocs: yup sorry typo. fixing now

x/stakeibc/keeper/lsm.go
riley-stride: It's a good callout. I would err on the side of clarity rather than granularity here, since we prob won't need fractional percentage point precision e.g. 1.5%

x/stakeibc/keeper/msg_server_lsm_liquid_stake_test.go
```diff
@@ -128,15 +144,17 @@ func (s *KeeperTestSuite) TestLSMLiquidStake_Successful_NoExchangeRateQuery() {
...
-	s.Require().Equal(int64(9_000_000), hostZone.Validators[0].SlashQueryProgressTracker.Int64(), "slash query progress")
+	s.Require().Equal(expectedQueryProgress.Int64(), hostZone.Validators[0].SlashQueryProgressTracker.Int64(), "slash query progress")
 }
 
 func (s *KeeperTestSuite) TestLSMLiquidStake_Successful_WithExchangeRateQuery() {
 	tc := s.SetupTestLSMLiquidStake()
 
 	// Increase the liquid stake size so that it breaks the query checkpoint
-	tc.validMsg.Amount = sdk.NewInt(3_000_000)
+	queryProgressSlack := tc.queryCheckpoint.Sub(tc.initialQueryProgress) // largest stake amount before query
```
riley-stride, line 156: nit
```suggestion
	queryProgressSlack := tc.queryCheckpoint.Sub(tc.initialQueryProgress) // remaining amount that can be staked in one message before a slash query is issued
```

x/stakeibc/keeper/lsm_test.go
```diff
@@ -262,7 +311,12 @@ func (s *KeeperTestSuite) TestShouldCheckIfValidatorWasSlashed() {
 		s.App.StakeibcKeeper.SetParams(s.Ctx, params)
 
 		validator := types.Validator{SlashQueryProgressTracker: tc.progress}
-		actualShouldQuery := s.App.StakeibcKeeper.ShouldCheckIfValidatorWasSlashed(s.Ctx, validator, tc.stakeAmount)
+		actualShouldQuery := s.App.StakeibcKeeper.ShouldCheckIfValidatorWasSlashed(
```
riley-stride, line 314: Can we test the case where Bob can give himself a little more slack by liquid staking bit by bit, but really not that much, e.g.
- TVL is 1000
- Bob liquid stake 99, which brings TVL to 1099, no query is issued. 
- Bob liquid stakes 10 more, no query is triggered. Checkpoint math: 99+10 = 109
- Bob liquid stakes 10 more, which issues the query

riley-stride (approved):
Once a test for the "bit by bit" Bob liquid staking case is added I'm signed off on this. Clean code!

x/stakeibc/keeper/lsm.go
riley-stride: It's a good flag that the checkpoint can be reduced via unbondings.

Could you expand on this example? I rewrote my understanding of your example. As it's written below I think it's expected behavior. Basically unbondings can lower the checkpoint.

ex: I think in your example someone is liquid staking 200 (based on "oldprogress 900 new progress 1100")
- Progress 900
- totalHostZoneStake 100000 --> checkpoint 1000
- Unbonding 10000
- totalHostZoneStake 90000 --> checkpoint 900
- LSM liquid stake 200 -> Progress 1100 ->  newprogress/checkpoint = 1.22 -> Slash Query issued

ethan-stride: My point is that Slash Query would not get issued.  It is only issued when the old/checkpoint is different from new/checkpoint but because of the change in checkpoint, now both of these are 1

ethan-stride: The way to fix these issues is to calculate once and store the value for checkpoint at the time of the last slash query instead of computing the checkpoint dynamically each time we do the check.

sampocs: yeah this is a good point, lemme think on it for a bit

sampocs: So while I do think it's possible to miss a query because of this, I don't think this is a huge concern considering the following: 
  1.  The TVL only has the potential to decrease when we submit the unbonding ICAs which is only once per 3-5 days
  2. The unbonding amount would have to decrease more than the amount of new stake we've had for that epoch (which is 100% possible and likely will happen at some point, but in practice has not happened yet)
  3. The worst case impact of this is that the query interval is twice as wide as it should be, but the 1% threshold is already so conservative so a bump to 2% is not concerning
  
So considering the compounding unlikeliness of this happening, with the small margin of impact, I think it's fine to safely ignore this. But curious to hear your guy's thoughts.

ethan-stride: I disagree but maybe I don't have the full info.  I think a malicious attacker could indefinitely keep us from correctly checking the slash query and the 3-5 day window makes it more complicated but not impossible.  The point of the example I was showing is that when the amount is just about to overflow the checkpoint, unbonding any can put both onto the over 1.0 ratio case.  But there is no reason they can't do this again and again.  Also keep in mind that these accumulators are per validator while the total amount which is going into the checkpoint computation is across the whole host zone.

sampocs: Can you elaborate on how a malicious attacker could game this? Keep in mind, a slash would have had to occur here for the attack to be relevant

sampocs: and even if it was manipulated such that the old progress was decreased, a LS of more than 1% of TVL would still trip the query

x/stakeibc/keeper/lsm_test.go
sampocs: chatted offline, decided to not move forward with this test because the action that triggered the change in TVL is agnostic to this test

x/stakeibc/keeper/lsm.go
ethan-stride: Specifically I disagree with statement 2, I think they can cause this with an arbitrarily small unbonding, it doesn't have to be greater than the amount which is being new staked.

So the reason we need to do the ICQ at all is a validator could have been slashed and we need to know the new exchange rate in order to give out the right amount of stTokens to new deposits. The nature of the attack is that attacker notices (or causes) a validator to be slashed before the protocol has yet noticed and then by keeping the amounts correct they can deposit arbi.  For simplicity imagine not many other people are also depositing or unbonding from this target validator. So even if unbonding ICAs only happen every 3-5 days, they know when the next 3-5 day unbonding ICA is going to happen and by waiting until shortly before it they also know how much.  They can then use LSM tokens to add as much value to the specific slashed validator so that it is right before triggering the ICQ, when the unbonding ICAs happen the total amount delegated across the whole hostzone will decrease.  Even a small decrease in the total will lead to the "checkpoint" going down and the old/checkpoint and new/checkpoint now falling onto above the line on the next deposit.  They can then do this same thing again and again to continue to deposit onto this validator before we have checked the ICQ to see if it is slashed.  

You are correct that other people can come in and mess up the attackers plans like if another individual does more than 1% in a single transaction to this specific validator.

x/stakeibc/keeper/host_zone.go
```diff
@@ -158,7 +158,19 @@ func (k Keeper) IncrementValidatorSlashQueryProgress(
...
+	queryThreshold := sdk.NewDecWithPrec(int64(params.ValidatorSlashQueryThreshold), 2) // percentage
+	checkpoint := queryThreshold.Mul(sdk.NewDecFromInt(hostZone.TotalDelegations)).TruncateInt()
+
+	// Prevent division by 0 error if either the threshold or TVL is 0
+	if checkpoint.IsZero() {
+		return nil
+	}
+
 	validator.SlashQueryProgressTracker = validator.SlashQueryProgressTracker.Add(amount)
+	validator.SlashQueryInterval = validator.SlashQueryProgressTracker.Quo(checkpoint)
```
ethan-stride, line 172: Given the conversation offline earlier, I think this is secure but we might want to change the name to something which makes more sense in this context like previous_progress_tracker to mean this ratio which we are saving

ethan-stride (commented):
LGTM pending seeing the other version and we can decide which is cleaner

riley-stride:
discussed offline, LGTM

riley-stride (commented):
discussed offline, LGTM

x/stakeibc/keeper/host_zone.go
```diff
@@ -158,18 +158,30 @@ func (k Keeper) IncrementValidatorSlashQueryProgress(
...
-	validator.SlashQueryProgressTracker = validator.SlashQueryProgressTracker.Add(amount)
-	validator.SlashQueryInterval = validator.SlashQueryProgressTracker.Quo(checkpoint)
+	// Optionally re-calculate the checkpoint
+	// Threshold of 1% means once 1% of TVL has been breached, the query is issued
+	if shouldUpdateCheckpoint {
+		params := k.GetParams(ctx)
+		queryThreshold := sdk.NewDecWithPrec(int64(params.ValidatorSlashQueryThreshold), 2) // percentage
+		checkpoint := queryThreshold.Mul(sdk.NewDecFromInt(hostZone.TotalDelegations)).TruncateInt()
+
+		validator.SlashQueryCheckpoint = checkpoint
```
ethan-stride, line 183: looks good to me.

x/stakeibc/keeper/host_zone.go
```diff
@@ -215,14 +227,19 @@ func (k Keeper) AddValidatorToHostZone(ctx sdk.Context, chainId string, validato
 		valWeight = minWeight
 	}
 
+	// Determine the slash query checkpoint for LSM liquid stakes
+	params := k.GetParams(ctx)
+	queryThreshold := sdk.NewDecWithPrec(int64(params.ValidatorSlashQueryThreshold), 2) // percentage
+	checkpoint := queryThreshold.Mul(sdk.NewDecFromInt(hostZone.TotalDelegations)).TruncateInt()
```
ethan-stride, line 233: This is the initialization only for a new validator... make sense just confirming.

ethan-stride (commented):
LGTM

riley-stride (approved):
LGTM

x/stakeibc/keeper/host_zone.go
sampocs: no longer relevant with new implementation - resolving

PR #762, opened 2023-05-07
[LSM] Query validator exchange rate after adding validator

sampocs:
We now need to store each validator's exchange rate on the host zone struct. We can easily accomplish this for new validators by querying it as soon as they're added.

## Brief Changelog
* Added step to query validator exchange rate after adding validator
* Added query check in unit test

### DIFF
--- a/x/stakeibc/keeper/msg_server_add_validators_test.go
+++ b/x/stakeibc/keeper/msg_server_add_validators_test.go
@@ -1,41 +1,69 @@
 package keeper_test
 
 import (
+	"fmt"
+
 	sdkmath "cosmossdk.io/math"
 	_ "github.com/stretchr/testify/suite"
 
 	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/cosmos/cosmos-sdk/types/bech32"
+	stakingtypes "github.com/cosmos/cosmos-sdk/x/staking/types"
+	ibctesting "github.com/cosmos/ibc-go/v5/testing"
 
 	"github.com/Stride-Labs/stride/v9/x/stakeibc/types"
-	stakeibctypes "github.com/Stride-Labs/stride/v9/x/stakeibc/types"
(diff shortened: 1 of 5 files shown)


riley-stride:
What's the outcome if the query to the val's exch rate is sent off successfully but its callback fails? I think the val's exch rate on the host zone struct would be nil. Does this cause unintended consequences later?

x/stakeibc/keeper/msg_server_add_validators_test.go
```diff
@@ -66,6 +95,26 @@ func (s *KeeperTestSuite) TestAddValidators_Successful() {
 	for i := 0; i < 3; i++ {
 		s.Require().Equal(*tc.expectedValidators[i], *hostZone.Validators[i], "validators %d", i)
 	}
+
+	// Confirm ICQs were submitted
+	queries := s.App.InterchainqueryKeeper.AllQueries(s.Ctx)
```
riley-stride, line 100: lines 100-117 are some solid testing 💯

riley-stride (approved):
Nice testing in the queries! 

As long as we're not making any crucial assumption that the val exch rate on the hz struct is non-nil, this is good to go imo. per my comment I think failed query responses could result in nil val exch rates on host zone structs, which isn't caught by this code

sampocs:
> What's the outcome if the query to the val's exch rate is sent off successfully but its callback fails? I think the val's exch rate on the host zone struct would be nil. Does this cause unintended consequences later?

I think it's fine cause we handle this case explicitly in the exchange rate callback. 

Separate question - if we don't happen to have an exchange rate stored for the validator and we trigger the slash query from an LSM liquid stake, do you think we should submit the delegator shares query? 

Right now we check if the exchange rate changed in order to determine whether to submit it, but if the exchange rate is nil, we just treat it as if the exchange rate has not changed

PR #766, opened 2023-05-10
[LSM] Documentation around error case in transfer callback

sampocs:
Closes: #XXX

Resolution to a previous Informal Systems audit finding [IF-STRIDE-RECORDS-IBCTOKENSLOSTONSTRIDE](https://github.com/Stride-Labs/informal-systems-audit/blob/trunk/2022/Q4/report/src/findings/IF-STRIDE-RECORDS-IBCTOKENSLOSTONSTRIDE.md).

~Previously, if an error was thrown while running the callback for OnAckPacket or OnTimeoutPacket, the function short circuited and the packet was never sent down to the transfer stack. This prevented tokens from being refunded. To address this, we need to make sure that the transfer module's OnAckPacket or OnTimeoutPacket is called regardless of whether the callback succeeded.~

After further discussion, we've decided to keep the current behavior of throwing an error if the callback fails as it ensure that the bank and records module remain in a consistent state. This PR simply cleans up and documents each function.

## Brief Changelog
* Moved IBC callbacks to ibc.go file
* Added comments around motivation for propagating an error 

## Author's Checklist

I have...

## Reviewers Checklist

I have...

      - [ ] not applicable
      - [ ] jira ticket `XXX` 
      - [ ] README.md 
      - [ ] not documented

### DIFF
--- a/x/icacallbacks/icacallbacks.go
+++ b/x/icacallbacks/icacallbacks.go
@@ -79,6 +79,7 @@ func UnpackAcknowledgementResponse(ctx sdk.Context, logger log.Logger, ack []byt
 	case *channeltypes.Acknowledgement_Error:
 		logger.Error(fmt.Sprintf("acknowledgement error: %s", response.Error))
 		return &types.AcknowledgementResponse{Status: types.AckResponseStatus_FAILURE, Error: response.Error}, nil
+
 	default:
 		return nil, errorsmod.Wrapf(channeltypes.ErrInvalidAcknowledgement, "unsupported acknowledgement response field type %T", response)
 	}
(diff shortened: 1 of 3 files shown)


asalzmann (approved):
Notes from our discussion

Summary of ack errors (should throw in both cases)
1. Ack can't be parsed
- throw an error, we don't know whether to refund tokens or not
2. Callback errors
- users transfers will flow through - don't need to worry about this case
- if the records module throws an error, it seems better not to refund the tokens, to keep state consistent/change as little as possible. We'll need an upgrade to fix state anyway. Records is broken, tokens are frozen.

Summary of timeout errors (should throw an error if a callback fails)
1. If a callback fails
- A user has not sent the token (user transfers have no callbacks)
- The records module is in a bad state, therefore freezing (not refunding) the tokens is ok, using similar logic to the ack errors above (keep records/bank consistent, change as little as possible)

LGTM

PR #768, opened 2023-05-10
[LSM] Move LSM Token transfer to end blocker

sampocs:
Closes: #XXX

Previously, if there was a timeout when transferring the LSM token to the host zone, it was retried immediately in the transfer callback. However, at this point in the code, the deposit account has not yet had the LSM token refunded yet, which will cause an insufficient balance error when trying to submit this transfer. The refund does not happen until it moves onto the Transfer module (which is the next module in the middleware stack). 

Additionally, bundling IBC transfer with the callback in the first place adds unwanted complexity when it comes to gracefully handling errors (it was flagged as a TODO to consider separating these).

As a result, the actual transfer submission has moved to the endblocker, similar to how the detokenization works. 

## Brief Changelog
* Added status TRANSFER_QUEUE to LSMTokenDeposits
* Added end blocker function `TransferAllLSMDeposits` to loop queued deposit records and submit the IBC transfer
* Updated LSMLiquidStake transaction to simply add the record in status TRANSFER_QUEUE instead of submitting the transfer
* Updated transfer callback to set the record back to status TRANSFER_QUEUE instead of resubmitting the transfer

## Author's Checklist

I have...

## Reviewers Checklist

I have...

      - [ ] not applicable
      - [ ] jira ticket `XXX` 
      - [ ] README.md 
      - [ ] not documented


x/stakeibc/keeper/lsm.go
```diff
@@ -186,6 +186,42 @@ func (k Keeper) ShouldCheckIfValidatorWasSlashed(
...
+			if err := k.RecordsKeeper.IBCTransferLSMToken(
+				ctx,
+				deposit,
+				hostZone.TransferChannelId,
+				hostZone.DepositAddress,
+				hostZone.DelegationIcaAddress,
+			); err != nil {
+				k.Logger(ctx).Error(fmt.Sprintf("Unable to submit IBC Transfer of LSMToken for %v%s on %s: %s",
+					deposit.Amount, deposit.Denom, hostZone.ChainId, err.Error()))
+				continue
```
ethan-stride, line 216: Should we set the LSMTokenDeposit to the TRANSFER_FAILED state if there is an error here?  This isn't a timeout it is an actual error

ethan-stride: It a transfer times out the first time but then actually has a problem, will it always continue and be retried each end blocker forever?

ethan-stride (commented):
LGTM

x/stakeibc/keeper/lsm.go
sampocs: I was pondering this as well. To clarify this is just if the _submission_ failed - meaning the transfer never got off the ground on Stride. The situation where we previously change the status to TRANSFER_FAILED is if the transfer left Stride successfully but failed on the hub. 

That said, I think this would only fail if either (a) the LSM token is no longer in the deposit account or (b) an invalid denom is specified. Either case is unlikely and would indicate a larger problem, so I'm actually thinking it might make sense to label these as FAILED as well

riley-stride:
> The refund does not happen until it moves onto the Transfer module (which is the next module in the middleware stack).
In the current implementation (before this PR is merged), when _does_ the refund happen, is it the same tx as the attempt to resubmit the transfer? If so, the main issue is that, in one tx, we _first_ attempt to resubmit the transfer, _then_ lower down the middleware stack get the refund?

sampocs:
Yeah that's exactly it.

The way to think about it is this is all one large OnAckPacket transaction that goes down the stack from autopilot to records to transfer. So we wind up having:
   - autopilot: does nothing
   - records: attempts to resubmit packet and fails because the tokens had been burned when it was first submitted
   - transfer: refunds tokens (but we never get here cause the above fails)

x/stakeibc/keeper/lsm.go
riley-stride: Agreed that this would indicate a larger problem. Maybe we can just add an event emission so that indexing would alert contributors.

riley-stride (approved):
lgtm

ethan-stride:
lgtm

PR #770, opened 2023-05-11
Automated version increment [Part 1]

sampocs:
Closes: #XXX

Each upgrade, we need to increment Stride's version in all the import statements and in different `Version` variables. This process could easily be automated.
This PR introduces a github action that does the sed replacements and opens a new pull request with the changes. The action must be invoked manually.

**Note**: This will have to be merged to main before it can be fully tested since you can't run GH workflows that haven't been merged. Once this is merged, I can continue debugging in a side branch

## Brief Changelog
* Added version.sh script to sed replace the module name, all import paths, and the version references
* Added GH action that:
  * Accepts user input for the version definitions
  * Creates a new branch 
  * Updates the version
  * Commits and pushes the changes 
  * Opens a pull request

### DIFF
--- a/.github/workflows/version.yml
+++ b/.github/workflows/version.yml
@@ -15,43 +15,47 @@ jobs:
     steps:
       - name: Get versions
         run: |
-          echo "old_version=${{ github.event.inputs.old_version }}" >> $GITHUB_ENV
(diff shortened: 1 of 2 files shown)


riley-stride (approved):
this is sweet

PR #772, opened 2023-05-13
[LSM] Add delegation_changes_in_progress counter on validator

sampocs:
Closes: #XXX

In the delegator shares ICQ callback, we need to check if there was any overlap with an ICA that updates the delegation field on the validator (see #759 for full context). Previously, we were using deposit and epoch unbonding records to determine if there was an ICA in progress at the time of the query callback, but this does not cover the case where a delegation changes due to a rebalancing. 

To address this, this PR adds a field to the validator that increments every time a relevant ICA is submitted, and decrements when in the callback for that ICA. Then, the delegator shares query callback can simply check that there are no ICAs in progress using this field, rather than looping across all the records. This field also helps outline this assumption more explicitly.

## Brief Changelog
* Added delegation_changes_in_progress field on the validator
* Added increment/decrement helper functions
* Incremented before delegate, undelegate, rebalance, and detokenize ICAs
* Decremented in the callback for each of the above ICAs
* Changed `CheckDelegationChangedDuringQuery` to reference the delegation_changes_in_progress field instead of looping records

## Author's Checklist

I have...

## Reviewers Checklist

I have...

      - [ ] not applicable
      - [ ] jira ticket `XXX` 
      - [ ] README.md 
      - [ ] not documented


x/stakeibc/keeper/host_zone.go
```diff
@@ -189,6 +189,36 @@ func (k Keeper) IncrementValidatorSlashQueryProgress(
 	return nil
 }
 
+// Increments the number of validator delegation changes in progress by 1
+// Note: This modifies the original host zone struct. The calling function must Set this host zone
+// for changes to persist
+func (k Keeper) IncrementValidatorDelegationChangesInProgress(hostZone *types.HostZone, validatorAddress string) error {
```
ethan-stride, line 195: If we compute the rebalance ICAs all at once but then send them in small batches of 5 messages, would we want to add 5 to the count or just one and then when we get the call back we would also decrement 1.   This counter is per validator, so some batches of 5 messages in rebalance ICAs would be several to the same validator like [A->B. A->C, A->D, B->C, B->E] so would this be A.changes+=3 B.changes+=2?  Should we update the validators who are receiving since their delegation is changing even if they didn't get the message so it would be A.changes+=3 B.changes+=3 C.changes+=2 D.changes+=1 E.changes +=1

x/stakeibc/keeper/host_zone.go
```diff
@@ -189,6 +189,36 @@ func (k Keeper) IncrementValidatorSlashQueryProgress(
 	return nil
 }
 
+// Increments the number of validator delegation changes in progress by 1
+// Note: This modifies the original host zone struct. The calling function must Set this host zone
```
ethan-stride, line 193: Why would it be bad to set the host zone inside this function so we don't have this footgun calling convention?

x/stakeibc/keeper/icacallbacks_detokenize.go
```diff
@@ -25,8 +25,19 @@ func DetokenizeCallback(k Keeper, ctx sdk.Context, packet channeltypes.Packet, a
 		return errorsmod.Wrapf(types.ErrUnmarshalFailure, "unable to unmarshal detokenize callback: %s", err.Error())
 	}
 	chainId := detokenizeCallback.Deposit.ChainId
+	deposit := detokenizeCallback.Deposit
```
ethan-stride, line 28: Sorry for stupid question but it is on the detokenize not on the transfer complete callback which the amount we could query of the native token in delegation account would change?

x/stakeibc/keeper/icacallbacks_rebalance.go
```diff
@@ -48,6 +48,21 @@ func RebalanceCallback(k Keeper, ctx sdk.Context, packet channeltypes.Packet, ac
...
+	// Regardless of failure/success/timeout, indicate that this ICA has completed
+	hostZone, found := k.GetHostZone(ctx, rebalanceCallback.HostZoneId)
+	if !found {
+		return errorsmod.Wrapf(sdkerrors.ErrKeyNotFound, "Host zone not found: %s", rebalanceCallback.HostZoneId)
+	}
+	for _, rebalancing := range rebalanceCallback.Rebalancings {
+		if err := k.DecrementValidatorDelegationChangesInProgress(&hostZone, rebalancing.SrcValidator); err != nil {
+			return err
+		}
+		if err := k.DecrementValidatorDelegationChangesInProgress(&hostZone, rebalancing.DstValidator); err != nil {
```
ethan-stride, line 60: ok good, this answers my previous question about would we apply this to the receiving validator as well.

x/stakeibc/keeper/icqcallbacks_delegator_shares_test.go
```diff
@@ -272,79 +219,24 @@ func (s *KeeperTestSuite) TestDelegatorSharesCallback_Retry_DelegationChange() {
...
 	s.Require().NoError(err, "no error expected during delegator shares callback")
 
 	// Confirm the validator's delegation was not modified
-	hostZone, found := s.App.StakeibcKeeper.GetHostZone(s.Ctx, tc.hostZone.ChainId)
+	actualHostZone, found := s.App.StakeibcKeeper.GetHostZone(s.Ctx, HostChainId)
 	s.Require().True(found, "host zone found")
 
-	initialDelegation := hostZone.Validators[tc.valIndexQueried].Delegation
-	s.Require().Equal(initialDelegation.Int64(), hostZone.Validators[tc.valIndexQueried].Delegation.Int64(), "validator delegation")
+	initialDelegation := actualHostZone.Validators[tc.valIndexQueried].Delegation
```
ethan-stride, line 235: I think you want the initialHostZone here, you are checking the same literal value against itself which will always be true on next line.

x/stakeibc/keeper/lsm_test.go
```diff
@@ -422,10 +422,11 @@ func (s *KeeperTestSuite) TestDetokenizeLSMDeposit() {
 	s.Require().NotEmpty(delegationICAAddress, "ICA Address should not be empty")
 
 	// Build the host zone and deposit (which are arguments to detokenize)
-	hostZone := types.HostZone{
+	initialHostZone := types.HostZone{
 		ChainId:              HostChainId,
 		DelegationIcaAddress: delegationICAAddress,
 		ConnectionId:         ibctesting.FirstConnectionID,
+		Validators:           []*types.Validator{{DelegationChangesInProgress: 0}},
```
ethan-stride, line 429: I think you get this for free given the field type but never hurts to make sure by setting it explicitly

x/stakeibc/keeper/unbonding_records_get_host_zone_unbondings_msgs_test.go
```diff
@@ -132,6 +132,26 @@ func (s *KeeperTestSuite) CheckUnbondingMessages(tc UnbondingTestCase, expectedU
...
 	}
 
+	// Check the delegation change in progress was incremented from each that had an unbonding
+	actualHostZone, found := s.App.StakeibcKeeper.GetHostZone(s.Ctx, HostChainId)
+	s.Require().True(found, "host zone should have been found")
+
+	for _, actualValidator := range actualHostZone.Validators {
+		validatorUnbonded := false
+		for _, unbondedVal := range expectedUnbondings {
+			if actualValidator.Address == unbondedVal.Validator {
```
ethan-stride, line 142: This is right given the types in the unbondings, the naming bugs me because it looks like comparing an Address to a Validator struct

ethan-stride (commented):
Mostly looks good, few questions and one error in a test.

x/stakeibc/keeper/lsm_test.go
sampocs: yeah you do, but I think it's good to indicate in the tests which fields are relevant

x/stakeibc/keeper/icqcallbacks_delegator_shares_test.go
sampocs: nice catch! done

x/stakeibc/keeper/icacallbacks_detokenize.go
sampocs: yeah transfer only updates the balance, detokenization updates the delegation

x/stakeibc/keeper/host_zone.go
sampocs: This goes back to that thing I mentioned earlier about where a lot of this is triggered from hooks where we don't revert state if an error is thrown, so it's good to push the set operations as far downstream as possible

We'll come back and fix this once we add the state revert code, and once we add the validators as a stand alone prefix store

x/stakeibc/keeper/host_zone.go
sampocs: I think it's easiest to just increment both source and destination validator for each redelegation message

x/stakeibc/keeper/unbonding_records_get_host_zone_unbondings_msgs_test.go
sampocs: do you have a suggestion or just noting it?

ethan-stride: The only way to make it clearer to a reader would be to use an extra variable with the word address in it so it is clear its an address compared to address like:  expectedValidatorAddress := unbondedVal.Validator   but probably not work it, I will remove the comment

ethan-stride:
LGTM!

x/stakeibc/keeper/icqcallbacks_delegator_shares.go
```diff
@@ -134,52 +127,27 @@ func DelegatorSharesCallback(k Keeper, ctx sdk.Context, args []byte, query icqty
 //
 // We can prevent Case #1 by checking if the delegation total on the validator has changed
 //   while the query was in flight
-// We can prevent Case #2 by checking if there are any delegation unbonding records
-//   in state IN_PROGRESS (meaning an ICA is in flight)
+// We can prevent Case #2 by checking if the validator has a delegation change in progress
 func (k Keeper) CheckDelegationChangedDuringQuery(
```
riley-stride, line 131: To sanity check the blocking loogic:

The check in this PR will ensure that there are no ICAs in flight when the ICQ comes back. 

It will NOT prevent a case like this:
- ICA begins, lands on host
- ICQ begins, hits the host
- ICA callback
- ICQ callback

But the above case should be prevented by the ICQ blocking the ICA, right?

x/stakeibc/keeper/icqcallbacks_delegator_shares_test.go
```diff
@@ -272,79 +219,24 @@ func (s *KeeperTestSuite) TestDelegatorSharesCallback_Retry_DelegationChange() {
 func (s *KeeperTestSuite) TestDelegatorSharesCallback_Retry_DelegationICAInProgress() {
 	tc := s.SetupDelegatorSharesICQCallback()
 
-	// Add a deposit record that makes it appear as if a delegation ICA is in progress
-	s.App.RecordsKeeper.SetDepositRecord(s.Ctx, recordstypes.DepositRecord{
-		Id:         100,
-		HostZoneId: HostChainId,
-		Status:     recordstypes.DepositRecord_DELEGATION_IN_PROGRESS,
-	})
+	// Update the validator's delegation change ICA counter to show a change is in progress
```
riley-stride, line 222: nice to see these simplify

riley-stride (approved):
lgtm. exceptionally clear commits and separation of logic!

x/stakeibc/keeper/icqcallbacks_delegator_shares.go
sampocs: The case you described actually doesn't pose any risk because the acks landed on stride in the same order that they were landed on the host (i.e. ICA land, ICQ land, ICA ack, ICQ ack)

The main thing that matters in these permutations is the landing and ack'ing. The two concerning scenarios are shown [here](https://github.com/Stride-Labs/stride/blob/a8305beaf396a8541aebe041ddfd649f21b98382/x/stakeibc/keeper/icqcallbacks_delegator_shares.go#L120C1-L130)

riley-stride: Ah right, thanks for clarifying

PR #773, opened 2023-05-15
[LSM] Move rebalance trigger to stride epoch

sampocs:
Closes: #XXX

Rebalancing should be triggered once per unbonding period. Identifying the day to rebalance is easily determined by modding the day epoch number with the unbonding period. However, the rebalancing should not be triggered alongside any unbondings, otherwise the unbondings could cause the redelegations to fail.

This PR moves the rebalance trigger to the stride epoch (so as not to overlap with the day epoch), but ensures this is only run once per day.

To determine the proper stride epoch, we use the following: `stride_epoch_number % 4 == 0`. 

Looking at the mainnet state at the time of writing this, we are on stride epoch 1017, and this epoch overlaps the day/unbonding epoch. That implies that `stride_epoch_number % 4 == 1` identifies the stride epoch that overlaps the day/unbonding epoch, and `stride_epoch_number % 4 == 0` identifies the stride epoch before unbonding.

**NOTE**: This makes assumptions about the interplay between the stride and day epoch (i.e. that there are always 4 stride epochs in 1 day epoch) - so it is probably not the best long term solution. But our epochs could probably use some cleanup in general, so this feels good enough for the time being.


ethan-stride (approved):
LGTM

riley-stride (approved):
good to go imo if we use go constant + invariant instead of hardcoded 4

riley-stride (approved):

PR #777, opened 2023-05-20
[LSM] LSMLiquidStakeEnabled flag on host zone

sampocs:
Since not all host zones will have LSM enabled, we need a flag on the host zone to indicate whether they have it enabled and whether LSM liquid stakes are supported.

## Brief Changelog
* Added bool field `LsmLiquidStakeEnabled` on host zone
* Added check in LSM liquid stake that it's enabled
* Extended register host zone command to include parameter

## Testing Instructions
* Start dockernet
```bash
make start-docker
```
* Confirm LSM is currently enabled by viewing the field `lsm_liquid_stake_enabled` on the host zone
```bash
```
* Build a proposal file to disable LSM (call it `proposal.json`)
```json
{
    "description": "Proposal to disable LSM",
    "hostZone": "LSM",
    "enabled": false,
    "deposit": "10000001ustrd"
}
```

* Submit the proposal and vote. Make sure to vote before before the voting period ends (30 seconds)
```
strided tx gov submit-legacy-proposal toggle-lsm proposal.json --from val1 --gas auto -y 
```

* Wait for the proposal to pass and then view the host zone again, now `lsm_liquid_stake_enabled` should be set to false
```bash
```

* Update the proposal.json to the following (to re-enable LSM)
```json
{
    "description": "Proposal to enable LSM",
    "hostZone": "LSM",
    "enabled": true,
    "deposit": "10000001ustrd"
}
```
* Submit the proposal and vote
```
strided tx gov submit-legacy-proposal toggle-lsm proposal.json --from val1 --gas auto -y 
```

* Wait for the proposal to pass and then view the host zone again, now `lsm_liquid_stake_enabled` should be set to true once again
```bash
```


proto/stride/stakeibc/host_zone.proto
```diff
@@ -52,7 +52,7 @@ message HostZone {
     (gogoproto.customtype) = "github.com/cosmos/cosmos-sdk/types.Dec",
     (gogoproto.nullable) = false
   ];
+  bool lsm_liquid_stake_enabled = 27;
   bool halted = 19;
   reserved 4, 5, 6, 7, 14, 15, 16;
-  // next id: 27
```
ethan-stride, line 57: Instead of removing this shouldn't you update it to say next id 28
I understand the highest id in the current proto is 27 but it could be 28 or higher, like the next id could be 29 there is no way to know just looking at the current proto the full history

x/stakeibc/keeper/lsm.go
```diff
@@ -125,6 +126,10 @@ func (k Keeper) GetHostZoneFromLSMTokenPath(ctx sdk.Context, path string) (types
 	// Confirm the channel is from one of Stride's supported host zones
 	for _, hostZone := range k.GetAllHostZone(ctx) {
 		if hostZone.TransferChannelId == channelId {
+			if !hostZone.LsmLiquidStakeEnabled {
```
ethan-stride, line 129: Will this error bubble up through the function calls and ultimately prevent stride from accepting the LSM token in the first place if it is from a host zone we don't have marked as enabled?

x/stakeibc/keeper/msg_server_register_host_zone.go
```diff
@@ -92,12 +92,13 @@ func (k msgServer) RegisterHostZone(goCtx context.Context, msg *types.MsgRegiste
...
-		DepositAddress:     depositAddress.String(),
-		MinRedemptionRate:  msg.MinRedemptionRate,
-		MaxRedemptionRate:  msg.MaxRedemptionRate,
+		RedemptionRate:        sdk.NewDec(1),
+		LastRedemptionRate:    sdk.NewDec(1),
+		UnbondingPeriod:       msg.UnbondingPeriod,
+		DepositAddress:        depositAddress.String(),
+		MinRedemptionRate:     msg.MinRedemptionRate,
+		MaxRedemptionRate:     msg.MaxRedemptionRate,
+		LsmLiquidStakeEnabled: msg.LsmLiquidStakeEnabled,
```
ethan-stride, line 101: We can't call this function if the host zone is already registered, so it is only on the startup of the chain correct? Meaning the chain has to be restarted to update the LSM status for any other chain? Is this the desired behavior or should it be more like a param we can update without needing to restart our chain?  Presumably once a 3rd party chain goes from false to true, they are not going in the other direction ever by removing the LSM module...

ethan-stride (commented):
LGTM

x/stakeibc/keeper/lsm.go
ethan-stride: I guess that asks the question, if we are getting an otherwise valid LSM NFT from a chain we think is not enabled... are we using the flag to really turn off LSM capabilities from some chains at that point for some strategic reason?

riley-stride: It's a good question. 

You would expect tokens from chains who do not have LSM enabled to fail `IsValidIBCPath`. 

But I don't think we can be sure. Maybe a host zone with LSM disabled gives a rogue smart contract access to mint tokens of denoms that satisfy `IsValidIBCPath`. 

I view this line as a sanity check for that case. 

Maybe we should do this check before `IsValidIBCPath`, wdyt @sampocs?

riley-stride (changes requested):
Looking good. Should we add a `SetLsmLiquidStakeStatus` msgServer to invoke when the Hub adds LSM?

x/stakeibc/keeper/lsm.go
sampocs: `IsValidIBCPath` is just a regex [expression](https://github.com/Stride-Labs/stride/pull/777/files/4017b18573d4fd13b4cb0616bfc5ff309f197ad7#diff-9d3cc274196539a649a91697864567e3ce6340cd5ec5d5543cb94c3bd3e0fed7R27). 

We wouldn't want to accept valid LSM NFTs for a non-LSM enabled host zone cause we would essentially be holding their token and doing nothing with it. However, this would only occur in the downtime between when a host zone upgrades to include LSM and when enable it on our end

x/stakeibc/keeper/msg_server_register_host_zone.go
sampocs: We should enable it for the hub in the software upgrade that adds LSM. For other host zones, it probably makes the most sense to expose a governance function to toggle it

x/stakeibc/keeper/lsm.go
ethan-stride: So related to my other question below, we only foresee moving from not yet enabled to enabled.  There is no reason we would ever want to move the other direction where even though in fact that chain does have the LSM module, we as stride want to turn off the LSM ability for strategic reasons -- maybe their validators are unhappy about LSM for one reason or another or we don't trust them after we somehow get a whole bunch of LSM tokens stuck which won't detokenize or something else weird.  This isn't foreseen as a feature kill switch, only intended in the one way upgrade process

sampocs:
@ethan-stride @riley-stride Added gov route to enable LSM for host's

x/stakeibc/handler.go
```diff
@@ -62,3 +64,18 @@ func NewHandler(k keeper.Keeper) sdk.Handler {
 		}
 	}
 }
+
+// Handles stakeibc gov proposals
+func NewStakeibcProposalHandler(k keeper.Keeper) govtypes.Handler {
+	return func(ctx sdk.Context, content govtypes.Content) error {
+		switch c := content.(type) {
+		case *types.AddValidatorsProposal:
+			return k.AddValidatorsProposal(ctx, c)
```
riley-stride, line 73: Why is AddValidatorsProposal new here?

riley-stride (approved):
gov route looks good to me

x/stakeibc/handler.go
sampocs: I consolidated`proposal_handler.go` and `handler.go` (and then removed `proposal_handler.go`)

PR #778, opened 2023-05-20
[LSM] Persistent LSM Integration Test

sampocs:
Adds an LSM integration test for the "happy path" that should be included in our CI. 

## Brief Changelog
* Added LSM liquid stake test that:
  * Delegates
  * Tokenizes
  * Transfers to stride
  * LSM Liquid stakes
* Added helper functions to config.sh
* Added a user account to genesis to perform liquid stakes with


dockernet/config.sh
```diff
@@ -354,17 +357,21 @@ WAIT_FOR_STRING() {
...
 
   main_cmd=$(GET_VAR_VALUE ${chain}_MAIN_CMD)
-  initial_balance=$($main_cmd q bank balances $address --denom $denom | grep amount)
+  initial_balance=$($main_cmd q bank balances $address --denom $denom | grep amount | NUMBERS_ONLY)
   for i in $(seq $max_blocks); do
-    new_balance=$($main_cmd q bank balances $address --denom $denom | grep amount)
+    new_balance=$($main_cmd q bank balances $address --denom $denom | grep amount | NUMBERS_ONLY)
+    balance_change=$(echo "$new_balance - $initial_balance" | bc)
 
     if [[ "$new_balance" != "$initial_balance" ]]; then
```
ethan-stride, line 376: Doesn't seem like minimum_change defined on 366 or balance_change on 374 are being used in this func?  Should this line be something like "$balance_change" >= "$minimum_change"

dockernet/config.sh
```diff
@@ -374,6 +381,28 @@ WAIT_FOR_BALANCE_CHANGE() {
...
 
+# Sleep until the total delegation amount has changed
+# Optionally provide a minimum amount it must change by  (to ignore interest)
+WAIT_FOR_DELEGATION_CHANGE() {
+  chain_id=$1
+  minimum_change=${2:-1} # defaults to 1
+
+  max_blocks=30
+
+  $STRIDE_MAIN_CMD q stakeibc show-host-zone $chain_id
```
ethan-stride, line 392: Is this supposed to be going into a variable or is it going to be echo to print out?

dockernet/tests/integration_tests.bats
```diff
@@ -199,6 +193,56 @@ setup_file() {
   assert_equal "$stake_diff" "1"
 }
 
+@test "[INTEGRATION-BASIC-$CHAIN_NAME] LSM liquid stake" {
+  if [[ "$CHAIN_NAME" != "LSM" ]]; then
```
ethan-stride, line 197: After the LSM is an accepted module to the hub, then GAIA in general will have LSM as part of it... will we still have the LSM as a separate docker build at that point?  Will this test update in the future or do we plan to have a vanilla GAIA and an LSM specific hub as separate dockers even after LSM is accepted?

dockernet/tests/integration_tests.bats
```diff
@@ -199,6 +193,56 @@ setup_file() {
...
+
+  lsm_token_ibc_denom=$(GET_IBC_DENOM $STRIDE_TRANSFER_CHANNEL ${validator_address}/${record_id})
+
+  # get initial balances
+  sttoken_balance_start=$($STRIDE_MAIN_CMD q bank balances $staker_address_on_stride --denom st$HOST_DENOM | GETBAL)
+  lsm_token_balance_start=$($STRIDE_MAIN_CMD q bank balances $staker_address_on_stride --denom $lsm_token_ibc_denom | GETBAL)
+
+  # LSM-liquid stake
+  $STRIDE_MAIN_CMD tx stakeibc lsm-liquid-stake $STAKE_AMOUNT $lsm_token_ibc_denom --from $USER_ACCT -y 
+  WAIT_FOR_BLOCK $STRIDE_LOGS 2
```
ethan-stride, line 230: Seems like a good place to use the WAIT_FOR_BALANCE_CHANGE() since that is what the next lines explicitly care about

dockernet/tests/integration_tests.bats
```diff
@@ -199,6 +193,56 @@ setup_file() {
...
+  # tokenize shares
+  $HOST_MAIN_CMD tx staking tokenize-share $validator_address ${TRANSFER_AMOUNT}${HOST_DENOM} $staker_address_on_host --from $USER_ACCT -y --gas 1000000
+  WAIT_FOR_BLOCK $STRIDE_LOGS 2
+
+  # get the record id from the tokenized share record
+  record_id=$($HOST_MAIN_CMD q staking last-tokenize-share-record-id | awk '{print $2}' | tr -d '"')
+
+  # transfer LSM tokens to stride
+  $HOST_MAIN_CMD tx ibc-transfer transfer transfer $HOST_TRANSFER_CHANNEL \
+    $staker_address_on_stride ${TRANSFER_AMOUNT}${validator_address}/${record_id} --from $USER_ACCT -y
```
ethan-stride, line 218: Store the ${validator_address}/${record_id} in a semantically meaningful name like lsm_token_host_denom

ethan-stride (commented):
LGTM, as good as this type of test can look, always going to be hard to read but that is the nature of BASH scripting.  You did a good job breaking things into functions/macros where possible

dockernet/config.sh
ethan-stride: Similar comment for the next function Wait for delegation change as well

dockernet/config.sh
riley-stride: +1, unclear what this line is doing

dockernet/config.sh
riley-stride: Actually looks like `minimum_change` is defined as a param in all these fns but not used in any of the fns

dockernet/tests/integration_tests.bats
riley-stride: imo once lsm is part of the gaia codebase, we should just use the latest gaia binary, rather than have two dockers

riley-stride: maybe we can comment around areas of dockernet that will need to change once lsm is in the standard gaia codebase

riley-stride: perhaps we can check whether the chain has LSM enabled in a different way, maybe by checking whether it has a x/staking globalCap param and that it's nonzero

riley-stride (approved):
You did a solid job breaking out helper functions here.

Ethan has some good recommendations to clean it up a bit more imo!

Do you think we should integration test the slash query case?

dockernet/config.sh
sampocs: good catch! updated

dockernet/config.sh
sampocs: how do you guys not see how important this line is???

sampocs: Kidding! I think I accidentally copy/pasted this line as I was writing the command a few lines below. Removed

dockernet/tests/integration_tests.bats
sampocs: once LSM is in gaia, we'll remove the LSM host zone entirely and update gaia to the latest version. I don't think we need to flag it now - it'll just be a find + delete of everything LSM in dockernet

dockernet/tests/integration_tests.bats
sampocs: done

dockernet/tests/integration_tests.bats
sampocs: done

sampocs:
> Do you think we should integration test the slash query case?

I was considering that too, but the tricky thing is that we'd then have to issue two LSM-liquid stakes (since the first one doesn't initiate a query) which would lengthen the tests a bit. 

I guess we could only test the second one up until the minting of stTokens which would only extend it by like 10sec?

sampocs:
@riley-stride @ethan-stride new IG test is in there now

ethan-stride (approved):
LGTM

riley-stride (approved):
lgtm

PR #788, opened 2023-05-24
Automated changelog [Part 1]

sampocs:
Closes: #XXX

Each upgrade, we need to update the changelog to include all the commits since the last release. This process could easily be automated.
This PR introduces a github action that does the changelog updates and opens a new pull request with the changes. The action must be invoked manually.

Note: This will have to be merged to main before it can be fully tested since you can't run GH workflows that haven't been merged. Once this is merged, I can continue debugging in a side branch

## Brief Changelog
* Added `changelog.sh` script to automatically update the changelog based on the git history
  * This updates the main changelog with both on-chain and off-chain changes and the creates the upgrade specific changelog for just on-chain changes
* Added `changelog.yml` workflow to run the above script and open up a PR with the change


shellvish (approved):
This is huge, thank you for doing this!

PR #789, opened 2023-05-24
Automate binary versioning

sampocs:
Closes: #XXX

We currently hard code the stride version and update it each time we release. However, this can easily be automated. This PR updates the makefile and code to pass the commit hash and version at build time. It also adds a new version command that prints the commit hash.

## Brief Changelog
* Fixed version propagation in Makefile
* Added version CLI command
* Changed dockernet build script to use make build
* Removed app.go and config.go version replacements in the version workflow
* Updated submodules


shellvish (approved):
You're on a roll!

PR #813, opened 2023-06-02
[LSM] Remove cap on number of validators

sampocs:
Closes: #XXX

We currently have a cap on the number of validators for a host zone that's configured with a param. However, with LSM we'll be adding most of the validators in a set so this cap is no longer necessary.

## Brief Changelog
* Removed check on number of validators when adding a validator or changing a validator's weight
* Removed `SafetyNumValidator` param


x/stakeibc/keeper/msg_server_change_validator_weight.go
```diff
@@ -15,28 +13,17 @@ func (k msgServer) ChangeValidatorWeight(goCtx context.Context, msg *types.MsgCh
...
-		if validator.GetAddress() == msg.ValAddr {
-
-			// when changing a weight from 0 to non-zero, make sure we have space in the val set for this new validator
-			if validator.Weight == 0 && msg.Weight > 0 {
-				err := k.ConfirmValSetHasSpace(ctx, validators)
-				if err != nil {
-					return nil, errorsmod.Wrap(types.ErrMaxNumValidators, "cannot set val weight from zero to nonzero on host zone")
-				}
-			}
+		if validator.Address == msg.ValAddr {
```
ethan-stride, line 21: Do we really want to allow every validator?  Even if we now will have most of them for the possible LSM token locations do we want the ability to have a blacklist of validators we just don't trust because they have been slashed so many times in the past?  Also it isn't possible for fake validators to exist in this pool right so no need to do some sort of check here that they are real?

ethan-stride (approved):
LGTM
very simple and mechanical changes.
Just discussion around if we would want a way to remove specific validators from the pool

x/stakeibc/keeper/msg_server_change_validator_weight.go
sampocs: Validators will only be added and removed through governance so there's no need to blacklist, we can just make sure to not include them in the proposal!

PR #816, opened 2023-06-06
Check if binaries exist during dockernet startup

sampocs:
Closes: #XXX

One of the most common issues people have with starting up dockernet is forgetting to build the binary for a host zone. This PR adds a check during the dockernet startup to alert the user if the binary does not exist.


ststkent (approved):
LGTM!

PR #818, opened 2023-06-07
IBC Rate Limiting Audit Fixes

sampocs:
This PR addresses two issues from the informal audit (note these are in a private repo):
* [The outflow is not rolled back if the packet times out or if the acknowledgement is not successful ](https://github.com/informalsystems/audit-stride/issues/6)
* [Channels with rate limiters could potentially have bottlenecks.](https://github.com/informalsystems/audit-stride/issues/4)

## Diagram
[Excalidraw](https://excalidraw.com/#room=1ee7a76b12c6fdca32c1,tuMfxmGnZsaGNMtRLNBMPw
)
### Overview
Currently when packets are sent, the outflow is incremented by the transfer amount; however, if the packet fails or times out, it is not decremented. This opens the door for someone to clog up the rate limit by sending packets with a short timeout. 

### Solution
The obvious solution here is to decrement the outflow during the timeout or ack failure. But the complexity lies in that we need to make sure that we're in the same rate limit time window as the original packet send, otherwise the outflow will have already been reset to 0 and any additional decrement would be incorrect.  In other words, if the rate limit was reset before the packet response was relayed (but after it was sent), then the outflow should not be modified during the ack.

### Implementation
This was implemented by saving down the sequence numbers of packets that have been sent in the given rate limit window, and then checking for the presence of that sequence number when handling the acknowledgment or timeout. More specifically, the high level changes were:
```
SendPacket:           + Stores sequence number for packet
ResetRateLimit:       + Clears all sequence numbers for pending packets
OnAckFailure/Timeout: + Check if sequence exists, and if so, decrement the outflow
                      + Removes sequence number (not entirely necessary, but decreases the size of the state)
```

### Changelog
* Added function `UndoSendPacket` which checks if the packet's sequence number exists in the pending store (i.e. the packet was sent in the current time window), and if so, decrements the outflow
* Added the following keeper functions to keep track of the pending packets:
  * `SetPendingSendPacket`
  * `RemovePendingSendPacket`
  * `CheckPacketSendDuringCurrentQuota`
  * `RemoveAllChannelPendingSendPackets`
* Modified `SendRateLimitedPacket` to store a pending packet
* Modified `ResetRateLimit` to remove all pending packets
* Implemented `TimeoutRateLimitedPacket` which calls `UndoSendPacket`
* Implemented `AcknowledgeRateLimitedPacket` which checks if the ack failed, and if so, calls `UndoSendPacket`
* Added helper function `ParsePacketInfo` to group some redundant packet parsing logic that was required in each IBC callback

### Overview
In Stride's stakeibc flow, a single transfer is issued every epoch with all the aggregate liquid stake deposits from the previous epoch window. This transfer is large enough that it would likely exceed the rate limit quota depending on how strict it was set.

#### Solution
Transfers to or from Stride module account's or ICA's should be whitelisted such that they are not included in the `Inflow/Outflow` calculation and cannot be rate limited.

### Implementation
This was implemented by adding a list of whitelisted addresses to the store, and adding a check to Send/Receive packet for whether the sender or receiver is in the list of addresses.

### Changelog
* Added keeper functions for the whitelist:
  *  `AddAddressToWhitelist`
  * `RemoveAddressFromWhitelist`
  * `IsAddressWhitelisted`
  * `GetAllWhitelistedAddresses`
* Modified `CheckRateLimitAndUpdateFlow` to check if the sender or receiver is in the whitelist, and if so, allows the packet to pass without updating the flow
* Modified stakeibc to whitelist each module account when registering a host zone

## Additional Misc Changelog
* Updated and added new unit tests
* Updated readme
* Added TODO in upgrade handler to white list host zone addresses
* Added queries for blacklisted denoms and whitelisted addresses

### Ack Failure/Timeout
* Set the host zones as `(GAIA JUNO OSMO)` in config.sh
* Start dockernet
```bash
make start-docker
```
* Go into `dockernet/scripts/ratelimit/run_all_tests.sh` and comment out all the tests besides the last one "test_send_failures"
* In a separate terminal window, run the following to constantly print the `ustrd` rate limit
```bash
while true; do echo "=======================" && build/strided --home dockernet/state/stride1 q ratelimit rate-limit channel-2 --denom ustrd && sleep 1; done
```
* Finally, run the ack timeout/failure tests which will simulate a timed out transfer and a failed transfer. In your other terminal window, you'll see the Outflow temporarily increase along the rate limit, and then shortly after get reset back to 0 as the timeout/failure is returned to stride. You'll have to look closely or you might miss it!
```
bash dockernet/scripts/ratelimit/run_all_tests.sh
```

### Address Whitelist
* After running the setup above, run the following in a different terminal window to display the atom rate limit flow
```bash
while true; do echo "=======================" && build/strided --home dockernet/state/stride1 q ratelimit rate-limit channel-2 --denom ustrd && sleep 1; done
```
* Run the dockernet scripts to transfer and liquid stake
```bash
bash dockernet/scripts/1.sh
bash dockernet/scripts/2.sh
```
* Confirm the rate limit flow was updated from the original transfer in script 1, but was not modified from the epochly transfer


x/ratelimit/keeper/rate_limit.go
```diff
@@ -151,41 +187,95 @@ func (k Keeper) GetAllRateLimits(ctx sdk.Context) []types.RateLimit {
...
+
+	iterator := store.Iterator(nil, nil)
+	defer iterator.Close()
+
+	pendingPackets := []string{}
+	for ; iterator.Valid(); iterator.Next() {
+		key := iterator.Key()
+
+		channelId := string(key[:types.PendingSendPacketChannelLength])
+		channelId = strings.TrimRight(channelId, "\x00") // removes null bytes from suffix
```
asalzmann, line 227: Why are there null bytes?

x/ratelimit/keeper/rate_limit.go
```diff
@@ -151,41 +187,95 @@ func (k Keeper) GetAllRateLimits(ctx sdk.Context) []types.RateLimit {
 	return allRateLimits
 }
 
+// Sets the sequence number of a packet that was just sent
+func (k Keeper) SetPendingSendPacket(ctx sdk.Context, channelId string, sequence uint64) {
+	store := prefix.NewStore(ctx.KVStore(k.storeKey), types.PendingSendPacketPrefix)
+	key := types.GetPendingSendPacketKey(channelId, sequence)
+	store.Set(key, []byte{1})
```
asalzmann, line 194: What's the `[]byte{1}` needed for?

x/ratelimit/keeper/rate_limit.go
```diff
@@ -54,36 +56,69 @@ func (k Keeper) UpdateFlow(rateLimit types.RateLimit, direction types.PacketDire
...
 	// If there's no rate limit yet for this denom, no action is necessary
 	rateLimit, found := k.GetRateLimit(ctx, denom, channelId)
 	if !found {
-		return nil
+		return false, nil
+	}
+
+	// Check if the sender or receiver are white listed
+	// If so, return a success without modifying the quota
+	if k.IsAddressWhitelisted(ctx, packetInfo.Sender) || k.IsAddressWhitelisted(ctx, packetInfo.Receiver) {
```
asalzmann, line 83:
```suggestion
	if k.IsAddressWhitelisted(ctx, packetInfo.Sender) && direction == types.PACKET_SEND || k.IsAddressWhitelisted(ctx, packetInfo.Receiver) && directino == types.PACKET_RECEIVE {
```

I think this condition is more specific, it's unnecessarily broad right now (e.g. sending to stride's module account address on a counterparty chain would be whitelisted)

x/ratelimit/keeper/packet.go
```diff
@@ -102,65 +112,120 @@ func ParseDenomFromRecvPacket(packet channeltypes.Packet, packetData transfertyp
...
+func ParsePacketInfo(packet channeltypes.Packet, direction types.PacketDirection) (RateLimitedPacketInfo, error) {
 	var packetData transfertypes.FungibleTokenPacketData
-	if err := json.Unmarshal(data, &packetData); err != nil {
-		return err
+	if err := json.Unmarshal(packet.GetData(), &packetData); err != nil {
+		return RateLimitedPacketInfo{}, err
+	}
+
+	var channelID, denom string
+	if direction == types.PACKET_SEND {
```
asalzmann, line 134: nit: would `types.PACKET_ORIGIN_STRIDE` and `types.PACKET_ORIGIN_HOST` be clearer? (don't have a strong view)

x/ratelimit/keeper/packet.go
```diff
@@ -102,65 +112,120 @@ func ParseDenomFromRecvPacket(packet channeltypes.Packet, packetData transfertyp
...
 
-	denom := ParseDenomFromRecvPacket(packet, packetData)
+	// Parse the denom, channelId, and amount from the packet
+	packetInfo, err := ParsePacketInfo(packet, types.PACKET_SEND)
+	if err != nil {
+		return err
+	}
+
+	// If the ack was successful, remove the pending packet
+	if ackResponse.Status == icacallbacktypes.AckResponseStatus_SUCCESS {
```
asalzmann, line 211: Should this check if the packet exists before removing it? e.g. what if it's just a normal ack (from a channel that has no rate limit)

dockernet/config.sh
```diff
@@ -196,6 +196,7 @@ OSMO_DENOM=$OSMO_DENOM
 OSMO_RPC_PORT=26357
 OSMO_MAIN_CMD="$OSMO_BINARY --home $DOCKERNET_HOME/state/${OSMO_NODE_PREFIX}1"
 OSMO_RECEIVER_ADDRESS='osmo1w6wdc2684g9h3xl8nhgwr282tcxx4kl06n4sjl'
+OSMO_FEES="--fees 10000uosmo"
```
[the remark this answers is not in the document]
sampocs: whoops, it's not needed anymore, removed

dockernet/config.sh
```diff
@@ -16,7 +16,7 @@ TX_LOGS=$DOCKERNET_HOME/logs/tx.log
 KEYS_LOGS=$DOCKERNET_HOME/logs/keys.log
 
 # List of hosts enabled
-HOST_CHAINS=() 
+HOST_CHAINS=(GAIA JUNO OSMO)
```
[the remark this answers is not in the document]
sampocs: done

x/ratelimit/README.md
```diff
@@ -55,6 +55,10 @@ Using the example above, let's say we created a 24 hour rate limit on `ibc/D24B4
 
 The module also contains a blacklist to completely halt all IBC transfers for a given denom. There are keeper functions to add or remove denoms from the blacklist; however, these functions are not exposed externally through transactions or governance, and they should only be leveraged internally from the protocol in extreme scenarios.
 
+## Address Whitelist
+
+There is also a whitelist to exclude module account's and ICAs. Stride periodically bundles liquid staking deposits and transfers in a single transaction at the top of the epoch. Without a whitelist, this transfer would trip the rate limit.
```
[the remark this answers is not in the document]
sampocs: done

x/ratelimit/types/keys.go
```diff
@@ -18,7 +20,21 @@ func KeyPrefix(p string) []byte {
...
-	PathKeyPrefix      = KeyPrefix("path")
-	RateLimitKeyPrefix = KeyPrefix("rate-limit")
-	BlacklistKeyPrefix = KeyPrefix("blacklist")
+	PathKeyPrefix             = KeyPrefix("path")
+	RateLimitKeyPrefix        = KeyPrefix("rate-limit")
+	PendingSendPacketPrefix   = KeyPrefix("pending-send-packet")
+	DenomBlacklistKeyPrefix   = KeyPrefix("denom-blacklist")
+	AddressWhitelistKeyPrefix = KeyPrefix("address-blacklist")
+
+	PendingSendPacketChannelLength int = 16
```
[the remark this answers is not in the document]
sampocs: that's just the length channel part of the byte array. 16 was arbitrary, but it has to be long enough such that when you convert a channel-id to a byte array, it does not exceed that length

x/ratelimit/keeper/rate_limit.go
sampocs: The length of the channel ID is fixed at 16 (related to above comment).

If you serialize a channelId of "channel-0", you get 
```
[99 104 97 110 110 101 108 45 48]
```

So then it gets zero-padded to 16 length
```
[99 104 97 110 110 101 108 45 48 0 0 0...]
```

And when you convert back to a string those zero's show up in the string
```
channel-0\x00\x00\0xx
```

Fwiw though, this is only used for export genesis - it's a pretty insignificant function

x/ratelimit/keeper/rate_limit.go
sampocs: All we really need to store is the key, the value doesn't matter as long as it's non-empty.

The implementation can be thought of more as a a set in python, rather than a map

x/ratelimit/keeper/packet.go
sampocs: the stored pending packets is only used for rate limiting to determine if the packet was sent in the same rate limit window. This code isn't deleting the actual packet, just the record that it was sent this window. It has no affect on other modules

x/ratelimit/README.md
```diff
@@ -125,6 +129,11 @@ For a more detailed explanation, see the[ ICS-20 ADR](https://github.com/cosmos/
     - (1) Remove Prefix: `transfer/channel-Z/ujuno`
     - (2) Hash: `ibc/...`
 
+## Packet Failures and Timeouts
+When a transfer is sent, the `Outflow` for the corresponding rate limit is incremented. Consequently, if the transfer fails on the host or times out, the change in `Outflow` must be reverted. However, the decrement is only necessary if the acknowledgement or timeout is returned in the same quota window that the packet was originally sent from.
+
+To keep track of whether the packet was sent in the same quota, the sequence number of all pending packets are stored. This is implemented by recording the sequence number of a SendPacket as it is sent, and then removing that list of sequence numbers each time the rate limit is reset at the end of the quota. Additionally, the sequence numbers are also removed when after an acknowledgement or timeout (a step that is not entirely necessary, but does reduce the size of the state).
```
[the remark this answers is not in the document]
sampocs: Since we use the channel-ID on stride, I think channel ID + sequence number is unique. I don't think it's possible to have two different ports on the same channel-id
My mental model is that a port is really just an alias for channel

x/ratelimit/keeper/packet.go
asalzmann: That makes sense, but I think we only set the pending packet if we update a flow
```
	if updatedFlow {
		k.SetPendingSendPacket(ctx, packetInfo.ChannelID, packet.Sequence)
	}
```

However, this will always try and delete a packet (unless I'm missing something). If you call `Delete` on a key that doesn't exist, does it throw an error?

x/ratelimit/README.md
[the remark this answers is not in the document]
asalzmann: Reading the IBC code, it seems the sequence number is unique for a channel/port pair, not a channel. I think we assume that only channel-id/transfer packets are sent across the channel, which seems safe because there's no code path that can send channel-id/other-port packets across the channel (and if that's true, then sequence number is unique per channel).

However, I think sending packets across the channel from another port is possible (we should verify this). Channel/ports definitely don't map 1:1, for example you can have two transfer channels: channel-1 and channel-2 on connection-1.

If it is possible for packets from other ports to be sent across the channel, it seems like using the full uniqueness key is a bit more defensive

x/stakeibc/module_ibc.go
```diff
@@ -98,22 +98,25 @@ func (im IBCModule) OnChanOpenAck(
 	}
 
 	// Set ICA account addresses
+	icaAddressAdded := true
 	switch {
-	// withdrawal address
 	case portID == withdrawalAddress:
 		zoneInfo.WithdrawalAccount = &types.ICAAccount{Address: address, Target: types.ICAAccountType_WITHDRAWAL}
```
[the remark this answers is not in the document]
sampocs: done

x/ratelimit/keeper/rate_limit.go
sampocs: done

x/ratelimit/keeper/packet.go
sampocs: I don't have a strong view either but I think I'd prefer to keep it as is

x/ratelimit/README.md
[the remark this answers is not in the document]
asalzmann: Actually this seems wrong
```
interface ChannelEnd {
  state: ChannelState
  ordering: ChannelOrder
  counterpartyPortIdentifier: Identifier
  counterpartyChannelIdentifier: Identifier
  connectionHops: [Identifier]
  version: string
}
```
A `ChannelEnd` has a `counterpartyPortIdentifier`, so it seems channels do map onto a single port on each chain?

x/ratelimit/keeper/packet.go
sampocs: Oh right, sorry I misunderstood your original question. Yeah this was intentional since calling Delete on a key that doesn't exist does not error 

I figured it was simpler to just call Remove regardless, but if you think it would help readability, I can add the check that it exists first before removing

x/ratelimit/README.md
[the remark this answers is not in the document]
asalzmann: https://github.com/cosmos/ibc/blob/main/spec/core/ics-004-channel-and-packet-semantics/README.md

x/ratelimit/keeper/packet.go
asalzmann: sgtm!

x/ratelimit/keeper/rate_limit.go
asalzmann: Ah I see! Thanks for the explanation

x/ratelimit/keeper/packet.go
asalzmann: Gotcha! Makes sense, either way seems fine to me

asalzmann (approved):

x/ratelimit/README.md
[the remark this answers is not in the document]
sampocs: chatted offline, in agreement that channel's do map to a single port so channel ID is sufficient as a unique identifier

sampocs:
Added the testing instructions to the PR description

x/ratelimit/keeper/rate_limit.go
sampocs: @asalzmann thinking about this a bit more, I'm wondering if we should just remove the ICA accounts from the whitelist now that there's the extra clause that specifies the packet type? 

The epochly transfer would be a SEND_PACKET with Sender == module account, and the prop 8 transfer would be a RECV_PACKET with Receiver == module account. I'm not sure there's any other cases to worry about

riley-stride:
I believe autopilot liquid staking routes funds through the module account. If we whiltelist that account to not be restricted by rate limiting when it performs protocol-initiated ibc-transfers to the delegation account, have we also unintentionally removed rate limiting from autopilot liquid staking?

x/ratelimit/keeper/rate_limit.go
riley-stride: That seems cleaner to me. Agree that those are the only two cases. Note: I believe prop 8 rev goes into a separate module account, that will need to be whitelisted.

x/stakeibc/keeper/msg_server_register_host_zone.go
```diff
@@ -74,6 +74,7 @@ func (k msgServer) RegisterHostZone(goCtx context.Context, msg *types.MsgRegiste
 	if err := utils.CreateModuleAccount(ctx, k.accountKeeper, zoneAddress); err != nil {
 		return nil, errorsmod.Wrapf(err, "unable to create module account for host zone %s", chainId)
 	}
+	k.RatelimitKeeper.AddAddressToWhitelist(ctx, zoneAddress.String())
```
riley-stride, line 77: note to remember to add whitelisting to upgrade handler

x/stakeibc/module_ibc.go
```diff
@@ -98,22 +98,25 @@ func (im IBCModule) OnChanOpenAck(
...
 		zoneInfo.DelegationAccount = &types.ICAAccount{Address: address, Target: types.ICAAccountType_DELEGATION}
 	case portID == redemptionAddress:
 		zoneInfo.RedemptionAccount = &types.ICAAccount{Address: address, Target: types.ICAAccountType_REDEMPTION}
 	default:
+		icaAddressAdded = false
 		ctx.Logger().Error(fmt.Sprintf("Missing portId: %s", portID))
 	}
 
+	if icaAddressAdded {
+		im.keeper.RatelimitKeeper.AddAddressToWhitelist(ctx, address)
```
riley-stride, line 117: Is this going to whitelist all ICA addresses? Seems safer to only do it for those we need to, wdyt?

x/ratelimit/keeper/rate_limit.go
riley-stride: Actually an even slightly safer design imo would be 

Whitelist 
- Both must be true: (SENDER == DELEGATION_ICA) && (RECEIVER == MAIN STAKEIBC MODULE ACCOUNT)
- Both must be true: (SENDER == WITHDRAWAL_ICA) && (RECEIVER == PROP8 STAKEIBC MODULE ACCOUNT)

With the "either or" whitelist approach and whitelisting the main stakeibc mod acct and the delegation ICA, I think we've disabled rate limits along these paths:
- transfers from stakeibc module account to malicious cosmoshub address
- transfers from delegation acct to malicious stride address

riley-stride (changes requested):
Fix #1 looks pretty good to me, added a couple questions around fix #2

Awesome unit test coverage on all of this

sampocs:
> I believe autopilot liquid staking routes funds through the stakeibc module account. If we whiltelist that account to not be restricted by rate limiting when it performs protocol-initiated ibc-transfers to the delegation account, have we also unintentionally removed rate limiting from autopilot liquid staking? Or are we only whitelisting the host ICA accounts (seems like the right approach)?

Sorry @riley-stride, just noticed this! I don't think autopilot LS uses a module account - I think it sends it to the user's stride address

x/ratelimit/keeper/packet.go
sampocs: cool, I'll just leave it as is

x/ratelimit/keeper/rate_limit.go
```diff
@@ -288,42 +287,45 @@ func (k Keeper) GetAllBlacklistedDenoms(ctx sdk.Context) []string {
...
+func (k Keeper) RemoveAddressWhitelist(ctx sdk.Context, sender, receiver string) {
 	store := prefix.NewStore(ctx.KVStore(k.storeKey), types.AddressWhitelistKeyPrefix)
-	key := types.KeyPrefix(address)
+	key := types.GetAddressWhitelistKey(sender, receiver)
 	store.Delete(key)
 }
 
-// Check if an address is currently whitelisted
-func (k Keeper) IsAddressWhitelisted(ctx sdk.Context, address string) bool {
+// Check if an address pair is currently whitelisted
```
riley-stride: Can we call them routes to make clear that they're directional? 

For example, if I read that (A,B) was a whitelisted pair, I would assume that both A->B and B->A would be excluded from rate limiting, but in reality only A->B is excluded

x/ratelimit/keeper/rate_limit_test.go
```diff
@@ -288,11 +292,14 @@ func (s *KeeperTestSuite) processCheckRateLimitAndUpdateFlowTestCase(tc checkRat
 			continue
 		}
 
-		if action.addToWhitelist != "" {
-			s.App.RatelimitKeeper.AddAddressToWhitelist(s.Ctx, action.addToWhitelist)
+		if action.addToWhitelist {
```
riley-stride, line 295: Do we check somewhere that if A-> B is whitelisted, then A-> C and C-> B both get rate limited?

x/stakeibc/module_ibc.go
```diff
@@ -101,16 +102,25 @@ func (im IBCModule) OnChanOpenAck(
 	switch {
 	case portID == withdrawalAddress:
 		zoneInfo.WithdrawalAccount = &types.ICAAccount{Address: address, Target: types.ICAAccountType_WITHDRAWAL}
-		im.keeper.RatelimitKeeper.AddAddressToWhitelist(ctx, address)
+
 	case portID == feeAddress:
 		zoneInfo.FeeAccount = &types.ICAAccount{Address: address, Target: types.ICAAccountType_FEE}
-		im.keeper.RatelimitKeeper.AddAddressToWhitelist(ctx, address)
+		rewardCollectorAddress := im.keeper.AccountKeeper.GetModuleAccount(ctx, types.RewardCollectorName).GetAddress()
```
riley-stride, line 108: I like the approach of setting these in zone registration. Let's just remember to set them in our upgrade handler for existing zones :)

x/ratelimit/keeper/rate_limit_test.go
sampocs: it's covered in `TestAddressWhitelist`!

x/stakeibc/module_ibc.go
sampocs: yup, there's a TODO in upgrades.go for this!

x/ratelimit/keeper/rate_limit.go
sampocs: Hmm that's a good point. But I think Address/Account should still be specified, otherwise `Route` may easily be conflated with a channel, and `Path` is already used to describe the channel

proto/stride/ratelimit/ratelimit.proto
```diff
@@ -49,3 +49,8 @@ message RateLimit {
   Quota quota = 2;
   Flow flow = 3;
 }
+
+message WhitelistedAddressPair {
+  string sender = 1;
```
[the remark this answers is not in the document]
sampocs: I think that's just left over boilerplate from people copy pasting haha. I think I'd prefer to leave it out until we have a good reason for it

proto/stride/ratelimit/ratelimit.proto
```diff
@@ -49,3 +49,8 @@ message RateLimit {
   Quota quota = 2;
   Flow flow = 3;
 }
+
+message WhitelistedAddressPair {
```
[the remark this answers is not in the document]
sampocs: My vote is option 1, I don't see any reason to add a restriction here since either a) the other chains don't have a stride address and a transfer would not be possible or b) they do have a stride address, in which case having a restriction here would break things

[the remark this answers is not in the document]
asalzmann: Sgtm - should we just add a comment above the data structure linking to this discussion?

PR #837, opened 2023-06-20
[LSM] Fail liquid stake on slash query timeout

sampocs:
Closes: #XXX

After discussing the LSM frontend UX, we realized it would make more sense to fail liquid stake transactions after the query timed out, rather than continually retrying it. This PR refactors ICQ so that a timed out query can be passed into a callback.

## Brief Changelog
* Added `TimeoutPolicy` to query to specify the action that should be taken after a timeout as either:
  * Reject the query response and do nothing else
  * Retry the query request
  * Execute the query callback
* Added helper `HasTimedOut` function on query struct to determine if the query timed out
* Removed the automatic retry on timeout that was in the end blocker
* Updated the retry code to use a unique query ID
* Refactored/cleaned up the validator exchange rate callback into sub-functions
* Added timeout handling in validator exchange rate callback to fail the LSM tx
* Leveraged the existing QueryValidatorExchangeRate function in LSM to submit the ICQ 

Set the existing ICQs with the following policies:
|            Query Type            |     Timeout Policy     | Timeout Duration |
|:--------------------------------:|:----------------------:|:----------------:|
|        Withdrawal Balance        |   REJECT_QUERY_REQUEST  |      End of epoch      |
|            Fee Balance           |   REJECT_QUERY_REQUEST  |      End of epoch      |
| Validator Exchange Rate (Not-LSM) |   REJECT_QUERY_REQUEST  |      24 hours      |
| Validator Exchange Rate (LSM)    | EXECUTE_QUERY_CALLBACK |     5 minutes    |
|         Delegator Shares         |   RETRY_QUERY_REQUEST  |      1 hour      |


sampocs:
@riley-stride @ethan-stride I don't really have a strong opinion on the timeout policy / duration for withdrawal balance and fee balance queries. Lmk your thoughts!

x/interchainquery/keeper/msg_server.go
```diff
@@ -93,6 +94,34 @@ func (k Keeper) VerifyKeyProof(ctx sdk.Context, msg *types.MsgSubmitQueryRespons
...
+		"QUERY TIMEOUT - QueryId: %s, TTL: %d, BlockTime: %d", query.Id, query.TimeoutTimestamp, ctx.BlockHeader().Time.UnixNano()))
+
+	switch query.TimeoutPolicy {
+	case types.TimeoutPolicy_REJECT_QUERY_RESPONSE:
+		k.Logger(ctx).Info(utils.LogICQCallbackWithHostZone(query.ChainId, query.CallbackId, "Rejecting query"))
+		return false, nil
+
+	case types.TimeoutPolicy_RETRY_QUERY_REQUEST:
+		k.Logger(ctx).Info(utils.LogICQCallbackWithHostZone(query.ChainId, query.CallbackId, "Retrying query..."))
+		return false, k.RetryICQRequest(ctx, query)
```
ethan-stride, line 114: So this function can actually fire off a query retry in this case... 
The "check" function was a pure functional computation except for this case which can have side effect so maybe we should name it handle timedout queries? If it is ok for this function to do the work then why return a boolean about if the callback should be executed instead of executing the callback right here?

ethan-stride, line 114: If we were going to have the max retries so it doesn't infinite loop we would make it a 3rd parameter on RetryICQRequest which gets decremented each time it is called? Or were you thinking of keeping this count down on the query object itself? or alternatively make the number of retries left saved as part of the callback?
Also when the number of retries are used up, how do you specify to reject or run the callback at that point?

x/stakeibc/keeper/icqcallbacks_validator_exchange_rate.go
```diff
@@ -37,26 +36,69 @@ func ValidatorExchangeRateCallback(k Keeper, ctx sdk.Context, args []byte, query
...
+	// If we are in the LSMLiquidStake callback, finish the transaction
+	if inLSMLiquidStakeCallback {
+		if err := k.LSMSlashQueryCallback(ctx, hostZone, query, validatorWasSlashed); err != nil {
+			return errorsmod.Wrapf(err, "unable to finish LSM liquid stake")
+		}
+	}
+
+	// If the validator was slashed, we'll have to issue a delegator shares query to determine
+	// the magnitude of the slash
+	if validatorWasSlashed {
```
ethan-stride, line 75: The comments help, but these cases feel very hard to understand.  It is really a state machine, maybe we can make it easier to understand as explicit named states and the transitions between them instead of these multiple cases with different returns?

x/stakeibc/keeper/icqcallbacks_validator_exchange_rate.go
```diff
@@ -37,26 +36,69 @@ func ValidatorExchangeRateCallback(k Keeper, ctx sdk.Context, args []byte, query
...
-		return errorsmod.Wrapf(types.ErrValidatorNotFound, "no registered validator for address (%s)", queriedValidator.OperatorAddress)
+		return false, errorsmod.Wrapf(types.ErrValidatorNotFound, "no registered validator for address (%s)", queriedValidator.OperatorAddress)
 	}
 	previousSharesToTokensRate := validator.InternalSharesToTokensRate
 
 	// If the validator's delegation shares is 0, we'll get a division by zero error when trying to get the exchange rate
 	//  because `validator.TokensFromShares` uses delegation shares in the denominator
 	if queriedValidator.DelegatorShares.IsZero() {
-		return errorsmod.Wrapf(types.ErrDivisionByZero,
+		return false, errorsmod.Wrapf(types.ErrDivisionByZero,
```
ethan-stride, line 101: This isn't a new issue with this PR but if I am understanding right, we won't be able to check if the validator was slashed or not if we don't have any shares there yet.  This is a problem because a 0 weight validator which we will still accept LSM tokens for could have 0 of our delegator shares on it; but then we won't be able to detect slashes and the users on that validator can then cash out the full amount to stTokens and be overpaid by the protocol... is this just currently unavoidable risk since we only detect slash via delegator shares changes?

x/stakeibc/keeper/icqcallbacks_validator_exchange_rate.go
```diff
@@ -72,97 +114,81 @@ func ValidatorExchangeRateCallback(k Keeper, ctx sdk.Context, args []byte, query
...
+func (k Keeper) LSMSlashQueryTimeout(ctx sdk.Context, hostZone types.HostZone, query icqtypes.Query) error {
+	var callbackData types.ValidatorExchangeRateQueryCallback
+	if err := proto.Unmarshal(query.CallbackData, &callbackData); err != nil {
+		return errorsmod.Wrapf(err, "unable to unmarshal validator exchange rate callback data")
+	}
+	lsmLiquidStake := *callbackData.LsmLiquidStake
 
-		EmitFailedLSMLiquidStakeEvent(ctx, hostZone, lsmTokenDeposit, errorMessage)
-		k.Logger(ctx).Error(errorMessage)
+	k.FailLSMLiquidStake(ctx, hostZone, lsmLiquidStake, "query timed out")
```
ethan-stride, line 144: Where are we checking if the query did actually time out?  This method should be called trigger timeout or something if the intention is it is a helper to force a failed query

x/stakeibc/keeper/msg_server_submit_tx.go
```diff
@@ -358,9 +359,28 @@ func (k Keeper) GetLightClientTimeSafely(ctx sdk.Context, connectionID string) (
 }
 
 // Submits an ICQ to get a validator's exchange rate
-func (k Keeper) QueryValidatorExchangeRate(ctx sdk.Context, chainId, validatorAddress string) error {
+func (k Keeper) QueryValidatorExchangeRate(
+	ctx sdk.Context,
+	chainId string,
+	validatorAddress string,
+	callbackDataBz []byte,
+	aggressiveTimeout bool,
```
ethan-stride, line 367: The aggressiveTimeout param feels like code smell of leaky abstraction.  How this method is arrived at (LSM liquid stake or a periodic accounting check) affects its inside implementation behavior through this variable... so someone reading this function has to be aware of the higher level complexities of how it can be called instead of having all context by just looking at the inside of the function itself.  Why not make two versions of this function called LiquidStakingQueryValidatorExchangeRate and PeriodicQueryValidatorExchange rate and avoid the aggressiveTimeout var and many case checkings inside and how we call to it?

ethan-stride (commented):
Somethings are cleaner but there are some functions which feel like they are not separated to the right level and bring in too much context of how they are being called to...  I like the ability to have the different timeout behaviors and I think a max retries is a good idea, that is definitely an improvement but the function which is doing the checking and triggering I think is very confusing.  It could be as simple as splitting it up into smaller more atomic functions or it might need a different organization like a literal state machine instead of lots of cases to check.

x/interchainquery/keeper/msg_server.go
```diff
@@ -93,6 +94,34 @@ func (k Keeper) VerifyKeyProof(ctx sdk.Context, msg *types.MsgSubmitQueryRespons
 	return nil
 }
 
+// Checks if the query timed out, and if so, checks the timeout policy
+// Returns whether the callback should be executed (only occurs if explicitly indicated)
+func (k Keeper) CheckForQueryTimeout(ctx sdk.Context, query types.Query) (executeCallback bool, err error) {
```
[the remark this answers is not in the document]
riley-stride: Could we either name this more clearly, maybe `CheckForQueryTimeoutAndExecutePolicy`, or pull the policy execution out of this function?

x/interchainquery/keeper/msg_server.go
```diff
@@ -150,19 +179,7 @@ func (k msgServer) SubmitQueryResponse(goCtx context.Context, msg *types.MsgSubm
...
-	}
-	if query.TimeoutTimestamp < currBlockTime {
-		k.Logger(ctx).Error(utils.LogICQCallbackWithHostZone(query.ChainId, query.CallbackId,
-			"QUERY TIMEOUT - QueryId: %s, TTL: %d, BlockTime: %d", query.Id, query.TimeoutTimestamp, ctx.BlockHeader().Time.UnixNano()))
-		return &types.MsgSubmitQueryResponseResponse{}, nil
-	}
-
 	// Immediately delete the query so it cannot process again
-	// We don't want to delete timed-out queries because they'll be retried
 	k.DeleteQuery(ctx, query.Id)
```
riley-stride, line 183: Moving the timeout check below this line changes the bevahior: now all queries will be deleted even if timed out. I think this is desired, but want to double check with you that that's intentional + newly introduced behavior.

x/interchainquery/keeper/keeper.go
```diff
@@ -77,17 +78,9 @@ func (k *Keeper) RetryICQRequest(ctx sdk.Context, query types.Query) error {
 	k.Logger(ctx).Info(utils.LogWithHostZone(query.ChainId,
 		"Queuing ICQ Retry - Query Type: %s, Query ID: %s", query.CallbackId, query.Id))
 
-	if err := k.ValidateQuery(ctx, query); err != nil {
```
riley-stride, line 80: Why are we removing this validation?

proto/stride/interchainquery/v1/genesis.proto
```diff
@@ -7,6 +7,12 @@ import "cosmos_proto/cosmos.proto";
 
 option go_package = "github.com/Stride-Labs/stride/v9/x/interchainquery/types";
 
+enum TimeoutPolicy {
+  REJECT_QUERY_RESPONSE = 0;
```
riley-stride, line 11: nit: do we need this enum entry if it's not used? maybe there's a future use case for it

x/interchainquery/keeper/msg_submit_query_response_test.go
```diff
@@ -40,15 +39,18 @@ func (s *KeeperTestSuite) SetupMsgSubmitQueryResponse() MsgSubmitQueryResponseTe
...
-		ConnectionId:    s.TransferPath.EndpointA.ConnectionID,
-		QueryType:       types.BANK_STORE_QUERY_WITH_PROOF,
-		RequestData:     append(data, []byte(HostChainId)...),
-		TimeoutDuration: time.Duration(1_000_000_000),
+		Id:               expectedId,
+		CallbackId:       "withdrawalbalance",
+		CallbackModule:   "stakeibc",
+		ChainId:          HostChainId,
+		ConnectionId:     s.TransferPath.EndpointA.ConnectionID,
+		QueryType:        "store/bank", // intentionally leave off key to skip proof
```
riley-stride, line 50: why no proof, for easier testing?

x/interchainquery/keeper/msg_submit_query_response_test.go
```diff
@@ -91,33 +94,64 @@ func (s *KeeperTestSuite) TestMsgSubmitQueryResponse_UnknownId() {
...
+	// set timeout to be expired and set the policy to reject
+	tc.query.TimeoutTimestamp = uint64(1)
+	tc.query.TimeoutPolicy = types.TimeoutPolicy_REJECT_QUERY_RESPONSE
+	s.App.InterchainqueryKeeper.SetQuery(s.Ctx, tc.query)
+
+	_, err := s.GetMsgServer().SubmitQueryResponse(tc.goCtx, &tc.validMsg)
+	s.Require().NoError(err)
 
-	// set timeout to be expired
+	// check that the query original query was deleted,
```
riley-stride, line 108: nit
```suggestion
	// check that the original query was deleted,
```

x/interchainquery/keeper/msg_submit_query_response_test.go
```diff
@@ -91,33 +94,64 @@ func (s *KeeperTestSuite) TestMsgSubmitQueryResponse_UnknownId() {
...
 	s.Require().NoError(err)
-	s.Require().NotNil(resp)
 
-	// check that the query was not deleted (since the query timed out)
+	// check that the query original query was deleted,
+	// but that a new one was created for the retry
 	_, found := s.App.InterchainqueryKeeper.GetQuery(s.Ctx, tc.query.Id)
-	s.Require().True(found)
+	s.Require().False(found, "original query should be removed")
+	s.Require().Len(s.App.InterchainqueryKeeper.AllQueries(s.Ctx), 1, "there should be one new query")
```
riley-stride, line 128: Can we also check the attributes of the new query are what we expect them to be?

x/interchainquery/keeper/msg_submit_query_response_test.go
```diff
@@ -91,33 +94,64 @@ func (s *KeeperTestSuite) TestMsgSubmitQueryResponse_UnknownId() {
...
+func (s *KeeperTestSuite) TestMsgSubmitQueryResponse_Timeout_ExecuteCallback() {
 	tc := s.SetupMsgSubmitQueryResponse()
 
+	// set timeout to be expired and set the policy to retry
+	tc.query.TimeoutTimestamp = uint64(1)
+	tc.query.TimeoutPolicy = types.TimeoutPolicy_EXECUTE_QUERY_CALLBACK
 	s.App.InterchainqueryKeeper.SetQuery(s.Ctx, tc.query)
 
 	// rather than testing by executing the callback in its entirety,
 	//   check by invoking it without a registered host zone and catching the appropriate error
```
riley-stride, line 140: just out of curiosity, why test it via this failure path? if possible to test the success case, I think we should! possible the proof checking logic will break the unit test...

x/interchainquery/keeper/msg_submit_query_response_test.go
```diff
@@ -91,33 +94,64 @@ func (s *KeeperTestSuite) TestMsgSubmitQueryResponse_UnknownId() {
...
+	s.Require().ErrorContains(err, "no registered zone for queried chain ID", "callback was invoked")
+}
+
+func (s *KeeperTestSuite) TestMsgSubmitQueryResponse_FindAndInvokeCallback() {
+	tc := s.SetupMsgSubmitQueryResponse()
+
+	s.App.InterchainqueryKeeper.SetQuery(s.Ctx, tc.query)
+
+	// The withdrawal balance test is already covered in it's respective module
+	// For this test, we just want to check that the callback function as invoked
```
riley-stride, line 151: nit
```suggestion
	// For this test, we just want to check that the callback function is invoked
```

x/stakeibc/keeper/msg_server_submit_tx.go
riley-stride: +1, aggressiveTimeout could be more clear

x/stakeibc/keeper/msg_server_submit_tx.go
```diff
@@ -358,9 +359,28 @@ func (k Keeper) GetLightClientTimeSafely(ctx sdk.Context, connectionID string) (
...
 	k.Logger(ctx).Info(utils.LogWithHostZone(chainId, "Submitting ICQ for validator exchange rate to %s", validatorAddress))
 
+	// If this query is executed from LSM, there should be a more aggressive timeout since it's UX blocking,
+	// and, in the event of a timeout, we should still enter the callback so we can alert the user that the query failed
+	// If this query is not for an LSM liquid stake, we can have a more relaxed timeout
+	var timeoutDuration time.Duration
+	var timeoutPolicy icqtypes.TimeoutPolicy
+	if aggressiveTimeout {
+		timeoutDuration = LSMSlashQueryTimeout
+		timeoutPolicy = icqtypes.TimeoutPolicy_EXECUTE_QUERY_CALLBACK
```
riley-stride, line 378: If I were invoking `QueryValidatorExchangeRate`, I would be surprised to learn that setting `aggressiveTimeout` determines the timeout policy

x/stakeibc/keeper/icqcallbacks_validator_exchange_rate.go
```diff
@@ -12,7 +11,6 @@ import (
 
 	"github.com/Stride-Labs/stride/v9/utils"
 	icqtypes "github.com/Stride-Labs/stride/v9/x/interchainquery/types"
-	recordstypes "github.com/Stride-Labs/stride/v9/x/records/types"
 	"github.com/Stride-Labs/stride/v9/x/stakeibc/types"
 )
```
riley-stride, line 16: Might be due to the diffs not showing clear changes: I'm having a hard time parsing the logic changes in this file (even when reviewing commit by commit), and these changes are sensitive. Would you mind doing a 1-2 min loom on the changes in this file?

riley-stride (changes requested):
Solid progress on this. I need to better understand the changes to `icqcallbacks_validator_exchange_rate.go` before approving.

sampocs:
@ethan-stride 

> Somethings are cleaner but there are some functions which feel like they are not separated to the right level and bring in too much context of how they are being called to... I like the ability to have the different timeout behaviors and I think a max retries is a good idea, that is definitely an improvement but the function which is doing the checking and triggering I think is very confusing. It could be as simple as splitting it up into smaller more atomic functions or it might need a different organization like a literal state machine instead of lots of cases to check.

I agree it's still not great, but it's a very tricky flow. I thought about it for awhile, but once I got into the refactoring, I found it quite difficult to try and organize things coherently. If you have a concrete suggestion, I'm all ears! Or even better, if you're visualizing something specific, feel free to open a PR to this one. But I'm not sure how to address this comment otherwise

x/interchainquery/keeper/msg_server.go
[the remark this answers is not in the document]
sampocs: This is a control flow function mostly meant to break up the main SubmitQueryResponse function - I don't think it would make sense on the query type. 

I could maybe just move it back into `SubmitQueryResponse` ? It gets a bit sloppy with the branching though

```go
if query.HasTimedOut(ctx) {
	k.Logger(ctx).Error(utils.LogICQCallbackWithHostZone(query.ChainId, query.CallbackId,
		"QUERY TIMEOUT - QueryId: %s, TTL: %d, BlockTime: %d", query.Id, query.TimeoutTimestamp, ctx.BlockHeader().Time.UnixNano()))

	switch query.TimeoutPolicy {
	case types.TimeoutPolicy_REJECT_QUERY_RESPONSE:
		k.Logger(ctx).Info(utils.LogICQCallbackWithHostZone(query.ChainId, query.CallbackId, "Rejecting query"))
		return nil

	case types.TimeoutPolicy_RETRY_QUERY_REQUEST:
		k.Logger(ctx).Info(utils.LogICQCallbackWithHostZone(query.ChainId, query.CallbackId, "Retrying query..."))
		return k.RetryICQRequest(ctx, query)

	case types.TimeoutPolicy_EXECUTE_QUERY_CALLBACK:
		k.Logger(ctx).Info(utils.LogICQCallbackWithHostZone(query.ChainId, query.CallbackId, "Executing callback..."))
	
	        if err := k.InvokeCallback(ctx, msg, query); err != nil {
		       k.Logger(ctx).Error(utils.LogICQCallbackWithHostZone(query.ChainId, query.CallbackId,
			"Error invoking ICQ callback, error: %s, QueryId: %s, QueryType: %s, ConnectionId: %s, QueryRequest: %v, QueryReponse: %v",
			err.Error(), msg.QueryId, query.QueryType, query.ConnectionId, query.RequestData, msg.Result))
		        return  err
                }
                return nil
	}

	default:
		return false, fmt.Errorf("Unsupported query timeout policy: %s", query.TimeoutPolicy.String())
	}
}

if err := k.InvokeCallback(ctx, msg, query); err != nil {
	k.Logger(ctx).Error(utils.LogICQCallbackWithHostZone(query.ChainId, query.CallbackId,
	    "Error invoking ICQ callback, error: %s, QueryId: %s, QueryType: %s, ConnectionId: %s, QueryRequest: %v, QueryReponse: %v",
	    err.Error(), msg.QueryId, query.QueryType, query.ConnectionId, query.RequestData, msg.Result))
	return  err
}
return nil
```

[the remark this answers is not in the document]
sampocs: (side note, I can move the callback error log to InvokeCallback)

x/interchainquery/keeper/msg_server.go
```diff
@@ -93,6 +94,34 @@ func (k Keeper) VerifyKeyProof(ctx sdk.Context, msg *types.MsgSubmitQueryRespons
 	return nil
 }
 
+// Checks if the query timed out, and if so, checks the timeout policy
+// Returns whether the callback should be executed (only occurs if explicitly indicated)
+func (k Keeper) CheckForQueryTimeout(ctx sdk.Context, query types.Query) (executeCallback bool, err error) {
+	if !query.HasTimedOut(ctx) {
+		return true, nil
```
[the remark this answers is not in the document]
sampocs: > So if the query hasn't timed out we always want to execute the callback?
correct

This is in the msg server tx so this is only called when a query is submitted. Are you saying that your nervous we might accidentally call this function elsewhere? If so, I think that's probably not a good reason to remove it - but it could be solved with a better name.

I'd also be fine with removing it and using something like the approach in the above comment since it seemed to have confused both you and RIley

x/interchainquery/keeper/msg_server.go
sampocs: yeah I hear ya. The original motivation was so that we wouldn't have to call InvokeCallback two different times. But I see the confusion, so maybe it's best to remove this function or do the refactor above

x/interchainquery/keeper/msg_server.go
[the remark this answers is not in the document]
sampocs: we could also take what's above and throw it in a function so then we have

```
if query.HasTimedOut(ctx) {
    k.HandleTImeout(...)
} else {
  k.InvokeCallback()
}
```

I think that's my vote

x/interchainquery/keeper/msg_server.go
sampocs: yeah an extra attribute with the count on the query object is how I was thinking about it

If all retries are used up and it still times out, we should reject it.

I think if we add the retry count, it might not make sense to remove the timeout policy

x/stakeibc/keeper/icqcallbacks_validator_exchange_rate.go
sampocs: can you double click on this - maybe with some pseudo code?

x/stakeibc/keeper/icqcallbacks_validator_exchange_rate.go
sampocs: https://github.com/Stride-Labs/stride/pull/837/files/a2287f26558eb3f22a544ddb259168464a6077de#diff-10ab5e08a468a7d9d65c2b1bc816f16246abc31f96c5a56a056952fd70308424R44-R49

x/stakeibc/keeper/icqcallbacks_validator_exchange_rate.go
sampocs: really good call out - I hadn't thought about that! 

To clarify thought, we detect slashes using the exchange rate query - irrespective of whether the number of delegated tokens changed - so we would be able to correctly identify that the validator was slashed. Additionally, zero-weight validators can have a delegated balance before a rebalancing.

So the concerning case is just the first delegation after a rebalancing to a zero-weight validator - in which case the RR is not adjusted

x/stakeibc/keeper/msg_server_submit_tx.go
sampocs: Objective is to avoid having to duplicate code across two functions which then puts a burden on the programmer to keep them in sync.

Maybe a middle ground is something like

```go
func SubmitValidatorExchangeRateICQ(callbackArgs, timeoutPolicy, timeoutDuration) {
   // get host zone
   // get validator
  // build query format
  // submit ICQ
}

func LSMQueryValidatorExchangeRate() {
   SubmitValidatorExchangeRateICQ(...)
}
func PeriodicQueryValidatorExchangeRate() {
   SubmitValidatorExchangeRateICQ(...)
}
```

x/interchainquery/keeper/msg_server.go
sampocs: yup that's right and it's how it's done on mainnet right now

An earlier commit to LSM changed to behavior to _not_ delete the query so that it can be scooped & retried in the end blocker. But that change is essentially reverted by this PR

x/interchainquery/keeper/keeper.go
sampocs: it's handled by `SubmitICQRequest`

proto/stride/interchainquery/v1/genesis.proto
sampocs: I was originally using it and then changed it last minute haha. This is our current behavior on mainnet. Once the three of us land on timeout policies for each callback, if this is not used we can remove it

x/stakeibc/keeper/icqcallbacks_validator_exchange_rate.go
sampocs: for sure!

x/stakeibc/keeper/msg_server_submit_tx.go
sampocs: do you think [this](https://github.com/Stride-Labs/stride/pull/837#discussion_r1245645683) would be better

x/interchainquery/keeper/msg_submit_query_response_test.go
sampocs: done

x/interchainquery/keeper/msg_submit_query_response_test.go
sampocs: this test belongs to the ICQ module and is just meant to test the generic case of timeout. So I don't think it makes sense to bundle the logic from one of our stakeibc callbacks here (we'd essentially be duplicating the other queries unit test and would have to mock out a ton of state)

That said, I don't love the that this just checks for a random chainId error so if you have a suggestion on how to clean that up, i'm all ears!

x/interchainquery/keeper/msg_submit_query_response_test.go
sampocs: done

x/interchainquery/keeper/msg_submit_query_response_test.go
sampocs: done

x/interchainquery/keeper/msg_server.go
ethan-stride: Extra attribute on the query object makes sense to me and is the cleanest option compared to the other things I was thinking like recursive calls to the function and the count param only existing in the function state.

Not only do I like the additional specification power the timeout policy gives, I think we need the timeout policy for what happens in the end when the retries run out.  Instead of 3 policy options (run_callback, fail, or retry)  We can have a policy and a number of retries.  While retries is >0 we retry but then the policy specifies what happens at the end.  And of course the number of retries can start at 0 to specify always run the callback or always fail gracefully

x/stakeibc/keeper/msg_server_submit_tx.go
ethan-stride: Yes exactly, no need to repeat the core logic but this also splits the behaviors into separate functions with better naming so it is clearer IMO how the function was arrived at compared to before where the condition checking internal to the function was doing this routing

x/interchainquery/keeper/msg_server.go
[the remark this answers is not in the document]
sampocs: yeah your mental model is right. It's unclear to me whether it's intentionally set up that way to fall into an OOP paradigm or just because it turns out that almost everything needs access to the store.  But I've found it's usually best to just throw everything under the keeper so it's easier to find down the road. 

For instance, our GetValidatorFromHostZone function just takes a list of validators so it doesn't need access to the keeper, and thus is defined as just a standard function,. But I've found myself often doing `k.` and searching for it, thinking I'm getting the name wrong, only to find it's just not a part of the keeper.

I generally use the following framework, and find that I usually run into one of four scenarios:
1. Stateful logic (extends keeper)
2. Business/control flow/main logic (extends keeper)
3. Property like functions on a data structure (extend individual data type in `types`)
4. Stateless utility functions <- this is the one where I often wind up throwing it in the keeper for the reasoning above and because it usually is part of a flow in (2)

x/interchainquery/keeper/msg_server.go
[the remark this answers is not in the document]
sampocs: yeah I think part of the confusion was also that the execution was kinda half baked in that it only executes the query retry but did not execute the callbacks.

Sounds like we're on the same page that the refactor is better

x/stakeibc/keeper/msg_server_submit_tx.go
sampocs: 🤝

x/stakeibc/keeper/icqcallbacks_validator_exchange_rate.go
ethan-stride: So what I was thinking was for the 2x2 square you talk through in the loom, you would essentially have states like LSM_TRIGGERED_SLASH_DETECTED, PERIODIC_SLASH_DETECTED, LSM_TRIGGERED_NO_SLASH, PERIODIC_NO_SLASH etc. and there would be a handler for each of those cases, but now I am realizing all 4 of these cases are mutually exclusive and there is no event which moves us from one to another of these states.  I was hoping that you could have different actions in separate functions based on the state but now I see that what makes this hard to cleanly split up is the async call structure which means the handler code has to occur inside the callback.  I still think it is confusing but my suggested solution isn't going to be any better.  This was to avoid the multiple if statements with different exits from the same function because it seemed like those required a lot of comments and context about how the callback was arrived at to make sense, but I don't know if it really is better described with states now that I think about it deeper.

x/interchainquery/types/query.go
```diff
@@ -1,9 +1,18 @@
...
 // the timeout timestamp
-func (q Query) HasTimedOut(ctx sdk.Context) bool {
-	return q.TimeoutTimestamp < uint64(ctx.BlockTime().UnixNano())
+func (q Query) HasTimedOut(currentBlockTime time.Time) bool {
+	return q.TimeoutTimestamp < uint64(currentBlockTime.UnixNano())
+}
+
+// Prints an abbreviated query description for logging purposes
+func (q Query) Description() string {
+	return fmt.Sprintf("QueryId: %s, QueryType: %s, ConnectionId: %s, QueryRequest: %v",
```
ethan-stride, line 16: This makes a lot more sense to me to have this helper on the query.  I get that it makes the logging have to split up the string generation and log output steps but I still like this better personally.

sampocs: yeah I can't stand the long log statements

x/stakeibc/keeper/icqcallbacks_validator_exchange_rate.go
sampocs: sounds like you're joining the club of having an epiphany that sounds good but then watching it break down as you get closer to the implementation haha

x/stakeibc/keeper/msg_server_submit_tx.go
riley-stride: I like this solution

x/stakeibc/keeper/msg_server_submit_tx.go
riley-stride: Lgtm

x/interchainquery/keeper/msg_submit_query_response_test.go
riley-stride: It feels less robust than it should be, but I don't have a concrete proposal for a solution -- I can't think of a way to bolster the test case. Maybe we leave it for now unless @ethan-stride has any ideas.

x/interchainquery/keeper/msg_server.go
riley-stride: On board with this solution!

riley-stride (approved):
lgtm, loom was super helpful
going to ask for a few more reviews on this per ethan's message

x/interchainquery/keeper/msg_submit_query_response_test.go
sampocs: leaving as is

sampocs:
@riley-stride @ethan-stride 

### Addressing MaxRetries
Coming back to this finally... I know we had talked about adding a max retries cap on the query - possible I'm forgetting things from our earlier convo, but this is my opinion now:

I think our queries fall into 3 camps:
1. **Epochly low-stakes queries** (withdrawal balance, fee balance, etc)
  * The main thing that matters here is that we do not have multiple of the same query in flight at the same time. It's also not a big deal if these timeout and we miss an epoch, and the relaying of them has been quite successful. 
  * For these, I don't think we need to over-optimize with retries - I think we can just set a 5 hour timeout and ignore them if they timeout
2. **UX blocking queries** (e.g. delegator shares prevents new LSM liquid stakes to that validator) 
  * I don't see any reason to cap the retries here considering we're in a tough spot if we stop issuing these queries and liquid stakes are still blocked (we'd probably need an upgrade)
3. **Action-on-timeout queries** (e.g. validator exchange rate query for LSM)
  * Here the policy is obvious as we need to execute the callback

Given this, I don't think we need max retries. I've updated the policies in the PR description.

### Remaining (unrelated) work in this PR
1. Delete original query during a retry so there are not multiple in flight
2. Block re-submitting DelegatorShares query if one is in progress

x/stakeibc/keeper/icacallbacks_reinvest.go
```diff
@@ -115,15 +115,18 @@ func ReinvestCallback(k Keeper, ctx sdk.Context, packet channeltypes.Packet, ack
...
 	query := icqtypes.Query{
 		ChainId:         chainId,
 		ConnectionId:    hostZone.ConnectionId,
 		QueryType:       icqtypes.BANK_STORE_QUERY_WITH_PROOF,
 		RequestData:     queryData,
 		CallbackModule:  types.ModuleName,
 		CallbackId:      ICQCallbackID_FeeBalance,
-		TimeoutDuration: time.Hour,
-		TimeoutPolicy:   icqtypes.TimeoutPolicy_RETRY_QUERY_REQUEST,
+		TimeoutDuration: timeoutDuration,
```
riley-stride, line 128: I believe this timeout should be a relative amount of time into the future from now (e.g. time.Hour). But here it's being set to the absolute time one block before the start of the next epoch. Is that right?

x/stakeibc/keeper/icacallbacks_reinvest_test.go
```diff
@@ -121,7 +130,7 @@ func (s *KeeperTestSuite) TestReinvestCallback_Successful() {
 	s.Require().Equal(HostChainId, query.ChainId, "query chain ID")
 	s.Require().Equal(ibctesting.FirstConnectionID, query.ConnectionId, "query connection ID")
 	s.Require().Equal(icqtypes.BANK_STORE_QUERY_WITH_PROOF, query.QueryType, "query type")
-	s.Require().Equal(time.Hour, query.TimeoutDuration, "query timeout duration")
+	s.Require().Equal(tc.initialState.durationUntilNextEpoch, query.TimeoutDuration, "query timeout duration")
```
riley-stride, line 133: In the test, this looks like a relative time

x/stakeibc/keeper/icacallbacks_reinvest.go
sampocs: it's being set to the relative time before the epoch. The calculation is:
relativeTimeout = epochEndTime - currentTime

riley-stride: Oh duh, I missed `.sub(ctx.blocktime())`

riley-stride: All good here!

riley-stride (approved):

x/stakeibc/keeper/icacallbacks_reinvest_test.go
riley-stride: Resolved

x/stakeibc/keeper/icacallbacks_reinvest.go
sampocs: There's actually a TImeout field on the query object that represents the absolute time and is set `SubmitICQRequest`

One thing we could do it allow the specification of _either_ Timeout or TimeoutDuration

sampocs: might make this code a little simpler, but the submission code a little messier

x/stakeibc/keeper/icqcallbacks_validator_exchange_rate.go
ethan-stride: You had asked if the LSM representing shares would fix this... I don't think it actually would because the core issue is that we don't have any shares on a given 0 weight val until someone comes in from out of the blue with an LSM liquid stake request.  In the shares representation, as long as we have an accurate exchange rate everything should be good... maybe for LSM coming from a 0 weight validator where we previously believed there to be zero shares we always trigger the second ICQ to get an updated exchange rate no matter how small the LSM stake amount is to account for this?

ethan-stride (approved):
LGTM if we think through the edge case about how to handle LSM liquid stakes coming from validators which are 0 weight and we thought had no shares from the state stride side tracks.

x/stakeibc/keeper/icqcallbacks_validator_exchange_rate.go
sampocs: So I think there's two main concerns here at the end of the day:
1. Are we minting the correct number of stTokens / was a user able to escape a slash?
2. Are we updating our redemption rate correctly?

For (1), if the exchange rate is updated, we always mint the correct number of stTokens now because the slash would have been reflected in the exchange rate.  If the exchange rate _is not_ updated, we do not mint the correct number of stTokens, but the magnitude of this error is capped by the checkpoint. 

For (2), the redemption rate is basically getting at the ratio of stTokens to native tokens. When calculating the number of native tokens, we need to sum the tokens that are currently staked. For zero-weight validators with respect to LSM, this can come from two sources: (a) the LSM liquid stakes that have not been detokenized yet, and (b) the liquid stakes that have been detokenized and is living as a native stake.

For (a), we can use the validator exchange rate to get the true value of that LSM token. For (b), we just want to make sure the amount we think we have delegated is correct. I believe if there was stake on that validator, the slash query should adjust it correctly. And if there was no stake on that validator, it's not important to the redemption rate equation.

Does that all track with you?

sampocs: @ethan-stride friendly bump on this, I think it's the last comment blocking the merge

x/stakeibc/keeper/icacallbacks_reinvest.go
sampocs: leaving this as is

x/stakeibc/keeper/icqcallbacks_validator_exchange_rate.go
ethan-stride: Yes, for (1) which is really what I was worried about, I think the key thing is that the magnitude of the error is capped by the checkpoint.  I was thinking about the zero weight validators being more likely to not have been updated recently but that shouldn't matter because the checkpoint sets a max cap on the potential error.  
For (2) I wasn't worried about the redemption rate so much as the exchange rate per validator.  If at some point we give out the "wrong" number of stTokens we still need to use the real number of stTokens in existence when doing all redemption rate computations going forward.  I don't disagree anywhere and we also had discussed part of this issue on the other related PRs, I'll close this issue.

PR #844, opened 2023-06-29
[LSM] Prevent ICAs to validator's with slash query in progress

sampocs:
## Context
A race condition can occur between a slash query and an ICA that modifies a validator's delegation. This manifests itself in two flavors:
```
//	Case 1)
//	         ICQ Lands on Host                                          ICQ Ack on Stride
//	                             ICA Lands on Host    ICA Ack on Stride
//	Case 2)
//	         ICA Lands on Host                                          ICA Ack on Stride
//	                             ICQ Lands on Host    ICQ Ack on Stride
```

We identify if the overlap was possible by checking the following _in the ICQ Ack_:
* check If the delegation amount in Stride's record keeping changed since the query was submitted (case #&ZeroWidthSpace;1)
* check if there there is currently an active ICA in progress (case #&ZeroWidthSpace;2)

However, as if this wasn't complex enough already, there is a small edge case where an undelegation and delegation could have simultaneously occurred with equal amounts in between the slash query. This would be an example of case #1, but where our first check is unable to identify the overlap because the net delegation did not actually change.

```
Ex:
ICQ submitted                  
      ICA Undelegate lands on host           [true delegation -10]
ICQ lands on host                            [looks like a slash]    
              ICA delegate lands on host     [true delegation +10]
      ICA undelegate Ack                     [record keeping delegation -10]
             ICA delegate Ack                [record keeping delegation +10]
ICQ ack on Stride                            [delegation in record keeping did not change 
                                              → query records delegation amount less than record keeping
                                              → incorrectly processed as slash]
```

While the likelihood of this edge case is low, the impact is severe. The solution in this PR is to exclude validator's with a slash query in progress when submitting ICAs.

## Brief Changelog
* Updated GetTargetValAmts to exclude validators with a slash query in progress (this function is called during each ICA)
* Added checks each time we access the map from  GetTargetValAmts to make sure the validator is present
* Removed `GetTotalValidatorDelegations` function since we have this value on the host zone
* Changed GetTotalValidatorWeight to take a list of validators instead of the host zone (so a subset of validators could be provided)
* **Removes sorting of validators on host zone struct** - I don't think there should be any ramifications of this
* **Note**: With this change, we should still unbond and delegate the full amount, the distribution will just be slightly off (but will be corrected during rebalancing).

## Impact described in psuedo-code
Considering it's a bit tough to review this without diving deeper into the code, here's some more detail on the changes to each ICA type. <br>
For context, this is all a consequence of the change to `GetTargetValAmts` which now (a) ignores validators with a slash query in progress and (b) does not sort the array on the host zone struct (it sorts a copy)

### Delegation
Before:
```python
targets = GetTargetValAmtsForHostZone(delegationAmount) # includes all validators
for val in hostZone.Validators: # note: this is sorted from the above function
    valAmount = targets[val]
    if valAmount > 0:
       msgs.append(MsgDelegate{...})
```
After:
```python
targets := GetTargetValAmtsForHostZone(delegationAmount)  # only returns validator w/o slash query in progress
for val in hostZone.Validators: # no longer sorted from above function
   if val not in targets:   # added check since not all vals are in `targets`
     continue

   valAmount = targets[val]
   if valAmount > 0:
      msgs.append(MsgDelegate{...})
```
### Undelegation
Before:
```python
balancedDelegations = GetTargetValAmtsForHostZone(currentDelegation - unbondAmount) # includes all validators

capacities = []
for val in hostZone.Validators:
   balanced = balancedDelegations[val]
   capacities.append({...})

capacityPrioritized = SortCapacity(capacities) # overrides the sorting done by GetTargetValAmtsForHostZone

for val in capacityPrioritized:
   msgs.append(MsgUndelegate{...})
```
After:
```python
balancedDelegations = GetTargetValAmtsForHostZone(currentDelegation - unbondAmount) # excludes slash query in progress vals

capacities = []
for val in hostZone.Validators:
   if val not in balancedDelegations: # added check since not all vals are in `balanceDelegations`
     continue

   balanced = balancedDelegations[val]
   capacities.append({...})

capacityPrioritized = SortCapacity(capacities) 

for val in capacityPrioritized:
   msgs.append(MsgUndelegate{...})
```
### Redelegate
Before:
```python
targets = GetTargetValAmtsForHostZone(totalStakedBalance) # includes all validators

rebalancings = []
for val in hostZone.Validators:
   targetDelegation = targets[val]
   rebalancings.append({DelegationChange: targetDelegation - val.CurrentDelegation})

for rebalancing in rebalancings:
   msgs.append(MsgRedelegate{...}) 
```

After:
```python
targets = GetTargetValAmtsForHostZone(totalStakedBalance) # excludes vals with slash query in progress

rebalancings = []
for val in hostZone.Validators:
   if val not in targets:  # added check since not all vals are in `targets`
      continue

   targetDelegation = targets[val]
   rebalancings.append({DelegationChange: targetDelegation - val.CurrentDelegation})

for rebalancing in rebalancings:
   msgs.append(MsgRedelegate{...})

### CODE REVIEW

File: x/stakeibc/keeper/validator_selection.go [hunk 1]
```diff
@@ -244,26 +249,33 @@ func (k Keeper) GetValidatorDelegationDifferences(ctx sdk.Context, hostZone type
...
+	if !totalDelegation.IsPositive() {
 		return nil, errorsmod.Wrapf(sdkerrors.ErrInvalidRequest,
-			"Cannot calculate target delegation if final amount is less than or equal to zero (%v)", finalDelegation)
+			"Cannot calculate target delegation if final amount is less than or equal to zero (%v)", totalDelegation)
+	}
+
+	// Ignore any validators with a slash query in progress
+	validators := []types.Validator{}
+	for _, validator := range hostZone.Validators {
+		if !validator.SlashQueryInProgress {
```
ethan-stride, line 265: Still thinking through is this really what we want to do, by doing the ICQ check deep inside this function and filtering out validators instead of at the time of specific ICA kick off could potentially affect all ICAs?  In the past I thought we had said we wouldn't ever want to block ICAs needed for standard route liquid stakes by the ICQ but we would want to block LSM liquid stakes (since we need an accurate exchange rate to deal with the shares representation).

Is there a guarantee that these helpers are only going to be called in the LSM related routes but not the pre-existing rebalance type functions?  Also, by filtering out validators, the balancing logic is going to be slightly different.  Instead of computing how much should go to that validator but then waiting to execute it until ICQs are clear, any mass which would have gone to that validator will now be rebalanced over to other validators.

x/stakeibc/keeper/validator_selection.go
```diff
@@ -244,26 +249,33 @@ func (k Keeper) GetValidatorDelegationDifferences(ctx sdk.Context, hostZone type
 	return delegationDeltas, nil
 }
 
-// This will get the target validator delegation for the given hostZone
-// such that the total validator delegation is equal to the finalDelegation
-// output key is ADDRESS not NAME
-func (k Keeper) GetTargetValAmtsForHostZone(ctx sdk.Context, hostZone types.HostZone, finalDelegation sdkmath.Int) (map[string]sdkmath.Int, error) {
+// This will split a total delegation amount across validators, according to weights
+// It returns a map of each portion, key'd on validator address
+// Validator's with a slash query in progres are excluded
```
ethan-stride, line 254: nit: "progres" -> "progress"

x/stakeibc/keeper/validator_selection.go
sampocs: >  In the past I thought we had said we wouldn't ever want to block ICAs needed for standard route liquid stakes by the ICQ but we would want to block LSM liquid stakes

This is true, but we decided against it eventually considering how dangerous it is to have conflicts and the fact that it only affect ICAs _to that validator_. The case we want to avoid is something like skipping a full epochly delegation.

Re: rebalancing, this is something I realized late last night too. I think we have to subtract out the validators that have a slash query in progress when we get the total

x/stakeibc/keeper/validator_selection.go
sampocs: done

x/stakeibc/keeper/validator_selection.go
ethan-stride: Yes I think this is critical or any validator with slash query in progress will lose all their delegation in a rebalance!  Just saw the update and it looks good, just avoiding those validators with slash query in progress from the total since they will be avoided in the rebalance logic.

ethan-stride: From our offline conversation about this, I think we need to apply the same exclusion logic on the weights when computing the TargetValAmounts so that only the relative weights of validators which do not have a slash query in progress are considered in that computation.

sampocs: aren't they already excluded [here](https://github.com/Stride-Labs/stride/pull/844/files#diff-7eb0c0f79f6e24ba948f073cca1f9f6848f614e735f052b4714b82657237ecebR280)?

ethan-stride: Yes sorry, I see we build the list of validators and then pass it as the incoming argument so any with ICQ active are already excluded.  Then I think this issue is resolved.

ethan-stride:
LGTM!  This will be much cleaner to think about in the future by splitting it up this way to filter at the validator level instead of carrying the ICQ status all the way until ICAs need to be performed.

x/stakeibc/keeper/validator_selection.go
```diff
@@ -244,26 +258,33 @@ func (k Keeper) GetValidatorDelegationDifferences(ctx sdk.Context, hostZone type
...
+	validators := []types.Validator{}
+	for _, validator := range hostZone.Validators {
+		if !validator.SlashQueryInProgress {
+			validators = append(validators, *validator)
+		}
 	}
 
 	// Sum the total weight across all validators
-	totalWeight := k.GetTotalValidatorWeight(hostZone)
+	totalWeight := k.GetTotalValidatorWeight(validators)
```
riley-stride, line 280: Assume we would like to delegate across 5 validators equally, each with 20% weight. But val1 has an ICQs in flight — would `totalWeight` here be 80%, and therefore would each of val2 through val5 receive 25% instead of 20%? 

Thinking about whether we'd want to give each of va2 through val5 20% weight, and keep what would have gone to the first val undelegated, to be delegated next time, OR delegate everything to val2 through val5, some of which would then be redelegated to val1 on the next rebalancing.

x/stakeibc/keeper/unbonding_records.go
```diff
@@ -92,11 +92,12 @@ func (k Keeper) GetTotalUnbondAmountAndRecordsIds(ctx sdk.Context, chainId strin
...
-//  and their fair portion of the total stake based on their weights
-//  (i.e. their balanced delegation)
+// and their fair portion of the total stake based on their weights
+// (i.e. their balanced delegation)
+//
 // Validators with a balanced delegation less than their current delegation
-//  are already at a deficit, are not included in the returned list,
-//  and thus, are will not incur any unbonding
+// are already at a deficit, are not included in the returned list,
+// and thus, are will not incur any unbonding
```
riley-stride, line 100: nit
```suggestion
// and thus, will not incur any unbonding
```

x/stakeibc/keeper/msg_server_submit_tx.go
```diff
@@ -61,21 +61,27 @@ func (k Keeper) DelegateOnHost(ctx sdk.Context, hostZone types.HostZone, amt sdk
...
 		splitDelegations = append(splitDelegations, &types.SplitDelegation{
 			Validator: validator.Address,
-			Amount:    relativeAmount.Amount,
+			Amount:    relativeAmount,
 		})
 	}
 	k.Logger(ctx).Info(utils.LogWithHostZone(hostZone.ChainId, "Preparing MsgDelegates from the delegation account to each validator"))
 
+	if len(msgs) == 0 {
+		return errorsmod.Wrap(sdkerrors.ErrInvalidRequest, "Target delegation amount was 0 for each validator")
```
riley-stride, line 82: is this even possible? i don't think delegateOnHost would get called in this scenario... unless maybe the only val that was going to be delegated to happened to have an ICQ in progress?

x/stakeibc/keeper/validator_selection.go
riley-stride: I must be misunderstanding this logic, based on your back n forth with Ethan in the PR comments. 

To help me undestand, maybe a simple question: in the case I described above, would the delegations made to val2 through val5 be their intended amounts, or slightly more (since val1 is excluded from the denominator when calculating the weights)?

x/stakeibc/keeper/validator_selection.go
```diff
@@ -244,26 +258,33 @@ func (k Keeper) GetValidatorDelegationDifferences(ctx sdk.Context, hostZone type
...
 	if totalWeight == 0 {
 		return nil, errorsmod.Wrapf(types.ErrNoValidatorWeights,
 			"No non-zero validators found for host zone %s", hostZone.ChainId)
 	}
 	k.Logger(ctx).Info(utils.LogWithHostZone(hostZone.ChainId, "Total Validator Weight: %d", totalWeight))
 
-	// sort validators by weight ascending, this is inplace sorting!
-	validators := hostZone.Validators
+	// sort validators by weight ascending
 	sort.SliceStable(validators, func(i, j int) bool { // Do not use `Slice` here, it is stochastic
```
riley-stride, line 288: To avoid any uncertainty about the effect of no longer sorting the validators, why not move this sorting block up, above the block that filters out validators who have a slash query in progress? That way the vals on the HZ get sorted here too, and the filtering still applies on the copy of the val array

riley-stride (commented):
My main question is whether we should exclude from the weight calc in `GetTargetValAmtsForHostZone` the weight of the val who has an ICQ in flight

x/stakeibc/keeper/msg_server_submit_tx.go
sampocs: shouldn't be possible in practice, but technically if each validator had a query in progress, it could occur

x/stakeibc/keeper/validator_selection.go
sampocs: > Assume we would like to delegate across 5 validators equally, each with 20% weight. But val1 has an ICQs in flight — would totalWeight here be 80%, and therefore would each of val2 through val5 receive 25% instead of 20%?

yep, that's right, delegations would be slightly more

riley-stride: Is your thinking that it's preferable for everything to be delegated and earning rewards but slightly off of the target weights, vs on the target weights but some is undelegated?

x/stakeibc/keeper/validator_selection.go
sampocs: I think we should remove the sorting unless we can think of a specific reason we'd need it. It's a dangerous operation (remember we once had a chain halt since it was nondeterministic). Informal also specially recommended we remove it for the case of unbondings.

x/stakeibc/keeper/validator_selection.go
sampocs: yeah exactly. I went back and forth on this a bit, but holding back delegations and undelegations feels like a dangerous game. Holding back delegations leads to less rewards. And (maybe this is too dramatic), but holding back undelegations inches slowly towards insolvency

x/stakeibc/keeper/validator_selection.go
riley-stride: I can't think of why we need it, but it _is_ currently being sorted on mainnet with no issues. So it feels to me that removing the sort could be the change that causes nondeterminism, rather than keeping it. 

The scenario we def want to avoid is re-ordering the vals and saving them back to the HZ without a deterministic sort. Don't think that's happening here even if we remove the sort, so we should be safe. 

Agree we should get to the bottom of why we originally sorted inplace... @asalzmann @shellvish do you recall why we sorted inplace here?

sampocs: I don't think the inplace sort was intentional. I think it was a consequence of us calling it off the host zone struct which has a list of pointers

x/stakeibc/keeper/validator_selection.go
riley-stride: Yeah that makes sense, I'm on board with this. Nice use of duration mismatch haha

x/stakeibc/keeper/validator_selection.go
shellvish: Looking through this again, I don't _think_ we need to sort at all. The only impact I can think of is the validator that gets the "overflow" will change, although I don't think this has a meaningful affect in practice.

The reason I think it shouldn't matter is that the output of this function is a map, which isn't ordered to begin with. So, as long as the contents of the map are the same, we shouldn't need to care about the order in which they were generated. 

The one thing that makes me nervous though is that we specifically are doing a reverse sort. I'm not sure why we chose to do that, which makes me think there might have been a reason for it. This was also done around launch though so it's very possible it was just an extra precaution because it makes the function easier to reason about.

@asalzmann might have a better memory of why we chose to sort!

sampocs: I think the non-deterministic part here was the sorting function itself. I think we used to use `slice.Sort` (which turned out to be non-deterministic) and then it was switched to a combo of `slice.SortStable` + reversing the list first (which is deterministic).

Also, worth noting, this has since been updated so that the validator name is [used as a tie breaker ](https://github.com/Stride-Labs/stride/blob/670369ef5a3af070da94ddbd7ed5b0dd832b7f7c/x/stakeibc/keeper/validator_selection.go#L287-L294)when two validator's have the same weight. So *technically*, we don't even need to use SortStable anymore (although we should still just to be safe)

sampocs: @shellvish considering the new tie breaker - do you still think the overflow validator changes? I don't think it does

shellvish: I think that's right re:tiebreaker, but I think it does change the overflow validator (because, I believe, we don't sort the validators on the Host Zone struct). Regardless, I do think we can remove the sorting entirely given the map struct, but would be good to understand why we sorted in the first place!

jstr1121:
> My main question is whether we should exclude from the weight calc in `GetTargetValAmtsForHostZone` the weight of the val who has an ICQ in flight

How rebalancing work if a ICQ for the val in flight? I not rebalance that validator, wouldn't it break the weight state more than rebalancing with previous ratio?

sampocs:
@jstr1121
> How rebalancing work if a ICQ for the val in flight? I not rebalance that validator, wouldn't it break the weight state more than rebalancing with previous ratio?

The validator's delegations will just remain unchanged (i.e. it will be excluded from rebalancing)

jstr1121 (approved):

x/stakeibc/keeper/validator_selection.go
sampocs: leaving removed

PR #852, opened 2023-07-13
Ethan/liquid gov draft WIP (Do not merge)

ethan-stride:
This WIP is a very rough draft of liquid gov

The main idea of this iteration is that a single token can be used in multiple simultaneous votes.

Instead of statically locking tokens, which would constantly have to update when they were locked until if they were used in a new vote after the first one, we store the votes with the time the tokens involved would become available and then from the current votes we can dynamically compute how many tokens should be available for withdraw and at what times.

Currently contains
- user tx triggers ICQ to load specific proposal data from foreign hub
- user tx allows address to deposit stTokens into escrow for voting
- user tx allows address to cast a vote (yes/no/veto) with the amount of tokens being used
- user tx allows address to withdraw stTokens if they are not currently being used to vote

Currently working on and testing (but not included in this WIP snapshot)
- time triggered ICA to tally and cast votes at end of each stride epoch
- Need to account for how many total tokens stride has locked and abstain correct fraction

Adding shortly:
- user tx endpoint to cancel/remove a current setting of their vote (if vote is still in active time window)
- query end points to report how many tokens will become available at what time(s)
- query end points to report more detail about proposals (like human readable strings, current tally)
- auto fetch of proposals so users don't need to trigger it
- keeper logic unit tests, integration tests
- more specific error types and descriptive human readable error messages

Other improvements in process:
- better naming of fields for this to be a stand-alone module and be consistent with newer coding style
- better types, especially for the golang native time.Time vs the protobuf timestamppb.Timestamp
- had started from stride v9, had to upgrade to v10, but now we are on v11 and underlying sdk version is also changed...


ethan-stride:
Superseded by https://github.com/Stride-Labs/stride/pull/878 which is based off of stride v11

PR #857, opened 2023-07-18
[LSM] Replace LSM dockernet host with gaia and lsm support

sampocs:
* Removed LSM dockernet host 
* Updated GAIA host to use one with LSM support


ethan-stride (approved):
LGTM

PR #858, opened 2023-07-18
[LSM] Updated stToken and redemption rate calculation (LSM Token now represents shares)

sampocs:
Closes: #XXX

Previously each LSM Token was denominated in tokens, which made it hard to calculate the true value of the token since a validator could have been slashed after tokenization, but before the liquid stake. A new change to LSM was added such that each LSM Token now maps 1:1 with a delegator share instead (see [this](https://github.com/cosmos/cosmos-sdk/pull/16747) for more context). 

This PR introduces the subsequent changes to the Stride side.

## Brief Changelog
* Updated stToken calculation
```
Previously:
stTokenAmount = LSMTokenAmount / RedemptionRate

Now:
stTokenAmount = LSMTokenAmount * ValidatorExchangeRate / RedemptionRate
```
* Updated redemption rate calculation to determine the value of the escrowed LSM tokens 
```
Previously:
Sum(LSMTokenAmount for each LSMToken)

Now:
Sum(LSMTokenAmount * ValidatorExchangeRate for each LSMToken)
```
* Move stToken calculation into it's own function
* Added unit tests


x/stakeibc/keeper/lsm.go
```diff
@@ -151,13 +151,20 @@ func (k Keeper) GetValidatorFromLSMTokenDenom(denom string, validators []*types.
 	}
 	validatorAddress := split[0]
 
-	// Confirm validator is in Stride's validator set and does not have an active slash query in flight
+	// Confirm the validator:
+	//  1. is in Stride's validator set
```
ethan-stride, line 155: We are going to expand the stride validator set to include basically all known (6 month+) validators so this will still work on the 0 weighted ones correct?

x/stakeibc/keeper/lsm.go
```diff
@@ -151,13 +151,20 @@ func (k Keeper) GetValidatorFromLSMTokenDenom(denom string, validators []*types.
...
+	//  2. does not have an active slash query in flight
+	//  3. has a known exchange rate
 	for _, validator := range validators {
 		if validator.Address == validatorAddress {
 			if validator.SlashQueryInProgress {
 				return types.Validator{}, errorsmod.Wrapf(types.ErrValidatorWasSlashed,
 					"validator %s was slashed, liquid stakes from this validator are temporarily unavailable", validator.Address)
 			}
+			if validator.InternalSharesToTokensRate.IsNil() || validator.InternalSharesToTokensRate.IsZero() {
+				return types.Validator{}, errorsmod.Wrapf(types.ErrValidatorExchangeRateNotKnown,
```
ethan-stride, line 165: So an LSM liquid stake tx from a validator we don't yet have the rate for will fail... but should we also try to kick off an async (ie. the user is not waiting for it) ICQ to see if we can now get the exchange rate?

ethan-stride (commented):
When the chain starts up, we do a bunch of ICQ to get the initial exchange rates for each validator right, we don't just start them off at 1.0 like in the tests correct?  I think we should have a lower bar for firing off a slash ICQ check for 0 weight validators the first time we see delegation coming from them, because those 0 weight validators could in practice go much longer without an ICQ hitting them so there is more chance for them to have been slashed.

x/stakeibc/keeper/lsm.go
sampocs: yup, 0 weight vals are included in the set

sampocs: i'll add a clarifying comment

x/stakeibc/keeper/lsm.go
sampocs: If we don't have the exchange rate yet, that should indicate that there's a query in progress (it's possible if we get a liquid stake the second after a validator was added)

sampocs:
@ethan-stride 
> I think we should have a lower bar for firing off a slash ICQ check for 0 weight validators the first time we see delegation coming from them

Imo, I don't really see a reason to deviate from our normal cadence for 0-weight validators. The magnitude of impact is the same as non-zero weight validators (capped at a percent of TVL) and if we can make it less likely that there was an unregistered slash by submitting the queries manually a more regular cadence.

riley-stride:
Nit: `ValidatorExchRate` might be worth renaming. I like `XtoY` terminology, because it makes clear the directionality, so that if the variable is g.t. 1 or l.t. 1 you immediately know which of shares or tokens is worth more.

x/stakeibc/keeper/lsm.go
riley-stride: Oh do we kick off an ICQ right when a new val is added?

riley-stride: And do validators start with nil exchrate or exchrate=1?

riley-stride: If we both start them at nil exchrate and kick off an ICQ by default upon adding them, then this logic makes sense. otherwise I don't understand the case where the exchrate would be nil or zero.

sampocs: It’s set to nil by default when the val is added and then the query for the exchange rate is submitted immediately

x/stakeibc/keeper/lsm.go
```diff
@@ -166,6 +173,25 @@ func (k Keeper) GetValidatorFromLSMTokenDenom(denom string, validators []*types.
...
+//	StTokens = LSMTokenShares * Validator Exchange Rate / Redemption Rate
+//
+// Note: in the event of a slash query, these tokens will be minted only if the
+// validator's exchange rate did not change
+func (k Keeper) CalculateLSMStToken(liquidStakedShares sdkmath.Int, lsmLiquidStake types.LSMLiquidStake) sdk.Coin {
+	hostZone := lsmLiquidStake.HostZone
+	validator := lsmLiquidStake.Validator
+
+	lsmTokenTokens := sdk.NewDecFromInt(liquidStakedShares).Mul(validator.InternalSharesToTokensRate)
+	stDenom := types.StAssetDenomFromHostZoneDenom(hostZone.HostDenom)
```
riley-stride, line 188: - pico nit: move the denom line down below the calc ends to make the calc logic slightly easier to inspect
- pico nit: why `lsmTokenTokens` instead of `lsmTokens`?

x/stakeibc/keeper/hooks.go
```diff
@@ -236,11 +237,18 @@ func (k Keeper) GetUndelegatedBalance(chainId string, depositRecords []recordsty
...
 		if deposit.Status != recordstypes.LSMTokenDeposit_DEPOSIT_PENDING {
-			total = total.Add(deposit.Amount)
+			validator, _, found := GetValidatorFromAddress(hostZone.Validators, deposit.ValidatorAddress)
+			if !found {
+				k.Logger(ctx).Error(fmt.Sprintf("Validator %s found in LSMTokenDeposit but no longer exists", deposit.ValidatorAddress))
+				continue
+			}
+			liquidStakedShares := deposit.Amount
+			liquidStakedTokens := sdk.NewDecFromInt(liquidStakedShares).Mul(validator.InternalSharesToTokensRate)
+			total = total.Add(liquidStakedTokens.TruncateInt())
```
riley-stride, line 251: nit: do we get better precision if we add to the total in `Dec` space, then truncate it outside the loop after all have been added? not sure

x/stakeibc/keeper/hooks.go
```diff
@@ -236,11 +237,18 @@ func (k Keeper) GetUndelegatedBalance(chainId string, depositRecords []recordsty
 
 // Returns the total delegated balance that's stored in LSM tokens
 // These are identified by any status besides "DEPOSIT_PENDING"
```
riley-stride, line 239: Could you please add a comment here explaining why PENDING is the only status that's not counted?

x/stakeibc/keeper/lsm_test.go
```diff
@@ -202,9 +206,86 @@ func (s *KeeperTestSuite) TestGetValidatorFromLSMTokenDenom() {
...
+			validatorExchangeRate: sdk.MustNewDecFromStr("0.75"),
+			redemptionRate:        sdk.OneDec(),
+			expectedStAmount:      sdkmath.NewInt(750), // 1000 * 0.75 / 1
+		},
+		{
+			name:                  "non-one exchange rate, non-one redemption rate",
+			liquidStakedShares:    sdkmath.NewInt(1000),
+			validatorExchangeRate: sdk.MustNewDecFromStr("0.75"),
+			redemptionRate:        sdk.MustNewDecFromStr("1.25"),
+			expectedStAmount:      sdkmath.NewInt(600), // 1000 * 0.75 / 1.25 = 600
```
riley-stride, line 261: If the user stakes 1 share, the val exchrate is < 1 and the RR is >= 1, do they get 0 stTokens? Ik it's unlikely anyone would input 1e-6 on the frontend, but to sanity check, would that case throw an error?

x/stakeibc/keeper/update_redemption_rates_test.go
```diff
@@ -271,33 +271,55 @@ func (s *KeeperTestSuite) TestGetRedemptionRate_DepositRecords() {
...
+		{Address: "valC", InternalSharesToTokensRate: sdk.MustNewDecFromStr("0.5")},
+	}
+
+	// Total: 1 + 2 + 3 + 4 + 5 + 6 + 7 + 8 + 9 + 10 = 65
 	lsmDeposits := []recordtypes.LSMTokenDeposit{
-		// Total: 1 + 2 + 3 + 4 + 5 + 6 + 7 + 8 + 9 + 10 = 65
-		{ChainId: HostChainId, Status: recordtypes.LSMTokenDeposit_TRANSFER_IN_PROGRESS, Amount: sdk.NewInt(1)},
-		{ChainId: HostChainId, Status: recordtypes.LSMTokenDeposit_TRANSFER_IN_PROGRESS, Amount: sdk.NewInt(2)},
+		// ValidatorA Exchange Rate 1.0
+		{ChainId: HostChainId, Status: transferInProgress, Amount: sdk.NewInt(1), ValidatorAddress: "valA"}, // 1 * 1.0 = 1
```
riley-stride, line 289: Should these tests cover the TRANSFER_QUEUE state?

riley-stride (approved):
Looking good! My comments are nit

x/stakeibc/keeper/lsm.go
sampocs: great suggestion on first bullet. 

`lsmTokenTokens` is kind of a typo - I used to have to variables `lsmTokensShares` and `lsmTokenTokens` but the shares one was replaced with `liquidStakedShares`. I'm going to update it to `liquidStakedTokens`

x/stakeibc/keeper/hooks.go
sampocs: hmm, good question. I think we do get more precision, but it might be less "accurate" since we want to replicate the number we're using when giving out stTokens

x/stakeibc/keeper/hooks.go
sampocs: done

x/stakeibc/keeper/lsm_test.go
sampocs: good thought. It would convert to 0 stTokens but then the liquid stake would fail [here](https://github.com/Stride-Labs/stride/blob/lsm/x/stakeibc/keeper/msg_server_lsm_liquid_stake.go#L78)

x/stakeibc/keeper/update_redemption_rates_test.go
sampocs: good call, done

sampocs:
@riley-stride 
> Nit: ValidatorExchRate might be worth renaming. I like XtoY terminology, because it makes clear the directionality, so that if the variable is g.t. 1 or l.t. 1 you immediately know which of shares or tokens is worth more.

Do you mean any place in particular? It's a bit of a mixture right now.

Fwiw, I thought we should change it as well up until recently where I saw the SDK actually has a mix of referencing it as exchange rate vs TokensToShares. 

A good frame work might be: at a high level, refer to it as exchange rate as it's shorter, but still to the point, and the direction isn't necessary. Whereas for instances where it's used more specifically (i.e. in an equation), use TokensToShares

Wdty?

Ex: keep query name as `QueryValidatorExchangeRate`, but keep field on validator named as `InternalTokensToShares`

riley-stride:
> @riley-stride
> 
> > Nit: ValidatorExchRate might be worth renaming. I like XtoY terminology, because it makes clear the directionality, so that if the variable is g.t. 1 or l.t. 1 you immediately know which of shares or tokens is worth more.
> 
> Do you mean any place in particular? It's a bit of a mixture right now.
> 
> Fwiw, I thought we should change it as well up until recently where I saw the SDK actually has a mix of referencing it as exchange rate vs TokensToShares.
> 
> A good frame work might be: at a high level, refer to it as exchange rate as it's shorter, but still to the point, and the direction isn't necessary. Whereas for instances where it's used more specifically (i.e. in an equation), use TokensToShares
> 
> Wdty?
> 
> Ex: keep query name as `QueryValidatorExchangeRate`, but keep field on validator named as `InternalTokensToShares`

I don't see good reason to use different terms, the SDK's approach seems unnecessarily confusing, would advocate for us to stick to a single descriptive & clear term like InternalTokensToShares, but don't feel that strongly. On board with whichever path you decide to take here!

sampocs:
@riley-stride 
> I don't see good reason to use different terms, the SDK's approach seems unnecessarily confusing, would advocate for Stride to stick to a single descriptive & clear term like InternalTokensToShares, but don't feel that strongly. On board with whichever path you decide to take here!

I hear ya and agree for the most part! I do think it's beneficial to not be overly verbose, so I think there might be some examples where using ExchangeRate makes sense, but I'll try and convert most (if not all) of them. Also we can remove the `Internal` and just to `TokensToSharesRate` which is not that much longer than `ExchangeRate`

sampocs:
@riley-stride After looking through the examples, I agree with you more, I think the use of both of them is confusing. Going to replace all exchange rate mentions

sampocs:
I'll push it to a separate PR

PR #861, opened 2023-07-19
Cache context in hooks and end blocker 

sampocs:
## Context
The code that runs in the end blocker does not propagate an error, so if an error occurs halfway through a function, there is the potential for partial state changes from the beginning of the function to be persisted. 

The solution is to use a temporary cache context for the function, that allows any partial state changes to be thrown out in the event of an error. The function is available in utils already - this PR just utilizes it for each hook/endblocker function.

**I'd recommend reviewing the redemption sweep refactor commit's separately.**

## Questions for reviewer
* If we fail to submit a detokenization ICA, should we mark the record as failed or leave it be so that it gets retried?

## Changelog
* Added `ApplyFuncIfNoError` wrapper to the following hooks or end blocker functions:

| Hook/EndBlocker Function              | Wrapped Function |
|---------------------------------------|------------------|
| `InitiateAllHostZoneUnbondings`       | `UnbondFromHostZone` |
| `SweepAllUnbondedTokens`              | `SweepUnbondedTokensForHostZone` (previously `SweepAllUnbondedTokensForHostZone`) |
| `SetWithdrawalAddress`                | `SetWithdrawalAddressOnHost` |
| `TransferExistingDepositsToHostZones` | `IBCTransferNativeTokens` |
| `StakeExistingDepositsOnHostZones`    | `DelegateOnHost` |
| `ReinvestRewards`                     | `SubmitWithdrawalBalanceICQ` (previously `UpdateWithdrawalBalance`) |
| `RebalanceAllHostZones`               | `RebalanceDelegationsForHostZone` |
| `AllocateHostZoneReward`              | `LiquidStake` |
| `TransferAllLSMDeposits`              | `IBCTransferLSMToken` |
| `DetokenizeAllLSMDeposits`            | `DetokenizeLSMDeposit` |

* Renamed `UpdateWithdrawalBalance` to `SubmitWithdrawalBalanceICQ`

### DIFF
--- a/x/stakeibc/keeper/hooks.go
+++ b/x/stakeibc/keeper/hooks.go
@@ -251,6 +251,8 @@ func (k Keeper) GetTotalTokenizedDelegations(ctx sdk.Context, chainId string) sd
 	return sdk.NewDecFromInt(total)
 }
 
+// Reinvests staking rewards by querying the accumulated rewards from the withdrawal account,
+// and then sending them to the delegation account on the query callback, where they will be staked
 func (k Keeper) ReinvestRewards(ctx sdk.Context) {
 	k.Logger(ctx).Info("Reinvesting tokens...")
 
@@ -269,7 +271,11 @@ func (k Keeper) ReinvestRewards(ctx sdk.Context) {
 		}
 		k.Logger(ctx).Info(utils.LogWithHostZone(hostZone.ChainId, "BlockTime for host zone: %d", blockTime))
 
-		err = k.UpdateWithdrawalBalance(ctx, hostZone)
+		// Submit a withdrawal balance ICQ to determine the amount of rewards
+		// The rewards will get to the delegation account in the callback
+		err = utils.ApplyFuncIfNoError(ctx, func(ctx sdk.Context) error {
+			return k.SubmitWithdrawalBalanceICQ(ctx, hostZone)
+		})
 		if err != nil {
 			k.Logger(ctx).Error(fmt.Sprintf("Error updating withdrawal balance for host zone %s: %s", hostZone.ConnectionId, err.Error()))
 			continue
--- a/x/stakeibc/keeper/hooks.go
+++ b/x/stakeibc/keeper/hooks.go
@@ -139,9 +139,12 @@ func (k Keeper) SetWithdrawalAddress(ctx sdk.Context) {
 	k.Logger(ctx).Info("Setting Withdrawal Addresses...")
 
 	for _, hostZone := range k.GetAllActiveHostZone(ctx) {
-		err := k.SetWithdrawalAddressOnHost(ctx, hostZone)
+		err := utils.ApplyFuncIfNoError(ctx, func(ctx sdk.Context) error {
+			return k.SetWithdrawalAddressOnHost(ctx, hostZone)
+		})
 		if err != nil {
 			k.Logger(ctx).Error(fmt.Sprintf("Unable to set withdrawal address on %s, err: %s", hostZone.ChainId, err))
+			continue
 		}
 	}
 }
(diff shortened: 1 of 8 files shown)


riley-stride:
Reviewed, looks like it does not make logical changes. And the refactored code is much more readable and clear. 

But I think we should get another set of eyes from either Vishal or Aidan (maybe both), given it's a refactor of core logic.

1. For each hook/endblocker fn, how did you decide which fn to wrap?

2. Could we put the table from the PR description into a README within the codebase?

3. 
> If we fail to submit a detokenization ICA, should we mark the record as failed or leave it be so that it gets retried?
My first thought is to leave it so that it's retried. Are you thinking that if it failed once it will always fail, so it's preferable to avoid the infinite retry loop and instead mark it failed?

sampocs:
@riley-stride 
1. usually the setup for each of these was an outer function like `DoSomethingForAllHostZones` and then within that there was a loop over each host zone and then it called a function like `DoSomethingForHostZone`. I wrapped the host zone specific function in these cases with the thinking that each host zone was an atomic unit
2. More than happy to add it to a README if you think it makes sense, but imo, I'm not sure how much value that adds. I think of this as just any other line of code - doesn't feel like something we need to single out
3. Yup exactly! It's mainly just a question of whether the retry would succeed or fail / whether we might get into an infinite loop. The only case where I know for sure a retry would help is if the channel is closed, but we catch that case above. Other than that, if feels like a failure is pretty deterministic?

x/stakeibc/keeper/hooks.go
```diff
@@ -265,7 +271,11 @@ func (k Keeper) ReinvestRewards(ctx sdk.Context) {
 		}
 		k.Logger(ctx).Info(utils.LogWithHostZone(hostZone.ChainId, "BlockTime for host zone: %d", blockTime))
 
-		err = k.UpdateWithdrawalBalance(ctx, hostZone)
+		// Submit a withdrawal balance ICQ to determine the amount of rewards
+		// The rewards will get to the delegation account in the callback
+		err = utils.ApplyFuncIfNoError(ctx, func(ctx sdk.Context) error {
+			return k.SubmitWithdrawalBalanceICQ(ctx, hostZone)
```
ethan-stride, line 277: This is just a more accurate rename of the function which does the same work, but what about the "second half" of the logic here found in the callback (in keeper/icqcallbacks_withdrawl_balance.go)  Should we also wrap that in the same type of rollback cache?

ethan-stride (commented):
LGTM, this is going to be super useful going forward.  Only question is around the one which fires off an ICQ since the callback can also have error states on which we might want to roll back but right now this is not caught if the ICQ send doesn't error itself.

riley-stride:
> @riley-stride
> 
> 1. usually the setup for each of these was an outer function like `DoSomethingForAllHostZones` and then within that there was a loop over each host zone and then it called a function like `DoSomethingForHostZone`. I wrapped the host zone specific function in these cases with the thinking that each host zone was an atomic unit
> 2. More than happy to add it to a README if you think it makes sense, but imo, I'm not sure how much value that adds. I think of this as just any other line of code - doesn't feel like something we need to single out
> 3. Yup exactly! It's mainly just a question of whether the retry would succeed or fail / whether we might get into an infinite loop. The only case where I know for sure a retry would help is if the channel is closed, but we catch that case above. Other than that, if feels like a failure is pretty deterministic?

1. Makes sense
2. In light of (1) I agree it's pretty clear, so no need to put anything in a README
3. Given the determinism, probably makes sense to fail them.

x/stakeibc/keeper/hooks.go
sampocs: the callbacks are processed as tx's from the relayer so they already atomically succeed/fail - no need for cache wrapper!

sampocs:
open

sampocs:
Open

sampocs:
open

sampocs:
open

sampocs:
open

sampocs:
open

PR #865, opened 2023-07-21
[LSM] Rename validator exchange rate to sharesToTokens rate

sampocs:
Closes: #XXX

There are both occurrences of validator "SharesToTokensRate" and "ExchangeRate" in our code base, which has been flagged as being confusing. Considering SharesToTokens is more precise and descriptive, this PR removes all occurrences of exchange rate.


riley-stride (approved):
Looks good to me! Ty for doing this in a separate PR and splitting it up by commits, made for a much easier review

x/stakeibc/keeper/icqcallbacks_delegator_shares.go
```diff
@@ -167,9 +172,9 @@ func (k Keeper) CheckForSlash(
 	chainId := hostZone.ChainId
 	validator := hostZone.Validators[valIndex]
 
-	// Calculate the number of tokens delegated (using the internal exchange rate)
+	// Calculate the number of tokens delegated (using the internal sharesToTokensRate)
 	// note: truncateInt per https://github.com/cosmos/cosmos-sdk/blob/cb31043d35bad90c4daa923bb109f38fd092feda/x/staking/types/validator.go#L431
-	delegatedTokens = queriedDelegation.Shares.Mul(validator.InternalSharesToTokensRate).TruncateInt()
+	delegatedTokens = queriedDelegation.Shares.Mul(validator.SharesToTokensRate).TruncateInt()
```
ethan-stride, line 177: Was the previous name including "Internal" to signify it is the stride side tracked version of the value rather than the true value which exists on the foreign hub? Getting rid of this "Internal" won't cause confusion because everything on the validator data structure is technically the stide side belief about the foreign hub's validators?

ethan-stride (approved):
LGTM

x/stakeibc/keeper/icqcallbacks_delegator_shares.go
sampocs: tbh I'm not entirely sure why it was used. I assume the thinking was that it's an implicit rate that's determined by using the `TokensToShares` helper function on the validator, which references the total number of shares and tokens (i.e. there's no field on the validator called "ExchanageRate" or "TokensToSharesRate")

In any event, I don't think it adds anything to the variable name

PR #866, opened 2023-07-21
[LSM] LSMLiquidStake events for frontend

sampocs:
Closes: #XXX

## Context 
It doesn't seem possible to retrieve the TransactionComplete bool that's returned from an lsm liquid stake (that indicates if the slash query was submitted). This PR add the relevant information to the events for the frontend.

## Brief Changelog
* Consolidated LSMLiquidStake events into a single event type `lsm_liquid_stake`
* Added an additional attribute `TransactionStatus` with values `success`, `pending` or `failed`
* Added a transaction ID to the deposit record so that, if there's a slash query, the frontend can find the final tx response

### Successful Liquid Stake
```json
{
  "type": "lsm_liquid_stake",
  "attributes": [
    {
      "key": "module",
      "value": "stakeibc"
    },
    {
      "key": "liquid_staker",
      "value": "stride1x92tnm6pfkl3gsfy0rfaez5myq5zh99a6a2w0p"
    },
    {
      "key": "host_zone",
      "value": "GAIA"
    },
    {
      "key": "native_base_denom",
      "value": "uatom"
    },
    {
      "key": "validator",
      "value": "cosmosvaloper1uk4ze0x4nvh4fk0xm4jdud58eqn4yxhrdt795p"
    },
    {
      "key": "native_ibc_denom",
      "value": "ibc/457E522AB5A620091319195479CADC0638D42A24B49B420BB728F424E9CA60A1"
    },
    {
      "key": "lsm_token_base_denom",
      "value": "cosmosvaloper1uk4ze0x4nvh4fk0xm4jdud58eqn4yxhrdt795p/1"
    },
    {
      "key": "native_amount",
      "value": "1000000"
    },
    {
      "key": "sttoken_amount",
      "value": "1000000"
    },
    {
      "key": "lsm_liquid_stake_tx_id",
      "value": "feccd34167c2ce1f78d448258aa1a57c1fe8b4554676d5ef71cf83f5f4c2adb2"
    },
    {
      "key": "transaction_status",
      "value": "success"
    }
  ]
}
```

### Pending Liquid Stake
```json
{
  "type": "lsm_liquid_stake",
  "attributes": [
    {
      "key": "module",
      "value": "stakeibc"
    },
    {
      "key": "liquid_staker",
      "value": "stride1x92tnm6pfkl3gsfy0rfaez5myq5zh99a6a2w0p"
    },
    {
      "key": "host_zone",
      "value": "GAIA"
    },
    {
      "key": "native_base_denom",
      "value": "uatom"
    },
    {
      "key": "validator",
      "value": "cosmosvaloper1uk4ze0x4nvh4fk0xm4jdud58eqn4yxhrdt795p"
    },
    {
      "key": "native_ibc_denom",
      "value": "ibc/457E522AB5A620091319195479CADC0638D42A24B49B420BB728F424E9CA60A1"
    },
    {
      "key": "lsm_token_base_denom",
      "value": "cosmosvaloper1uk4ze0x4nvh4fk0xm4jdud58eqn4yxhrdt795p/1"
    },
    {
      "key": "native_amount",
      "value": "1000000"
    },
    {
      "key": "sttoken_amount",
      "value": "1000000"
    },
    {
      "key": "lsm_liquid_stake_tx_id",
      "value": "8b328e47ac41cb2ad11ac84e8979eb36cf4e49ba815fa33be0f1c963b3f8f81a"
    },
    {
      "key": "transaction_status",
      "value": "pending"
    }
  ]
}
```

### Failed Liquid Stake
```json
{
  "type": "lsm_liquid_stake",
  "attributes": [
    {
      "key": "module",
      "value": "stakeibc"
    },
    {
      "key": "liquid_staker",
      "value": "stride1x92tnm6pfkl3gsfy0rfaez5myq5zh99a6a2w0p"
    },
    {
      "key": "host_zone",
      "value": "GAIA"
    },
    {
      "key": "native_base_denom",
      "value": "uatom"
    },
    {
      "key": "validator",
      "value": "cosmosvaloper17kht2x2ped6qytr2kklevtvmxpw7wq9rarvcqz"
    },
    {
      "key": "native_ibc_denom",
      "value": "ibc/19825915130745DAC8CBC51A6DBE4FC7644463CF4254CD46D49B15AABEE73FB8"
    },
    {
      "key": "lsm_token_base_denom",
      "value": "cosmosvaloper17kht2x2ped6qytr2kklevtvmxpw7wq9rarvcqz/1"
    },
    {
      "key": "native_amount",
      "value": "1000000"
    },
    {
      "key": "sttoken_amount",
      "value": "998685"
    },
    {
      "key": "lsm_liquid_stake_tx_id",
      "value": "8b328e47ac41cb2ad11ac84e8979eb36cf4e49ba815fa33be0f1c963b3f8f81a"
    },
    {
      "key": "transaction_status",
      "value": "failed"
    },
    {
      "key": "error",
      "value": "validator was slashed, failing LSMLiquidStake"
    }
  ]
}
```


riley-stride:
lgtm, solid work coordinating with kier on this!

riley-stride (approved):

x/stakeibc/keeper/lsm.go
```diff
@@ -87,6 +90,15 @@ func (k Keeper) ValidateLSMLiquidStake(ctx sdk.Context, msg types.MsgLSMLiquidSt
 	}, nil
 }
 
+// Generates a unique ID for the LSM token deposit so that, if a slash query is issued,
+// the query callback can be joined back with this tx
+// The key in the store for an LSMTokenDeposit is chainId + denom (meaning, there
```
ethan-stride, line 95: I remember talking about this possibly being an issue way back in early June, this fixes it so that it is still a singleton value behind the key (instead of needing to iterate an array or something) but many can be in progress at once, even from the same staker if they are on different blocks.  So does this also fix the other small issue where a user might try it with a fraction of their IBC'ed LSM tokens before doing the rest but it would have to complete the ICAs before they could do the second part?

ethan-stride (approved):
LGTM, 
I don't have great background on how the frontend uses these types of event but everything there looked reasonable. Biggest change I see here to the functionality is the key definition for LSM deposits and then needing to keep an id for it in the data

x/stakeibc/keeper/lsm.go
sampocs: Right now there can only be one LSM liquid stake for a given denom in progress at once. So the case where a user splits up their tokens and tries to lsm-liquid-stake while they still have an earlier stake in progress will still fail. 

The change above does not modify anything related to this limitation or how we store down the LSM token deposit - it only adds an ID that can be used by the event

ethan-stride: Oh, I see now.  If we were to apply this same ID logic here to the keeper would that fix this issue by allowing for more than one at a time to be in flight?  Would it be complicated by the way we would want to call it back with the getters since we wouldn't always know the original block height?

sampocs: good question! yeah it would resolve the issue. But imo, it's more a feature than an issue cause if a user has a stuck LSM liquid stake with a denom, I don't think we want to keep allowing LSM liquid stakes with that denom

And yeah that's right, it could potentially get more complicated when we have to use the getters since we don't have the height - although, to be fair, I don't think call the getter directly that often outside of unit tests

PR #873, opened 2023-07-31
[LSM] Batch send delegation ICAs

sampocs:
## Context
Send delegation ICA messages in batches so as not to exceed the gas limit

## Changelog
* Batched ICA messages into groups of 30 before submitting
* Updated delegate callback to decrement from the deposit record instead of removing (and remove if the remaining amount is 0)
* Added unit tests


sampocs:
This is actually not needed since validators outside the designated stride set will have zero weight and will not have a delegation message

PR #883, opened 2023-08-08
updated dockernet host zone versions 

sampocs:
* Updated osmo to their latest version with a fork that disables fees
* Fixed stargaze dockerfile
* Confirmed integration tests run on every dockernet host


shellvish (approved):
Thank you for doing this!

PR #800, opened 2023-05-25
add packet-forward-middleware

DongLieu:
## Brief Changelog
- add packet-forward-middleware
- write upgrade handler for packet-forward-middleware module
- write unit test for upgrade handler

## Author's Checklist

run all tests passed
golangci-lint run --fix

PR #889, opened 2023-08-09
add pfm

asalzmann:
closes https://github.com/Stride-Labs/stride/pull/800

**Summary**
Adds PFM by following the integration guide: https://github.com/cosmos/ibc-apps/blob/main/middleware/packet-forward-middleware/docs/integration.md

Tracking a few integration issues with the PFM team:
https://github.com/cosmos/ibc-apps/issues/146

I asked about these discrepancies on the issue (waiting on a response from Reece) for 1/2
* Why is app.GetSubspace in the integration docs but not the code?
* Why is the authority not in the integration docs but in the code? 
* Should we import app.PacketForwardKeeper in app.TransferKeeper? => Not required, just pass nil


SecretSaturn:
hey @asalzmann,

Stride's UX would really profit a lot from having PFM enabled.
We at Secret network would like to bridge in a lot of stride assets directly from osmosis and PFM would ease the UX quite a lot! :)

app/app.go
```diff
@@ -478,17 +484,36 @@ func NewStrideApp(
...
+	app.PacketForwardKeeper = packetforwardkeeper.NewKeeper(
+		appCodec,
+		keys[packetforwardtypes.StoreKey],
+		// app.GetSubspace(packetforwardtypes.ModuleName), TODO: why is this in the integration docs but not in code?
+		app.TransferKeeper, // will be zero-value here, reference is set later on with SetTransferKeeper.
+		app.IBCKeeper.ChannelKeeper,
+		app.DistrKeeper,
+		app.BankKeeper,
+		app.IBCKeeper.ChannelKeeper,
+		authtypes.NewModuleAddress(govtypes.ModuleName).String(), // TODO: this is not in the integration docs, but a comment says the authority should be the gov module
```
sampocs, line 498: the authority is new to SDK 47 for gov txs. I imagine the doc might have just been out of date

app/app.go
```diff
@@ -478,17 +484,36 @@ func NewStrideApp(
 	)
 	ratelimitModule := ratelimitmodule.NewAppModule(appCodec, app.RatelimitKeeper)
 
+	// Initialize the packet forward middleware Keeper
+	// It's important to note that the PFM Keeper must be initialized before the Transfer Keeper
+	app.PacketForwardKeeper = packetforwardkeeper.NewKeeper(
+		appCodec,
+		keys[packetforwardtypes.StoreKey],
+		// app.GetSubspace(packetforwardtypes.ModuleName), TODO: why is this in the integration docs but not in code?
+		app.TransferKeeper, // will be zero-value here, reference is set later on with SetTransferKeeper.
```
sampocs, line 493: should this be nil?

app/app.go
```diff
@@ -478,17 +484,36 @@ func NewStrideApp(
 	)
 	ratelimitModule := ratelimitmodule.NewAppModule(appCodec, app.RatelimitKeeper)
 
+	// Initialize the packet forward middleware Keeper
+	// It's important to note that the PFM Keeper must be initialized before the Transfer Keeper
+	app.PacketForwardKeeper = packetforwardkeeper.NewKeeper(
+		appCodec,
+		keys[packetforwardtypes.StoreKey],
+		// app.GetSubspace(packetforwardtypes.ModuleName), TODO: why is this in the integration docs but not in code?
```
sampocs, line 492: I don't see GetSubspace in the latest integration doc

app/app.go
```diff
@@ -699,14 +724,29 @@ func NewStrideApp(
...
+	// RecvPacket, message that originates from core IBC and goes down to app, the flow is the other way
+	// channel.RecvPacket -> packetforward.OnRecvPacket -> transfer.OnRecvPacket
+
 	// Stack three contains
-	// - IBC
+	// - core ibc
 	// - autopilot
 	// - records
 	// - ratelimit
+	// - pfm
```
sampocs, line 739: what's your rationale for the ordering here (not objecting, just curious)? Also we may want to look through the error cases in pfm and confirm there's no scenario where it short circuits before getting to the rest of the stack

app/app.go
```diff
@@ -478,17 +484,36 @@ func NewStrideApp(
...
+		app.IBCKeeper.ChannelKeeper,
+		authtypes.NewModuleAddress(govtypes.ModuleName).String(), // TODO: this is not in the integration docs, but a comment says the authority should be the gov module
+	)
+
 	// Create Transfer Keepers
 	app.TransferKeeper = ibctransferkeeper.NewKeeper(
 		appCodec, keys[ibctransfertypes.StoreKey],
 		app.GetSubspace(ibctransfertypes.ModuleName),
+		// TODO: Do we need to import the app.PacketForwardKeeper here?
 		app.RatelimitKeeper, // ICS4Wrapper
```
sampocs, line 506: based on the below stack ordering, I think we'll want to set app.PacketForwardKeeper here and then set app.RatelimitKeeper as the ICS4Wrapper of the PacketForwardKeeper
```go
app.RatelimitKeeper = (
   app.ChannelKeeper,
)   

app.PacketForwardKeeper = (
    app.RatelimitKeeper,
)

app.TransferKeeper = (
   app.PacketForwardKeeper
)
```

app/upgrades.go
```diff
@@ -257,6 +257,12 @@ func (app *StrideApp) setupUpgradeHandlers(appOpts servertypes.AppOptions) {
 			Added: []string{evmosvestingtypes.ModuleName},
 		}
 	}
+	// TODO: uncomment when v17 upgrade is ready
+	// case "v17":
+	// 	storeUpgrades = &storetypes.StoreUpgrades{
+	// 		// Add PFM store key
+	// 		Added: []string{routertypes.ModuleName},
```
sampocs, line 264: thank you for adding this! I think it will be `packetforwardtypes` though

app/app.go
```diff
@@ -758,6 +798,7 @@ func NewStrideApp(
 		ibc.NewAppModule(app.IBCKeeper),
 		params.NewAppModule(app.ParamsKeeper),
 		claim.NewAppModule(appCodec, app.ClaimKeeper),
+		packetforward.NewAppModule(app.PacketForwardKeeper, nil), // What should legacySubspace be set to and why is it required? https://github.com/cosmos/ibc-apps/issues/146
```
sampocs, line 801: nit: clarify comment now the issue has seen responses

app/app.go
asalzmann: app.GetSubspace or app.TransferKeeper?

If it's `app.TransferKeeper`, I didn't change anything (but my guess is it's typed so we can't pass `nil`)

app/app.go
asalzmann: removed

app/app.go
asalzmann: removed

app/app.go
asalzmann: it feels conceptually closest to ibc transfers, which is why I put it at the bottom (so no strong reason - we can put it elsewhere)

app/upgrades.go
asalzmann: updated

app/app.go
asalzmann: done

app/app.go
asalzmann: I see there's a todo above this line "Implement ICS4Wrapper in Records and pass records keeper here" - is that relevant?

I'm not too familiar with the `ICS4Wrapper`, are you? I implemented based on your suggestion.

app/app.go
sampocs: I think it's supposed to be `nil` per the integration [docs](https://github.com/cosmos/ibc-apps/pull/147/files#diff-6264bd95a51d36427689e0470a4c3300334d62154d1e89baed20b0a00a9323bfR59)

It's set after the keeper is created [[here](https://github.com/cosmos/ibc-apps/pull/147/files#diff-6264bd95a51d36427689e0470a4c3300334d62154d1e89baed20b0a00a9323bfR80)]

app/app.go
sampocs: The wrapper is how you define traffic _up_ the stack. 

Traffic _down_ the stack is defined [[here](https://github.com/Stride-Labs/stride/blob/main/app/app.go#L703-L713)] based on the ordering with which each middleware is wrapped. But traffic _up_ the stack travels through these ICS4Wrappers. 

Right now, our upwards stack is technically just `Transfer` [[→](https://github.com/Stride-Labs/stride/blob/main/app/app.go#L485)] `RateLimit` [[→](https://github.com/Stride-Labs/stride/blob/main/app/app.go#L477)] `Core IBC (Channel)` (no records or autopilot)

---

As to the TODO you're referencing, that was originally added because it technically felt more more correct to have the up and down stack be mirror images of each other. But on second though, considering records and Autopilot doesn't need to override anything during the ICS4 calls (SendPacket, WriteAcknowledgement), I don't think it's necessary to include them in the stack. 

Curious to hear your thoughts though!

---

All that said, at the very least, we need to make sure to include PFM in this upward stack (per original comment)

asalzmann: Ah, gotcha. Thanks for explaining that. 

I agree, they should mirror each other but don't need to since Autopilot doesn't need to override anything during the ICS4 calls.

> All that said, at the very least, we need to make sure to include PFM in this upward stack (per original comment)

Is [this line](https://github.com/Stride-Labs/stride/pull/889/files#diff-0f1d2976054440336a576d47a44a37b80cdf6701dd9113012bce0e3c425819b7R503) sufficient?

app/app.go
asalzmann: Updated

app/app.go
sampocs: Yup! that and [this](https://github.com/Stride-Labs/stride/pull/889/files#diff-0f1d2976054440336a576d47a44a37b80cdf6701dd9113012bce0e3c425819b7R495) are all that's needed

asalzmann:
@sampocs are you signed off on this PR?

Is the main outstanding task to review interaction between autopilot / PFM?

sampocs:
I’m signed off on all the comments, but remaining tasks are:
* think about the interaction between autopilot snd pfm
* test it in dockernet if you havent already
* fix whatever the build error is that’s causing CI to fail

asalzmann:
[ ] think about the interaction between autopilot snd pfm
[ ] test it in dockernet if you havent already
[ ] fix whatever the build error is that’s causing CI to fail

sampocs (approved):

PR #899, opened 2023-08-22
Added oracle transfer channel to icaoracle MsgInstantiateOracle and register aminos

sampocs:
## Context and Purpose
As it stands, the stToken denom passed to the ICA Oracle Contract is as it appears on Stride (e.g. stuosmo). However, since the contract lives on Osmosis, it makes more sense to store the denom as it appears on Osmosis (e.g. ibc/{hash(transfer/channel-X/stuosmo)}. This is accomplished by including the Stride <> Osmosis channel ID when instantiating the contract. The contract will then use the channel ID to calculate the denom hash before writing it to the store (see [oracle-PR-12](https://github.com/Stride-Labs/ica-oracle/pull/12)).

This PR also adds the `Type` function to each oracle transaction to fix ledger transactions.

## Brief Changelog
* Added a channel ID field to both the stride `InstantiateOracle` tx, and to the contract instantiation message
* Updated the `InstantiateOracle` msg to pass the field to the contract 
* Renamed the `RedemptionRate` attributes fields to have `sttoken_denom` instead of `denom`
* Fixed icaoracle module codec and added Type functions for ledger support


shellvish (approved):
Looks great, thank you for doing this!

PR #900, opened 2023-08-22
Remove max slash safety check

sampocs:
## Context and Purpose
In the slash query callback, there's a check that makes sure the slash is not greater than 10%. However, if there was a large undelegation after the slash (and before the slash was registered), it was appear as if the slash percentage was larger than it was.

Ex:
|                                        | Delegation in Record Keeping | True Delegaation |
|----------------------------------------|:----------------------------:|:----------------:|
| Initial Delegation                     |           1,000,000          |     1,000,000    |
| 5% Slash                               |           1,000,000          |      950,000     |
| Undelegation or Rebalancing of 600,000 |            400,000           |      350,000     |

Then when the slash is registered, it appears to be a slash of `(400,000 - 350,000) / 400,000 = 12.5%` 

## Testing
Follow the steps below first on `lsm` to confirm the error can be recreated. Then try the same process on this branch to confirm it's been resolved. 

1. Start dockernet and run the setup script
```bash
git submodule sync --recursive
make start-docker build=sgr
bash dockernet/scripts/lsm/setup.sh
```
2. LSM Liquid stake twice with 30 seconds in between
```bash
build/strided --home dockernet/state/stride1 tx stakeibc lsm-liquid-stake 1500000 ibc/19825915130745DAC8CBC51A6DBE4FC7644463CF4254CD46D49B15AABEE73FB8 --from staker --gas auto -y

sleep 30

build/strided --home dockernet/state/stride1 tx stakeibc lsm-liquid-stake 1500000 ibc/19825915130745DAC8CBC51A6DBE4FC7644463CF4254CD46D49B15AABEE73FB8 --from staker --gas auto -y
```
3. Slash the validator
```
docker pause dockernet-gaia2-1   // it might be  dockernet_gaia2_1 on linux

sleep 30
 
docker unpause dockernet-gaia2-1   // it might be  dockernet_gaia2_1 on linux
```
4. Wait for rebalancing to occur by checking the stride logs (you can search for the word "Rebalancing", cap sensitive, and wait for it to appear for the second time
5. Submit another LSM liquid stake
```bash
build/strided --home dockernet/state/stride1 tx stakeibc lsm-liquid-stake 1000000 ibc/19825915130745DAC8CBC51A6DBE4FC7644463CF4254CD46D49B15AABEE73FB8 --from staker --gas auto -y
```
6. Check `state.log`. If running on the `lsm` branch, one valdiator should be stuck with `slash_query_in_progress` set to true. When running on this branch, that field should be false for all validators.

## Changelog
* Removed `SafetyMaxSlash` parameter
* Removed check in slash query callback


x/stakeibc/keeper/icqcallbacks_delegator_shares_test.go
```diff
@@ -348,19 +348,6 @@ func (s *KeeperTestSuite) TestDelegatorSharesCallback_WeightOverfow() {
 	s.Require().Regexp(expectedErrMsg, err.Error())
 }
 
-func (s *KeeperTestSuite) TestDelegatorSharesCallback_SlashGtTenPercent() {
```
riley-stride, line 351: Should we leave this and just expect success or is that redundant? Maybe we could add a GT 10% example to the success case. Will leave to you.

riley-stride (approved):

x/stakeibc/keeper/icqcallbacks_delegator_shares_test.go
sampocs: I think we should gut it. Doesn't make sense imo to have this anymore since we don't have specific branching based on the slash magnitude!

ethan-stride (approved):
LGTM went through the tests by hand on LSM branch where it gets stuck on the max slash threshold and on this PR where it no longer gets stuck.

PR #910, opened 2023-08-25
LSM Migration for v14

sampocs:
* Migrated stakeibc host zone and validator types
* Migrated stakeibc params
* Cleared out pending queries
* Enabled LSM gaia


x/stakeibc/migrations/v3/convert.go
```diff
@@ -0,0 +1,86 @@
...
+	newstakeibctypes "github.com/Stride-Labs/stride/v13/x/stakeibc/types"
+)
+
+var (
+	// The threshold, denominated in percentage of TVL, of when a slash query should
+	// be submitted (1 => 1%)
+	ValidatorSlashQueryThreshold uint64 = 1
+	// The exchange rate here does not matter since it will be updated after the slash query
+	// Setting it to this value makes it easier to verify that we've submitted the query
+	DefaultExchangeRate = sdk.MustNewDecFromStr("1.000000000000001")
```
riley-stride, line 17: Won't this rate be used in the RR if we hit an epoch before a first slash query to this validator?

riley-stride: (Separate) nit: I think exchRate g.t. 1 is impossible so might make sense to make the default (even if unused) something like 0.999 instead.

x/stakeibc/migrations/v3/convert.go
```diff
@@ -0,0 +1,86 @@
...
+
+// Converts an old validator data type to the new schema
+// Changes are as follows:
+//   - Added SlashQueryProgressTracker field
+//   - Added SlashQueryCheckpoint field
+//   - Added DelegationsInProgress field
+//   - Added SlashQueryInProgress field
+//   - InternalExchangeRate is now a decimal named SharesToTokensRate
+//   - DelegationAmt renamed to Delegation
+func convertToNewValidator(oldValidator oldstakeibctypes.Validator, totalDelegations sdkmath.Int) newstakeibctypes.Validator {
```
riley-stride, line 28: Looks good once the questions re `DefaultExchangeRate` are addressed.
Confirming that you haven't missed any proto fields on `Validator`.

x/stakeibc/migrations/v3/convert.go
```diff
@@ -0,0 +1,86 @@
...
+	queryThreshold := sdk.NewDecWithPrec(int64(ValidatorSlashQueryThreshold), 2) // percentage
+	slashQueryCheckpoint := queryThreshold.Mul(sdk.NewDecFromInt(totalDelegations)).TruncateInt()
+
+	sharesToTokensRate := DefaultExchangeRate
+	if oldValidator.InternalExchangeRate != nil && !oldValidator.InternalExchangeRate.InternalTokensToSharesRate.IsNil() {
+		sharesToTokensRate = oldValidator.InternalExchangeRate.InternalTokensToSharesRate
```
riley-stride, line 34: The names `sharesToTokensRate` and `InternalTokensToSharesRate` imply that these values would be 1/x inverses. Is that the case? If so let's invert. If not, can we add a comment here to clarify?

x/stakeibc/migrations/v3/convert.go
```diff
@@ -0,0 +1,86 @@
...
+func convertToNewHostZone(oldHostZone oldstakeibctypes.HostZone) newstakeibctypes.HostZone {
+	var validators []*newstakeibctypes.Validator
+	for _, oldValidator := range oldHostZone.Validators {
+		newValidator := convertToNewValidator(*oldValidator, oldHostZone.StakedBal)
+		validators = append(validators, &newValidator)
+	}
+
+	return newstakeibctypes.HostZone{
+		ChainId:               oldHostZone.ChainId,
+		ConnectionId:          oldHostZone.ConnectionId,
```
riley-stride, line 66: pico nit: reorder bech32prefix and connectionid to align with the .proto file ordering

x/stakeibc/migrations/v3/convert.go
```diff
@@ -0,0 +1,86 @@
...
+		Bech32Prefix:          oldHostZone.Bech32Prefix,
+		TransferChannelId:     oldHostZone.TransferChannelId,
+		IbcDenom:              oldHostZone.IbcDenom,
+		HostDenom:             oldHostZone.HostDenom,
+		UnbondingPeriod:       (oldHostZone.UnbondingFrequency - 1) * 7,
```
riley-stride, line 71: This arithmetic seems right, but given how sensitive it is and the fact that we've debated the formula in the past (though I think we're all aligned on it now) let's make sure we are all in sync on this before merging.

riley-stride:
Quick review on the convert functions. Added some comments. Going to review the rest of the PR now.

x/stakeibc/migrations/v3/convert.go
sampocs: It's only used in the redemption rate when summing up LSMTokenDeposits, but we should run the query on all validators before LSM is even live on the hub

sampocs: also, agreed, 0.9999 probably makes more sense

x/stakeibc/migrations/v3/convert.go
```diff
@@ -0,0 +1,86 @@
...
+	}
+}
+
+// Converts an old host zone data type to the new schema
+// Changes are as follows:
+//   - ICA Accounts are now strings
+//   - Address has been renamed to DepositAddress
+//   - UnbondingFrequency has been changed to UnbondingPeriod
+//   - StakedBal has been renamed to TotalDelegations
+//   - Removed blacklisted validators
```
riley-stride, line 56: Where is this implemented below?

x/stakeibc/migrations/v3/convert.go
riley-stride: Cool this makes sense

x/stakeibc/migrations/v3/convert_test.go
```diff
@@ -0,0 +1,188 @@
...
+	unbondingPeriod := uint64(21)
+
+	halted := true
+
+	valAddress := "val"
+	valDelegation := sdk.NewInt(5)
+	valWeight := uint64(6)
+	totalDelegations := sdk.NewInt(1_000_000)
+	slashCheckpoint := sdk.NewInt(10_000) // 1% of TVL
+	sharesToTokensRate := sdk.MustNewDecFromStr("0.99")
```
riley-stride, line 86: Might make sense to add expected test values for a `InternalTokensToSharesRate`& `sharesToTokensRate` conversion.

x/stakeibc/migrations/v3/convert_test.go
```diff
@@ -0,0 +1,188 @@
...
+	feeAddress := "fee"
+	delegationAddress := "delegation"
+	redemptionAddress := "redemption"
+
+	redemptionRate := sdk.NewDec(1)
+	lastRedemptionRate := sdk.NewDec(2)
+	minRedemptionRate := sdk.MustNewDecFromStr("0.95")
+	maxRedemptionRate := sdk.MustNewDecFromStr("1.25")
+	unbondingFrequency := uint64(4)
+	unbondingPeriod := uint64(21)
```
riley-stride, line 77: I recall seeing unit tests for the `UP=(UF-1 * 7)` conversion formula. If we have those somewhere prob not necessary to add them here. Otherwise might be good to add a few unit tests around that conversion here.

app/upgrades/v14/upgrades.go
```diff
@@ -252,6 +285,39 @@ func MigrateAccount1(ctx sdk.Context, evk evmosvestingkeeper.Keeper, sk stakingk
...
+}
+
+// Enable LSM liquid stakes for Gaia
+func EnableLSMForGaia(ctx sdk.Context, k stakeibckeeper.Keeper) error {
+	hostZone, found := k.GetHostZone(ctx, GaiaChainId)
+	if !found {
+		return stakeibctypes.ErrHostZoneNotFound.Wrapf(GaiaChainId)
+	}
+
+	hostZone.LsmLiquidStakeEnabled = true
```
riley-stride, line 316: LSM will be turned on on Stride before it's live on the Hub. Thinking this through: in theory Stride will be willing to accept LSM IBC vouchers for LSMLiquidStake as soon as this upgrade goes live. But no such vouchers will exist, so LSMLiquidStakes will all fail until Hub has LSM enabled. Is that right?

x/stakeibc/keeper/host_zone.go
```diff
@@ -207,8 +207,10 @@ func (k Keeper) DecrementValidatorDelegationChangesInProgress(hostZone *types.Ho
 		return errorsmod.Wrapf(types.ErrValidatorNotFound, "validator %s not found", validatorAddress)
 	}
 	if validator.DelegationChangesInProgress == 0 {
-		return errorsmod.Wrapf(types.ErrInvalidValidatorDelegationUpdates,
-			"cannot decrement the number of delegation updates if the validator has 0 updates in progress")
+		// TODO: Revert after upgrade v14
+		// return errorsmod.Wrapf(types.ErrInvalidValidatorDelegationUpdates,
+		// 	"cannot decrement the number of delegation updates if the validator has 0 updates in progress")
+		return nil
```
riley-stride, line 213: As I understand it this is the 0 floor. In theory this line should not be reachable, so imo we should emit an event here and log before returning nil, so that we can trace any issues. Then in v14 we revert.

x/stakeibc/migrations/v3/convert.go
shellvish: Verified with all the existing host zones that this formula lines up

x/stakeibc/migrations/v3/convert.go
sampocs: I think the old name was inaccurate cause it's value is always < 1. I'll add a comment to clarify

x/stakeibc/migrations/v3/convert.go
sampocs: done

x/stakeibc/migrations/v3/convert.go
sampocs: the blacklisted validators? It happens implicitly since the field is not on the new struct

x/stakeibc/migrations/v3/convert_test.go
sampocs: commented above - the original name was misleading, there's no conversion

x/stakeibc/migrations/v3/convert_test.go
sampocs: we've got a suite of tests on the reverse [here](https://github.com/Stride-Labs/stride/blob/main/x/stakeibc/types/host_zone_test.go). Imo, I think this test case is sufficient since all our mainnet unbonding frequencies are either 3, 4, or 5

app/upgrades/v14/upgrades.go
sampocs: exactly!

riley-stride (approved):

x/stakeibc/keeper/host_zone.go
sampocs: good call, added a log

PR #913, opened 2023-08-28
Bump version v13 to v14

asalzmann:
**Summary**
Bump version from v13 to v14

Also, add a script that safe deletes *pb.go files that are *not* part of previous migrations. This lets anyone easily remove and regenerate *pb.go files to check work. 

**Test plan**
Existing unittests


shellvish (approved):
Thanks for doing this!

PR #929, opened 2023-09-07
community pool ibc transfer and ICA callback triggers

ethan-stride:

Implements the logic needed to sweep in specific denom tokens from the communityPoolDepositAddress ICA, liquid stake or redeem them if they are the right denom type(s), sweep out tokens back to the communityPoolReturnAddress ICA, and call fund community poo from the return addressl on an epochly cadence.

Higher order functions to do this logic across hostZones or do error checking for a specific hostZone's community pool are in community_pool.go    
Lower level helper functions to build the needed ibc transfer commands, autopilot commands, etc. are in transfer.go or in their own callback files.

## Brief Changelog

  - Implemented IBCTransferCommunityPoolICATokensToStride with high level error checking
  - Created helper IBCTransferCommunityPoolTokens in transfer.go which builds and sends the transfer msg from the foreign zone triggered from ICA msgs
  -- Notice which transfer channels are used and which denoms (ibc denoms on the foreign chain) are being checked
  - Added autopilot command to atomically liquid stake or redeem stake the ibc transferred funds based on which denom
 - Implemented CommunityPoolTransferCallback, registered in icacallbacks, added needed callback arg data types
 - Created helper IBCReturnAllCommunityPoolTokens in transfer.go to send everything in holding address to return ICA 
 - Created helpers to epochly trigger sweep in from deposit ICA and sweep out to return ICA for all hostzones with pools
 - Added types, handlers, rpc end points for new tx where user can trigger immediate partial sweep and liquid stake
 - Implemented ICQCommunityPoolDepositICABalance to check for hostDenom and the ibc denom of stHostDenom
 - Implemented the icqcallback_pool_deposit_balance, depending on the denom will trigger token transfer with autopilot memo to stake, redeem, or do nothing.
 - Implemented the ICA msg call to fund-community-pool with the new denoms on hostZone after the transfer has finished

## Author's Checklist
- There are no known errors, but I haven't actually run it yet against any foreign hub until GAIA has a community pool added
- Currently, the pooltypes are missing, installing them breaks the sdk version. Except for this, everything builds without error. 
- Still working to address several of the comments, cleaning up how errors propagate and are logged
- This should have all the logic end to end for the community pool flow.  Working on unit tests for all the callback triggers

## Open Questions / Uncertain Aspects

It is super important to be very clear about which direction messages are travelling and which denom is being used.  In transfer.go there is a helper function GetIbcDenomOnHostZone(strideDenom, hostZone) which computes the denom on a foreign host zone after a token has been transferred from stride.  In review please make sure that this is correct since many things like the call to fund-community-pool or knowing when the ibc denom is really stDenom and can be unstaked by autopilot depend on this logic.

The MsgFundCommunityPool ICA call relies on the sdk module at x/protocolpool but it seems that the best version I can find is very very new: https://pkg.go.dev/cosmossdk.io/x/protocolpool?tab=versions and it needs sdk v51+ which we haven't upgraded to yet. Is there a better version to import?

One open question is should we remove the tx which a user can call to trigger the sweep in for a specific denom by hand?  Sam pointed out that it creates a race condition, if the epochly ICQ fires, the tx happens, then the ICQ callback could try to operate on tokens which are no longer there.  This tx was included for 2 reasons originally: so that we could test by hand on small amounts, and so that if any other denom accidentally got put in the deposit ICA that it wouldn't be stuck there forever.  Leaning towards only expected DYDX or stDYDX to ever show up in the deposit ICA and just getting rid of this tx so all logic is epochly triggered.


sampocs:
open

x/stakeibc/keeper/transfer.go
```diff
@@ -0,0 +1,102 @@
...
+func (k Keeper) IBCReturnAllCommunityPoolTokens(ctx sdk.Context, communityPoolHostZone types.HostZone) error {
+	// Use bankKeeper to see all coin types and amounts currently in the stride-side holding address
+	req := banktypes.NewQueryAllBalancesRequest(sdk.AccAddress(communityPoolHostZone.CommunityPoolHoldingAddress), nil)
+	resp, err := k.bankKeeper.AllBalances(ctx, req)
+	if err != nil {
+		return err
+	}
+
+	memo := ""
+	errors := make([]error, 0)
```
sampocs, line 67: rather than using an errors array, I would recommend making a sub-function for the transfer of a given token and then this outer function is something like:

```go
for _, coin := range resp.Balances {
   if err := k.Subfunction(...); err != nil {
      k.Logger(ctx).Error(err.Wrapf("error submitting ibc transfer for %v %s", coin.Amount, coin.Denom).Error())
      continue
  }
}
```

And then the only error returned to this outer function is from [this](https://github.com/Stride-Labs/stride/pull/929/files#diff-dbc8d2b8a50b35a2b45a2483916e27d4256f669d24fa8a0e69ec4378dd0f1771R63). That way, one bad denom doesn't cause this whole function to error

x/stakeibc/keeper/transfer.go
```diff
@@ -0,0 +1,102 @@
...
+)
+
+// Transfers tokens from the community pool deposit ICA account to the host zone holding address for that pool
+func (k Keeper) IBCTransferCommunityPoolTokens(ctx sdk.Context, token sdk.Coin, communityPoolHostZone types.HostZone, shouldLiquidStake bool) error {
+
+	// The memo may contain autopilot commands to atomically liquid stake tokens when transfer succeeds
+	//  both transfer+liquid stake will succeed and stTokens will end in the stride side holding address, 
+	//  or neither will and the original base tokens will return to the foreign deposit ICA address
+	memoCommands := ""
+	autopilotCmd := "{ \"autopilot\": { \"receiver\": \"%s\",  \"stakeibc\": { \"action\": \"LiquidStake\" } } }"
```
sampocs, line 21: It might make sense to import this from autopilot. It's not quite as clean, but it does protect against the case where we change the schema

```go
import autopilottypes "github.com/Stride-Labs/stride/v14/x/autopilot/types"

autopilotMetadata := autopilottypes.RawPacketMetadata{
    Autopilot: {
       Receiver: communityPoolHostZone.CommunityPoolHoldingAddress,
       Stakeibc: {
         Action: "LiquidStake"
       }
    }
}

autopilotCmdBz, err := json.Marshal(autopilotMetadata)
autopilotCmd := string(autopilotCmdBz)
```

You might need to pull out the AutopilotMetadata struct since I believe it's private.

x/stakeibc/keeper/transfer.go
```diff
@@ -0,0 +1,102 @@
...
+	if shouldLiquidStake {
```
sampocs, line 22: I think we can remove this parameter for the time being since we're only allowing LS's from the host denom

x/stakeibc/keeper/community_pool.go
```diff
@@ -1,12 +1,104 @@
...
 )
 
+// For each hostZone with a valid community pool, trigger the ICQs and ICAs to transfer deposited tokens to holding address
+// Since ICQs and ICAs take time to complete, it is almost certain tokens swept in and processed will be swept out in a later epoch
+func (k Keeper) SweepInAllDepositedCommunityPoolTokens(ctx sdk.Context) error {
+	hostZones := k.GetAllActiveHostZone(ctx)
+	for _, hostZone := range hostZones {
+		if hostZone.CommunityPoolDepositIcaAddress != "" &&
+			hostZone.CommunityPoolHoldingAddress != "" &&
+			hostZone.CommunityPoolReturnIcaAddress != "" {
```
sampocs, line 20: nit: We should add a helper function on the host zone type to validate these

```go
// types/host_zone.go

func (h HostZone) ValidateCommunityPoolAddresses() error {
   if h.CommunityPoolDepositIcaAddress == "" {
          return types.ErrICANotRegistered.Wrapf("community pool deposit ICA not registered")
   }
   ...
}
```

(out of scope point - we should also add this for our other ICA addresses at some point!)

x/stakeibc/client/cli/tx_community_pool_liquid_stake.go
```diff
@@ -0,0 +1,50 @@
...
+
+func CmdCommunityPoolLiquidStake() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "community-pool-liquid-stake [community-pool-host-zone] [amount] [token-denom]",
+		Short: "IBC transfer and liquid stake tokens from community pool host zone",
+		Args:  cobra.ExactArgs(3),
+		RunE: func(cmd *cobra.Command, args []string) (err error) {
+			communityPoolHostZoneId := args[0]
+			amount, found := sdk.NewIntFromString(args[1])
+			if !found {
```
sampocs, line 24: nit
```suggestion
			amount, ok := sdk.NewIntFromString(args[1])
			if !ok {
```

x/stakeibc/keeper/community_pool.go
```diff
@@ -1,12 +1,104 @@
...
-	k.Logger(ctx).Info("Transfering tokens from community pool to Stride...")
+// then as an atomic action, liquid stake the tokens with Autopilot commands in the ibc message memos
+func (k Keeper) IBCTransferCommunityPoolICATokensToStride(ctx sdk.Context, communityPoolHostZoneId string, token sdk.Coin) error {
+	k.Logger(ctx).Info(fmt.Sprintf("Transfering %d %s tokens from community pool deposit ICA to Stride hub holding address", token.Amount.Int64(), token.Denom))
+
+	// TODO: add safety check here on if the amount is greater than a threshold to avoid many, frequent, small transfers
+	//       threshold might be a config we tune to avoid ddos type attacks, for now using 0 as hard coded threshold
+	if token.Amount.LTE(sdkmath.ZeroInt()) {
+		k.Logger(ctx).Error(fmt.Sprintf("[IBCTransferCommunityPoolICATokensToStride] The amount %v to transfer did not meet the minimum threshold!", token.Amount))
+		return errors.New("Transfer Amount below threshold!")
```
sampocs, line 58: let's add a ticket to go through at the end and make new error types! also, as I mentioned on the call, feel free to remove these error logs and just log them from the upstream caller

proto/stride/stakeibc/tx.proto
```diff
@@ -50,6 +52,17 @@ message MsgLSMLiquidStake {
 }
 message MsgLSMLiquidStakeResponse { bool transaction_complete = 1; }
 
+message MsgCommunityPoolLiquidStake {
```
sampocs, line 55: we discussed this on slack - I think you can go ahead and remove this tx cause we can run the test without it!

x/stakeibc/keeper/community_pool.go
```diff
@@ -1,12 +1,104 @@
...
+
+	sdkmath "cosmossdk.io/math"
 	sdk "github.com/cosmos/cosmos-sdk/types"
+
+	"github.com/Stride-Labs/stride/v14/utils"
 )
 
+// For each hostZone with a valid community pool, trigger the ICQs and ICAs to transfer deposited tokens to holding address
+// Since ICQs and ICAs take time to complete, it is almost certain tokens swept in and processed will be swept out in a later epoch
+func (k Keeper) SweepInAllDepositedCommunityPoolTokens(ctx sdk.Context) error {
```
sampocs, line 15: Our hooks generally do not return an error cause they're run in the begin/end block, so you can remove the error return type here. And then make sure to still catch the errors from `ICQCommunityPoolDepositICABalances` and `IBCReturnAllCommunityPoolTokens`, but you can just log them

x/stakeibc/keeper/community_pool.go
```diff
@@ -1,12 +1,104 @@
...
+
 // ibc transfers tokens from the foreign hub community pool deposit ICA address onto Stride hub
-func (k Keeper) IBCTransferCommunityPoolICATokensToStride(ctx sdk.Context, token sdk.Coin) error {
-	k.Logger(ctx).Info("Transfering tokens from community pool to Stride...")
+// then as an atomic action, liquid stake the tokens with Autopilot commands in the ibc message memos
+func (k Keeper) IBCTransferCommunityPoolICATokensToStride(ctx sdk.Context, communityPoolHostZoneId string, token sdk.Coin) error {
+	k.Logger(ctx).Info(fmt.Sprintf("Transfering %d %s tokens from community pool deposit ICA to Stride hub holding address", token.Amount.Int64(), token.Denom))
+
+	// TODO: add safety check here on if the amount is greater than a threshold to avoid many, frequent, small transfers
+	//       threshold might be a config we tune to avoid ddos type attacks, for now using 0 as hard coded threshold
```
sampocs, line 55: very good call out here, we should add this for all liquid stakes

x/stakeibc/keeper/community_pool.go
```diff
@@ -1,12 +1,104 @@
...
+		return err
+	}
+
+	k.Logger(ctx).Info(utils.LogWithHostZone(hostZone.ChainId, "[IBCTransferCommunityPoolICATokensToStride] Transfer community pool tokens to Stride successfully initiated!"))
 
 	return nil
 }
+
+// Check if incoming denom is the base denom of a known hostZone for which Stride is able to liquid stake
+func (k Keeper) DenomCanLiquidStake(ctx sdk.Context, denom string) bool {
```
sampocs, line 96: per discussion this morning, I think you can remove this

x/stakeibc/keeper/transfer.go
```diff
@@ -0,0 +1,102 @@
...
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	banktypes "github.com/cosmos/cosmos-sdk/x/bank/types"
+	ibctypes "github.com/cosmos/ibc-go/v7/modules/apps/transfer/types"
+	clienttypes "github.com/cosmos/ibc-go/v7/modules/core/02-client/types"
+
+	"github.com/Stride-Labs/stride/v14/x/stakeibc/types"
+)
+
+// Transfers tokens from the community pool deposit ICA account to the host zone holding address for that pool
+func (k Keeper) IBCTransferCommunityPoolTokens(ctx sdk.Context, token sdk.Coin, communityPoolHostZone types.HostZone, shouldLiquidStake bool) error {
```
sampocs, line 15: nit: I respect the descriptive variable names 🤝 , but for these functions I think it might make sense to change `communityPoolHostZone` → `hostZone`, and `communityPoolHostZoneId` to `hostZoneId` since there's only one host zone in question in the function and it makes the code slightly more concise

x/stakeibc/keeper/community_pool.go
```diff
@@ -1,12 +1,104 @@
...
+
+	// TODO: add safety check here on if the amount is greater than a threshold to avoid many, frequent, small transfers
+	//       threshold might be a config we tune to avoid ddos type attacks, for now using 0 as hard coded threshold
+	if token.Amount.LTE(sdkmath.ZeroInt()) {
+		k.Logger(ctx).Error(fmt.Sprintf("[IBCTransferCommunityPoolICATokensToStride] The amount %v to transfer did not meet the minimum threshold!", token.Amount))
+		return errors.New("Transfer Amount below threshold!")
+	}
+
+	hostZone, hostZoneFound := k.GetHostZone(ctx, communityPoolHostZoneId)
+	if !hostZoneFound || hostZone.Halted {
```
sampocs, line 62: impressive catch making sure to check the hz isn't halted here!

(side note - out of scope of this PR) I wonder if we should add a function `GetActiveHostZone` like
```go
func (k Keeper) GetActiveHostZone(ctx sdk.Context, hostZoneId string) (types.HostZone, error) {
    hostZone, found := k.GetHostZone(ctx, hostZoneId) 
    if !found {
       return types.ErrHostNotFound.Wrap(hostZoneId)
    }
    if host.Halted {
       return types.ErrHostHalted.Wrap(hostZoneId)
    }
    return nil, hostZone
}
```

That way, we don't have to (a) duplicate the hostNotFound error that we have in every function and (b) worry about forgetting to check if a host is halted

x/stakeibc/keeper/community_pool.go
```diff
@@ -1,12 +1,104 @@
 package keeper
 
 import (
+	"errors"
```
sampocs, line 4: I found some of these function names a bit confusing at first glance - I'm wondering if we can improve some of them (although, I'm struggling to find good alternatives). 

I think the main source of confusion was the ICA in the name and the similarity with the transfer function. So maybe something like:
```
ICQCommunityPoolDepositICABalances → ICQCommunityPoolDepositBalance
IBCTransferCommunityPoolICATokensToStride → TransferAndStakeCommunityPoolTokens
```
wdyt?

sampocs (commented):
Beautiful work here Ethan!! 

Considering this phase 1 PR was more manageable than I was expecting, I went ahead and added nit comments as well. I'll review the transfer / ICQ code again once the changes are up.

x/stakeibc/keeper/abci.go
```diff
@@ -44,4 +44,7 @@ func (k Keeper) EndBlocker(ctx sdk.Context) {
 	// Submit an IBC transfer or detokenization ICA for all queued LSM Deposits across each host
 	k.TransferAllLSMDeposits(ctx)
 	k.DetokenizeAllLSMDeposits(ctx)
+	// Transfers in and out of tokens for hostZones which have community pools
+	k.SweepInAllDepositedCommunityPoolTokens(ctx)
+	k.SweepOutAllReturningCommunityPoolTokens(ctx)
```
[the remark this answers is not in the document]
ethan-stride: Yeah correct, we only really want these once per stride epoch, I'll move them

x/stakeibc/keeper/transfer.go
```diff
@@ -0,0 +1,102 @@
...
+	req := banktypes.NewQueryAllBalancesRequest(sdk.AccAddress(communityPoolHostZone.CommunityPoolHoldingAddress), nil)
+	resp, err := k.bankKeeper.AllBalances(ctx, req)
+	if err != nil {
+		return err
+	}
+
+	memo := ""
+	errors := make([]error, 0)
+
+	for _, foundCoin := range resp.Balances {
```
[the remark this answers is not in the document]
ethan-stride: For this particular transfer which is from stride "holding" module account back to the return ICA, we don't need any token type check because all tokens which are found in the holding account should go back to the return ICA, and eventually get fund-community-pool called on them regardless of their denom.

Tokens end up in the holding address after a transfer and possible autopilot action (stake, redeem, or no action).  So there could be the native denom, the stNativeDenom, or other weird 3rd party denoms (or all of these possibilities at once) in the holding address and it doesn't matter because they all are going back to the return ICA and then the pool each stride epoch.

x/stakeibc/keeper/transfer.go
ethan-stride: Yep good point.  This subfunction wrap suggestion is even more pertinent now because there is more complicated logic around registering a callback and saving args.  I was only doing the errors array for the purpose that a single failure wouldn't stop the rest of the attempted sends but that should still be possible with your method as well.  I am still going through more generally and fixing the errors to use errormod and only log at the top level instead of all through the call stack

x/stakeibc/keeper/transfer.go
ethan-stride: Good point, this is actually how I was trying to do the PFM meta data originally but I couldn't get it to import the right library so I gave up and just did the json route.  But since autopilot is a stride module we don't need to import from an external hosted location I will take your advice here and use the hard types instead of json

x/stakeibc/keeper/transfer.go
ethan-stride: So I have updated this in the recent push -- when staking we have the hostDenom, when redeeming we have the stHostDenom (or rather the ibc denom for it on the hostZone), but for ones triggered by the tx route, they can have any denom.  Thus the autopilot actions are more like (LiquidStake, RedeemStake, NoAction) in the current implementation

x/stakeibc/keeper/community_pool.go
ethan-stride: Yeah good point.  I am still going through and doing the errorsmod passing instead of logging at every level but it will take a bit.  Maybe I should do a small refactor with only fixing the error passing to log at the top level.

proto/stride/stakeibc/tx.proto
ethan-stride: So there are two purposes.  One is to run a test, the second is if any other denom besides DYDX or stDYDX makes it into the deposit ICA, this tx would allow them to eventually return those funds to the community pool.  Otherwise, any 3rd party denom which is in the deposit ICA would be stuck there.  

So if somehow there is a 3rd party token like ATOM in the pool and it mistakenly gets put into the deposit ICA, then running this tx would transfer (without stake or unstake autopilot action) it to the holding module address on Stride.  Then on the next epoch, everything regardless of denom in the holding address will be transferred to the return ICA and eventually be used to fund-community-pool.

I do kind of agree that the low likelihood but real race condition between the epochly checks and the tx that you pointed out might be justification to just remove it completely.  Maybe we just say it is their mistake if any tokens other than DYDX or stDYDX ever show up in the deposit ICA and don't try to address it in this iteration.

x/stakeibc/keeper/community_pool.go
ethan-stride: Yes, we do this pattern everywhere and not terribly consistently.  This suggested function is a good idea in a lot of places in the codebase, maybe if we ever refactor stakeibc like I want to this would be a useful helper since it normally needs to be checked when going from an id to a hostzone variable often in callbacks

x/stakeibc/keeper/community_pool.go
ethan-stride: Yeah, I think I want to remove all the ICQ, IBC, ICA references in function names.  

ICQs would be named [NounWhichIsBeingLookedUp]Query so for example DepositSharesQuery.  Their callback would be in a file called icqcallbacks_deposit_shares.go and the function would be named DepositSharesQueryCallback

IBC doesn't really add anything, I would name these functions something like 
Transfer[AddressFrom][Tokentype]To[AddressTo]  eg. TransferDepositTokensToHolding  or at a higher level
Transfer[ZoneFrom][Tokentype]To[ZoneTo]  eg.  TransferStrideStakedTokensToCommunityPoolZone
idk... maybe it is too long and not that consistent but I do think it is clear which direction things are moving and where you expect them to end up which is important at a glance.

I think part of the problem with ICA is that we often use it to refer to the noun (account address) but sometimes use ICA to describe the commands which the tx to the foreign chain triggers.  I know I am inconsistent on this, but the rest of our code is as well.  

In a lot of the comments I make the distinction that all these functions are really doing is sending a msg which triggers the start of something, but because it is async and it is really a chain of callbacks where the rest of the action will take place, I kind of don't like function names which make it seem like those actions happen to completion inside the function (eg. TransferAndStakeCommunityPoolTokens)

x/stakeibc/keeper/transfer.go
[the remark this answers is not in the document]
sampocs: makes sense, thanks for the explanation!

x/stakeibc/keeper/transfer.go
sampocs: makes sense!

proto/stride/stakeibc/tx.proto
sampocs: yeah from our discussion the other day, I think we're all aligned that if someone accidentally places it there, it's kind of on them. There are a few attack vectors with this tx (see slack). 

But even attack vectors aside, txs in general have a lot of boilerplate so I don't think we should bear the burden of having to add more complexity to our codebase to support this case when we could always refund through an upgrade if needed! (and realistically no one is going to accidentally send funds here haha)

shellvish (commented):
Looks good to me, left some small comments! Will think a bit more about the transfer flow (specifically the error cases), just to make sure I'm not missing anything.

Overall great job, really exciting to see this come together!

x/stakeibc/keeper/transfer.go
```diff
@@ -0,0 +1,214 @@
...
+		})
+	}
+	return nil
+}
+
+
+// helper function to find the ibc denom on the foreign chain of tokens after transfer from stride  
+func (k Keeper) GetDenomOnHostZone(strideDenom string, hostZone types.HostZone) (ibcDenom string) {
+	// we use the hostZone.TransferChannelId because direction is stride to hostZone
+	sourcePrefix := transfertypes.GetDenomPrefix(transfertypes.PortID, hostZone.TransferChannelId)
```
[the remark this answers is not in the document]
ethan-stride: You are correct, I just talked about it with Sam.  I need to get the counter party channel instead.  Good catch!

x/stakeibc/keeper/hooks.go
```diff
@@ -35,6 +35,8 @@ func (k Keeper) BeforeEpochStart(ctx sdk.Context, epochInfo epochstypes.EpochInf
 		k.CleanupEpochUnbondingRecords(ctx, epochNumber)
 		// Create an empty unbonding record for this epoch
 		k.CreateEpochUnbondingRecord(ctx, epochNumber)
+		// Transfers in and out of tokens for hostZones which have community pools
+		k.SweepAllCommunityPoolTokens(ctx)
```
shellvish, line 39: Nit: Let's do this on the Stride epoch!

x/stakeibc/keeper/transfer.go
```diff
@@ -0,0 +1,214 @@
...
+	// Verify that the holding address exists
+	if hostZone.CommunityPoolHoldingAddress == "" || hostZone.CommunityPoolDepositIcaAddress == "" {
+		return errors.New("Invalid holding address or deposit address, cannot build valid ICA commands")
+	}
+
+	// The memo may contain autopilot commands to atomically liquid stake/redeem tokens when transfer succeeds
+	//  both transfer+liquid stake will succeed and tokens will end in the stride side holding address, 
+	//  or neither will and the original base tokens will return to the foreign deposit ICA address
+	memoCommands := ""
+	autopilotMetadata := autopilottypes.RawPacketMetadata{
```
sampocs, line 41: I think we may need to specify a redemption address here if I'm not mistaken

x/stakeibc/keeper/transfer.go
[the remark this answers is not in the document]
sampocs: +1 - I think this should be the channel on the foreign chain

x/stakeibc/keeper/transfer.go
```diff
@@ -0,0 +1,214 @@
...
+			CallbackId:   ICACallbackID_CommunityPoolReturn,
+			CallbackArgs: callbackArgsBz,
+		})
+	}
+	return nil
+}
+
+
+// helper function to find the ibc denom on the foreign chain of tokens after transfer from stride  
+func (k Keeper) GetDenomOnHostZone(strideDenom string, hostZone types.HostZone) (ibcDenom string) {
```
sampocs, line 201: Since this will only work with tokens that are created on Stride - and the only tokens created on stride our stTokens, I think we could rename this to something like: `GetStTokenIBCDenomOnHost`

sampocs: ah I see you actually have that function below. My rec would be to probably just call the above directly instead

x/stakeibc/keeper/transfer.go
```diff
@@ -0,0 +1,214 @@
...
+	} else {
+		//return errorsmod.Wrapf(errors.New(""), "denom is %s", k.GetDenomOnHostZone(coin.Denom, hostZone))
+		result := fmt.Sprintf("Successfully submitted ibctransfer for %+v with response %+v", 
+				coin, msgTransferResponse)
+		k.Logger(ctx).Info(result)
+
+		// If there was no error in sending the transfer msg, store what the transferred denom will be in callback with amount
+		callbackArgs := types.CommunityPoolReturnTransferCallback{
+			HostZoneId: hostZone.ChainId,
+			DenomStride: coin.Denom,
```
sampocs, line 177: DenomStride is only used for logging, so I think you can remove it from the callback data

sampocs (commented):
Gosh darn this is some good looking code!! Apologies for the late review

x/stakeibc/keeper/transfer.go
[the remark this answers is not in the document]
sampocs: whoops sorry, sent that on a stale page before I saw ethan's response

x/stakeibc/keeper/community_pool.go
```diff
@@ -1,12 +1,138 @@
...
+	callbackData := types.CommunityPoolDepositQueryCallback{
+		Denom: denom,
+	}
+	callbackDataBz, err := proto.Marshal(&callbackData)
+	if err != nil {
+		return errorsmod.Wrapf(err, "unable to marshal community pool deposit balance callback data %v", callbackData)
+	}
+
+
+	// Timeout query at end of epoch
```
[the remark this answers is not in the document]
sampocs: I don't think there's any race condition here! If the packet times out, the channel closes anyways

x/stakeibc/keeper/icqcallbacks_pool_deposit_balance.go
```diff
@@ -0,0 +1,70 @@
...
+	if depositDenom == communityPoolHostZone.HostDenom {
+		autoPilotAction = LiquidStake
+	}
+	ibcStDenom := k.GetStakedHostTokenDenomOnHostZone(communityPoolHostZone)
+	if depositDenom == ibcStDenom {
+		autoPilotAction = RedeemStake
+	}
+	transferCoin := sdk.NewCoin(depositDenom, depositBalanceAmount)
+
+	return k.TransferCommunityPoolTokens(ctx, communityPoolHostZone, transferCoin, autoPilotAction)
```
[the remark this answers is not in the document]
sampocs: Thinking through this a bit more, I think ApplyFuncIfNoError makes more sense here because this functionality is mostly used one-off, we probably don't have to worry as much about monitoring for stuck messages. 

I also think the stateful solution gets a bit messy when you have epochs and have to index each record by the epoch ID

x/stakeibc/keeper/hooks.go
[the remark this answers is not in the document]
ethan-stride: It was supposed to be the stride epoch and I put it in the wrong place, per discussions yesterday morning we might actually put it in the even faster hourly epoch since there are more steps than in the original plan

x/stakeibc/keeper/transfer.go
```diff
@@ -0,0 +1,214 @@
...
+	memoCommands := ""
+	autopilotMetadata := autopilottypes.RawPacketMetadata{
+		Autopilot: &struct{
+			Receiver string                  				`json:"receiver"`
+			Stakeibc *autopilottypes.StakeibcPacketMetadata `json:"stakeibc,omitempty"`
+			Claim    *autopilottypes.ClaimPacketMetadata    `json:"claim,omitempty"`
+		}{
+			Receiver: hostZone.CommunityPoolHoldingAddress,
+			Stakeibc: &autopilottypes.StakeibcPacketMetadata{
+				Action: "LiquidStake",
```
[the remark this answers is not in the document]
ethan-stride: No, the memoCommands string is initialized to "" and only gets overwritten with the jsonified autopilotMetadata if we are in the LiquidStake or Redeem stake action cases.  The memoCommands variable is what is used when building the MsgTransfer so no action just is a blank memo.

But this no longer matters since this checking logic will happen in a different function in the 3 epoch plan and we no longer will allow denoms other than DYDX or stDYDX for now so it should always be either a LiquidStake or RedeemStake action

x/stakeibc/keeper/transfer.go
```diff
@@ -0,0 +1,214 @@
...
+			return errorsmod.Wrapf(err, "Couldn't build autopilot json for %v", autopilotMetadata)
+		}
+
+		memoCommands = string(autopilotCmdBz)
+		k.Logger(ctx).Info(fmt.Sprintf("[TransferCommunityPoolTokens] Transferring %v %s and then redeeming stake with memo %s", 
+			token.Amount, token.Denom, memoCommands))
+	} else {
+		k.Logger(ctx).Info(fmt.Sprintf("[TransferCommunityPoolTokens] Transferring %v %s with no additional action", 
+			token.Amount, token.Denom))
+	}
```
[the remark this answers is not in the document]
ethan-stride: Sure that makes sense.  Of course under the 3 epoch plan autopilot no longer matters so this logic is eliminated

x/stakeibc/keeper/transfer.go
```diff
@@ -0,0 +1,214 @@
...
+	transferChannel, found := k.IBCKeeper.ChannelKeeper.GetChannel(ctx, ibctypes.PortID, hostZone.TransferChannelId)
+	if !found {
+		return errorsmod.Wrapf(channeltypes.ErrChannelNotFound, "transfer channel %s not found", hostZone.TransferChannelId)
+	}
+	incomingTransferChannelId := transferChannel.Counterparty.ChannelId
+
+	// one timeout for ICA command ibc messages to arrive, other timeout for subsequent transfer message itself 
+	ibcTransferTimeoutNanos := k.GetParam(ctx, types.KeyIBCTransferTimeoutNanos)
+	icaTimeoutTimestamp := uint64(ctx.BlockTime().UnixNano()) + ibcTransferTimeoutNanos
+	transferTimeoutTimestamp := icaTimeoutTimestamp + ibcTransferTimeoutNanos
```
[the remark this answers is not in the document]
ethan-stride: Ok, that certainly simplifies things.  The weird case is if the ICQ is triggered at the start of the epoch and takes almost the whole epoch before the callback kicks off the transfer so instead of a fixed time window it hits the epoch end very shortly after it is kicked off -- but given stride epoch length this seems unlikely and even if it fails the tokens will be processed in the next epoch trigger anyway

x/stakeibc/keeper/transfer.go
ethan-stride: You mean specifically for the RedeemStake case or even for the LiquidStake?  There is another field which I wasn't initializing in the autopilot struct called "ClaimPacketMetadata" which contains a stride address so perhaps this is where the redemption address is supposed to be included?

x/stakeibc/keeper/transfer.go
ethan-stride: Yep, and in the 3 epoch plan, we don't need to store this at all since the 3rd epoch triggers an ICQ on known denoms instead of needing to remember the amount and denom for the fundcommunitypool call

x/stakeibc/keeper/transfer.go
ethan-stride: This one was written when we were thinking any denom might make it to the holding address, but even with just DYDX and stDYDX, this was trying to unwind the DYDX denom as well -- ie. RedeemStake would output an ibc representation of DYDX in the holding and then this would return "DYDX" because it is going across the channel in the opposite direction to return to its host zone.

Doesn't matter anymore because the 3 epoch plan separates the 2nd epoch transfer to return and the 3rd epoch ICQ->fundcommunitypool.  The ICQ is checking known denoms so this more general function is no longer needed and like you said we only need the ability to figure out the ibc hash for the stDYDX denom on the DYDX chain now.

x/stakeibc/keeper/icqcallbacks_pool_deposit_balance.go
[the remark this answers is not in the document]
ethan-stride: Yep as we discussed, this wouldn't solve the issue of an ica failure or timeout but it would prevent the callback from hanging in the endblocker so I will use the ApplyFuncIfNoError

x/stakeibc/keeper/community_pool.go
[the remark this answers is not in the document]
ethan-stride: Still thinking about this -- I don't think we need to worry about the timeout packet racing with the ICA because the timeout policy is to reject query response so there would be no ICA triggered, but Sam and I have been discussing that the timeout for the the transfer which the ICA triggers on the other chain should be the epoch boundary for the 3 epoch plan to be easier to reason about.  

Therefore we should timeout this ICQ with at least enough time remaining in the epoch for the ICA and the transfer triggered by it to complete.  So maybe much earlier like 50% of the epoch or something for this ICQ timeout -- still should be plenty of time in most cases if we use the stride epochs

x/stakeibc/keeper/icqcallbacks_pool_deposit_balance.go
```diff
@@ -0,0 +1,70 @@
...
+
+	// Confirm the balance is greater than zero for now...
+	// ...perhaps use a positive threshold in the future to avoid work when transfer would be small
+	if depositBalanceAmount.LTE(sdkmath.ZeroInt()) {
+		k.Logger(ctx).Info(utils.LogICQCallbackWithHostZone(chainId, ICQCallbackID_CommunityPoolDepositBalance,
+			"No need to transfer tokens -- not enough found %v %s", depositBalanceAmount, depositDenom))
+		return nil
+	}
+
+	// Transfer the tokens and potentially use autopilot to operate on them (stake/redeem) on landing
```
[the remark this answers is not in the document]
ethan-stride: As clarified in the meeting, autopilot is currently atomic for this type of action so the transfer would be rolled back if the stake action had an issue and funds would remain in the Deposit ICA (even though this will no longer be true for future versions of autopilot with multiple chained actions).  With the new 3 epoch plan, we won't need to worry about this anymore since autopilot is no longer needed.

ethan-stride:
It was already pretty stateless, but the new 3 epoch plan means we don't even need to store down the marshalled callback data for these transfer functions to use since they will no longer even need callback functions.  The new plan simplifies and even eliminates so many functions!

x/stakeibc/keeper/transfer.go
sampocs: yeah that's what I meant - but no longer relevant anyways with the new plan!

x/stakeibc/keeper/community_pool.go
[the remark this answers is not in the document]
sampocs: My mental model is that either the relayer is on and both the ICQ and ICA will go through, or the relayer is off and neither will go through. 

That said, I think it makes the most sense to just time them both out at the end of the epoch. That way, if the relayer comes back on 3/4ths of the way through the epoch, they both get relayed (whereas with the 50/50 split, neither would)

sampocs:
> We'll also have to remember to add support for these channels to RestoreICAAccount

Disregard - the current implementation covers it!

x/stakeibc/keeper/community_pool.go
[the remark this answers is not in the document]
sampocs: am i good to mark this resolved?

x/stakeibc/keeper/transfer.go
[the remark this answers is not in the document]
sampocs: Yeah that's definitely the relevant edge case, but I think it's pretty unlikely and even if it does occur, we can recover from it next epoch anyways like you said

riley-stride (changes requested):

x/stakeibc/keeper/transfer.go
```diff
@@ -0,0 +1,146 @@
...
+	memo := ""
+	var msgs []proto.Message
+	msgs = append(msgs, transfertypes.NewMsgTransfer(
+		transfertypes.PortID,
+		counterpartyChannelId, // for transfers of communityPoolHostZone -> Stride
+		token,
+		hostZone.CommunityPoolDepositIcaAddress, // ICA controlled address on community pool zone
+		destinationHoldingAddress,               // Stride address, unique to each community pool / hostzone
+		clienttypes.Height{},
+		endEpochTimestamp,
```
riley-stride, line 68: Does this msg type not need both timeoutHeight and timeoutTimestamp — it looks like it's just getting a Timestamp only.

x/stakeibc/keeper/transfer.go
```diff
@@ -0,0 +1,146 @@
...
+func (k Keeper) TransferHoldingToCommunityPoolReturn(ctx sdk.Context, hostZone types.HostZone, coin sdk.Coin) error {
+	memo := ""
+	strideEpochTracker, found := k.GetEpochTracker(ctx, epochstypes.STRIDE_EPOCH)
+	if !found {
+		return errorsmod.Wrapf(types.ErrEpochNotFound, epochstypes.STRIDE_EPOCH)
+	}
+	endEpochTimestamp := uint64(strideEpochTracker.NextEpochStartTime)
+
+	// build and send an IBC message for each coin to transfer all back to the hostZone
+	msg := transfertypes.NewMsgTransfer(
```
riley-stride, line 103: Just one coin, right?
```suggestion
	// build and send an IBC message for the coin to transfer all back to the hostZone
```

x/stakeibc/keeper/transfer.go
```diff
@@ -0,0 +1,146 @@
...
+	}
+
+	result := fmt.Sprintf("Successfully submitted ibctransfer for %+v with response %+v",
+		coin, msgTransferResponse)
+	k.Logger(ctx).Info(result)
+
+	return nil
+}
+
+// given a hostZone with native denom, returns the ibc denom on the zone for the staked stDenom
```
riley-stride, line 126: Please add unit tests for this, especially the counterparty channel logic

x/stakeibc/keeper/hooks.go
```diff
@@ -80,6 +80,9 @@ func (k Keeper) BeforeEpochStart(ctx sdk.Context, epochInfo epochstypes.EpochInf
 		if epochNumber%StrideEpochsPerDayEpoch == 0 {
 			k.RebalanceAllHostZones(ctx)
 		}
+
+		// Transfers in and out of tokens for hostZones which have community pools
```
riley-stride, line 84: What's the reason behind running it at this point in the hooks.go flow? Why not earlier / later? I don't have well formed thoughts on where it should be run, mostly curious.

dockernet/config/ica_host.json
```diff
@@ -17,6 +17,7 @@
                     "/cosmos.staking.v1beta1.MsgCancelUnbondingDelegation",
                     "/cosmos.distribution.v1beta1.MsgWithdrawDelegatorReward",
                     "/cosmos.distribution.v1beta1.MsgSetWithdrawAddress",
```
riley-stride, line 19: Is MsgFundCommunityPool enabled in the "add ICA to DYDX" PR?

x/stakeibc/keeper/community_pool.go
```diff
@@ -1,12 +1,248 @@
...
+		err = k.QueryCommunityPoolBalance(ctx, hostZone, types.ICAAccountType_COMMUNITY_POOL_RETURN, stIbcDenom)
+		if err != nil {
+			k.Logger(ctx).Error(utils.LogWithHostZone(hostZone.ChainId, "Querying stHostDenom %s in return - %s", stIbcDenom, err.Error()))
+		}
+	}
+}
+
+// ICQ specific denom for balance in the deposit ICA or return ICA on the community pool host zone
+// Depending on account type and denom, discovered tokens are transferred to Stride or funded to the pool
+func (k Keeper) QueryCommunityPoolBalance(
```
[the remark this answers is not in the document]
sampocs: agreed that would be better!

x/stakeibc/keeper/transfer.go
sampocs: only needs one of either height or timestamp!

x/stakeibc/keeper/transfer.go
sampocs: There’s a test [here](https://github.com/Stride-Labs/stride/blob/d0e01c74ab7b7a375bfa45bebd764197924a720b/x/stakeibc/keeper/transfer_test.go#L175) already - lmk if you were picturing something else!

x/stakeibc/keeper/hooks.go
sampocs: The ordering of these functions doesn’t matter since they’re all emitted in the same block

but there’s a different question as to whether we should submit these hourly or stride-epochly

pros of hourly is it completes the cycle faster, but con is that we’re then issuing a ton of pointless ICQs because of how infrequent this feature is used

I think my vote is probably to change it to hourly and add a feature flag, but I don't have a strong opinion. Curious to hear your thoughts!

dockernet/config/ica_host.json
sampocs: There aren’t any messages specified in that PR cause it will go in the upgrade handler, but Aidan did send a list of messages to allow that included this one

x/stakeibc/keeper/transfer.go
riley-stride: Missed that one, nice! Resolving

x/stakeibc/keeper/transfer.go
riley-stride: This was somehow recommended on the wrong line.

x/stakeibc/keeper/hooks.go
ethan-stride: I agree, hourly is a lot more ICQ to an address which will be empty 99.9% of the time. 3 hours vs 18 hours isn't that different of an experience for the community pool.  The only thing which would change the UX is if we decide to keep records in the future so that we can add query endpoints to report how many funds are in flight at any moment and what stage they are in.

sampocs (approved):
🚢 !

PR #934, opened 2023-09-14
Use detokenization amount from tx response

sampocs:
Closes: #XXX

## Context 
Similar to unbondings, detokenization ICAs have the potential to return slightly less tokens than were specified in the tx due to rounding imprecision in the SDK (see #933 for more context). This can create a small difference in stride's internal record keeping (on the order of < 10uatom). 

However, the number of tokens that winds up being detokenized is part of the tx response, so we can just use that value instead, which will prevent the rounding imprecision from being reflected in record keeping.

## Brief Changelog
* Updated detokenize callback to use the response from the `RedeemTokensForShares` message

## Author's Checklist

I have...

## Integration Tests
```bash
integration_tests.bats
 ✓ [INTEGRATION-BASIC-GAIA] host zones successfully registered
 ✓ [INTEGRATION-BASIC-GAIA] liquid stake mint and transfer
 ✓ [INTEGRATION-BASIC-GAIA] LSM liquid stake
 ✓ [INTEGRATION-BASIC-GAIA] LSM liquid stake with slash query
 ✓ [INTEGRATION-BASIC-GAIA] packet forwarding automatically liquid stakes
 ✓ [INTEGRATION-BASIC-GAIA] rewards are being distributed to stakers
```

      - [ ] not applicable
      - [ ] jira ticket `XXX` 
      - [ ] README.md 
      - [ ] not documented

### DIFF
--- a/x/stakeibc/keeper/icacallbacks_detokenize.go
+++ b/x/stakeibc/keeper/icacallbacks_detokenize.go
@@ -1,6 +1,8 @@
 package keeper
 
 import (
+	"fmt"
+
 	errorsmod "cosmossdk.io/errors"
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	"github.com/cosmos/gogoproto/proto"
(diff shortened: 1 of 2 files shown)


x/stakeibc/keeper/icacallbacks_detokenize.go
```diff
@@ -61,8 +63,21 @@ func (k Keeper) DetokenizeCallback(ctx sdk.Context, packet channeltypes.Packet,
 	// If the ICA succeeded, remove the token deposit
 	k.RecordsKeeper.RemoveLSMTokenDeposit(ctx, deposit.ChainId, deposit.Denom)
 
+	// Determine the actual number of tokens that were turned to native stake
+	// (this can be slightly different than the amount initiated in the redeem tokens tx
+	// due a precision error in the SDK)
+	if len(ackResponse.MsgResponses) != 1 {
+		return fmt.Errorf("Invalid number of messages (%d) in detokenize response: %v",
```
riley-stride, line 70: Do we ever hit this?

riley-stride (approved):

x/stakeibc/keeper/icacallbacks_detokenize.go
sampocs: We shouldn't cause only one msg is submitted at a time
https://github.com/Stride-Labs/stride/blob/main/x/stakeibc/keeper/lsm.go#L296

(Also confirmed in dockernet there is only 1 tx)

riley-stride: makes sense, resolved

PR #936, opened 2023-09-17
add rounding calibration function

sampocs:
Closes: #XXX

## Author's Checklist

I have...

- [ ] Run and PASSED locally all GAIA integration tests
- [ ] If the change is contentful, I either:
    - [ ] Added a new unit test OR 
    - [ ] Added test cases to existing unit tests
- [ ] OR this change is a trivial rework / code cleanup without any test coverage

If skipped any of the tests above, explain.

## Reviewers Checklist

*All items are required. Please add a note if the item is not applicable and please add
your handle next to the items reviewed if you only reviewed selected items.*

I have...

- [ ] reviewed state machine logic
- [ ] reviewed API design and naming
- [ ] manually tested (if applicable)
- [ ] confirmed the author wrote unit tests for new logic
- [ ] reviewed documentation exists and is accurate

## Documentation and Release Note

  - [ ] Does this pull request introduce a new feature or user-facing behavior changes? 
  - [ ] Is a relevant changelog entry added to the `Unreleased` section in `CHANGELOG.md`?
  - [ ] This pull request updates existing proto field values (and require a backend and frontend migration)? 
  - [ ] Does this pull request change existing proto field names (and require a frontend migration)?
  How is the feature or change documented? 
      - [ ] not applicable
      - [ ] jira ticket `XXX` 
      - [ ] specification (`x/<module>/spec/`) 
      - [ ] README.md 
      - [ ] not documented

PR #950, opened 2023-10-11
Cleanup v15

riley-stride:
Closes: #XXX

**UNDELEGATEHOST** [#935](https://github.com/Stride-Labs/stride/pull/935)
- Remove undelegatehost tx

**CALIBRATION** [#936](https://github.com/Stride-Labs/stride/pull/936)
- Remove admin gate on calibration tx
- Limit adjustment to record keeping by 3000microdenom
- Do not block ICAs (there is no need to since we cap the record keeping change)


x/stakeibc/keeper/msg_server_submit_tx.go
```diff
@@ -526,11 +526,6 @@ func (k Keeper) SubmitCalibrationICQ(ctx sdk.Context, hostZone types.HostZone, v
 		return errorsmod.Wrapf(err, "unable to marshal delegator shares callback data")
 	}
```
sampocs, line 527: it wont let me highlight it, but we can also remove the callback data from this

x/stakeibc/keeper/icqcallbacks_callibrate_delegation.go
```diff
@@ -54,36 +57,6 @@ func CalibrateDelegationCallback(k Keeper, ctx sdk.Context, args []byte, query i
...
-
-		if err := k.InterchainQueryKeeper.RetryICQRequest(ctx, query); err != nil {
-			return errorsmod.Wrapf(err, "unable to resubmit delegator shares query")
-		}
-		return nil
-	}
-
-	// If there was no ICA/ICQ overlap, update the validator to indicate that the query
-	//  is no longer in progress (which will unblock LSM liquid stakes to that validator)
-	validator.SlashQueryInProgress = false
```
sampocs, line 85: I think I agree with removing this flag completely from the calibration query (i.e. not setting it before the submission and ignoring it in the callback). 

But one downside is that we we still need to set back the queries that are currently stuck. We can do this in the upgrade handler, but there's the potential that another gets stuck before the upgrade occurs

x/stakeibc/keeper/icqcallbacks_callibrate_delegation.go
```diff
@@ -97,6 +70,12 @@ func CalibrateDelegationCallback(k Keeper, ctx sdk.Context, args []byte, query i
 	}
 
 	delegationChange := validator.Delegation.Sub(delegatedTokens)
+	// if the delegation change is more than the calibration threshold constant, log and throw an error
+	if delegationChange.Abs().GTE(CalibrationThreshold) {
+		k.Logger(ctx).Info(utils.LogICQCallbackWithHostZone(chainId, ICQCallbackID_Calibrate,
```
sampocs, line 75:
```suggestion
		k.Logger(ctx).Error(utils.LogICQCallbackWithHostZone(chainId, ICQCallbackID_Calibrate,
```

x/stakeibc/keeper/icqcallbacks_callibrate_delegation.go
```diff
@@ -97,6 +70,12 @@ func CalibrateDelegationCallback(k Keeper, ctx sdk.Context, args []byte, query i
...
+			"Delegation change is GT CalibrationThreshold, failing calibration callback"))
+		return errorsmod.Wrapf(types.ErrCalibrationThresholdExceeded, "calibration threshold %v exceeded, attempted to calibrate by %v ", CalibrationThreshold, delegationChange)
```
sampocs, line 77: I think we should just log an error, but return `nil` here so that the tx succeeds and the query doesn't stay stuck.

x/stakeibc/keeper/icqcallbacks_callibrate_delegation.go
```diff
@@ -11,6 +11,9 @@ import (
 	"github.com/Stride-Labs/stride/v15/x/stakeibc/types"
 )
 
+// CalibrationThreshold is the max amount of tokens by which a calibration can alter internal record keeping of delegations
+var CalibrationThreshold = sdk.NewInt(1000)
```
sampocs, line 15: Injective actually has the potential to reach this pretty quick cause of the granular microdenom. We can either raise this or just commit to running calibration frequently. Wdyt?

x/stakeibc/keeper/msg_server_submit_tx.go
sampocs: another option is to share the same callback as the delegator_shares query and add a "calibration" bool field to the callback data to identify which one we're in. Then we can cut back on all the duplicated code

sampocs (commented):
Looks good, thanks for doing this! Couple small discussion point below:

x/stakeibc/keeper/icqcallbacks_callibrate_delegation.go
riley-stride: Raised to 100,000 (0.1 macrodenom for 1e-6 chains). Still may get hit fairly quickly on 1e-18 chains, but gives us more slack and we can do calibrations.

x/stakeibc/keeper/msg_server_submit_tx.go
riley-stride: Would you recommend the bool field? Seems cleaner, but I would prefer to not futz with the delegator_shares query as it's somewhat sensitive logic.

x/stakeibc/keeper/icqcallbacks_callibrate_delegation.go
sampocs: This feels a bit more dangerous - I think I'd advocate for keeping it at 1000 for now and monitoring injective over the next couple weeks before we push this live and then re-evaluate.

We could also just remove this tx altogether and issue it epochly

x/stakeibc/keeper/icqcallbacks_callibrate_delegation.go
riley-stride: done

x/stakeibc/keeper/icqcallbacks_callibrate_delegation.go
riley-stride: done

x/stakeibc/keeper/icqcallbacks_callibrate_delegation.go
riley-stride: I don't recall the convo about removing this from the calibration query. Could you point me to that?

x/stakeibc/keeper/icqcallbacks_callibrate_delegation.go
riley-stride: That makes sense. I'll leave it at 1000.

x/stakeibc/keeper/msg_server_submit_tx.go
sampocs: yeah I think my vote is also to not touch it cause it's a pretty gnarly callback as is. I would just remove the callback data from the calibration query and call it a day. We can always try and clean it up later!

x/stakeibc/keeper/icqcallbacks_callibrate_delegation.go
sampocs: I don't recall having a convo about it, I just saw it in this PR haha

x/stakeibc/keeper/icqcallbacks_callibrate_delegation.go
sampocs: Looks like you set it to 100,000 in a recent commit, so I think you need to drop it back down to 1000

sampocs: looking at injective, they got off by ~2200 over the course of probably 2 weeks. 

Let's do either like 3k or 10k

riley-stride: Set to 3k

x/stakeibc/keeper/icqcallbacks_callibrate_delegation.go
sampocs: sorry for confusion - I'm suggesting I agree with the change.

---

I think there's two ways to do this:
1. Keep setting flag during calibration query
    * Pros: we can run a query to set all the current "true" flags back to false
    * Cons: opens the door for them to get stuck again
2. Do not set flags at all during calibration query (implementation in this PR)
    * Pros: we don't risk the flag getting stuck to true
    * Cons: we need to manually set current stuck flags back to false

---

But I'm saying I agree with option 2 - I don't think we actually have real pending queries, so we should be able to just set all flags back to false in the upgrade handler

x/stakeibc/keeper/icqcallbacks_delegator_shares_test.go
```diff
@@ -375,3 +375,24 @@ func (s *KeeperTestSuite) TestDelegatorSharesCallback_PrecisionError() {
 	expectedValDelegation := tc.hostZone.Validators[tc.valIndexQueried].Delegation.Add(precisionErrorTokens)
 	s.Require().Equal(expectedValDelegation.Int64(), validator.Delegation.Int64(), "validator delegation amount")
 }
+
+func (s *KeeperTestSuite) TestCalibrateDelegationCallback_Successful() {
```
asalzmann, line 379: Should this also test the case where an update tries to change the delegation by greater than `CalibrationThreshold`?

x/stakeibc/keeper/icqcallbacks_callibrate_delegation.go
```diff
@@ -97,6 +70,12 @@ func CalibrateDelegationCallback(k Keeper, ctx sdk.Context, args []byte, query i
 	}
 
 	delegationChange := validator.Delegation.Sub(delegatedTokens)
+	// if the delegation change is more than the calibration threshold constant, log and throw an error
+	if delegationChange.Abs().GTE(CalibrationThreshold) {
+		k.Logger(ctx).Error(utils.LogICQCallbackWithHostZone(chainId, ICQCallbackID_Calibrate,
+			"Delegation change is GT CalibrationThreshold, failing calibration callback"))
+		return nil
```
asalzmann, line 77: Just double checking - no state is written above this line right? Doesn't look like it to me, but since this doesn't throw an error, state above this line could be written

x/stakeibc/keeper/icqcallbacks_callibrate_delegation.go
```diff
@@ -11,6 +11,9 @@ import (
 	"github.com/Stride-Labs/stride/v15/x/stakeibc/types"
 )
 
+// CalibrationThreshold is the max amount of tokens by which a calibration can alter internal record keeping of delegations
+var CalibrationThreshold = sdk.NewInt(3000)
```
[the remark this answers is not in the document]
riley-stride: 3000. I just fixed the PR description to say 3000. If you want details, see the thread in comments above with @sampocs

x/stakeibc/keeper/icqcallbacks_callibrate_delegation.go
riley-stride: Good flag. Correct, no state is written in the callback above this line.

x/stakeibc/keeper/icqcallbacks_delegator_shares_test.go
riley-stride: Added a test for this. Required some additional scaffolding; if you have ideas on how to set this test up without that scaffolding please lmk @sampocs!

x/stakeibc/keeper/icqcallbacks_callibrate_delegation.go
riley-stride: Sounds good! Seems reasonable to flip those flags in the upgrade handler

x/stakeibc/keeper/msg_server_submit_tx.go
riley-stride: @sampocs gentle bump

x/stakeibc/keeper/icqcallbacks_callibrate_delegation.go
sampocs: Tracking it [here](https://github.com/Stride-Labs/stride/issues/1003) - marking this resolved!

x/stakeibc/keeper/msg_server_submit_tx.go
sampocs: Sorry not sure how I missed this earlier! You can just set it to `[]byte{}`!

riley-stride: Thanks! Done [here](https://github.com/Stride-Labs/stride/pull/950/commits/299631329f48e246358fc2dd45d07c4591ff65ed).

x/stakeibc/keeper/icqcallbacks_delegator_shares_test.go
sampocs: Simplified the setup and added some more coverage [[here](https://github.com/Stride-Labs/stride/pull/1013)]

x/stakeibc/keeper/icqcallbacks_callibrate_delegation.go
sampocs: I raised this to 5000 in my other PR - I think it'd be good to have a little more leeway

x/stakeibc/keeper/msg_server_submit_tx.go
```diff
@@ -500,10 +500,6 @@ func (k Keeper) SubmitCalibrationICQ(ctx sdk.Context, hostZone types.HostZone, v
 	if hostZone.DelegationIcaAddress == "" {
 		return errorsmod.Wrapf(types.ErrICAAccountNotFound, "no delegation address found for %s", hostZone.ChainId)
 	}
-	validator, _, found := GetValidatorFromAddress(hostZone.Validators, validatorAddress)
```
sampocs: I think we should actually leave this in, but ignore the returned validator
```go
_, _, found := GetValidatorFromAddress(hostZone.Validators, validatorAddress)
if !found {
	return errorsmod.Wrapf(types.ErrValidatorNotFound, "no registered validator for address (%s)", validatorAddress)
}
```

That way the tx fails if someone specifies a validator address that's not in our set

sampocs (changes requested):

x/stakeibc/keeper/msg_server_submit_tx.go
riley-stride: Done [here](https://github.com/Stride-Labs/stride/pull/950/commits/06144f37cf93f7317781c37a82952089bba21d08).

sampocs (approved):
lgtm!

x/stakeibc/keeper/icqcallbacks_delegator_shares_test.go
sampocs: the other PR was merged into here - going to mark this resolved!

shellvish (approved):
Looks great to me, thank you for doing this!

x/stakeibc/keeper/icqcallbacks_callibrate_delegation.go
```diff
@@ -4,13 +4,15 @@ import (
...
 	stakingtypes "github.com/cosmos/cosmos-sdk/x/staking/types"
-	"github.com/cosmos/gogoproto/proto"
 
 	"github.com/Stride-Labs/stride/v16/utils"
 	icqtypes "github.com/Stride-Labs/stride/v16/x/interchainquery/types"
 	"github.com/Stride-Labs/stride/v16/x/stakeibc/types"
 )
 
+// CalibrationThreshold is the max amount of tokens by which a calibration can alter internal record keeping of delegations
+var CalibrationThreshold = sdk.NewInt(5000)
```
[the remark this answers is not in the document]
sampocs: yeah I would have assumed that as well. 

It's totally it's possible that it's caused by something else on their chain, but unfortunately it's closed source :(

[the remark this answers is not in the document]
sampocs: my understanding from other chains is that's it's only due to the rounding error - cause on other chains, it only differs by a couple utokens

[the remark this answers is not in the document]
sampocs: thinking about this a little more - I've noticed that often when a validator is slashed, it seems to happen in two increments (i.e. the shares_to_tokens_rate changes twice)

Very possible that's what happened here and we would have been able to resolve it with a normal slash tx (instead of the calibration query). 

Either way, I don't have any concern with 5000 and we can lower it later if get more intel on injective!

x/stakeibc/keeper/icqcallbacks_delegator_shares_test.go
riley-stride: Looks great, thanks for jumping in on this @sampocs 🙌

PR #955, opened 2023-10-16
phase 2: auto-trade reward token revenue via routes in params

ethan-stride:
This PR is the logic for phase 2 of the community pools where foreign ibc denoms can be treated as revenue.

## Tech Spec and Diagrams
https://docs.google.com/document/d/1U9ah95DABkg_ofSGgkDnhzr4wyJdHCCBExCKCX-t_WM/edit
https://app.excalidraw.com/l/9UFOCMAZLAI/2ErXWeealJE

### Phase 2: Automatically convert revenue of foreign "reward" denom to host denom
There are 4 logical steps kicked off by epochly triggered ICQs:
1. ICQ the reward denom in the withdrawal address on hostZone, transfer all found tokens with unwinding through the reward chain to a trade ICA on the trade chain (Osmosis). ICQ is number 1 in the excalidraw, transfer is number 2.
2. ICQ the reward denom in the trade ICA, callback fires ICA messages to perform a swap to the host chain's denom up to the allowed max swap size. ICQ is number 3 in the excalidraw, swap ICA is number 4.
3. ICQ the host chain denom in the trade ICA, transfer all found tokens back to the withdrawal ICA on the hostZone.  ICQ is number 5 in the excalidraw, transfer is number 6.

The first and last ICQs here are done on a stride epoch cadence, because the swap is limited by a max swap size to prevent slippage, the swaps can happen in smaller more frequent triggers so the middle ICQ is done on an hourly cadence.

Also for safety, a min acceptable output amount is computed using the last known swap price.  There is an independent hourly ICQ to get this "spot price" for the pool (which is a ratio of input to output denom in that pool) and this value is persisted to the TradeConfig type inside the TradeRoute information on the keeper state.  Swaps use whatever value is found (which could be up to 1 hour out of date) when computing the minimum expected output for the swap.  If a swap fails, the tokens just remain in the trade ICA and will be attempted again on the next epochly call to the swap action.

There are governance gated transactions to create, delete, and update the TradeRoute data stored on the keeper state.  The most common expected use case is to update which pool the TradeRoute uses to do the swap or to change the max swap size.

## Testing
```
bash ./dockernet/tests/run_all_tests.sh
integration_tests.bats
 ✓ [INTEGRATION-BASIC-GAIA] host zones successfully registered
 ✓ [INTEGRATION-BASIC-GAIA] liquid stake mint and transfer
 ✓ [INTEGRATION-BASIC-GAIA] LSM liquid stake
 ✓ [INTEGRATION-BASIC-GAIA] LSM liquid stake with slash query
 ✓ [INTEGRATION-BASIC-GAIA] packet forwarding automatically liquid stakes
 ✓ [INTEGRATION-BASIC-GAIA] rewards are being distributed to stakers

11 tests, 0 failures
```


sampocs:
open

x/stakeibc/keeper/reward_converter.go
```diff
@@ -0,0 +1,442 @@
...
+	}
+
+	// See if pool swap spot price has been set to a valid ratio (string representing a float like "10.203")
+	// If there is a valid spot price, use it to set a floor for the acceptable minimum output tokens
+	// 5% slippage is allowed so multiply by 0.95 the expected spot price to get the target minimum
+	// minOut is the minimum number of route.TargetDenomOnTradeZone we must receive or the swap will fail
+	minOut := sdk.ZeroInt()
+	if route.SpotPrice != "" {
+		if spotPrice, err := sdk.NewDecFromStr(route.SpotPrice); err == nil {
+			slippageRatio := sdk.NewDecWithPrec(95, 2) // 0.95 to allow 5% loss to slippage
```
sampocs, line 156: should we change slippage to be a param on the TradeRoute?

x/stakeibc/keeper/hooks.go
```diff
@@ -83,10 +84,20 @@ func (k Keeper) BeforeEpochStart(ctx sdk.Context, epochInfo epochstypes.EpochInf
...
 		k.ProcessAllCommunityPoolTokens(ctx)
+
+		// Do transfers for all reward and swapped tokens defined by the trade routes every stride epoch
+		k.TransferAllRewardTokens(ctx)
 	}
 	if epochInfo.Identifier == epochstypes.MINT_EPOCH {
 		k.AllocateHostZoneReward(ctx)
 	}
+	if epochInfo.Identifier == epochstypes.HOUR_EPOCH {
+		k.UpdateAllSwapPrices(ctx) // update swap spot price frequently
```
sampocs, line 95: considering there's a bit of uncertainty w.r.t. the cadence of these two, it might make sense to decouple them and start them both off at the hourly epoch until we look into some more data. If we need to shorten the cadence beyond that, we should probably create a more granular epoch type rather than using an interval to join the two

x/stakeibc/keeper/reward_converter.go
```diff
@@ -0,0 +1,442 @@
...
+
+	startingDenom := route.RewardDenomOnHostZone
+	sendTokens := sdk.NewCoin(startingDenom, amount)
+
+	withdrawlIcaAddress := route.HostToRewardHop.FromAccount.Address
+	unwindIcaAddress := route.HostToRewardHop.ToAccount.Address
+	tradeIcaAddress := route.RewardToTradeHop.ToAccount.Address
+
+	// Import and use pfm data structures instead of this JSON if we can determine consistent module version...
+	memoJSON := fmt.Sprintf(`"forward": {"receiver": "%s","port": "%s","channel":"%s","timeout":"10m","retries": 2}`,
```
sampocs, line 60: is there an option to specify a timeout timestamp here instead of a duration?

Also, it's possible it doesn't matter if it times out after the epoch, but feels easier to reason about if we have everything time out at once. It's possible there's a race condition I'm not thinking of

x/stakeibc/keeper/reward_converter.go
```diff
@@ -0,0 +1,442 @@
...
+	}
+
+	tradeIcaAccount := route.RewardToTradeHop.ToAccount
+	tradeTokens := sdk.NewCoin(route.RewardDenomOnTradeZone, amount)
+
+	// Prepare Osmosis GAMM module MsgSwapExactAmountIn from the trade account to perform the trade
+	// If we want to generalize in the future, write swap message generation funcs for each DEX type,
+	// decide which msg generation function to call based on check of which tradeZone was passed in
+	var msgs []proto.Message
+	if amount.GT(sdk.ZeroInt()) {
```
sampocs, line 172: I'm assuming this check was in here as a mistake cause we verify above that amount is greater than 0 and, if it was zero, we'd want to exit instead of continuing with sending an empty ICA

x/stakeibc/keeper/reward_converter.go
```diff
@@ -0,0 +1,442 @@
...
+	// minOut is the minimum number of route.TargetDenomOnTradeZone we must receive or the swap will fail
+	minOut := sdk.ZeroInt()
+	if route.SpotPrice != "" {
+		if spotPrice, err := sdk.NewDecFromStr(route.SpotPrice); err == nil {
+			slippageRatio := sdk.NewDecWithPrec(95, 2) // 0.95 to allow 5% loss to slippage
+			inputToOutputRatio := slippageRatio.Mul(spotPrice)
+			minOut = sdk.NewDecFromInt(amount).Mul(inputToOutputRatio).TruncateInt()
+		} else {
+			k.Logger(ctx).Error("Couldn't parse the spot price %s as a Decimal", route.SpotPrice)
+			// Don't allow a missing spot price to stop the swap, use zeroInt as the lower bound for now
```
sampocs, line 161: we'd only not have a price immediately after the TradeRoute is registered right? Feels like we could go ahead and throw an error here since this is unexpected

sampocs (commented):
Beautifully done once again!! 

One larger open question - should we handle the case where the price query fails? Would be a bit messy to keep track of how recent it is, but I'd be worried of the case where we just keeping using a stale value.

Feels like our options are:
* Include no staleness check
* Include a staleness check (would require additional state to track when the price was last updated)
* Update the price to indicate if the query failed (e.g. set it to zero/nil)

x/stakeibc/keeper/reward_converter.go
ethan-stride: Yes but it should be optional and have a default which we specify.  Similar to the min and max swap amounts, if they don't initialize those we can set them to good known defaults.  As discussed on a call, we might want to group all of these params in the route which are rules for how to trade into a new type called TradeConfigs because the update tx route only touches these but would not touch the denoms, hops etc once they are set.

x/stakeibc/keeper/hooks.go
ethan-stride: By putting them both in the hourly but possibly allowing them to have different "cadences" like all the examples in the stride epoch above with their own cadences, this should allow for maximum configuration.

x/stakeibc/keeper/reward_converter.go
ethan-stride: I could be wrong, but from [this code](https://github.com/cosmos/ibc-apps/blob/main/middleware/packet-forward-middleware/packetforward/ibc_middleware.go) which I believe is the most up to date version of pfm, it seems like all input values related to time are intended to be durations instead of absolute timestamps...  I agree it would be nice to enforce the epoch boundary but I don't think it should matter for this transfer because the ICQs either find the funds or they don't in which case they should find them on the next epoch

x/stakeibc/keeper/reward_converter.go
ethan-stride: Yes, it was a stand in for the min_swap_amount before that was defined on the traderoute.  Now that we do those checks before reaching this, I think we can remove it no issue.

x/stakeibc/keeper/reward_converter.go
ethan-stride: Yes true, although as we discussed, it could also happen that they change something on their side so queries break somehow and the data we have becomes out of date, or in the wrong format.  We want to add a lastUpdatedTimestamp so we know how fresh the price is.

x/stakeibc/keeper/reward_converter.go
ethan-stride: Also, we might want to have a variable on here about trade frequency which is essentially the cadence currently hard coded in the stride hourly modulus.  The reason this might need to update, if we lower the max_swap_amount then to move the same volume we would need to speed up the swap cadence.

x/stakeibc/keeper/hooks.go
ethan-stride: See comment above, the cadence might actually belong as part of the swap config variables on the tradeRoute

sampocs: I think if we're under the assumption that it's okay if the price is ~1hour stale because of how loose the threshold is, then I think it's simpler to just do these both at the top of the hour

I think should investigate the cadence a bit more with actual data, but until them, I think it makes sense to just keep it as simple as possible

x/stakeibc/keeper/reward_converter.go
sampocs: hmm, do you think it's worth setting up the two timeouts to guarantee this succeeds or times out by the end of the epoch?

i.e. first transfer has a timeout of currentTime+3hours, and then this autopilot message has timeout duration of 3 hours

I don't want to overcomplicate things, but it does give us a bit of peace of mind if we know things don't overlap epochs

sampocs: that said, if we have good confidence that the overlap doesn't matter (because as you said the ICQ either finds it or it doesn't), then maybe it's not worth it

x/stakeibc/keeper/reward_converter.go
sampocs: yeah sorry my comment was a bit stale. As we discussed on the call/slack, we'll just keep executing the trade even if the price is stale. This is implemented correctly in this branch, marking resolved

x/stakeibc/keeper/reward_converter.go
sampocs: From ethan on slack:
> No I would just make sure everything timeout at the epoch boundary. Most of the time it should ever even be close with 6hour stride epoch

x/stakeibc/keeper/hooks.go
sampocs: from ethan on slack: 
> True totally agree.

x/stakeibc/keeper/reward_converter.go
ethan-stride: The only one which matters for overlap is the swap ICQ and swap ICA command with itself.  The transfers and every other ICQ/ICA can overlap without issue, if the ICQ for the reward in the trade account happens while a previous ICA swap command is actively changing that amount by swapping, then we would have a race condition where they affect each other.  Even so, it would cause the ICQ to overestimate the amount available and the swap ICA commands it calls in the callback could fail due to trying to trade more funds than exist.  But no funds are lost, there is just a 1 epoch delay when the ICA fails and then it should work fine.  To be clear, with the epoch boundary timeouts, even this issue should be impossible currently

sampocs (approved):
🚢!

PR #961, opened 2023-10-25
Add claim rewards ica

riley-stride:
Closes: #XXX

## Author's Checklist

I have...

## Reviewers Checklist

I have...

      - [ ] not applicable
      - [ ] jira ticket `XXX` 
      - [ ] README.md 
      - [ ] not documented


sampocs (commented):
> Can we withdraw all rewards with a single msg (rather than the current multimsg approach)? I can't find a way in the SDK.

I can't find a message either. We probably have to do it by val

---

> Will the current multimsg fit in a block if it includes many validators? Should we cap the number of validators?

I'm not sure, but my vote would be to play it safe and batch it

---

>  Does the validator iteration look good?

Lgtm, but I think the main question here is whether the correct validator address is used.  But i'd imagine it uses the same address as MsgDelegate and other staking txs. And if you verified through the cli, it should be good!

---

> In this implementation we skip withdrawing rewards from 0-delegation validators. Is that the right approach?

I think it is! By the time they actually get to delegation 0 in our record keeping, the undelegation message should have been processed and _I think_ rewards should stop accruing while it's undelegating.

---

> Is executing the ICA at the top of every epoch (not day) the right cadence?

It probably doesn't matter either way. I don't think we need it that frequent but it's also low cost to just do it each epoch.

---

> Is there anything to unit test? Note we don't unittest SetWithdrawalAddressOnHost

I don't think we really need one; however, it's pretty low cost to add and can't hurt! You can just do it all it one function like:


```go
func (s *KeeperTestSuite) ClaimAccruedStakingRewardsOnHost () {
   // Create a delegation ICA channel for the ICA submission
   owner := types.FormatICAAccountOwner(HostChainId, types.ICAAccountType_Delegation)
   channelId, portId := s.CreateICAChannel(owner)

   // Create validators
   validators := types.Validator{}
   for i := 0; i < {some number much larger than batch size}; i++ {
       validators = append(validators, types.Validator{
            Address: fmt.Sprintf("val-%d", i),
            Delegation: i
       })
   } 
   // ^ for the above you could do something like set the delegation to 0 for a subset of them

   // Create host zone 
   hostZone := types.HostZone{
       ChainId: HostChainId,
       DelegationIcaAccount: "delegation",
       WithdrawalIcaAccount: "withdrawal",
       Validators: validators,
   }

   // Create epoch tracker for ICA timeout
   strideEpoch := types.EpochTracker{
      EpochIdentifier:    epochtypes.STRIDE_EPOCH,
       NextEpochStartTime: uint64(s.Coordinator.CurrentTime.UnixNano() + 30_000_000_000), // used for timeout
   }
   s.App.StakeibcKeeper.SetEpochTracker(s.Ctx, strideEpoch)

   // Get start sequence number to confirm ICA was set
   startSequence, found := s.App.IBCKeeper.ChannelKeeper.GetNextSequenceSend(s.Ctx, portId, channelId)
   s.Require().True(found)

   // Call claim accrued rewards to submit ICAs
   err := s.App.StakeibcKeeper.ClaimAccruedRewardsForHost(s.Ctx, hostZone)
   s.Require().NoError(err, "no error expected when accruing rewards")

   // Confirm sequence number incremented by the number of txs 
   // where the number of txs is equal: 
   // (total_validators - validators_with_zero_delegation) / batch_size
   expectedEndSequence := startSequence + XXX
   actualEndSequence, found := s.App.IBCKeeper.ChannelKeeper.GetNextSequenceSend(s.Ctx, portId, channelId)
   s.Require().True(found)
   s.Require().Equal(expectedEndSequence, actualEndSequence, "sequence number should have incremented")

   // Attempt to call it with a host zone without a delegation ICA address, it should fail
   invalidHostZone := hostZone 
   invalidHostZone.DelegationIcaAddress = ""
   err := s.App.StakeibcKeeper.ClaimAccruedRewardsForHost(s.Ctx, hostZone)
   s.Require().ErrorContains(err, "ICA account not found")
   
  //  same for withdrawal address

  // Attempt to call claim with an invalid connection ID on the host zone so the ica fails
}

```

x/stakeibc/keeper/msg_server_submit_tx.go
```diff
@@ -153,6 +157,55 @@ func (k Keeper) SetWithdrawalAddressOnHost(ctx sdk.Context, hostZone types.HostZ
...
+	k.Logger(ctx).Info(utils.LogWithHostZone(hostZone.ChainId, "Withdrawal Address: %s, Delegator Address: %s",
+		hostZone.WithdrawalIcaAddress, hostZone.DelegationIcaAddress))
+
+	validators := hostZone.Validators
+
+	// Build multi-message transaction to withdraw rewards from each validator
+	// batching txs into groups of ClaimRewardsICABatchSize messages, to ensure they will fit in the host's blockSize
+
+	// Calculate the number of batches
+	numBatches := (len(validators) + ClaimRewardsICABatchSize - 1) / ClaimRewardsICABatchSize
```
shellvish, line 177: maybe a dumb question, can we do `numBatches := len(validators) / ClaimRewardsICABatchSize` ?

x/stakeibc/keeper/msg_server_submit_tx.go
```diff
@@ -153,6 +157,55 @@ func (k Keeper) SetWithdrawalAddressOnHost(ctx sdk.Context, hostZone types.HostZ
...
+
+	// Iterate over the batches
+	for batch := 0; batch < numBatches; batch++ {
+		start := batch * ClaimRewardsICABatchSize
+		end := (batch + 1) * ClaimRewardsICABatchSize
+		if end > len(validators) {
+			end = len(validators)
+		}
```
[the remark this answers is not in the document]
shellvish: +1, this feels _slightly_ easier to follow imo!

x/stakeibc/keeper/msg_server_submit_tx.go
```diff
@@ -153,6 +157,55 @@ func (k Keeper) SetWithdrawalAddressOnHost(ctx sdk.Context, hostZone types.HostZ
...
+				continue
+			}
+			msg := &distributiontypes.MsgWithdrawDelegatorReward{
+				DelegatorAddress: hostZone.DelegationIcaAddress,
+				ValidatorAddress: val.Address,
+			}
+			msgs = append(msgs, msg)
+		}
+
+		_, err := k.SubmitTxsStrideEpoch(ctx, hostZone.ConnectionId, msgs, types.ICAAccountType_DELEGATION, "", nil)
```
shellvish, line 200: nit: because we skip assembling the msg if `val.Delegation.IsZero`, should we have a check here of `if len(msgs) > 0` then submit the Txs, and otherwise just continue with the loop?

x/stakeibc/keeper/msg_server_claim_accrued_staking_rewards_on_host_test.go
```diff
@@ -0,0 +1,87 @@
...
+	invalidHostZone = hostZone
+	invalidHostZone.WithdrawalIcaAddress = ""
+	err = s.App.StakeibcKeeper.ClaimAccruedStakingRewardsOnHost(s.Ctx, hostZone)
+	s.Require().ErrorContains(err, "ICA account not found")
+
+	// Attempt to call claim with an invalid connection ID on the host zone so the ica fails
+	invalidHostZone = hostZone
+	invalidHostZone.ConnectionId = ""
+	err = s.App.StakeibcKeeper.ClaimAccruedStakingRewardsOnHost(s.Ctx, hostZone)
+	s.Require().ErrorContains(err, "Failed to SubmitTxs")
```
shellvish, line 86: Really well commented + thorough test, thank you for making this!

shellvish (commented):
Looks great to me! I think there's one very nit `if` statement that we should add, otherwise this looks perfect

x/stakeibc/keeper/msg_server_submit_tx.go
riley-stride: Haha yes good point, I think so. But I removed this now in favor of the approach Sam recommended above.

x/stakeibc/keeper/msg_server_submit_tx.go
riley-stride: Good eye, added [here](https://github.com/Stride-Labs/stride/pull/961/commits/8b9393ac26fecb329ccc51f62eb5c69d05dd9969).

x/stakeibc/keeper/msg_server_claim_accrued_staking_rewards_on_host_test.go
```diff
@@ -0,0 +1,87 @@
...
+	"cosmossdk.io/math"
+	_ "github.com/stretchr/testify/suite"
+
+	epochtypes "github.com/Stride-Labs/stride/v16/x/epochs/types"
+	types "github.com/Stride-Labs/stride/v16/x/stakeibc/types"
+)
+
+// constant number of zero delegations
+const numZeroDelegations = 37
+const claimRewardsICABatchSize = 10
```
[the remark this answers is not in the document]
sampocs: I think this should work:

```go
import "github.com/Stride-Labs/stride/v16/x/stakeibc/keeper"

keeper.ClaimRewardsICABatchSize
```

sampocs (approved):

PR #964, opened 2023-11-03
new flow with three stride epoch triggers for in, transform, out -- no autopilot

ethan-stride:
This updates all the existing function to work with the new three epoch plan

The remaining work is unit tests and to fill in helpers to liquid stake and redeem from stride
These helpers would be called in TransformTokensInHolding based on the denoms discovered

These changes reduced the complexity of the callback data or in some cases removed it completely, removed the need for autopilot actions, and should make sweeps from stage to stage more robust to intermittent failures.  All while avoiding the need for state keeping records!

Open question: should we split holding address into two to keep denom and stDenom tokens always separate just in case the transfer out fails? In that specific case tokens could be confused if they were transferred in from deposit and need to be transformed, or already were transformed but had a problem being sent to the return ICA.

Open question: many of the messages have a timeout which is fixed length, but should we make these all aligned with the end of the next stride epoch?  This would make it easy to reason that the three epochly triggered call paths are not overlapping.


x/stakeibc/keeper/community_pool.go
```diff
@@ -20,59 +20,92 @@ import (
...
+// Depending on account type and denom, discovered tokens are transferred to Stride or funded to the pool
+func (k Keeper) QueryCommunityPoolBalance(ctx sdk.Context, 
+											hostZone types.HostZone, 
+											icaType types.ICAAccountType, 
+											denom string) error {
+
+	k.Logger(ctx).Info(utils.LogWithHostZone(hostZone.ChainId, 
+		"Building ICQ for %s balance in community pool %s address", denom, icaType.String()))
+
+	icaAddress := ""
```
sampocs, line 78: nit: use switch statement

x/stakeibc/keeper/transfer.go
```diff
@@ -109,13 +60,13 @@ func (k Keeper) TransferCommunityPoolDepositToHolding(ctx sdk.Context, hostZone
...
 		types.ICAAccountType_COMMUNITY_POOL_DEPOSIT,
-		icaTimeoutTimestamp,
+		endEpochTimestamp,
 		icaCallbackId,
 		icaCallbackData)
 	if err != nil {
 		return errorsmod.Wrapf(err, "Failed to SubmitTxs, Messages: %v, err: %s", msgs, err.Error())
 	}
-	k.Logger(ctx).Info("[TransferCommunityPoolTokens] Successfully sent ICA command to kick off remote ibc transfer from community pool deposit ICA")
+	k.Logger(ctx).Info("Successfully sent ICA command to kick off ibc transfer from deposit ICA to holding address")
```
sampocs, line 69: ty for removing the [FUNCTION NAME], we should get rid of that across the codebase too

x/stakeibc/keeper/transfer.go
```diff
@@ -147,43 +98,20 @@ func (k Keeper) TransferHoldingToCommunityPoolReturn(ctx sdk.Context, hostZone t
...
-	// we use the hostZone.TransferChannelId because direction is stride to hostZone
-	sourcePrefix := transfertypes.GetDenomPrefix(transfertypes.PortID, hostZone.TransferChannelId)
-	prefixedDenom := sourcePrefix + strideDenom
-
-	return transfertypes.ParseDenomTrace(prefixedDenom).IBCDenom()
-}
-
 // given a hostZone with native denom, returns the ibc denom on the zone for the staked stDenom
-func (k Keeper) GetStakedHostTokenDenomOnHostZone(hostZone types.HostZone) (ibcStakedDenom string) {
+func (k Keeper) GetStakedDenomOnHostZone(ctx sdk.Context, hostZone types.HostZone) (ibcStakedDenom string) {
```
sampocs, line 105: similarly, would recommend something like: `GetStIBCDenomOnHostZone` since "staked" sounds more like the native token

sampocs (approved):
beautiful work and thanks for knocking this out so quickly!

x/stakeibc/keeper/community_pool.go
```diff
@@ -20,59 +20,92 @@ import (
...
 	}
 }
 
-// ICQ specific denom for balance in the deposit ICA on the community pool host zone
-// The ICQ callback will call the transfer function with found token(s) as input
-func (k Keeper) QueryCommunityPoolDepositBalance(ctx sdk.Context, hostZone types.HostZone, denom string) error {
-	k.Logger(ctx).Info(utils.LogWithHostZone(hostZone.ChainId, "Submitting ICQ for %s in community pool deposit account balance", denom))
+// ICQ specific denom for balance in the deposit ICA or return ICA on the community pool host zone
+// Depending on account type and denom, discovered tokens are transferred to Stride or funded to the pool
+func (k Keeper) QueryCommunityPoolBalance(ctx sdk.Context,
```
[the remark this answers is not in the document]
sampocs: resolved in my PR (happened automatically with my auto-save formatting)

x/stakeibc/keeper/community_pool.go
ethan-stride: done

PR #968, opened 2023-11-06
community pool liquid stake integration test scripts

sampocs:
## Context
Added a couple small changes while doing manual integration tests. Successfully tested the staking and redemption flow.

## Changelog
* Fixed bug in AccAddress conversion in liquid stake and redeem stake
* Allowed redemptions to ICA accounts
* Created holding addresses in register host
* Removed `community_pool_address` from host zone
* Added stake/redeem/claim scripts
* Added community pool accounts to logs


proto/stride/stakeibc/host_zone.proto
```diff
@@ -36,8 +36,6 @@ message HostZone {
       [ (cosmos_proto.scalar) = "cosmos.AddressString" ];
   string community_pool_redeem_holding_address = 31
       [ (cosmos_proto.scalar) = "cosmos.AddressString" ];
-  string community_pool_address = 32
-      [ (cosmos_proto.scalar) = "cosmos.AddressString" ];
```
ethan-stride, line 40: Good catch, we originally thought this was needed for the call to FundCommunityPool but it is not so we should remove it!

dockernet/config/ica_host.json
```diff
@@ -17,6 +17,7 @@
                     "/cosmos.staking.v1beta1.MsgCancelUnbondingDelegation",
                     "/cosmos.distribution.v1beta1.MsgWithdrawDelegatorReward",
                     "/cosmos.distribution.v1beta1.MsgSetWithdrawAddress",
+                    "/cosmos.distribution.v1beta1.MsgFundCommunityPool",
```
ethan-stride, line 20: This is the whitelist of which msg types are allowed?  Specifically which msg types the host will listen for and in this case the host is DYDX and stride is the client

dockernet/scripts/community-pool-staking/README.md
```diff
@@ -0,0 +1,20 @@
+## Community Pool Staking Integration Tests
+* Set `HOST_CHAINS` to `(DYDX)` in `config.sh`
+* Start dockernet
+```bash
+make start-docker
+```
```
ethan-stride, line 6: call setup.sh to create the connections between dydx and noble?  Or is this not needed for phase1/3

ethan-stride (commented):
LGTM.  Was able to run the test locally and it worked smoothly.  Awesome to see it do each step in the balance logs and make it back to the pool

dockernet/scripts/community-pool-staking/README.md
sampocs: not needed for phase 1/3!

sampocs: but good call that we should update the readme when we go through stage2

PR #971, opened 2023-11-07
unit test for QueryCommunityPoolBalance

sampocs:


x/stakeibc/keeper/community_pool_test.go
```diff
@@ -0,0 +1,146 @@
...
+	icqtypes "github.com/Stride-Labs/stride/v14/x/interchainquery/types"
+	"github.com/Stride-Labs/stride/v14/x/stakeibc/keeper"
+	"github.com/Stride-Labs/stride/v14/x/stakeibc/types"
+)
+
+func (s *KeeperTestSuite) SetupQueryCommunityPoolBalance(icaAccountType types.ICAAccountType) types.HostZone {
+	// We need to register the transfer channel to initialize the light client state
+	s.CreateTransferChannel(HostChainId)
+
+	// Craete host zone
```
ethan-stride, line 18: nit: create

x/stakeibc/keeper/community_pool_test.go
```diff
@@ -0,0 +1,146 @@
...
+	hostZone := types.HostZone{
+		ChainId:                        HostChainId,
+		ConnectionId:                   ibctesting.FirstConnectionID,
+		CommunityPoolDepositIcaAddress: depositAddress.String(),
+		CommunityPoolReturnIcaAddress:  returnAddress.String(),
+	}
+	s.App.StakeibcKeeper.SetHostZone(s.Ctx, hostZone)
+
+	// Create epoch tracker for timeout
+	epochEndTime := uint64(1000)
```
ethan-stride, line 31: suggestion: instead of using an absolute value here, we should use the ctx.BlockTime and add something to it.

This is because in the query function we do 
`timeout := time.Unix(0, int64(strideEpochTracker.NextEpochStartTime))
timeoutDuration := timeout.Sub(ctx.BlockTime())`

So if this test is working, it is because BlockTime() is 0 in the test framework, just about any other BlockTime would lead to a negative duration

x/stakeibc/keeper/community_pool_test.go
```diff
@@ -0,0 +1,146 @@
...
+	// We need to register the transfer channel to initialize the light client state
+	s.CreateTransferChannel(HostChainId)
+
+	// Craete host zone
+	// We must use valid addresses for each ICA since they're serialized for the query request
+	depositAddress := s.TestAccs[0]
+	returnAddress := s.TestAccs[1]
+	hostZone := types.HostZone{
+		ChainId:                        HostChainId,
+		ConnectionId:                   ibctesting.FirstConnectionID,
```
ethan-stride, line 24: make the HostDenom a field and set it to Atom since later in the tests you assume Atom... this would allow you to properly check that it is hostZone.hostDenom instead of hard coding it

x/stakeibc/keeper/community_pool_test.go
```diff
@@ -0,0 +1,146 @@
...
+	s.Require().Equal(types.ModuleName, query.CallbackModule, "query callback module")
+	s.Require().Equal(keeper.ICQCallbackID_CommunityPoolBalance, query.CallbackId, "query callback id")
+
+	var actualCallbackData types.CommunityPoolBalanceQueryCallback
+	err := proto.Unmarshal(query.CallbackData, &actualCallbackData)
+	s.Require().NoError(err, "no error expected when unmarshalling callback data")
+
+	expectedCallbackData := types.CommunityPoolBalanceQueryCallback{
+		IcaType: icaAccountType,
+		Denom:   Atom,
```
ethan-stride, line 68: hostZone.hostDenom

x/stakeibc/keeper/community_pool_test.go
```diff
@@ -0,0 +1,146 @@
...
+	}
+	s.Require().Equal(expectedCallbackData, actualCallbackData, "query callabck data")
```
ethan-stride, line 70: Does this call DeeplyEqual under the hood?  Otherwise you aren't checking internal fields you are checking pointers....

x/stakeibc/keeper/community_pool_test.go
```diff
@@ -0,0 +1,146 @@
...
+	s.App.StakeibcKeeper.SetEpochTracker(s.Ctx, epochTracker)
+
+	return hostZone
+}
+
+// Helper function to verify the query that was submitted from the community pool balance query
+func (s *KeeperTestSuite) checkCommunityPoolQuerySubmission(
+	hostZone types.HostZone,
+	icaAccountType types.ICAAccountType,
+	expectedIcaAddress string,
```
ethan-stride, line 45: The hostZone and icaAccountType together should give enough info to compute the expectedIcaAddress so why include it as a 3rd input?  This would avoid the weird pattern of handing in 'Atom' as a filler value in the failure cases below?

ethan-stride (commented):
Very clean, a few more possible simplifications in the setup maybe

x/stakeibc/keeper/community_pool_test.go
sampocs: So this was actually intentional, because there isn't a validation that the time is greater than the current block time like there is in transfer (so I was trying to indicate in the test that a dummy value would work).

That said, I realize I should be validating the timeout was set in the query object and it doesn't make sense to use a negative time. I'll give it another pass!

x/stakeibc/keeper/community_pool_test.go
sampocs: i don't think we ever reference the host denom field from the struct, so I was leaving it off intentionally. We pass the denom in as a param in the query so I was just using the `Atom` field as input and validation

I also think we should try to use the default global variables when possible (e.g. HostChainId, Atom) so we don't have to pass those around to each function.

Lmk if you disagree with any of that!

x/stakeibc/keeper/community_pool_test.go
sampocs: same point as above, this isn't pulled off the host zone struct in the function, so I don't think we should use the hz struct to supply the value here

x/stakeibc/keeper/community_pool_test.go
sampocs: The callback data struct only has string fields so there shouldn't be any pointers!

I think the pointers are only relevant if you have a nested struct

x/stakeibc/keeper/community_pool_test.go
sampocs: i went back and forth on this myself. My end decision was that it made more sense to hard code the expected address when invoking the test, rather than redefine the switch logic that's in the actual code in this function. 

this was the test reads a bit more explicitly and we don't run the risk of reintroducing a bug into our test (if there was a bug in the og function)

x/stakeibc/keeper/community_pool_test.go
sampocs: done

x/stakeibc/keeper/community_pool_test.go
ethan-stride: Hmmm, good point since this covers any denom for the stake or unstake routes.  I agree we should use the testing framwork's defaults but I always worry when I see any hardcoded value like Atom but for this case I am convinced.

x/stakeibc/keeper/community_pool_test.go
ethan-stride: ok... but this introduces a new case to test (where the icaAddress provided doesn't match the type given) which is impossible since the logic in the function computes the address.  I don't think I agree with this philosophy yet.  Seems cleaner to avoid it completely and also avoid the test cases where you are passing filler values

x/stakeibc/keeper/community_pool_test.go
sampocs: just pushed up a new attempt - lmk what you think

x/stakeibc/keeper/community_pool_test.go
sampocs: yeah I think if there was conditional logic based on the denom in the query function (which I actually originally thought there was when I started writing this test), then we should be more explicit. But since it's agnostic to the denom, I think we can just use the default

x/stakeibc/keeper/community_pool_test.go
ethan-stride: I meant that it would be comparing the addresses where the structs are stored themselves... but after more research I see that you are right and golang handles this differently than C and structs made of only simple types can be fieldwise compared with the built in == operator.  If the internal fields were pointers that is when the DeeplyEqual would be needed.

x/stakeibc/keeper/community_pool_test.go
sampocs: hmm yeah that's a good point, re: address not matching type. I'll remove the param

and re: filler values, I think it's unavoidable - even if we remove the address string, the accountType is still a filler

sampocs: updated

x/stakeibc/keeper/community_pool_test.go
```diff
@@ -0,0 +1,168 @@
...
+		ChainId:                        HostChainId,
+		ConnectionId:                   ibctesting.FirstConnectionID,
+		CommunityPoolDepositIcaAddress: depositAddress.String(),
+		CommunityPoolReturnIcaAddress:  returnAddress.String(),
+	}
+	s.App.StakeibcKeeper.SetHostZone(s.Ctx, hostZone)
+
+	// Create epoch tracker for timeout
+	timeoutDuration := time.Second * 30
+	epochEndTime := uint64(s.Ctx.BlockTime().Add(timeoutDuration).UnixNano())
```
ethan-stride, line 40: more verbose but clearer what is happening and less room for issues +1

ethan-stride (commented):
LGTM

PR #973, opened 2023-11-07
community pool liquid stake unit test

sampocs:


x/stakeibc/keeper/community_pool_test.go
```diff
@@ -166,3 +175,137 @@ func (s *KeeperTestSuite) TestQueryCommunityPoolBalance_FailedQuerySubmission()
...
+
+	s.Require().Equal(startSequence, endSequence, "no transfer should have been initiated")
+}
+
+// Test liquid stake with an invalid stake holding address
+func (s *KeeperTestSuite) TestLiquidStakeCommunityPoolTokens_Failure_InvalidAddress() {
+	tc := s.SetupLiquidStakeCommunityPoolTokens()
+
+	invalidHostZone := tc.hostZone
+	invalidHostZone.CommunityPoolStakeHoldingAddress = "invalid"
```
ethan-stride, line 285: do you need to call s.App.StakeibcKeeper.SetHostZone(s.Ctx, invalidHostZone) after this modification?

ethan-stride (commented):
LGTM very clean

x/stakeibc/keeper/community_pool_test.go
sampocs: good question! we actually don't since the host zone is passed as a param instead of being looked up in this function

PR #974, opened 2023-11-08
several redeem community pool tests

ethan-stride:
Several tests for redeem stake

Definitely need a few more error cases based on epoch unbonding record setup, and foreign ICA addresses

### DIFF
--- a/x/stakeibc/keeper/community_pool_test.go
+++ b/x/stakeibc/keeper/community_pool_test.go
@@ -184,30 +184,30 @@ func (s *KeeperTestSuite) SetupRedeemCommunityPoolTokens() RedeemCommunityPoolTo
 	// Create relevant module and ICA accounts
 	depositAddress := s.TestAccs[0]
 	communityPoolHoldingAddress := s.TestAccs[1]
-	communityPoolReturnICAAddress := DelegationICAAddress // need an address on HostChain (starts cosmos)
+	communityPoolReturnICAAddress := HostICAAddress // need an address on HostChain (starts cosmos)
 
-	// native stTokens which will be redeemed, dummy tokens which should not be touched
-	initialNativeStTokens := sdk.NewInt(1000)
+	// stTokens which will be redeemed, dummy tokens which should not be touched
+	initialStTokens := sdk.NewInt(1000)
 	initialDummyTokens := sdk.NewInt(999)
 
-	// Fund the redeem holding address with native staked tokens and
+	// Fund the redeem holding address with stTokens and
 	// some dummy tokens that should not get touched while redeeming
 	stDenom := types.StAssetDenomFromHostZoneDenom(Atom)
-	s.FundAccount(communityPoolHoldingAddress, sdk.NewCoin(stDenom, initialNativeStTokens))
+	s.FundAccount(communityPoolHoldingAddress, sdk.NewCoin(stDenom, initialStTokens))
 	s.FundAccount(communityPoolHoldingAddress, sdk.NewCoin(Atom, initialDummyTokens))   // dummy token
 	s.FundAccount(communityPoolHoldingAddress, sdk.NewCoin(IbcAtom, initialDummyTokens)) // dummy token
 
 	// Create a host zone with valid addresses to perform the liquid stake
 	hostZone := types.HostZone{
-		ChainId:                           HostChainId,
-		Bech32Prefix:                      Bech32Prefix,
+		ChainId:                           HostChainId,	//GAIA
+		Bech32Prefix:                      Bech32Prefix, //cosmos
 		HostDenom:                         Atom,
 		IbcDenom:                          IbcAtom,
 		TransferChannelId:                 ibctesting.FirstChannelID,
 		CommunityPoolRedeemHoldingAddress: communityPoolHoldingAddress.String(),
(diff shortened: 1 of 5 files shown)


x/stakeibc/keeper/community_pool_test.go
```diff
@@ -166,3 +169,116 @@ func (s *KeeperTestSuite) TestQueryCommunityPoolBalance_FailedQuerySubmission()
...
+	communityPoolHoldingAddress  sdk.AccAddress
+}
+
+func (s *KeeperTestSuite) SetupRedeemCommunityPoolTokens() RedeemCommunityPoolTokensTestCase {
+	s.CreateTransferChannel(HostChainId)
+
+	// Create relevant module and ICA accounts
+	depositAddress := s.TestAccs[0]
+	communityPoolHoldingAddress := s.TestAccs[1]
+	communityPoolReturnICAAddress := DelegationICAAddress // need an address on HostChain (starts cosmos)
```
sampocs, line 187: wdyt about renaming this to `HostICAAddress` just so we can reuse it? I think it's only used in one other place

x/stakeibc/keeper/community_pool_test.go
```diff
@@ -166,3 +169,116 @@ func (s *KeeperTestSuite) TestQueryCommunityPoolBalance_FailedQuerySubmission()
...
+
+	// native stTokens which will be redeemed, dummy tokens which should not be touched
+	initialNativeStTokens := sdk.NewInt(1000)
+	initialDummyTokens := sdk.NewInt(999)
+
+	// Fund the redeem holding address with native staked tokens and
```
sampocs, line 193:
```suggestion
	// Fund the redeem holding address with stTokens and
```

x/stakeibc/keeper/community_pool_test.go
```diff
@@ -166,3 +169,116 @@ func (s *KeeperTestSuite) TestQueryCommunityPoolBalance_FailedQuerySubmission()
...
+		RedemptionRate:                    sdk.OneDec(),
+	}
+	s.App.StakeibcKeeper.SetHostZone(s.Ctx, hostZone)
+
+	// Create the epoch tracker and deposit records so the liquid stake succeeds
+	epochNumber := uint64(1)
+	epochTracker := types.EpochTracker{
+		EpochIdentifier:    epochtypes.DAY_EPOCH,
+		EpochNumber:        epochNumber,
+		NextEpochStartTime: uint64(s.Coordinator.CurrentTime.UnixNano() + 30_000_000), // dictates transfer timeout
```
sampocs, line 220: Double check me, but I think we can remove the next epoch start time cause it shouldnt be used by this function

x/stakeibc/keeper/community_pool_test.go
```diff
@@ -166,3 +169,116 @@ func (s *KeeperTestSuite) TestQueryCommunityPoolBalance_FailedQuerySubmission()
...
+	// Confirm there are no longer staked tokens in the holding address after redeem
+	stDenom := types.StAssetDenomFromHostZoneDenom(tc.hostZone.HostDenom)
+	stAtomBalance := s.App.BankKeeper.GetBalance(s.Ctx, tc.communityPoolHoldingAddress, stDenom)
+	s.Require().Zero(stAtomBalance.Amount.Int64(), "balance of redeem holidng address should be zero")
+
+	// Confirm the dummy tokens were untouched by the redeem call
+	atomBalance := s.App.BankKeeper.GetBalance(s.Ctx, tc.communityPoolHoldingAddress, Atom)
+	ibcAtomBalance := s.App.BankKeeper.GetBalance(s.Ctx, tc.communityPoolHoldingAddress, IbcAtom)	
+	s.Require().Equal(tc.initialDummyTokens.Int64(), atomBalance.Amount.Int64(), "Atom tokens should not be touched")
+	s.Require().Equal(tc.initialDummyTokens.Int64(), ibcAtomBalance.Amount.Int64(), "IbcAtom tokens should not be touched")
```
sampocs, line 259: I'd add a check here that a user redemption record was created

sampocs (commented):
Good stuff! Imo, I don't actually think you need any more failure cases since the redemption specific failure cases are covered in that test. Only things I would add are, I'd extend the success case to also check that:
* A user redemption record was created
* If you RedeemCommunityPoolTokens again (now that there are no more stTokens), no new redemption record is created

x/stakeibc/keeper/community_pool_test.go
ethan-stride: Sure that would be more general.  I was surprised in the createICAChannel testing function, when it registers and ICA, it seems to do it on the stride side instead of making an array of HostSide ICAs.  It would be easy to modify that function to do both (so add that it makes ICAs on the hostSide) and then we would have more than one potential address to test with (I can imagine we want to test sending ICA to ICA at somepoint)

sampocs: The tricky thing is that the host side is also technically a "stride chain" so it will still have a stride prefix.

I agree it would be nice to actually test the interaction of the two chains more in the unit tests. I think when we built this framework, it was a bit complex to setup but probably something we should revisit soon! But I think for now we can just scope the tests to the point at which the message is sent

sampocs (approved):
lgtm!

PR #978, opened 2023-11-14
whitelist ICA accounts from rate limit during host zone registration

sampocs:
Closes: #XXX

Any epochly transfers between stride and host zones should be whtielisted so they don't trip the rate limit.

## Brief Changelog
* Added the following whitelisted address pairs
  * Deposit address (on stride) to delegation ICA (on host)
  * Fee ICA (on host) to reward collector (on stride)
  * Community pool deposit ICA (on host) to community pool stake holding address (on stride)
  * Community pool deposit ICA (on host) to community pool redeem holding address (on stride)
  * Community pool stake holding address (on stride) to community pool return ICA
* Fixed variable names

### DIFF
--- a/x/stakeibc/ibc_middleware.go
+++ b/x/stakeibc/ibc_middleware.go
@@ -92,46 +92,46 @@ func (im IBCMiddleware) OnChanOpenAck(
 	ctx.Logger().Info(fmt.Sprintf("Found matching address for chain: %s, address %s, port %s", zoneInfo.ChainId, address, portID))
 
 	// addresses
-	withdrawalAddress, err := icatypes.NewControllerPortID(types.FormatICAAccountOwner(hostChainId, types.ICAAccountType_WITHDRAWAL))
+	withdrawalPortID, err := icatypes.NewControllerPortID(types.FormatICAAccountOwner(hostChainId, types.ICAAccountType_WITHDRAWAL))
 	if err != nil {
 		return err
 	}
-	feeAddress, err := icatypes.NewControllerPortID(types.FormatICAAccountOwner(hostChainId, types.ICAAccountType_FEE))
+	feePortID, err := icatypes.NewControllerPortID(types.FormatICAAccountOwner(hostChainId, types.ICAAccountType_FEE))
 	if err != nil {
 		return err
 	}
-	delegationAddress, err := icatypes.NewControllerPortID(types.FormatICAAccountOwner(hostChainId, types.ICAAccountType_DELEGATION))
+	delegationPortID, err := icatypes.NewControllerPortID(types.FormatICAAccountOwner(hostChainId, types.ICAAccountType_DELEGATION))
 	if err != nil {
 		return err
 	}
-	redemptionAddress, err := icatypes.NewControllerPortID(types.FormatICAAccountOwner(hostChainId, types.ICAAccountType_REDEMPTION))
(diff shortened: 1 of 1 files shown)


x/stakeibc/ibc_middleware.go
```diff
@@ -90,52 +91,93 @@ func (im IBCMiddleware) OnChanOpenAck(
...
-	case portID == communityPoolReturnAddress:
+	case portID == communityPoolReturnPortID:
 		zoneInfo.CommunityPoolReturnIcaAddress = address
 	default:
 		ctx.Logger().Error(fmt.Sprintf("Missing portId: %s", portID))
 	}
 
+	// Once the delegation channel is registered, whitelist epochly transfers so they're not rate limited
+	// Epochly transfers go from the deposit address to the delegation address
+	if portID == delegationPortID {
```
ethan-stride, line 142: Why are you whitelisting the delegation and the fee address here?  For most human kicked off staking actions we would want the security that having a rate limit gives normally.  I assume you are whitelisting these because our call to stake and redeem in the community pool route also goes through -- is there any way we can make just those calls unlimited but keep the normal staking actions limited?

ethan-stride (commented):
LGTM, I think the other part of this rate related security goal is that we ought to add a flag which enables the community pool feature or disables it entirely.  Most of the chains won't care about this, and even the one which we are building this for now, probably uses it a few times in the beginning then almost never again.  We should allow them to turn it off for security sake unless they intend to use it on purpose

x/stakeibc/ibc_middleware.go
sampocs: great question. So for context, I just copied over what we did for the other host zones awhile back. And the two non-community pool related whitelists are to make sure our normal epochly flow during regular liquid stakes don't trip the rate limit. 

The fee address here is whitelisted here because stToken rewards to STRD stakers go through that route. And then the delegation address is whitelisted cause that's where the epochly liquid stakes flow through. 

The part I can't quite remember is why we have the deposit → delegation whitelist when we only whitelist stTokens at the moment. But I think it may have just been precautionary in case we whitelisted native tokens in the future

x/stakeibc/ibc_middleware.go
```diff
@@ -90,52 +91,93 @@ func (im IBCMiddleware) OnChanOpenAck(
...
+			Sender:   zoneInfo.DepositAddress,
+			Receiver: zoneInfo.DelegationIcaAddress,
+		})
+	}
+
+	// Once the fee channel is registered, whitelist reward transfers so they're not rate limited
+	// Reward transfers go from the fee address to the reward collector
+	if portID == feePortID {
+		rewardCollectorAddress := im.keeper.AccountKeeper.GetModuleAccount(ctx, types.RewardCollectorName).GetAddress()
+		im.keeper.RatelimitKeeper.SetWhitelistedAddressPair(ctx, ratelimittypes.WhitelistedAddressPair{
```
riley-stride, line 153: To confirm, setting whitelisted address pairs works both when the sender is on stride and the receiver is off stride or when the sender if off stride and the receiver is on stride?

riley-stride (approved):

x/stakeibc/ibc_middleware.go
sampocs: yup!

shellvish (approved):
This looks great to me!

PR #979, opened 2023-11-15
dockernet community pool phase 2 scripts

sampocs:
Closes: #XXX

In order to test out reinvestment and the trade route, we needed additional dockernet setup.

## Brief Changelog
* Updated setup script to start relayers between dydx, noble and osmosis
* Added script to create an osmosis pool of dydx/usdc
* Added scripts to test phase 1/3 via a gov prop (both in the legacy and SDK 47 format)
* Added helper functions to find clients, connections, channels, and ibc denom's from the cli
* Reverted config.sh changes to defaults


dockernet/config/relayer_config_dydx_noble.yaml
```diff
@@ -34,6 +34,21 @@ chains:
       timeout: 20s
       output-format: json
       sign-mode: direct
+  osmo:
```
ethan-stride, line 37: Where can I see which version of Osmosis we have pulled into dockernet? Relevant because the GAMM module which we do the trade ICA tx at has slightly different APIs at different versions

dockernet/config/relayer_config_dydx_noble.yaml
```diff
@@ -43,3 +58,19 @@ paths:
     src-channel-filter:
       rule: ""
       channel-list: []
+  noble-osmo:
```
ethan-stride, line 61: These are directional in the direction we expect transfers to happen, we don't care about return messages so we don't need the other directions correct?

dockernet/scripts/community-pool-staking/create_pool.sh
```diff
@@ -0,0 +1,64 @@
...
+echo ">>> USDC to Osmosis:"
+$NOBLE_MAIN_CMD tx ibc-transfer transfer transfer $noble_to_osmo_channel $(OSMO_ADDRESS) ${LIQUIDITY}${USDC_DENOM} \
+    --from ${NOBLE_VAL_PREFIX}1 -y | TRIM_TX
+sleep 15
+
+echo ">>> Balances:"
+$OSMO_MAIN_CMD q bank balances $(OSMO_ADDRESS)
+
+echo -e "\nCreating dydx/usdc pool on osmosis..."
+pool_file=${STATE}/${OSMO_NODE_PREFIX}1/pool.json
```
ethan-stride, line 49: what is STATE here?

dockernet/scripts/community-pool-staking/reinvest.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+set -eu 
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+source ${SCRIPT_DIR}/../../config.sh
+
+echo ">>> Sending usdc tokens to withdrawal ICA to simulate rewards..." 
+$NOBLE_MAIN_CMD tx ibc-transfer transfer transfer channel-0 $(GET_ICA_ADDR DYDX withdrawal) 1000000${USDC_DENOM} \
```
ethan-stride, line 7: You know it will be noble's channel-0 for the transfer to dydx because of the order you made the connecting relayers?

dockernet/scripts/community-pool-staking/stake_proposal.sh
```diff
@@ -0,0 +1,55 @@
...
+        }
+      ]
+    }
+  ],
+  "deposit": "2000000000udydx"
+}
+EOF
+
+echo ">>> Submitting proposal to spend community pool tokens..."
+$DYDX_MAIN_CMD tx gov submit-proposal $proposal_file --from ${DYDX_VAL_PREFIX}1 -y | TRIM_TX
```
ethan-stride, line 31: Nice, this is a good test for phase 1/3 since this is how we know that community pool action will actually originate.  Is it the same type of gov proposal for each of the community pools (since we now know there could be more than one) or are some of them just module accounts which don't require the gov actions to kick them off?

dockernet/scripts/community-pool-staking/stake_proposal_legacy.sh
```diff
@@ -0,0 +1,43 @@
+#!/bin/bash
+set -eu 
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+source ${SCRIPT_DIR}/../../config.sh
+
+deposit_ica_account=$(GET_ICA_ADDR GAIA community_pool_deposit)
```
ethan-stride, line 6: You are testing this on GAIA because dydx no longer has the legacy gov?

dockernet/src/init_chain.sh
```diff
@@ -213,35 +229,34 @@ if [ "$CHAIN" == "STRIDE" ]; then
...
-        relayer_address=$($MAIN_CMD keys show $RELAYER_GAIA_ICS_ACCT --keyring-backend test -a | tr -cd '[:alnum:]._-')
-        $MAIN_CMD add-genesis-account ${relayer_address} ${VAL_TOKENS}${DENOM}
+        add_relayer_account "$RELAYER_GAIA_ICS_ACCT" "$RELAYER_GAIA_ICS_MNEMONIC"
     fi
-
-    if [ "$CHAIN" == "DYDX" ]; then
-        echo "$RELAYER_DYDX_NOBLE_MNEMONIC" | $MAIN_CMD keys add $RELAYER_DYDX_NOBLE_ACCT --recover --keyring-backend test >> $KEYS_LOGS 2>&1
-        relayer_address=$($MAIN_CMD keys show $RELAYER_DYDX_NOBLE_ACCT --keyring-backend test -a | tr -cd '[:alnum:]._-')
-        $MAIN_CMD add-genesis-account ${relayer_address} ${VAL_TOKENS}${DENOM}
+    if [ "$CHAIN" == "NOBLE" ]; then
```
ethan-stride, line 257: I know these are the relayer addresses not the init on the ICA addresses, but it reminded me about the reward zone ICA.   

If noble has the most up to date pfm module, then we don't need to create the reward zone unwinding ICA at all, if the address is input as "pfm" the most recent pfm module is supposed to be able to handle it if a non-legal address is given.  But we should test this in different versions of integration scripts by changing the TradeRoute data we initialize, one with valid address on the reward zone and one where the address is simply "pfm"

ethan-stride (commented):
LGTM

Lots of nice helper functions which will make these scripts more manageable in the future and nice integration tests for kicking off the community pool actions via gov

dockernet/config/relayer_config_dydx_noble.yaml
sampocs: [Here's](https://github.com/Stride-Labs/stride/blob/main/dockernet/dockerfiles/Dockerfile.osmo#L7) the version we use at the moment - but these dydx branches are a little stale and I think use a different version atm. I'll refresh everything after merging phase 1/3 to main

dockernet/config/relayer_config_dydx_noble.yaml
sampocs: they're actually bidirectional!

dockernet/scripts/community-pool-staking/create_pool.sh
sampocs: it's an env variable that defines the dockernet state directory (`dockernet/state`)

dockernet/scripts/community-pool-staking/reinvest.sh
sampocs: yep exactly! we could also try and determine it dynamically but it felt like doing a bit too much for a simple script like this

dockernet/scripts/community-pool-staking/stake_proposal_legacy.sh
sampocs: yup exactly

dockernet/scripts/community-pool-staking/stake_proposal.sh
sampocs: my understanding is the other "community pools" are just normal module accounts and they can do a gov prop to do a bank send - but it's functionally the same as a normal bank send

dockernet/src/init_chain.sh
sampocs: agreed, good call!

dockernet/scripts/community-pool-staking/reinvest.sh
ethan-stride: yeah I hear you, but the ordering that channels are created seems like one of the easiest things to break by accident.  And some of the code might still work, it would just be interacting with a different unintended chain so it would be harder to catch why certain integration tests suddenly start failing

PR #980, opened 2023-11-15
Adding new traderoute types and keepers

ethan-stride:
Updating the types to be more structured and saved on the keeper store instead of in params


x/stakeibc/keeper/trade_route_test.go
```diff
@@ -0,0 +1,119 @@
...
+		rewardICA := types.ICAAccount{
+			ChainId: rewardChain,
+			Type: types.ICAAccountType_UNWIND,
+		}
+		tradeICA := types.ICAAccount{
+			ChainId: tradeChain,
+			Type: types.ICAAccountType_TRADE,
+		}
+
+		host_reward_hop := types.TradeHop{
```
sampocs, line 37: nit: camel case for these hops

x/stakeibc/keeper/trade_route_test.go
```diff
@@ -0,0 +1,119 @@
...
+		rewardChain := strconv.Itoa(i+1) + "chain"
+		tradeChain := strconv.Itoa(i+2) + "chain"
+
+		hostICA := types.ICAAccount{
+			ChainId: hostChain,
+			Type: types.ICAAccountType_WITHDRAWAL,
+		}
+		rewardICA := types.ICAAccount{
+			ChainId: rewardChain,
+			Type: types.ICAAccountType_UNWIND,
```
sampocs, line 30: do we want to keep this as "UNWIND"?

proto/stride/stakeibc/trade_route.proto
```diff
@@ -2,9 +2,35 @@ syntax = "proto3";
...
+message TradeRoute {
+	string reward_denom_on_host_zone = 1;
+	string reward_denom_on_reward_zone = 2;
+	string reward_denom_on_trade_zone = 3;
+	string target_denom_on_trade_zone = 4;
+	string target_denom_on_host_zone = 5;
+
+	TradeHop host_to_reward_hop = 6;
+	TradeHop reward_to_trade_hop = 7;
+	TradeHop trade_to_host_hop = 8;
```
sampocs, line 26: should we make these nullable is false? I feel like nullable nested structs get kinda messy from protos

For instance, if nullable is false and we have an uninitialized struct, we can do `tradeRoute.host_to_reward_hop.transfer_channel_id` and it will return an empty string. Otherwise, we have to either first check that each field is parent field is not null or use the Getters

x/stakeibc/keeper/trade_route_test.go
```diff
@@ -0,0 +1,119 @@
...
+	"github.com/Stride-Labs/stride/v14/testutil/nullify"
+	"github.com/Stride-Labs/stride/v14/x/stakeibc/keeper"
+	"github.com/Stride-Labs/stride/v14/x/stakeibc/types"
+)
+
+func createNTradeRoute(keeper *keeper.Keeper, ctx sdk.Context, n int) []types.TradeRoute {
+	items := make([]types.TradeRoute, n)
+	for i := range items {
+
+		hostChain := strconv.Itoa(i) + "chain"
```
sampocs, line 20: I know we have a lot of tests scaffolded this way cause of ignite, but I don't think there's really any value in filling out the fields that aren't the keys, and it makes it a bit annoying every time we modify fields

sampocs (commented):
Thanks for getting this together so quickly!! Submitting this midway, then going to slack you about the unit tests!

proto/stride/stakeibc/trade_route.proto
```diff
@@ -2,9 +2,35 @@ syntax = "proto3";
 package stride.stakeibc;
 
 import "gogoproto/gogo.proto";
+import "stride/stakeibc/ica_account.proto";
 
 option go_package = "github.com/Stride-Labs/stride/v14/x/stakeibc/types";
 
+// Define new types less general and more specific to this task, old types to be removed in later PR
```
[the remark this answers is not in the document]
ethan-stride: Yep that is done in the other PR, removing the old types was responsible for 1700 of the removed LOC!

proto/stride/stakeibc/trade_route.proto
ethan-stride: Or we could define a function to ValidateTradeRoute which checks that everything is correctly initialized and call that in the start of each of the functions which uses a trade route.  I don't think the proto nullable = false actually does much to prevent the non-initialized mistake, it just prevents the runtime error from referring to a field on something which doesn't exist but an empty string is also very bad and would break things downstream

sampocs: yeah agreed on the ValidateTradeRoute thing! I still think it's a bit cleaner to not deal with points though. I always thought that was a pain when we had some on our host zone structs and I noticed other projects (like the client/connection/channel objects in ibc-go) just set nullable to false and it's always easier to deal with

sampocs: would not using pointers solve [this](https://github.com/Stride-Labs/stride/pull/980/files#diff-8f6f0a1d11c2bb7abe4951c1e628218eb98be41b99e1372df42ce56fdc4f917bR71-R75)?

x/stakeibc/keeper/trade_route_test.go
ethan-stride: You could change it to REWARD if you want, this is the main one which really doesn't matter because pfm means we never directly use this to send messages from Stride chain so the ICAAccountType is never used to figure out the owner, portID, etc

ethan-stride:
As agreed to previously, merging this into PR 955 so that there is a single PR for review and red team (even though the final few commits on this have not yet been approved).  Everything will be reviewed before that branch (PR 955) is merged to master and this PR 980 will still exist to show a history of the commits if reviewers want to see more detailed logical diffs at each step

PR #981, opened 2023-11-15
Properly Account for slash_query_in_progress in Unbondings

shellvish:
The unbonding algorithm currently doesn't properly account for slash_query_in_progress. It effectively assumes we want to re-distribute those validators stake to others before finding the "optimal" delegations after rebalancing.

This PR removes any delegations to slash_query_in_progress validators when considering stake to be distributed


x/stakeibc/keeper/unbonding_records.go
```diff
@@ -267,7 +267,16 @@ func (k Keeper) UnbondFromHostZone(ctx sdk.Context, hostZone types.HostZone) err
...
-	delegationAfterUnbonding := hostZone.TotalDelegations.Sub(totalUnbondAmount)
+	// first, get the total delegations _excluding_ validators with slash_query_in_progress
+	totalValidDelegationBeforeUnbonding := sdkmath.ZeroInt()
+	for _, validator := range hostZone.Validators {
+		if !validator.SlashQueryInProgress {
+			totalValidDelegationBeforeUnbonding = totalValidDelegationBeforeUnbonding.Add(validator.Delegation)
+		}
+	}
+	// then subtract out the amount to unbond
+	delegationAfterUnbonding := totalValidDelegationBeforeUnbonding.Sub(totalUnbondAmount)
```
sampocs, line 278: should we add a check here that `delegationAfterUnbonding` is not negative? possible it's already checked downstream

sampocs (approved):
sorry for the delay! This looks great, thanks for doing this!

x/stakeibc/keeper/unbonding_records.go
shellvish: I think it's checked in the next function (`k.GetTargetValAmtsForHostZone`) but happy to add an additional check if you think it would make things more clear / has better code hygiene!

PR #982, opened 2023-11-15
reward converter functions use trade route data structure

ethan-stride:
Updating all the functions in the reward converter to use the trade route data structure, simplifies tons of code and removes many unnecessary callback data types.

Having issues with the GAMM version so the just the trade ICA function is commented out, working on this in a separate commit.  This updates everything else and removes the old types which are no longer needed.

Also would be nice to use the proper PFM middleware data structures instead of hard coding the json memo but we have to pick a version of the module since parts of their message types change over different versions.


sampocs (approved):
lgtm! will review biz logic later

proto/osmosis/gamm/v1beta1/osmosis.proto
```diff
@@ -0,0 +1,37 @@
+syntax = "proto3";
+package osmosis.gamm.v1beta1;
+
+import "gogoproto/gogo.proto";
+import "amino/amino.proto";
+import "cosmos/base/v1beta1/coin.proto";
+import "osmosis/poolmanager/v1beta1/swap_route.proto";
+
+option go_package = "github.com/Stride-Labs/stride/v14/x/stakeibc/types";
+
```
sampocs, line 10: * Delete swap_route.proto
* Delete swap_route.pb.go
* Move SwapAmountInRoute to here
```proto
message SwapAmountInRoute {
  uint64 pool_id = 1 [ (gogoproto.moretags) = "yaml:\"pool_id\"" ];
  string token_out_denom = 2
      [ (gogoproto.moretags) = "yaml:\"token_out_denom\"" ];
}
```
* Update SwapAmountInRoute to use local
```proto
string sender = 1 [ (gogoproto.moretags) = "yaml:\"sender\"" ];
  repeated SwapAmountInRoute routes = 2
      [ (gogoproto.nullable) = false ];
```

PR #987, opened 2023-11-22
phase 2 - pool price query

sampocs:
Closes: #XXX

## Author's Checklist

I have...

## Reviewers Checklist

I have...

      - [ ] not applicable
      - [ ] jira ticket `XXX` 
      - [ ] README.md 
      - [ ] not documented


proto/osmosis/gamm/v1beta1/osmosis.proto
```diff
@@ -24,17 +26,63 @@ message MsgSwapExactAmountIn {
...
 message SwapAmountInRoute {
   uint64 pool_id = 1 [ (gogoproto.moretags) = "yaml:\"pool_id\"" ];
   string token_out_denom = 2
       [ (gogoproto.moretags) = "yaml:\"token_out_denom\"" ];
 }
+
+// A TwapRecord stores the most recent price of a pair of denom's
+message OsmosisTwapRecord {
+  uint64 pool_id = 1;
+  // Lexicographically smaller denom of the pair
```
ethan-stride, line 38: Interesting, so the definition of asset0 vs asset1 is based on the alphabetic ordering? Presumably on the ibc denoms? We will have to be careful about this

proto/stride/stakeibc/trade_route.proto
```diff
@@ -39,15 +41,32 @@ message TradeRoute {
   // Currently Osmosis is the only trade chain so this is an osmosis pool id
   uint64 pool_id = 9;
 
-  // Spot price is a decimal ratio of the input to output denom as a string
-  string spot_price = 10;
+  // price of the output denom relative to the input denom
+  // this is update periodically via an ICQ
+  string host_token_price = 10 [
```
ethan-stride, line 46: I don't love this name because it isn't clear what the host price is being priced against... or that it is the ibc denoms which will be involved etc.  Maybe a clearer name for what we are really doing would be would be recent_swap_ratio or something -- this also implies that it could be out of date because it makes a reader ask, how recent? But I think that is a good thing and just more self contained as a name

proto/stride/stakeibc/trade_route.proto
```diff
@@ -39,15 +41,32 @@ message TradeRoute {
...
+    (cosmos_proto.scalar) = "cosmos.Dec",
+    (gogoproto.customtype) = "github.com/cosmos/cosmos-sdk/types.Dec",
+    (gogoproto.nullable) = false
+  ];
+  // timestamp that the price was last updated
+  google.protobuf.Timestamp last_price_update_time = 11
```
ethan-stride, line 52: last is implied, since there are many time related types what about price_update_timestamp as a name?

proto/stride/stakeibc/trade_route.proto
```diff
@@ -39,15 +41,32 @@ message TradeRoute {
...
+      [ (gogoproto.nullable) = false, (gogoproto.stdtime) = true ];
+
+  // max acceptable slippage from the trade, (e.g. 0.05 for 5%)
+  // the latest price and slippage determine the min output tokens from the
+  // trade
+  string max_slippage_percentage = 12 [
```
ethan-stride, line 58: Maybe splitting hairs, slippage is specifically loss from the action of the trade itself because of  shallow volume pool or something, but we also want to use this to defend against high variance price swings -- so the last seen price vs the current price if it is too large of a difference in the wrong direction -- that technically has nothing to do with slippage but we would still want to disallow the trade.  Also percentage makes me think the input would be "5" to mean 5% but you want "0.05" so I might rename this to "max_allowed_swap_loss" or something similar.

x/stakeibc/keeper/icqcallbacks_pool_price.go
```diff
@@ -0,0 +1,95 @@
...
+	var ratioOfHostToRewardDenom sdk.Dec
+	if twapRecord.Asset0Denom == tradeRoute.TargetDenomOnTradeZone {
+		ratioOfHostToRewardDenom = twapRecord.P0LastSpotPrice
+	} else {
+		ratioOfHostToRewardDenom = twapRecord.P1LastSpotPrice
+	}
+
+	// We want the price of the host denom in terms of the reward denom,
+	// so we can use the ratio above to determine the price, via (1 / RatioOfHostToReward)
+	hostPrice := sdk.OneDec().Quo(ratioOfHostToRewardDenom)
```
ethan-stride, line 66: Why involve this extra math which could make weird rounding issues (and should add an extra div by 0 check), from the definitions can't you just reverse the p0 and the p1 spot prices in the if statement above to already get what you want?

x/stakeibc/keeper/reward_converter.go
```diff
@@ -133,57 +133,56 @@ func (k Keeper) TransferConvertedTokensTradeToHost(ctx sdk.Context, amount sdk.I
...
+	if route.MinSwapAmount.IsPositive() && route.MinSwapAmount.GT(rewardAmount) {
 		return nil
 	}
 
 	// If the max swap amount was not set it would be ZeroInt, if positive we need to compare to the amount given
 	//  then if max swap amount is LTE to amount full swap is possible so amount is fine, otherwise set amount to max
-	if route.MaxSwapAmount.IsPositive() && route.MaxSwapAmount.GT(amount) {
-		amount = route.MaxSwapAmount
+	if route.MaxSwapAmount.IsPositive() && route.MaxSwapAmount.GT(rewardAmount) {
+		rewardAmount = route.MaxSwapAmount
```
ethan-stride, line 146: Is this a possible weird race condition: the ICQ detects 1.5x the maxSwapAmount so we fire off a trade command with the amount being maxSwapAmount... afterwards there will be 0.5x maxSwapAmount in the trade ICA so the next swap command wouldn't need to modify it... but what if the swap ICA hasn't executed by the time that the next epoch ICQ on the reward ICA happens? So the ICQ still sees 1.5x maxSwapAmount and so it asks for more to be swapped than will currently be available.  Can we protect against this by making the swap ICA timeout on or slightly before the epoch boundaries?

x/stakeibc/keeper/reward_converter.go
```diff
@@ -133,57 +133,56 @@ func (k Keeper) TransferConvertedTokensTradeToHost(ctx sdk.Context, amount sdk.I
...
-	}
+	//
+	// To calculate minOut, we first convert the rewardAmount into units of HostDenom,
+	//   and then we multiply by (1 - maxSlippage)
+	//
+	// The price of the host token is denominated in units of reward tokens
+	// Meaning, a price of 1.1 implies 1.1 RewardTokens are needed to buy 1 HostToken
+	// So, to convert from units of RewardTokens to units of HostTokens,
+	// we divide the reward amount by the price:
+	//   AmountInHost = AmountInReward / PriceOfHostRelativeToReward
```
ethan-stride, line 160: This seems to me to be a much harder way to think about it.  I would just have ratio of host to reward because then you can just multiply the rewardAmount to get the expected hostAmount

x/stakeibc/keeper/reward_converter.go
```diff
@@ -133,57 +133,56 @@ func (k Keeper) TransferConvertedTokensTradeToHost(ctx sdk.Context, amount sdk.I
...
+	if route.HostTokenPrice == sdk.ZeroDec() {
```
ethan-stride, line 161: Careful on this check, in the code below for TradeRoute initialization you use sdk.OneDec to be the starting value, maybe we should use something literally impossible like a negative number sdk.Dec to explicitly mean not initialized since zero could be a legal value still... although we probably wouldn't want to do that trade if it were zero

ethan-stride: Sorry I see now that was just initialization for a test not for all traderoute setups.  Still we might want an impossible value to start out

ethan-stride (commented):
Looks really good.  I'm so glad that we found this data structure and were able to do a store ICQ with proof instead of going down a rabbit hole.

ethan-stride:
To answer the questions in the overall comment, I think the best option is a trade off between swaps still happening and the potential that the loss is too much and we achieve this by relaxing the allowable slippage when the price is out of date but only up to a point.  

The reasoning:  It would be very bad if swaps can't happen, because this will affect the staking ability downstream so all the holders of stDYDX would be missing out on value they should be getting.  It would be as bad for this particular asset as if the stride zone were down for the period of time that swaps can't happen.  So we definitely want to avoid that case.  By relaxing the allowed slippage loss a bit when the price is out of date we still can have swaps happen with high probability.  

It would also be bad on the other direction if each swap were losing a lot of value because the swap price is temporarily terrible but we can't see that without an up to date price.  By having a max we are willing to relax the slippage we protect against the worst downside movement and still guarantee that some value is coming back.

In practice, we expect the pool we select on Osmosis to be deep (ie. have lots of value for each denom on each side and have lots of activity so high volume) if this is the case then even "swapping blind" without the min value intelligently set shouldn't be that risky. It should return what the ICQ would have told us most of the time.  If the pool does become shallow, the purpose of the update-trade-route tx is for humans to be able to change the pool id to a better one so it would only be making value losing swaps for a short period of time until we were able to notice this.  Because of these properties, I think that the risk of us blocking ourselves from swapping when it would have been fine is higher that the risk of losing a lot of value from bad swaps.  The way to mitigate that is to allow for the slippage to be increased automatically so swaps still happen

proto/osmosis/gamm/v1beta1/osmosis.proto
sampocs: yup - we'll just have to sort them before submitting the request

sampocs:
Reposting our decision from standup on the open questions:

---

> How can we compute the TWAP from this struct? Seems like we might need another historical record. Should we just use spot price?

Decision: just use spot price. Set 5% slippage and 1 hour updates to start (we can always change via a gov prop if needed)
Long term: move to better oracle (e.g. Pyth)

> Should we try to execute the trade if the price is stale?

Decision: keep using the same slippage, but use off-chain monitoring to identify if the value is stale
We can always change the slippage if needed using a gov vote
Trades will stochastically succeed/fail based on how far off the current price is from the last price
Gives us more protection if the pool gets drained

proto/stride/stakeibc/trade_route.proto
sampocs: i'm good with that!

proto/stride/stakeibc/trade_route.proto
sampocs: yeah good points, I like "max_allowed_swap_loss"

sampocs: should we add a descriptor to indicate that it's a decimal? Maybe like "_rate" or "_ratio"?

x/stakeibc/keeper/icqcallbacks_pool_price.go
sampocs: yeah my thinking was that this would be a bit more readable (since seeing "if asset0, set price1" would be a bit confusing

But rounding issues and the div by 0 check are good reasons to not do the division. I'll update this!

x/stakeibc/keeper/reward_converter.go
sampocs: I think zero should be impossible unless the pool is drained, in which case I think we'd also want to reject the trade

I'd also be fine with -1, it's just a bit clunkier with sdkmath (I think you have to do `sdk.OneDec().MulInt64(-1)`)

x/stakeibc/keeper/reward_converter.go
sampocs: yeah I'm a bit torn. I think I biased in this direction cause I've seen from the validator exchange rate that the ratios can be cause for confusion. Also, since the revenue is in USDC, I thought it'd be confusing from a querying perspective to see something like "0.1" for the price of DYDX, when the price is actually $10

proto/stride/stakeibc/trade_route.proto
sampocs: from Ethan on slack:
> yeah rate is good for this one I think

proto/stride/stakeibc/trade_route.proto
sampocs: went with swap_price - pushed in latest commit

x/stakeibc/keeper/reward_converter.go
sampocs: as we discussed offline, moved forward with using the ratio-price (for lack of a better term). Pushed in the latest commit

x/stakeibc/keeper/reward_converter.go
sampocs: hmm, good question. But I don't think there's a race condition cause will have to have succeeded or failed before the next epoch, and the query isn't submitted until the next epoch

ethan-stride: Yes that was my thinking as well so currently it should be fine.  If we ever make the ICA controller work in the way occasionally discussed where it stores down the query this  block but doesn't send until the next block, that could re-open this issue on the rare case that the send block falls on the start of a new Stride epoch

proto/stride/stakeibc/trade_route.proto
```diff
@@ -49,7 +49,7 @@ message TradeRoute {
     (gogoproto.nullable) = false
   ];
   // timestamp that the price was last updated
-  google.protobuf.Timestamp last_price_update_time = 11
+  google.protobuf.Timestamp price_update_time = 11
```
ethan-stride, line 53: this is fine for now, but in the future I prefer using the word "timestamp" in the name when the type is a timestamp because there are so many possible time formats -- human readable, a difference -> duration, unix time in seconds, timestamp in nanos, etc. and in different places across the codebase we use time meaning any of these.

ethan-stride:
LGTM!

proto/stride/stakeibc/trade_route.proto
sampocs: the protobuf type is Timestamp, but the go type is time.Time

PR #989, opened 2023-11-23
phase 2 - implement create trade route

sampocs:
## Brief Changelog
* Removed create-trade-route cli command
* Gov-gated create-trade-route
* Implemented create-trade-route to validate and register ICAs
* Implemented create-trade-route validate basic
* Updated MsgProto to exclude unnecessary fields
* Added GetActiveHostZone helper
* Cleaned up GetChainIdFromConnecrtionId


proto/stride/stakeibc/tx.proto
```diff
@@ -210,35 +212,42 @@ message MsgResumeHostZone {
...
-    (gogoproto.nullable) = false
-  ];
-  string max_swap_amount = 21 [
-    (cosmos_proto.scalar) = "cosmos.Int",
-    (gogoproto.customtype) = "github.com/cosmos/cosmos-sdk/types.Int",
-    (gogoproto.nullable) = false
-  ];
+
+  // Connection IDs between stride and the other zones
+  string stride_to_reward_connection_id = 3;
```
ethan-stride, line 226: At least for now since the only interaction with the reward zone is via PFM, this reward connection ID technically is never used but still reasonable to include it I suppose.  In the future if we want to handle reward zones without PFM installed we would need to manually send them transfer ibc kick off commands for which the connection id would be needed

proto/stride/stakeibc/tx.proto
```diff
@@ -210,35 +212,42 @@ message MsgResumeHostZone {
...
-    (gogoproto.customtype) = "github.com/cosmos/cosmos-sdk/types.Int",
-    (gogoproto.nullable) = false
-  ];
-  string max_swap_amount = 21 [
-    (cosmos_proto.scalar) = "cosmos.Int",
-    (gogoproto.customtype) = "github.com/cosmos/cosmos-sdk/types.Int",
-    (gogoproto.nullable) = false
-  ];
+
+  // Connection IDs between stride and the other zones
```
ethan-stride, line 225: You are not including the stride_to_host_connection_id because we assume that while reward and trade chains might not be in the hostZones array, the "host zone" will be and so we can get the connection id (which is needed in ICA messages we send it) from the normal hostZone data structure?

proto/stride/stakeibc/tx.proto
```diff
@@ -210,35 +212,42 @@ message MsgResumeHostZone {
...
+  // ibc denom of the host's token on the trade zone (e.g. ibc/dydx on Osmosis)
+  string host_denom_on_trade = 11;
+  // the host zone's native denom (e.g. dydx on dYdX)
+  string host_denom_on_host = 12;
+
+  // The osmosis pool ID
+  uint64 pool_id = 13;
+
+  // min and max set boundaries of reward denom on trade chain we will swap
+  uint64 min_swap_amount = 14;
```
ethan-stride, line 249: So eventually these need to become sdk.Int types because they are directly compared to sdk.Coin amounts.  Is it just simpler to make the inputs and these protos with uint64 and do the convert to Int at the last second before compare?

x/stakeibc/client/cli/tx.go
```diff
@@ -40,7 +40,6 @@ func GetTxCmd() *cobra.Command {
 	cmd.AddCommand(CmdUndelegateHost())
 	cmd.AddCommand(CmdUpdateInnerRedemptionRateBounds())
 	cmd.AddCommand(CmdResumeHostZone())
-	cmd.AddCommand(CmdCreateTradeRoute())
```
ethan-stride, line 43: So it is still part of the "API" as a tx route but it isn't part of the CLI? Makes sense I hadn't seen an example of this before but of course those are separable

x/stakeibc/keeper/host_zone.go
```diff
@@ -35,6 +35,18 @@ func (k Keeper) GetHostZone(ctx sdk.Context, chainId string) (val types.HostZone
 	return val, true
 }
 
+// GetActiveHostZone returns an error if the host zone is not found or if it's found, but is halted
```
ethan-stride, line 38: Useful, we do this check everywhere so this should save many LOC across the codebase

x/stakeibc/keeper/keeper.go
```diff
@@ -106,24 +107,24 @@ func (k *Keeper) SetHooks(gh types.StakeIBCHooks) *Keeper {
 	return k
 }
 
-func (k Keeper) GetChainID(ctx sdk.Context, connectionID string) (string, error) {
-	conn, found := k.IBCKeeper.ConnectionKeeper.GetConnection(ctx, connectionID)
+// GetAuthority returns the x/icaoracle module's authority.
```
ethan-stride, line 110: icaoracle module's authority? Should this be stakeibc module's authority?

x/stakeibc/keeper/msg_server_create_trade_route.go
```diff
@@ -2,81 +2,187 @@ package keeper
...
+// >>> strided tx gov submit-proposal {proposal_file.json} --from wallet
+func (ms msgServer) CreateTradeRoute(goCtx context.Context, msg *types.MsgCreateTradeRoute) (*types.MsgCreateTradeRouteResponse, error) {
 	ctx := sdk.UnwrapSDKContext(goCtx)
-	k.Logger(ctx).Info(fmt.Sprintf("create trade route: %s", msg.String()))
+	if ms.authority != msg.Authority {
+		return nil, errorsmod.Wrapf(govtypes.ErrInvalidSigner, "invalid authority; expected %s, got %s", ms.authority, msg.Authority)
+	}
+
+	// Validate trade route does not already exist for this denom
+	_, found := ms.Keeper.GetTradeRoute(ctx, msg.RewardDenomOnHost, msg.HostDenomOnHost)
```
ethan-stride, line 60: Reasonable check for now, in the future if we allow more than one DEX we could allow multiple traderoutes for the same start and end denom, we would compare the best recent price to decide which one we want.  In that case this check would become more complex for also including the trade chain id... I know this problem isn't only here in this check, it would require changing the keepers as well.

x/stakeibc/keeper/msg_server_create_trade_route.go
```diff
@@ -2,81 +2,187 @@ package keeper
...
+	if hostZone.WithdrawalIcaAddress == "" {
+		return nil, errorsmod.Wrapf(types.ErrICAAccountNotFound, "withdrawal account not initialized on host zone")
 	}
 
+	// Register the new ICA accounts
 	hostICA := types.ICAAccount{
 		ChainId:      msg.HostChainId,
 		Type:         types.ICAAccountType_WITHDRAWAL,
-		ConnectionId: msg.HostConnectionId,
-		Address:      msg.HostIcaAddress,
```
ethan-stride, line 28: Why don't we include the address here as hostZone.WithdrawalIcaAddress since we just checked that it is initialized already directly above?

x/stakeibc/keeper/msg_server_create_trade_route.go
```diff
@@ -2,81 +2,187 @@ package keeper
...
+			Type:         icaAccountType,
+			ConnectionId: connectionId,
+			Address:      icaAddress,
+		}
+		return account, nil
+	}
+
+	// Otherwise, if there's no account already, register a new one
+	if err := k.ICAControllerKeeper.RegisterInterchainAccount(ctx, connectionId, owner, appVersion); err != nil {
+		return account, err
```
ethan-stride, line 184: So if there is no icaFound yet, we fire off an ICA command to register a new account and because this is async we return the account (types.ICAAccount) with no address specified yet... how does it eventually add the address so that the ICAAccount type is usable?

x/stakeibc/types/message_create_trade_route.go
```diff
@@ -1,88 +1,130 @@
...
+	ChannelIdRegex    = `^channel-\d+$`
+	IBCPrefix         = "ibc/"
+)
 
-func (msg *MsgCreateTradeRoute) Route() string {
-	return RouterKey
-}
+var (
+	_ sdk.Msg            = &MsgCreateTradeRoute{}
+	_ legacytx.LegacyMsg = &MsgCreateTradeRoute{}
```
ethan-stride, line 24: Can you explain what these two lines do?

ethan-stride (commented):
Looks good, much streamlined and lots of good helpers
I am a bit unclear where the authority gets its permissions defined.  I see that the module keeper now has an authority level which must match, and that the module account is being used as the authority address but how does the gov vote actually sign the tx in the right way to verify that it has the right permissions to execute this action?

ethan-stride:
Looks good, what about the update-trade-route and delete-trade-route cli commands?  I imagine we want to do the same gov gating logic on them, although there is no proposal type there right?  We expect that adding a new trade route is a configuration which will happen very very rarely but the update-trade-route needs to be able to update all the swap configs so things like pool-id if there is a higher volume pool we know about, the max-swap-amount, etc.

sampocs:
> but how does the gov vote actually sign the tx in the right way to verify that it has the right permissions to execute this action?

Low confidence here, but I think the idea is that the gov vote doesn't _technically_ sign this tx. If a user were to submit this tx normally, it would be rejected because the address doesn't match. So the only way to get this tx to succeed would be to have it originate directly from the protocol (instead of through a user), which I think is what the gov module does behind the scenes (i.e. constructs a MsgCreateTradeRoute and submits it directly)

> Looks good, what about the update-trade-route and delete-trade-route cli commands?

yeah was planning to tackle these in a separate PR once some of these other ones are merged!

proto/stride/stakeibc/tx.proto
sampocs: the connection ID is still used to create the ICA account in the first place! but agreed, it's never used for any actual ICA call

proto/stride/stakeibc/tx.proto
sampocs: exactly!

proto/stride/stakeibc/tx.proto
sampocs: Great call out here. I made them uint's intentionally cause I wasn't sure if the gov module could parse the json into sdk.Int's, but we should be able to test that easily and confirm!

x/stakeibc/client/cli/tx.go
sampocs: yup that's right! cause you initiate it through the CLI via the `submit-proposal` tx in the gov module instead of through stakeibc (code above)

x/stakeibc/keeper/keeper.go
sampocs: ah yeah, good catch! done

x/stakeibc/keeper/msg_server_create_trade_route.go
sampocs: yeah good point, but I'd imagine we'll move to the auction approach before we're in a scenario where we want to support multiple DEXs (low confidence though)

Either way, there'd be a bit of a refactor if we did decide to support multiple

PR #991, opened 2023-11-28
min_swap_amount also gates reward token transfers

ethan-stride:
Updating the starting transfer(s) in the route to use the pre-existing min_swap_amount TradeRoute field to determine if the transfer is worth it or reward tokens need to build up.  The parameter is part of the configurable fields in the update-trade-route tx.

This is to avoid small transfers and swaps given that there are known frictional transfer fees.

Also updated the amount type to use cosmossdk.io/math.Int instead of the deprecated cosmos-sdk/types.Int


x/stakeibc/keeper/reward_converter.go
```diff
@@ -5,6 +5,7 @@ import (
 	"time"
 
 	errorsmod "cosmossdk.io/errors"
+	math "cosmossdk.io/math"
```
sampocs, line 8: nit: i think we use `sdkmath` as the alias in most places in the codebase

sampocs (approved):
small change to the sdkmath alias, otherwise LGTM!

PR #992, opened 2023-11-28
phase 2 - misc proto renaming

sampocs:
## Changelog
* Renamed price_update_time → price_update_timestamp and changed type to int64
* Renamed target_denom → host_denom
* Renamed TradeRewardTokens to SwapRewardTokens


x/stakeibc/keeper/icqcallbacks_pool_price.go
```diff
@@ -55,19 +54,19 @@ func PoolPriceCallback(k Keeper, ctx sdk.Context, args []byte, query icqtypes.Qu
...
 
 	k.Logger(ctx).Info(utils.LogICQCallbackWithHostZone(chainId, ICQCallbackID_PoolPrice,
 		"Query response - price ratio of %s to %s is %s",
-		tradeRoute.RewardDenomOnTradeZone, tradeRoute.TargetDenomOnTradeZone, price))
+		tradeRoute.RewardDenomOnTradeZone, tradeRoute.HostDenomOnTradeZone, price))
 
 	// Update the price and time on the trade route data
 	tradeRoute.SwapPrice = price
-	tradeRoute.PriceUpdateTime = ctx.BlockTime()
+	tradeRoute.PriceUpdateTimestamp = ctx.BlockTime().Unix()
```
ethan-stride, line 69: Is this nanos, I thought it was seconds? Should it be nanos everywhere?

ethan-stride: ok I see in the protos this was intended to be seconds, so I think this is correct as intended but a lot of other places in the code base that we use timestamps (like the timestamps passed into the ICA msgs) are nanos

ethan-stride (commented):
LGTM

x/stakeibc/keeper/icqcallbacks_pool_price.go
ethan-stride: Also the epoch trackers are nanos so like strideEpochTracker.NextEpochStartTime is nanos I believe

sampocs: yeah a lot of other stuff is in nanos, but this shouldn't interact with anything else, so I didn't think it was necessary to make them consistent

i don't have a strong preference, I can switch it to nanos if you prefer

sampocs: sg, I'll do nano. 

int64 was just so there's no need to cast. I can do uint for that as well if you think it makes sense

ethan-stride:
LGTM

PR #995, opened 2023-11-28
phase 2 - added trade config

sampocs:
## Brief Changelog
* Added gov gate to update and delete txs
* Added TradeConfig data structure to provide better organization to TradeRoute
* Key'd TradeRoute by reward denom in it's native format instead of ibc format


x/stakeibc/keeper/msg_server_delete_trade_route.go
```diff
@@ -2,33 +2,28 @@ package keeper
...
+		return nil, errorsmod.Wrapf(govtypes.ErrInvalidSigner, "invalid authority; expected %s, got %s", ms.authority, msg.Authority)
 	}
 
-	routes := k.GetAllTradeRoutes(ctx)
-	for _, route := range routes {
-		if route.TargetDenomOnHostZone == msg.HostDenom && route.RewardDenomOnRewardZone == msg.RewardDenom {
-			k.RemoveTradeRoute(ctx, route.RewardDenomOnHostZone, route.TargetDenomOnHostZone)
-		}
-		// if no matching trade route was found for the given host-denom and reward-denom... do nothing
+	_, found := ms.Keeper.GetTradeRoute(ctx, msg.RewardDenom, msg.HostDenom)
```
ethan-stride, line 20: Why have this check that it exists?  If delete succeeds silently when it can't find the route and when it actually removes the route then it becomes idempotent which is often nice.

x/stakeibc/keeper/msg_server_create_trade_route.go
```diff
@@ -107,20 +109,36 @@ func (ms msgServer) CreateTradeRoute(goCtx context.Context, msg *types.MsgCreate
 		ToAccount:         hostICA,
 	}
 
+	// If a max allowed swap loss is not provided, use the default
+	maxAllowedSwapLossRate := msg.MaxAllowedSwapLossRate
+	if maxAllowedSwapLossRate == "" {
```
ethan-stride, line 114: This is nice, I think we should do the same thing for min_swap_amount and max_swap_amount defaulting to 0 and maxint64 if they come in as empty strings

ethan-stride (commented):
LGTM
The names are finally becoming consistent, it feels a lot more clean

x/stakeibc/keeper/msg_server_create_trade_route.go
sampocs: they're actually int types so they default to 0 if not provided!

x/stakeibc/keeper/msg_server_delete_trade_route.go
sampocs: I think it's just a slightly nicer UX in the event that you were to fat finger either of the denom's.

Since this is gov prop, you can imagine a scenario where we have a typo but don't find out for 3 days later until the prop passes and we notice it was never deleted. Whereas, if we add a check here, simulating this tx for the gov prop will fail if the input is invalid

x/stakeibc/keeper/msg_server_create_trade_route.go
ethan-stride: ok nice, but that isn't good for the max swap size, it would cause all the swaps to exit before sending ICAs right now

x/stakeibc/keeper/icqcallbacks_pool_price.go
```diff
@@ -25,9 +25,8 @@ import (
 //	P0LastSpotPrice gives the ratio of Asset0Denom / Asset1Denom
 //	P1LastSpotPrice gives the ratio of Asset1Denom / Asset0Denom
 //
-// When storing down the price, we want to denominate the price of the TargetDenom,
-// relative to the price of the RewardDenom, which means we have to take the inverse
-// from the response
+// When storing down the price, we want to store down the ratio of HostDenom.
```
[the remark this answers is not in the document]
sampocs: sorry this was a typo, I meant to say ration of HostDenom to RewardDenom

x/stakeibc/keeper/msg_server_delete_trade_route.go
ethan-stride: hmm ok.  Yes that is a good argument I'm sold

proto/stride/stakeibc/tx.proto
```diff
@@ -245,33 +246,59 @@ message MsgCreateTradeRoute {
...
-  ];
+  option (cosmos.msg.v1.signer) = "authority";
+  option (amino.name) = "stride/x/stakeibc/MsgUpdateTradeRoute";
+
+  // authority is the address that controls the module (defaults to x/gov unless
+  // overwritten).
+  string authority = 1 [ (cosmos_proto.scalar) = "cosmos.AddressString" ];
+
+  // The reward denom of the route in it's native form (e.g. usdc)
+  string reward_denom = 2;
```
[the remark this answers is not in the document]
sampocs: the key update was done in this PR as well!

x/stakeibc/keeper/msg_server_create_trade_route.go
sampocs: we have a validate basic check that confirms max > min, so I don't think a max of 0 is valid

ethan-stride: So then if they do not include any input value for min and max it will actually not be legal whereas the reasonable defaults feel like no min no max and no max would be implemented as a default value of maxint64 right?  I think we need a similar if statement to set this as the max swap default

sampocs: tis a good point

sampocs: should we do max int or just leave it as 0 and only do the check if if it's != 0?

ethan-stride: I would do max int so that it exists and the normal checks just work out of the box

sampocs: I'm realizing it might be an unintuitive UX to query the trade route and see 18446744073709551615

sampocs: whoops, meant to send that before your last message. Does that change your stance at all?

sampocs: also, I think if we do math.MaxUint, that'll still be a tiny number of tokens for host zones like injective/evmos with a more granular micro denom

ethan-stride: This also gives a way via update tx that humans can pause the transfers and the swapping.

If they set the min_swap to be a very high number, new funds would stop being transferred over to the trade route and if they set the max_swap to 0 they have effectively paused swaps from happening

I know there are checks that min_swap < max_swap but maybe we don't actually want or need those because setting it such that min_swap > max_swap is actually useful to pause activity in this way

ethan-stride: You could make it the more readable 10e19 for humans instead of the absolute max

sampocs: is there any reason we'd do that to pause instead of just removing the trade route?

ethan-stride: I understand your 0 explicit check solution and I think that would be fine if we don't think that the max actually being zero could be a useful case

ethan-stride: You are right we would probably just remove the trade route.  It could be something very niche, like update the min swap to stop new tranfers, wait 2-3 stride epochs for the current funds in flight in the system to be traded and returned to host, then remove the route.  It allows you to stop the transfers and swaps independently but I can't imagine a real usecase where you would want transfers into the trade ICA building up without swapping happening can you?

sampocs: i can't either, but I'm onboard with just doing a human readable number like 10e24

sampocs: I think we also should change the type to string cause max's for host zone's with 18 decimal tokens will exceed the uint limit

ethan-stride:
LGTM

x/stakeibc/keeper/msg_server_create_trade_route.go
sampocs: changed the type to sdk.Int in latest commit

PR #771, opened 2023-05-12
Autopilot enhancement

jstr1121:
## Context and purpose of the change

The goal of this PR is to support two more features on autopilot module

* RedeemStake as part of ibc transfer
* LiquidStake and IBC transfer back liquid staking token as part of ibc transfer

Note: This PR moved from https://github.com/Stride-Labs/stride/pull/626

## Brief Changelog

- MsgLiquidStake is returning received liquid staking token amount
- Add unit test for RedeemStake and LiquidStakeAndIBCTransfer
- Add integration test for RedeemStake and LiquidStakeAndIBCTransfer

## Author's Checklist

I have...

- [x] Run and PASSED locally all GAIA integration tests
- [x] If the change is contentful, I either:
    - [x] Added a new unit test OR 
    - [x] Added test cases to existing unit tests
- [ ] OR this change is a trivial rework / code cleanup without any test coverage

If skipped any of the tests above, explain.

## Reviewers Checklist

*All items are required. Please add a note if the item is not applicable and please add
your handle next to the items reviewed if you only reviewed selected items.*

I have...

- [ ] reviewed state machine logic
- [ ] reviewed API design and naming
- [ ] manually tested (if applicable)
- [ ] confirmed the author wrote unit tests for new logic
- [ ] reviewed documentation exists and is accurate

### How to test integration test on local

- Update `DefaultStakeibcActive` to `true` at `x/autopilot/types/params.go`
- Comment out all `skip "DefaultActive set to false, skip test"` on `dockernet/tests/integration_tests.bats`
- Install docker image for stride `make build-docker build=s`
- Start dockernet with `make start-docker`
- Run integration test with `make test-integration-docker`
- Ensure the result is as following
```
integration_tests.bats
 ✓ [INTEGRATION-BASIC-GAIA] host zones successfully registered
 ✓ [INTEGRATION-BASIC-GAIA] ibc transfer updates all balances
 ✓ [INTEGRATION-BASIC-GAIA] liquid stake mint and transfer
 ✓ [INTEGRATION-BASIC-GAIA] packet forwarding automatically liquid stakes
 ✓ [INTEGRATION-BASIC-GAIA] transfer stuatom to host chain
 ✓ [INTEGRATION-BASIC-GAIA] packet forwarding automatically liquid stake and ibc transfer stAsset to original network
 ✓ [INTEGRATION-BASIC-GAIA] tokens on GAIA were staked
 ✓ [INTEGRATION-BASIC-GAIA] packet forwarding automatically redeem stake
[...]

PR #996, opened 2023-11-28
Refresh autopilot LS + forward

asalzmann:
Closes: #771 

Add autopilot liquid stake + forward support. This allows autopilot liquid stakers to specify a receiving address and IBC channel, where stTokens will be forwarded after being liquid staked.

Open questions
* What should the IBC transfer timeout be?
* Should we use a hash for the sender address (similar to PFM) to avoid the security vulnerability discovered by [SCV](https://twitter.com/SCVSecurity/status/1682329758020022272)? 

Next steps
* Remove StrideAddress (it's unused)
* Implement redeem functionality

## Brief Changelog

  - Add integration test
  - Update autopilot keeper to have access to the transfer keeper
  - Add `IbcReceiver` and `TransferChannel` to `StakeibcPacketMetadata`
  - Add `IBCTransferStAsset` function, called at the end of `RunLiquidStake`

## Test plan
- Add integration test
    - TODO: add additional test for missing `IbcReceiver` and `TransferChannel` fields
- Manually test LS + forward to a different network


dockernet/tests/integration_tests.bats
```diff
@@ -152,6 +152,34 @@ setup_file() {
   assert_equal "$diff" $STAKE_AMOUNT
 }
 
+@test "[INTEGRATION-BASIC-$CHAIN_NAME] packet forwarding automatically liquid stake and ibc transfer stAsset to original network" {
```
sampocs, line 155: Have you run these tests all the way through and confirmed they pass? I'm have a slight recollection of needing to bump down the original autopilot test so that it didn't interfere with the other tests

side note: we discussed in platform prioritization, that moving these tests to a different framework is one of the highest priority items for the beginning of next year. So hopefully this will be less painful moving soon!

x/autopilot/types/parser.go
```diff
@@ -25,13 +25,21 @@ type ModuleRoutingInfo interface {
...
 }
 
+const LiquidStake = "LiquidStake"
+const RedeemStake = "RedeemStake"
+
 // Packet metadata info specific to Stakeibc (e.g. 1-click liquid staking)
 type StakeibcPacketMetadata struct {
-	Action        string `json:"action"`
-	StrideAddress string
+	// TODO: use a constant here
```
sampocs, line 33: is this referring to the constants you added above or something else?

x/autopilot/keeper/liquidstake.go
```diff
@@ -72,12 +72,47 @@ func (k Keeper) RunLiquidStake(ctx sdk.Context, addr sdk.AccAddress, token sdk.C
...
 
 	msgServer := stakeibckeeper.NewMsgServerImpl(k.stakeibcKeeper)
-	_, err := msgServer.LiquidStake(
+	result, err := msgServer.LiquidStake(
 		sdk.WrapSDKContext(ctx),
 		msg,
 	)
 	if err != nil {
-		return errorsmod.Wrapf(sdkerrors.ErrInsufficientFunds, err.Error())
+		return errorsmod.Wrapf(err, err.Error())
```
sampocs, line 80:
```suggestion
		return err
```

x/autopilot/keeper/liquidstake.go
```diff
@@ -72,12 +72,47 @@ func (k Keeper) RunLiquidStake(ctx sdk.Context, addr sdk.AccAddress, token sdk.C
...
+		SourceChannel: channelId,
+		Token:         stAsset,
+		// TODO: does this reintroduce the bug in PFM where senders can be spoofed?
+		// If so, should we instead call PFM directly to forward the packet?
+		// Or should we obfuscate the sender, making it a random address?
+		Sender:           sender,
+		Receiver:         packetMetadata.IbcReceiver,
+		TimeoutTimestamp: timeoutTimestamp,
+		// TimeoutHeight:    clienttypes.Height{},
+		// Memo:             "stTokenIBCTransfer",
```
sampocs, line 113: nit: uncomment or remove

x/autopilot/keeper/liquidstake.go
```diff
@@ -72,12 +72,47 @@ func (k Keeper) RunLiquidStake(ctx sdk.Context, addr sdk.AccAddress, token sdk.C
...
+		channelId = hostZone.TransferChannelId
+	}
+	transferMsg := &transfertypes.MsgTransfer{
+		SourcePort:    transfertypes.PortID,
+		SourceChannel: channelId,
+		Token:         stAsset,
+		// TODO: does this reintroduce the bug in PFM where senders can be spoofed?
+		// If so, should we instead call PFM directly to forward the packet?
+		// Or should we obfuscate the sender, making it a random address?
+		Sender:           sender,
```
sampocs, line 109: Based on our discussion yesterday, I think we should make a point to just not trust senders

Can you give some more details around these two solutions? Specifically:
* how complex is it to do the obfuscation ourselves?
* would using PFM avoid the need for us to implement our own callback?
* what do you consider the tradeoffs between these? my gut reaction is that using PFM might simplify the implementation but make things a bit harder to reason about

x/autopilot/keeper/liquidstake.go
```diff
@@ -72,12 +72,47 @@ func (k Keeper) RunLiquidStake(ctx sdk.Context, addr sdk.AccAddress, token sdk.C
...
+	if err != nil {
+		return err
+	}
+
+	return k.IBCTransferStAsset(ctx, result.StToken, addr.String(), hostZone, packetMetadata)
+}
+
+func (k Keeper) IBCTransferStAsset(ctx sdk.Context, stAsset sdk.Coin, sender string, hostZone *stakeibctypes.HostZone, packetMetadata types.StakeibcPacketMetadata) error {
+	ibcTransferTimeoutNanos := k.stakeibcKeeper.GetParam(ctx, stakeibctypes.KeyIBCTransferTimeoutNanos)
+	timeoutTimestamp := uint64(ctx.BlockTime().UnixNano()) + ibcTransferTimeoutNanos
```
sampocs, line 97: I don't think the timeout matters that much, but from a UX perspective, I think my vote would be to use the default transfer timeout which is 10minutes (`transfertypes.DefaultRelativePacketTimeoutTimestamp`)

sampocs (commented):
very clean! looks good so far, pending the sender change!

x/autopilot/module_ibc.go
```diff
@@ -196,10 +196,14 @@ func (im IBCModule) OnRecvPacket(
...
-		// Try to liquid stake - return an ack error if it fails, otherwise return the ack generated from the earlier packet propogation
-		if err := im.keeper.TryLiquidStaking(ctx, packet, newData, routingInfo); err != nil {
-			im.keeper.Logger(ctx).Error(fmt.Sprintf("Error liquid staking packet from autopilot for %s: %s", newData.Sender, err.Error()))
-			return channeltypes.NewErrorAcknowledgement(err)
+		switch routingInfo.Action {
+		case types.LiquidStake:
+			// Try to liquid stake - return an ack error if it fails, otherwise return the ack generated from the earlier packet propogation
+			if err := im.keeper.TryLiquidStaking(ctx, packet, newData, routingInfo); err != nil {
+				im.keeper.Logger(ctx).Error(fmt.Sprintf("Error liquid staking packet from autopilot for %s: %s", newData.Sender, err.Error()))
+				return channeltypes.NewErrorAcknowledgement(err)
```
sampocs, line 204: do we know if partial state changes get reverted when an ack error is thrown?

dockernet/config.sh
```diff
@@ -57,6 +57,13 @@ STWALK_DENOM="stuwalk"
 STEVMOS_DENOM="staevmos"
 STDYDX_DENOM="studydx"
 
+IBC_GAIA_CHANNEL_0_STATOM_DENOM='ibc/054A44EC8D9B68B9A6F0D5708375E00A5569A28F21E0064FF12CADC3FEF1D04F'
+IBC_GAIA_CHANNEL_1_STATOM_DENOM='ibc/8B21DA0E34A49AE151FEEBCCF3AFE1188E24BA8E19439FB93434DF6008E7E228'
+IBC_GAIA_CHANNEL_2_STATOM_DENOM='ibc/60CB7A5465C318C8F68F603D78721A2ECC1DA2D0E905C6AD9ACD1CAC3F0DB22D'
+IBC_GAIA_CHANNEL_3_STATOM_DENOM='ibc/0C0FD07C29EB075C18EA77B73CF9FCE68A268E0738C9F5B11D13E418AD889437'
```
sampocs, line 63: I think you can remove all besides channel 0. In dockernet, each host chain → stride should be channel 0

x/autopilot/keeper/liquidstake.go
```diff
@@ -72,12 +72,47 @@ func (k Keeper) RunLiquidStake(ctx sdk.Context, addr sdk.AccAddress, token sdk.C
...
+	channelId := packetMetadata.TransferChannel
```
riley-stride, line 98: Would it be safer to not take the channel as input from the user and restrict forwarding to chains Stride supports as host zones? In most cases, passing it in would be fine because it'd be handled on the frontend. But I could imagine integrations who want to use autopilot getting it wrong and consequently losing tokens on wrong channels / chains.

The downside of restricting to only host zones is that we can't LS + forward to chains like Neutron or other defi hubs we don't support for liquid staking yet. I lean towards keeping it as is, but curious to hear your thoughts on this.

x/autopilot/keeper/liquidstake.go
shellvish: nit: should we allow for a memo to be passed to the next chain? If this is technically challenging, absolutely don't think it needs to go into this version, but would allow for some more complex IBC-chaining (e.g. autopilot + swap on Osmosis)

x/autopilot/keeper/liquidstake.go
shellvish: The only potential worry I have about obfuscating the sender is what happens if the IBC transfer fails, does it get returned to the obfuscated sender?

shellvish (approved):
Looks great to me, thanks for doing this! Fwiw, I think this is also working on testnet now.

dockernet/config.sh
sampocs: handled in the redeem stake PR

x/autopilot/module_ibc.go
```diff
@@ -18,7 +18,7 @@ import (
 	ibcexported "github.com/cosmos/ibc-go/v7/modules/core/exported"
 )
 
-const MaxMemoCharLength = 256
+const MaxMemoCharLength = 512
```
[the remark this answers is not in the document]
sampocs: Good callout! This is just meant to protect against spam/state bloat. The original value was stolen from ibc-go [[ref](https://github.com/cosmos/ibc-go/blob/2bd29c08fd1fe50b461fc33a25735aa792dc896e/modules/apps/27-interchain-accounts/types/packet.go#L21-L22)]. All that should matter is that our expected autopilot messages will be under this limit

Upping it to 512 should be totally fine but we should make sure that it's big enough for a PFM message as well!

dockernet/tests/integration_tests.bats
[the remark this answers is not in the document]
sampocs: they're in #1012 - IG is already in there!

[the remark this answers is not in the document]
sampocs: might be a bit tough to add this to the current test framework since we often run tests with only one zone

I think we should definitely try to test this manually though!

x/autopilot/keeper/liquidstake.go
sampocs: +1 to the keeping current setup! 

It's definitely more error prone, but feels like a must have to be able to send to non-host zone chains.

Unfortunately, I do think this reinforces the fact that we should might need a more robust fallback strategy in the case of an ack failure though

sampocs: we discussed this on a call and decided on keeping the current setup! marking resolved

x/autopilot/module_ibc.go
sampocs: confirmed it does [[ref](https://github.com/cosmos/ibc-go/blob/v7.1.0/modules/core/keeper/msg_server.go#L433-L458)] - marking resolved

x/autopilot/keeper/liquidstake.go
sampocs: we discussed this on a call - decided to move forward with obfuscating the sender and using a fallback address. marking this resolved (since it will be implemented in a separate PR)

x/autopilot/types/parser.go
asalzmann: yes, removed

x/autopilot/keeper/liquidstake.go
asalzmann: done

x/autopilot/keeper/liquidstake.go
asalzmann: We use 30 min elsewhere in the codebase, and given the UX with the fallback address is pretty painful, imo we should keep the higher timeout (it might help in the case that lots of blocks are full). although 10 vs. 30 min are both arbitrary and we could go higher

x/autopilot/module_ibc.go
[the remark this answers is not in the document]
asalzmann: Should be big enough for most! We could increase it to ~2k to be safe or leave it, don't feel strongly

x/autopilot/types/parser.go
```diff
@@ -43,14 +51,18 @@ func (m StakeibcPacketMetadata) Validate() error {
...
+	case RedeemStake:
+	default:
 		return errorsmod.Wrapf(ErrUnsupportedStakeibcAction, "action %s is not supported", m.Action)
 	}
 
 	return nil
 }
 
 // Validate claim packet metadata includes the stride address
+// TODO: remove this function
```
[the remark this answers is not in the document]
asalzmann: Removed the comment

x/autopilot/keeper/liquidstake.go
asalzmann: > nit: should we allow for a memo to be passed to the next chain? If this is technically challenging, absolutely don't think it needs to go into this version, but would allow for some more complex IBC-chaining (e.g. autopilot + swap on Osmosis)

for now, we explicitly don't allow pfm / autopilot to interact. agree that it might be worth doing in the future, but I think it complicates the potential code paths we'd have to think through, so would vote to keep it simple for now!

dockernet/tests/integration_tests.bats
asalzmann: I think I did run them, but will run again after addressing comments!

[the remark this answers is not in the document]
asalzmann: Good point! Added a todo above to check this manually

x/autopilot/keeper/keeper.go
```diff
@@ -9,6 +9,7 @@ import (
 	"github.com/cosmos/cosmos-sdk/codec"
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	paramtypes "github.com/cosmos/cosmos-sdk/x/params/types"
+	ibctransferkeeper "github.com/cosmos/ibc-go/v7/modules/apps/transfer/keeper"
```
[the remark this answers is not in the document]
asalzmann: done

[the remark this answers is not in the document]
asalzmann: done

dockernet/tests/integration_tests.bats
sampocs: This comment is a bit stale, but I think this individual test succeeded, but it had caused an LSM test to fail.

It was resolved in a separate PR though!

x/autopilot/keeper/liquidstake.go
sampocs: The fallback address is only used for failures - timeouts are retried infinitely

dockernet/tests/integration_tests.bats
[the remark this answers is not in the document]
sampocs: I tested this manually the other week, but would be good to have someone else go through it as well!

sampocs:
@asalzmann fyi, unit tests for this are in #1041

x/autopilot/module_ibc.go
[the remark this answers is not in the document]
sampocs: I think increasing it to 2k makes sense!

A PFM transfer with 2 hops is roughly ~500 characters I think 512 is probably a bit too restrictive

dockernet/tests/integration_tests.bats
asalzmann: ah, thanks!

x/autopilot/keeper/liquidstake.go
asalzmann: This was changed and is being discussed here https://github.com/Stride-Labs/stride/pull/1038#discussion_r1443768561

x/autopilot/module_ibc.go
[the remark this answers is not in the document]
asalzmann: Increased to 2k

Fwiw there's a bit of discussion on this here https://twitter.com/zmanian/status/1743123853461889058

asalzmann:
Next steps for this PR
- Merge in other PRs that depend on this PR 
- Do manual test in description
- (maybe - unsure if needed) add another integration test

dockernet/tests/integration_tests.bats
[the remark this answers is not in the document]
sampocs: tested this offline as a group. resolving

sampocs:
(for the record)...

> Merge in other PRs that depend on this PR

We decided to merge this first

> Do manual test in description

Done and resolved comment

> (maybe - unsure if needed) add another integration test

We decided it wasn't necessary

PR #997, opened 2023-11-29
phase 2 - integration tests

sampocs:
## Chain Changelog
* Added brackets to pfm command
* Changed ICQ/trade cadence

## Dockernet Changelog
* Added add_trade_route script
* Added trade ICA to balances
* Created channels.log
* Added stride-noble relayer
* Upgraded osmosis

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -18,7 +18,7 @@
 	path = deps/juno
 	url = https://github.com/CosmosContracts/juno.git
 [submodule "deps/osmosis"]
-	# Commit: v16.1.1-no-fees
+	# Commit: v20.5.0-no-fees
 	path = deps/osmosis
 	url = https://github.com/Stride-Labs/osmosis.git
 [submodule "deps/stargaze"]
--- a/dockernet/config/relayer_config_dydx_noble.yaml
+++ b/dockernet/config/relayer_config_dydx_noble.yaml
@@ -73,6 +73,14 @@ paths:
     src-channel-filter:
       rule: ""
       channel-list: []
+  stride-noble:
+    src:
+      chain-id: STRIDE
+    dst:
+      chain-id: NOBLE
+    src-channel-filter:
+      rule: ""
+      channel-list: []
   dydx-noble:
     src:
       chain-id: DYDX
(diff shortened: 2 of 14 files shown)


dockernet/config/relayer_config_dydx_noble.yaml
```diff
@@ -73,6 +73,14 @@ paths:
     src-channel-filter:
       rule: ""
       channel-list: []
+  stride-noble:
```
ethan-stride, line 76: In theory do we really need this relayer?  Is it just for the ICA initial setup?
We never send any direct messages to the reward chain, it is only done via pfm...

x/stakeibc/keeper/reward_converter.go
```diff
@@ -67,7 +67,7 @@ func (k Keeper) TransferRewardTokensHostToTrade(ctx sdk.Context, amount sdkmath.
 	// Import and use pfm data structures instead of this JSON if we can determine consistent module version...
 	// This transfer channel id is a channel on the reward Zone for transfers to the trade zone
 	// (not to be confused with a transfer channel on Stride or the Host Zone)
-	memoJSON := fmt.Sprintf(`"forward": {"receiver": "%s","port": "%s","channel":"%s","timeout":"10m","retries": 2}`,
+	memoJSON := fmt.Sprintf(`"{ forward": {"receiver": "%s", "port": "%s", "channel":"%s", "timeout":"10m", "retries": 2} }`,
```
ethan-stride, line 70: Good catch, was this failing to parse before?

ethan-stride (commented):
LGTM
I know it is still a work in progress but everything make sense

.gitmodules
```diff
@@ -18,7 +18,7 @@
 	path = deps/juno
 	url = https://github.com/CosmosContracts/juno.git
 [submodule "deps/osmosis"]
-	# Commit: v16.1.1-no-fees
+	# Commit: v20.5.0-no-fees
```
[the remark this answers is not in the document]
sampocs: shouldn't we use the latest mainnet version? I assume v20.5 has the patch for this

dockernet/config/relayer_config_dydx_noble.yaml
sampocs: right - only used for the ICA creation

x/stakeibc/keeper/reward_converter.go
sampocs: I wasn't seeing an error for it, but I think it would have appeared on noble (and I doubt they have it logged by default)

I just noticed it visually when I was trying to recreate the transfer manually to confirm pfm was working in dockernet

.gitmodules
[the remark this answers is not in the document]
ethan-stride: Yeah normally I would think so, but I bet everything we care about would be sufficiently up to date in v20.2.0 and we don't know if they have the patch in yet.  I doubt this is the main cause of the problem but if we do see it dropping our incoming packets completely this might be something to try.

PR #1008, opened 2023-12-04
reward converter - added ICA address validation

sampocs:


x/stakeibc/keeper/reward_converter.go
```diff
@@ -77,6 +77,17 @@ func (k Keeper) BuildHostToTradeTransferMsg(
 	unwindIcaAddress := route.RewardAccount.Address
 	tradeIcaAddress := route.TradeAccount.Address
 
+	// Validate ICAs were registered
+	if withdrawlIcaAddress == "" {
+		return msg, errorsmod.Wrapf(types.ErrICAAccountNotFound, "no host account found for %s", route.Description())
+	}
+	if unwindIcaAddress == "" {
```
ethan-stride, line 84: Really weird but possible case: reward unwinding zone has newest pfm installed (so it would work even without an address being given -- any string "", or "pfm" ca be passed in for the address) but they haven't decided to turn on ICA access for some reason so we can't register an account.  We would be stopping here when everything would work in this transfer

ethan-stride (commented):
LGTM

x/stakeibc/keeper/reward_converter.go
sampocs: Hmm that’s a good point. But I guess it’s a question of whether we’d want to sent transfers through hosts without ICA. I think my preference would be to only send transfers to places where we own the intermediate address

PR #1009, opened 2023-12-05
Allow multiple redemptions per epoch

asalzmann:

Currently, and address can only redeem once per epoch. This was simpler to implement, but multiple redemptions per epoch are required to add redemptions to autopilot.

## Brief Changelog
* If a user has already redeemed, add the redemption amount to their `UserRedemptionRecord` (instead of creating a new one)
* If a user has already redeemed, don't append the `UserRedemptionRecord` ID to the `HostZoneUnbonding`

The rest of the logic should work without being changed, but please review this carefully and think about potential ways it could break.

## TODOs
* Integration test
* Unittest
* Think deeply about protocol design


x/stakeibc/keeper/msg_server_redeem_stake.go
```diff
@@ -84,20 +80,36 @@ func (k msgServer) RedeemStake(goCtx context.Context, msg *types.MsgRedeemStake)
...
+			EpochNumber: epochTracker.EpochNumber,
+			// claimIsPending represents whether a redemption is currently being claimed,
+			// contingent on the host zone unbonding having status CLAIMABLE
+			ClaimIsPending: false,
+		}
+		k.Logger(ctx).Info(fmt.Sprintf("UserRedemptionRecord not found - creating for %s", redemptionId))
 	}
+
 	// then add undelegation amount to epoch unbonding records
+	// ASSUMPTION: we don't need to update the EpochUnbondingRecord directly (only the HostZoneUnbonding)
```
sampocs, line 112: is this meant to point out that the HZU's are pointers? Cause we do update the EUR [here](https://github.com/Stride-Labs/stride/pull/1009/files#diff-1d851c7e8c08659458b50e0840e7e3668f32b24ea7d0a4280adb13bba3b70379R153-R158)

x/stakeibc/keeper/msg_server_redeem_stake.go
```diff
@@ -109,7 +121,10 @@ func (k msgServer) RedeemStake(goCtx context.Context, msg *types.MsgRedeemStake)
 		return nil, errorsmod.Wrapf(types.ErrInvalidHostZone, "host zone not found in unbondings: %s", hostZone.ChainId)
 	}
 	hostZoneUnbonding.NativeTokenAmount = hostZoneUnbonding.NativeTokenAmount.Add(nativeAmount)
-	hostZoneUnbonding.UserRedemptionRecords = append(hostZoneUnbonding.UserRedemptionRecords, userRedemptionRecord.Id)
+	if !userHasRedeemedThisEpoch {
+		// Only append if a UserRedemptionRecord to the HZU if it wasn't previously appended
+		hostZoneUnbonding.UserRedemptionRecords = append(hostZoneUnbonding.UserRedemptionRecords, userRedemptionRecord.Id)
```
sampocs, line 126: (obviously out of scope of this PR/upgrade) but we should remove the URRs from the EUR/HUR. Each URR is key'd by chainId/epoch number so if we need to grab the full list for an epoch, we can just do a prefix lookup

x/stakeibc/keeper/msg_server_redeem_stake.go
```diff
@@ -54,6 +49,7 @@ func (k msgServer) RedeemStake(goCtx context.Context, msg *types.MsgRedeemStake)
 
 	// construct desired unstaking amount from host zone
 	stDenom := types.StAssetDenomFromHostZoneDenom(hostZone.HostDenom)
+	// ASSUMPTION (CHECK ME): it doesn't matter that RRs can change _within_ an epoch (due to LSM)
```
sampocs, line 52: I double checked the code base and the only times we use the RR are in liquid stake and redeem. I believe this implies that in other places we just use the total number of tokens, and it wouldn't matter if that collection of tokens was built up from different rates.

Similarly, the only times we modify NativeTokenAmount are in redeem and claim, and the only times we modify StTokenAmount are in redeem and undelegate.

Additionally, in v18, if we end up doing the st → native conversion at the time that the unbonding is initiated (instead of during redemption), we will no longer even have to calculate this here

sampocs (commented):
lgtm! 

waiting to approve until the unit test is in here

x/stakeibc/keeper/msg_server_redeem_stake.go
riley-stride: I think we should always prefer to use the most up-to-date redemption rate because it will more accurately reflect ground truth NativeTokenAmount vs stTokenAmount. 

It's desirable to unstake an amount of tokens commensurate with the most up-to-date RR.

Now that the RR changes _within_ an epoch, that's better than when the RR updated only epochly.

x/stakeibc/keeper/msg_server_redeem_stake.go
```diff
@@ -109,7 +121,10 @@ func (k msgServer) RedeemStake(goCtx context.Context, msg *types.MsgRedeemStake)
 		return nil, errorsmod.Wrapf(types.ErrInvalidHostZone, "host zone not found in unbondings: %s", hostZone.ChainId)
 	}
 	hostZoneUnbonding.NativeTokenAmount = hostZoneUnbonding.NativeTokenAmount.Add(nativeAmount)
-	hostZoneUnbonding.UserRedemptionRecords = append(hostZoneUnbonding.UserRedemptionRecords, userRedemptionRecord.Id)
+	if !userHasRedeemedThisEpoch {
```
riley-stride, line 124: Can you explain how `userHasRedeemedThisEpoch` tells us whether the URR was already appended to the HZU? My understanding is that it tells us whether the URR exists (as it is the _found_ return value from `GetUserRedemptionRecord()`.

riley-stride (commented):
Looking good! Added one question. Waiting until integration and unit tests to approve.

x/stakeibc/keeper/msg_server_redeem_stake.go
```diff
@@ -84,20 +80,36 @@ func (k msgServer) RedeemStake(goCtx context.Context, msg *types.MsgRedeemStake)
...
+	// ----------------- UNBONDING RECORD KEEPING -----------------
+	// Fetch the record
+	senderAddr := sender.String()
+	redemptionId := recordstypes.UserRedemptionRecordKeyFormatter(hostZone.ChainId, epochTracker.EpochNumber, senderAddr)
+	userRedemptionRecord, userHasRedeemedThisEpoch := k.RecordsKeeper.GetUserRedemptionRecord(ctx, redemptionId)
+	if userHasRedeemedThisEpoch {
+		k.Logger(ctx).Info(fmt.Sprintf("UserRedemptionRecord found for %s", redemptionId))
+		// Add the unbonded amount to the UserRedemptionRecord
+		// The record is set below
+		userRedemptionRecord.Amount.Add(nativeAmount)
```
sampocs, line 93: need to set the amount on the record here
```suggestion
		userRedemptionRecord.Amount = userRedemptionRecord.Amount.Add(nativeAmount)
```

sampocs:
Just pushed up a unit test. I also tested manually and it worked as expected

I think adding this to our long standing integrations tests is actually not as straightforward as it seems. But imo, I think the unit test is sufficient coverage

sampocs (changes requested):
One other issue - the record is key'd by just stride address. So if someone redeems twice with the same stride address, but a different receiver address, the second address will get ignored

I think this leaves us with the original attack vector. 

First solution that comes to mind is the change the URR key, but that will require a painful migration. Curious if you guy can think of a better solution

x/stakeibc/keeper/msg_server_redeem_stake.go
shellvish: I think this is using an assumption that when the URR is created, it's always appended to the HZU. Might be good to double check that assumption is valid, but I think it's valid!

x/stakeibc/keeper/msg_server_redeem_stake_test.go
```diff
@@ -104,9 +104,19 @@ func (s *KeeperTestSuite) TestRedeemStake_Successful() {
 	user := tc.user
 	redeemAmount := msg.Amount
 
-	// get the initial unbonding amount *before* calling liquid stake, so we can use it to calc expected vs actual in diff space
-	_, err := s.GetMsgServer().RedeemStake(sdk.WrapSDKContext(s.Ctx), &msg)
-	s.Require().NoError(err)
+	// Split the message amount in 2, and call redeem stake twice (each with half the amount)
+	// This will check that the same user can redeem multiple times
```
shellvish, line 108: nit: should we also change the RR between the two `RedeemStake`s to validate we handle changing RRs correctly?

x/stakeibc/keeper/msg_server_redeem_stake.go
shellvish: One nit comment, the RRs can change _within_ an epoch outside of LSM too, because the URR are keyed daily and the redemption rate changes based on `stride_epochs` (i.e. 6 hours)

shellvish (commented):
Looks great to me! We should add integration tests on top of the unit test, then I'm all signed off 🙂 

Happy to add the integration tests too if that would be helpful!

x/stakeibc/keeper/msg_server_redeem_stake.go
asalzmann: yeah this was just pointing out pointers, removed

x/stakeibc/keeper/msg_server_redeem_stake.go
asalzmann: Good catch!

x/stakeibc/keeper/msg_server_redeem_stake.go
asalzmann: Good point @shellvish. I had thought we were keying on stride_epoch, surprised we're keying on day.

In any case, all of the above comments make sense to me! Thanks for sanity checking this approach.

x/stakeibc/keeper/msg_server_redeem_stake.go
asalzmann:
```
userRedemptionRecord, userHasRedeemedThisEpoch := k.RecordsKeeper.GetUserRedemptionRecord(ctx, redemptionId)
```

`userHasRedeemedThisEpoch` is true if a URR is found for the `redemptionId`, which is a key composed of chain id, epoch number (day), sender addr.

Before a user has redeemed, no record will be found. After a user has redeemed in an epoch (as defined by the key), the URR will be stored and found.

asalzmann: I don't think it has anything to do with the HZU - although @shellvish I might be missing something, can you double check this?

asalzmann:
> Just pushed up a unit test. I also tested manually and it worked as expected

@sampocs is this unittest sufficient, or should I add more?

> One other issue - the record is key'd by just stride address. So if someone redeems twice with the same stride address, but a different receiver address, the second address will get ignored
I think this leaves us with the original attack vector.
First solution that comes to mind is the change the URR key, but that will require a painful migration. Curious if you guy can think of a better solution

I think where we landed was to replace the sender address with the receiver - @sampocs did you happen to take notes on that call?

sampocs:
@asalzmann 

>  is this unittest sufficient, or should I add more?

Imo, it's sufficient - was there something else you thought should be included though?

> I think where we landed was to replace the sender address with the receiver - @sampocs did you happen to take notes on that call?

Yeah we landed on replacing sender with receiver. This is done in #1035 

Notes from that call are at the bottom of the PFM + Autopilot doc and are in the upgrade channel.

x/stakeibc/keeper/msg_server_redeem_stake.go
sampocs: Good callout @shellvish, I forgot they were key'd daily as well! 

Marking this comment resolved cause it appears we're all in agreement!

x/stakeibc/keeper/msg_server_redeem_stake.go
sampocs: I believe Riley is referring to [these lines](https://github.com/Stride-Labs/stride/blob/30ee082fc59c420367fe922bfced8c46fdf8aa2a/x/stakeibc/keeper/msg_server_redeem_stake.go#L124-L127C3) where we append the redemption record _id_ onto the HZU record.

And I agree with Vishal that it's just based on the assumption that we always append the ID to the HZU during the first redemption! I've confirmed this as well - in fact, as far as I can tell, the place this list is actually used is in the HZU query.

x/stakeibc/keeper/msg_server_redeem_stake_test.go
sampocs: decision: use non 1.0 RR, but keep RR consistent between messages

shellvish (approved):
LGTM!

sampocs (approved):

PR #1011, opened 2023-12-05
reward converter - add trade route ID to portId

sampocs:
## Context
We need to have a separate trade ICA for each trade route. Otherwise, if two trade routes shared a common denom, it could interfere with the queries.

## Brief Changelog
* Added `FormatTradeRouteICAOwner` and `FormatTradeRouteICAOwnerFromAccount` 
* Renamed `FormatICAAccountOwner` to `FormatHostZoneICAOwner`
* Updated trade route ICA owners to use the new function
* Added unit tests for the relevant types functions

**NOTE**: @ethan-stride, I'm not really loving the implementation of these owner functions. Curious if you can think of a better way to do it

### DIFF
--- a/app/apptesting/test_helpers.go
+++ b/app/apptesting/test_helpers.go
@@ -231,7 +231,10 @@ func (s *AppTestHelper) CreateICAChannel(owner string) (channelID, portID string
 	_, transferChannelExists := s.App.IBCKeeper.ChannelKeeper.GetChannel(s.Ctx, ibctesting.TransferPort, ibctesting.FirstChannelID)
 	if !transferChannelExists {
 		ownerSplit := strings.Split(owner, ".")
-		s.Require().Equal(2, len(ownerSplit), "owner should be of the form: {HostZone}.{AccountName}")
+		isHostZoneICA := len(ownerSplit) == 2
+		isTradeRouteICA := len(ownerSplit) == 3
+		s.Require().True(isHostZoneICA || isTradeRouteICA,
+			"owner should be either of the form: {chainId}.{AccountName} or {chainId}.{rewarDenom}-{hostDenom}.{accountName}")
 
 		hostChainID := ownerSplit[0]
 		s.CreateTransferChannel(hostChainID)
--- a/x/stakeibc/keeper/msg_server_create_trade_route.go
+++ b/x/stakeibc/keeper/msg_server_create_trade_route.go
@@ -79,17 +79,24 @@ func (ms msgServer) CreateTradeRoute(goCtx context.Context, msg *types.MsgCreate
 	}
 
 	// Register the new ICA accounts
+	tradeRouteId := types.GetTradeRouteId(msg.RewardDenomOnReward, msg.RewardDenomOnHost)
 	hostICA := types.ICAAccount{
 		ChainId:      msg.HostChainId,
 		Type:         types.ICAAccountType_WITHDRAWAL,
 		ConnectionId: hostZone.ConnectionId,
 		Address:      hostZone.WithdrawalIcaAddress,
 	}
-	unwindICA, err := ms.Keeper.RegisterTradeRouteICAAccount(ctx, msg.StrideToRewardConnectionId, types.ICAAccountType_CONVERTER_UNWIND)
+
+	unwindConnectionId := msg.StrideToRewardConnectionId
+	unwindICAType := types.ICAAccountType_CONVERTER_UNWIND
+	unwindICA, err := ms.Keeper.RegisterTradeRouteICAAccount(ctx, tradeRouteId, unwindConnectionId, unwindICAType)
 	if err != nil {
 		return nil, errorsmod.Wrapf(err, "unable to register the unwind ICA account")
 	}
-	tradeICA, err := ms.Keeper.RegisterTradeRouteICAAccount(ctx, msg.StrideToTradeConnectionId, types.ICAAccountType_CONVERTER_TRADE)
+
+	tradeConnectionId := msg.StrideToTradeConnectionId
+	tradeICAType := types.ICAAccountType_CONVERTER_TRADE
+	tradeICA, err := ms.Keeper.RegisterTradeRouteICAAccount(ctx, tradeRouteId, tradeConnectionId, tradeICAType)
 	if err != nil {
 		return nil, errorsmod.Wrapf(err, "unable to register the trade ICA account")
 	}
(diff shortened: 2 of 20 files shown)


app/apptesting/test_helpers.go
```diff
@@ -231,7 +231,10 @@ func (s *AppTestHelper) CreateICAChannel(owner string) (channelID, portID string
 	_, transferChannelExists := s.App.IBCKeeper.ChannelKeeper.GetChannel(s.Ctx, ibctesting.TransferPort, ibctesting.FirstChannelID)
 	if !transferChannelExists {
 		ownerSplit := strings.Split(owner, ".")
-		s.Require().Equal(2, len(ownerSplit), "owner should be of the form: {HostZone}.{AccountName}")
+		isHostZoneICA := len(ownerSplit) == 2
+		isTradeRouteICA := len(ownerSplit) == 3
+		s.Require().True(isHostZoneICA || isTradeRouteICA,
+			"owner should be either of the form: {chainId}.{AccountName} or {chainId}.{rewarDenom}-{hostDenom}.{accountName}")
```
ethan-stride, line 237: hmm, at first I thought the tradeRoute one seems redundant but chainId would be the tradeZone chain Id right, so then it could be independent of the rewardDenom and hostDenom so all of that information really is needed.  My second thought was that for consistency we could use the chainIds for the reward Zone and hostZone but then I realized this is for convenience of being able to look up the TradeRoute from keeper.  So everything makes sense and is in fact needed even though it seemed like it should be simplified at first glance

x/stakeibc/keeper/msg_server_create_trade_route.go
```diff
@@ -79,17 +79,24 @@ func (ms msgServer) CreateTradeRoute(goCtx context.Context, msg *types.MsgCreate
 	}
 
 	// Register the new ICA accounts
+	tradeRouteId := types.GetTradeRouteId(msg.RewardDenomOnReward, msg.RewardDenomOnHost)
```
ethan-stride, line 82: Second input should be HostDenomOnHost not RewardDenomOnHost

x/stakeibc/types/ica_account.go
```diff
@@ -1,5 +1,21 @@
...
+import fmt "fmt"
+
+// Helper function to build the host zone ICA owner in the form "{chainId}.{ICA_TYPE}"
+func FormatHostZoneICAOwner(chainId string, accountType ICAAccountType) (result string) {
 	return chainId + "." + accountType.String()
 }
+
+// Helper function to build the ICA owner for a trade route ICA
+// in the form "{chainId}.{rewardDenom}-{hostDenom}.{ICA_TYPE}"
+func FormatTradeRouteICAOwner(chainId, rewardDenom, hostDenom string, icaAccountType ICAAccountType) string {
```
ethan-stride, line 12: nit: add string as type for first two params for consistency

x/stakeibc/types/keys.go
```diff
@@ -48,8 +48,8 @@ func EpochTrackerKey(epochIdentifier string) []byte {
 }
 
 // Definition for the store key format based on tradeRoute start and end denoms
-func TradeRouteKeyFromDenoms(startDenom string, endDenom string) (key []byte) {
-	return []byte(startDenom + "-" + endDenom)
+func TradeRouteKeyFromDenoms(rewardDenom, hostDenom string) (key []byte) {
```
ethan-stride, line 51: nit: add string type to first param for consistency

ethan-stride (commented):
LGTM
Lots of places I would think, surely we can combine these two functions, or wow that seems wordy or there has to be a way to make the function take fewer inputs.... and each time when I would look at the minimum information needed to accomplish the goal you would be right.  It is well done but it doesn't feel elegant because this particular piece of code can't elegant.  Having ICAs which don't exist on a hostZone struct are just going to carry around a lot of extra fields no matter what, I think this is about the best we can do for the inherent complexity of this part of the system

x/stakeibc/keeper/msg_server_create_trade_route.go
sampocs: good catch!!

x/stakeibc/types/ica_account.go
sampocs: this was done intentionally - it's a common practice in go when you have params of the same type

x/stakeibc/types/keys.go
sampocs: same as above

sampocs:
@ethan-stride what are you thoughts on the `FormatTradeRouteICAOwnerFromAccount`? Does it make sense to pass the full account or instead do
```go
func FormatTradeRouteICAOwnerFromAccount(chainId, tradeRouteId string, icaAccountType ICAAccountType) string {
	return fmt.Sprintf("%s.%s.%s", chainId, tradeRouteId, icaAccountType.String())
}
```

it would muddy the invocation a bit
```go
// from 
owner := types.FormatTradeRouteICAOwnerFromAccount(route.GetRouteId(), route.TradeAccount)
// to
tradeAccount := route.TradeAccount
owner := types.FormatTradeRouteICAOwnerFromAccount(tradeAccount.ChainId, route.GetRouteId(), tradeAccount.Type)
```
But it would be a little more explicit about which fields are used instead of having to know that the ICAAccount struct needs both chainId and type filled out

x/stakeibc/keeper/msg_server_create_trade_route.go
sampocs: done

x/stakeibc/types/ica_account.go
ethan-stride: ok, not my favorite style thing but I see how it works

ethan-stride:
> @ethan-stride what are you thoughts on the `FormatTradeRouteICAOwnerFromAccount`? Does it make sense to pass the full account or instead do
> 
> ```go
> func FormatTradeRouteICAOwnerFromAccount(chainId, tradeRouteId string, icaAccountType ICAAccountType) string {
> 	return fmt.Sprintf("%s.%s.%s", chainId, tradeRouteId, icaAccountType.String())
> }
> ```
> 
> it would muddy the invocation a bit
> 
> ```go
> // from 
> owner := types.FormatTradeRouteICAOwnerFromAccount(route.GetRouteId(), route.TradeAccount)
> // to
> tradeAccount := route.TradeAccount
> owner := types.FormatTradeRouteICAOwnerFromAccount(tradeAccount.ChainId, route.GetRouteId(), tradeAccount.Type)
> ```
> 
> But it would be a little more explicit about which fields are used instead of having to know that the ICAAccount struct needs both chainId and type filled out

From discussion on Slack:  being explicit about which fields from the account are going into the function call makes it harder to mistakenly pass in an account which is not yet initialized without realizing it.  This makes the function clearer but calling it slightly more work -- overall a tradeoff which is probably worth it so the new proposed function params make sense.

ethan-stride:
LGTM I think it is ready

PR #1012, opened 2023-12-05
add RedeemStake to autopilot

asalzmann:
Closes: #771

Add `RedeemStake` to autopilot.

## Brief Changelog

* Add integration test for `RedeemStake`
* Add unittest for `RedeemStake`
* Integrate with autopilot module


Vvaradinov (approved):
LGTM! 👍🏻  Excited to test this with the outpost.

dockernet/tests/integration_tests.bats
```diff
@@ -152,6 +152,25 @@ setup_file() {
...
+
+#   # make sure stATOM balance decreased
+#   sttoken_balance_end=$($STRIDE_MAIN_CMD q bank balances $(STRIDE_ADDRESS) --denom st$HOST_DENOM | GETBAL)
+#   stibctoken_balance_end=$($HOST_MAIN_CMD q bank balances $HOST_VAL_ADDRESS --denom $IBC_GAIA_STATOM_DENOM | GETBAL)
+#   sttoken_balance_diff=$(($sttoken_balance_start-$sttoken_balance_end))
+#   stibctoken_balance_diff=$(($stibctoken_balance_end-$stibctoken_balance_start))
+#   assert_equal "$sttoken_balance_diff" "$PACKET_FORWARD_STAKE_AMOUNT"
+#   assert_equal "$stibctoken_balance_diff" "$PACKET_FORWARD_STAKE_AMOUNT"
+# }
+
```
sampocs, line 173: Yeah I think you can remove this. It's functionally the same as the first IG test

dockernet/tests/integration_tests.bats
```diff
@@ -180,6 +199,42 @@ setup_file() {
   assert_equal "$stibctoken_balance_diff" "$PACKET_FORWARD_STAKE_AMOUNT"
 }
 
+@test "[INTEGRATION-BASIC-$CHAIN_NAME] packet forwarding automatically redeem stake" {
+  skip "DefaultActive set to false, skip test"
```
sampocs, line 203: is this test still supposed to be inactive?

dockernet/tests/integration_tests.bats
```diff
@@ -180,6 +199,42 @@ setup_file() {
...
+  stibctoken_balance_end=$($HOST_MAIN_CMD q bank balances $HOST_VAL_ADDRESS --denom $IBC_GAIA_STATOM_DENOM | GETBAL)
+  stibctoken_balance_diff=$(($stibctoken_balance_start-$stibctoken_balance_end))
+  assert_equal "$stibctoken_balance_diff" "200"
+
+  WAIT_FOR_BLOCK $STRIDE_LOGS 5
+
+  # check that the tokens were transferred to the redemption account
+  AMOUNT=$($STRIDE_MAIN_CMD q records list-user-redemption-record  | grep -Fiw 'amount' | head -n 1 | grep -o -E '[0-9]+')
+  amount_positive=$(($AMOUNT > 0))
+  assert_equal "$amount_positive" "1"
```
sampocs, line 234: just double checking - is this the only redemption record at this point in the tests?

x/stakeibc/types/message_liquid_stake.go
```diff
@@ -32,6 +32,14 @@ func StAssetDenomFromHostZoneDenom(hostZoneDenom string) string {
...
 }
 
+func IsStAssetDenom(stAssetDenom string) bool {
+	return strings.HasPrefix(stAssetDenom, "st")
+}
+
+func HostZoneDenomFromStAssetDenom(stAssetDenom string) string {
+	return stAssetDenom[2:]
+}
+
```
sampocs, line 42: nit: this feels like a weird place to put these helpers. Might make more sense to put them in `types/host_zone.go`

x/autopilot/keeper/redeem_stake.go
```diff
@@ -0,0 +1,91 @@
...
+
+	"github.com/Stride-Labs/stride/v16/x/autopilot/types"
+	stakeibckeeper "github.com/Stride-Labs/stride/v16/x/stakeibc/keeper"
+	stakeibctypes "github.com/Stride-Labs/stride/v16/x/stakeibc/types"
+)
+
+func (k Keeper) TryRedeemStake(
+	ctx sdk.Context,
+	packet channeltypes.Packet,
+	newData transfertypes.FungibleTokenPacketData,
```
sampocs, line 21: nit: newData is kind of a confusing variable name here. `packetData` or `tokenPacketData` would probably be better

x/autopilot/keeper/redeem_stake.go
```diff
@@ -0,0 +1,91 @@
...
+		return err
+	}
+
+	msgServer := stakeibckeeper.NewMsgServerImpl(k.stakeibcKeeper)
+	_, err = msgServer.RedeemStake(
+		sdk.WrapSDKContext(ctx),
+		msg,
+	)
+	if err != nil {
+		return errorsmod.Wrapf(sdkerrors.ErrInsufficientFunds, err.Error())
```
sampocs, line 88:
```suggestion
		return err
```

x/autopilot/keeper/redeem_stake.go
```diff
@@ -0,0 +1,91 @@
...
+
+	strideAddress, err := sdk.AccAddressFromBech32(newData.Receiver)
+	if err != nil {
+		return errorsmod.Wrapf(sdkerrors.ErrInvalidAddress, "invalid receiver (%s) in autopilot memo", strideAddress)
+	}
+
+	return k.RunRedeemStake(ctx, strideAddress, packetMetadata.IbcReceiver, hostZoneDenom, token, []metrics.Label{})
+}
+
+func (k Keeper) RunRedeemStake(ctx sdk.Context, addr sdk.AccAddress, receiver string, hostZoneDenom string, token sdk.Coin, labels []metrics.Label) error {
```
sampocs, line 65: can this labels argument be removed?

x/autopilot/keeper/redeem_stake.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+
+	if err := token.Validate(); err != nil {
+		return err
+	}
+
+	strideAddress, err := sdk.AccAddressFromBech32(newData.Receiver)
+	if err != nil {
+		return errorsmod.Wrapf(sdkerrors.ErrInvalidAddress, "invalid receiver (%s) in autopilot memo", strideAddress)
+	}
```
sampocs, line 60: Isn't this address an address on the host zone?

sampocs: if so, rename the variable, but we should also make sure the conversion to AccAddress works. I assume it will considering it's just going to bytes - but it's also possible that it needs the bech prefix

sampocs (commented):
looks good! mostly nit's

So much easier to reason about without the transfer back haha

I'll review the unit tests tomorrow!

x/autopilot/keeper/redeem_stake_test.go
```diff
@@ -0,0 +1,253 @@
...
+	stakeibckeeper "github.com/Stride-Labs/stride/v16/x/stakeibc/keeper"
+	stakeibctypes "github.com/Stride-Labs/stride/v16/x/stakeibc/types"
+)
+
+func getRedeemStakeStakeibcPacketMetadata(address, ibcReceiver, transferChannel string) string {
+	return fmt.Sprintf(`
+		{
+			"autopilot": {
+				"receiver": "%[1]s",
+				"stakeibc": { "action": "RedeemStake", "ibc_receiver": "%[2]s", "transfer_channel": "%[3]s" }
```
sampocs, line 31: the transfer channel isn't used during redemptions right? Similarly, should we rename the ibc_receiver field?

x/autopilot/keeper/redeem_stake_test.go
```diff
@@ -0,0 +1,253 @@
...
+
+	strdDenom := "ustrd"
+	prefixedDenom = transfertypes.GetPrefixedDenom(packet.GetSourcePort(), packet.GetSourceChannel(), strdDenom)
+	strdIbcDenom := transfertypes.ParseDenomTrace(prefixedDenom).IBCDenom()
+
+	stAtomDenom := "stuatom"
+	prefixedDenom = transfertypes.GetPrefixedDenom(packet.GetSourcePort(), packet.GetSourceChannel(), stAtomDenom)
+	stAtomFullDenomPath := transfertypes.ParseDenomTrace(prefixedDenom).GetFullDenomPath()
+
+	addr1 := sdk.AccAddress(ed25519.GenPrivKey().PubKey().Address().Bytes())
```
sampocs, line 62: nit:
```suggestion
	addr1 := suite.TestAccs[0]
```

x/autopilot/keeper/redeem_stake_test.go
```diff
@@ -0,0 +1,253 @@
...
+		packetData       transfertypes.FungibleTokenPacketData
+		expSuccess       bool
+		expRedeemStake   bool
+	}{
+		{ // params not enabled
+			forwardingActive: false,
+			packetData: transfertypes.FungibleTokenPacketData{
+				Denom:    stAtomFullDenomPath,
+				Amount:   "1000000",
+				Sender:   "cosmos16plylpsgxechajltx9yeseqexzdzut9g8vla4k",
```
sampocs, line 75: nit: this address is used a bunch, you can pull it out into a variable

x/autopilot/keeper/redeem_stake_test.go
```diff
@@ -0,0 +1,253 @@
...
+	stAtomDenom := "stuatom"
+	prefixedDenom = transfertypes.GetPrefixedDenom(packet.GetSourcePort(), packet.GetSourceChannel(), stAtomDenom)
+	stAtomFullDenomPath := transfertypes.ParseDenomTrace(prefixedDenom).GetFullDenomPath()
+
+	addr1 := sdk.AccAddress(ed25519.GenPrivKey().PubKey().Address().Bytes())
+	testCases := []struct {
+		forwardingActive bool
+		recvDenom        string
+		packetData       transfertypes.FungibleTokenPacketData
+		expSuccess       bool
```
sampocs, line 67: imo, it's a bit more bullet proof to use an expectedError string here instead - prevents possible false positive errors from an inadequate test setup

x/autopilot/keeper/redeem_stake_test.go
```diff
@@ -0,0 +1,253 @@
...
+			expSuccess:     false,
+			expRedeemStake: false,
+		},
+	}
+
+	for i, tc := range testCases {
+		suite.Run(fmt.Sprintf("Case %d", i), func() {
+			packet.Data = transfertypes.ModuleCdc.MustMarshalJSON(&tc.packetData)
+
+			suite.SetupTest() // reset
```
sampocs, line 154: nit: I don't think this is required - should happen automatically when using `suite.Run`

x/autopilot/keeper/redeem_stake_test.go
```diff
@@ -0,0 +1,253 @@
...
+				WithdrawalAccount:     nil,
+				FeeAccount:            nil,
+				DelegationAccount:     nil,
+				RedemptionAccount:     nil,
+				IbcDenom:              atomIbcDenom,
+				HostDenom:             atomHostDenom,
+				RedemptionRate:        sdk.NewDec(1),
+				Address:               addr1.String(),
+				StakedBal:             sdk.NewInt(1000000),
+			})
```
sampocs, line 212: I think most of this state is not required for this test. I'm happy to take a pass at slimming it down if you want!

x/autopilot/keeper/redeem_stake_test.go
```diff
@@ -0,0 +1,253 @@
...
+
+			// mint coins to be spent on liquid staking
+			coins := sdk.Coins{sdk.NewInt64Coin(atomIbcDenom, 1000000)}
+			err := suite.App.BankKeeper.MintCoins(ctx, minttypes.ModuleName, coins)
+			suite.Require().NoError(err)
+			err = suite.App.BankKeeper.SendCoinsFromModuleToAccount(ctx, minttypes.ModuleName, addr1, coins)
+			suite.Require().NoError(err)
```
sampocs, line 219: what's this for?

x/autopilot/keeper/redeem_stake_test.go
```diff
@@ -0,0 +1,253 @@
...
+			msgServer := stakeibckeeper.NewMsgServerImpl(suite.App.StakeibcKeeper)
+			msg := stakeibctypes.NewMsgLiquidStake(addr1.String(), sdk.NewInt(1000000), atomHostDenom)
+			_, err = msgServer.LiquidStake(sdk.WrapSDKContext(suite.Ctx), msg)
+			suite.Require().NoError(err)
+
+			// send tokens to ibc transfer channel escrow address
+			escrowAddr := transfertypes.GetEscrowAddress(packet.DestinationPort, packet.DestinationChannel)
+			err = suite.App.BankKeeper.SendCoins(suite.Ctx, addr1, escrowAddr, sdk.Coins{sdk.NewInt64Coin(stAtomDenom, 1000000)})
+			suite.Require().NoError(err)
+			suite.App.TransferKeeper.SetTotalEscrowForDenom(suite.Ctx, sdk.NewInt64Coin(stAtomDenom, 1000000))
```
sampocs, line 231: is this required so the sttoken can be properly received

sampocs (commented):
Coverage on the tests looks good! But they are a bit bloated - I can take a pass at slimming them down if it'd help!

x/autopilot/keeper/redeem_stake.go
sampocs: disregard - I see that this actually is the stride address

x/autopilot/keeper/redeem_stake.go
sampocs: resolved in #1030

x/autopilot/keeper/redeem_stake.go
sampocs: resolved in #1030

x/autopilot/keeper/redeem_stake.go
sampocs: resolved in #1030

x/autopilot/keeper/redeem_stake_test.go
sampocs: resolved in #1030

x/autopilot/keeper/redeem_stake_test.go
sampocs: resolved in #1030

x/autopilot/keeper/redeem_stake_test.go
sampocs: resolved in #1030

x/autopilot/keeper/redeem_stake_test.go
sampocs: resolved in #1030

x/autopilot/keeper/redeem_stake_test.go
sampocs: resolved in #1030

x/autopilot/keeper/redeem_stake_test.go
sampocs: resolved in #1030

x/autopilot/keeper/redeem_stake_test.go
sampocs: wasn't necessary, resolved in #1030

dockernet/tests/integration_tests.bats
sampocs: resolved in #1031 (it was not needed)

dockernet/tests/integration_tests.bats
sampocs: re-activated it in #1031

x/autopilot/keeper/redeem_stake.go
shellvish, line 88: nit: the error here might be separate from `ErrInsufficientFunds` (e.g. receiver field is improperly formatted), but I think. Sam's suggestion fixes this!

x/autopilot/keeper/redeem_stake_test.go
```diff
@@ -0,0 +1,253 @@
...
+			expSuccess:     false,
+			expRedeemStake: false,
+		},
+		{ // invalid redeem receiver
+			forwardingActive: true,
+			packetData: transfertypes.FungibleTokenPacketData{
+				Denom:    stAtomFullDenomPath,
+				Amount:   "1000000",
+				Sender:   "cosmos16plylpsgxechajltx9yeseqexzdzut9g8vla4k",
+				Receiver: getRedeemStakeStakeibcPacketMetadata(addr1.String(), "xxx", ""),
```
shellvish, line 141: nit: do you think it's worth making the "xxx" here the properly formatted address for another chain (e.g. `osmo16plylpsgxechajltx9yeseqexzdzut9g0hvdry`)?

x/autopilot/module_ibc.go
```diff
@@ -203,7 +203,12 @@ func (im IBCModule) OnRecvPacket(
 				im.keeper.Logger(ctx).Error(fmt.Sprintf("Error liquid staking packet from autopilot for %s: %s", newData.Sender, err.Error()))
 				return channeltypes.NewErrorAcknowledgement(err)
 			}
-			// case types.RedeemStake: TODO: add redeem stake logic
+		case types.RedeemStake:
+			// Try to redeem stake - return an ack error if it fails, otherwise return the ack generated from the earlier packet propogation
+			if err := im.keeper.TryRedeemStake(ctx, packet, newData, routingInfo); err != nil {
```
shellvish, line 208: just checking - if this fails, do the tokens end up the receiver address on Stride? If so, it might be worth adding an integration test checking this (just to triple-check we avoid the worst-case scenario of this failing and the tokens getting burned)

shellvish (approved):
Looks great to me! A couple of nit suggestions, but overall very clean flow 🙂

x/autopilot/keeper/redeem_stake.go
sampocs: Agreed - I addressed this [here](https://github.com/Stride-Labs/stride/pull/1030/files#diff-6badb27c023a93b231bdecf85015f77643e5defa71a492747225acaf41d810f5R75-R76)!

x/autopilot/keeper/redeem_stake_test.go
sampocs: good call! I'm going to add a test case to the unit test branch

sampocs: added [here](https://github.com/Stride-Labs/stride/pull/1030/commits/d71a9407501c9994cf34dbabaf12f3ceb4e630ce)!

dockernet/tests/integration_tests.bats
```diff
@@ -303,8 +281,72 @@ setup_file() {
   assert_equal "$sttoken_balance_diff" "$PACKET_FORWARD_STAKE_AMOUNT"
 }
 
+@test "[INTEGRATION-BASIC-$CHAIN_NAME] autopilot liquid stake and transfer" {
```
riley-stride, line 284: Do you think we need to integration test redeeming twice from the same account in the same day epoch?

dockernet/tests/integration_tests.bats
```diff
@@ -303,8 +281,72 @@ setup_file() {
...
+  WAIT_FOR_BALANCE_CHANGE $CHAIN_NAME $HOST_VAL_ADDRESS $IBC_STTOKEN 
+
+  # make sure stATOM balance decreased
+  stibctoken_balance_end=$($HOST_MAIN_CMD q bank balances $HOST_VAL_ADDRESS --denom $IBC_STTOKEN | GETBAL)
+  stibctoken_balance_diff=$(($stibctoken_balance_start-$stibctoken_balance_end))
+  assert_equal "$stibctoken_balance_diff" "$redeem_amount"
+
+  WAIT_FOR_BLOCK $STRIDE_LOGS 5
+
+  # check that the tokens were transferred to the redemption account
```
riley-stride, line 342: Does this test check for moving tokens, or for creating the URR?

x/autopilot/module_ibc.go
riley-stride: +1 to Vishal's comment, afaiu this would only require adding a single check to the existing integration test

x/autopilot/keeper/redeem_stake_test.go
```diff
@@ -0,0 +1,430 @@
...
+	s.Require().True(found)
+	s.Require().True(len(hostZoneUnbonding.UserRedemptionRecords) > 0,
+		"user redemption record should have been created")
+
+	// check that tokens were escrowed
+	escrowBalance := s.App.BankKeeper.GetBalance(s.Ctx, depositAddress, redeemDenom)
+	s.Require().Equal(redeemAmount.Int64(), escrowBalance.Amount.Int64(), "tokens should have been escrowed")
+}
+
+func (s *KeeperTestSuite) TestTryRedeemStake() {
```
riley-stride, line 87: Should we add a test that tries to redeem twice by the same user in the same day? My sense is we want high confidence that case will work, and might be best to test it with a unit test vs integration test.

x/stakeibc/types/message_liquid_stake.go
```diff
@@ -32,6 +32,14 @@ func StAssetDenomFromHostZoneDenom(hostZoneDenom string) string {
 	return "st" + hostZoneDenom
 }
 
+func IsStAssetDenom(stAssetDenom string) bool {
```
riley-stride, line 35: Afaiu this check could fail with assets like `stride` where the prefix is `st`. If we ever have a hz with denom `stx` I think the current RedeemStake flow would allow those holding native `stx` tokens to redeem them for tokens `x` because `IsStAssetDenom` would pass, then `stx` would get converted to `x` by `HostZoneDenomFromStAssetDenom`. There might be logic in the middleware that would catch this and prevent this issue, but we should probably just fix it here at `IsStAssetDenom` by iterating the existing stTokens rather than checking that the first two chars in the string are "st".

riley-stride (commented):
Looking very good! Added some light comments, main ones are potential issues with `IsStAssetDenom` and `HostZoneDenomFromStAssetDenom`, and testing the case where you redeem twice in one day.

x/stakeibc/types/message_liquid_stake.go
sampocs: We actually already have a `CheckIsStToken` that does this exactly. Going to replace `IsStAssetDenom` with that

x/autopilot/keeper/redeem_stake_test.go
sampocs: There's a unit test for this in the PR that allows multiple redemptions per epoch! [[ref](https://github.com/Stride-Labs/stride/blob/ef7d0776a1ff04abff686a8de9c5758627a75100/x/stakeibc/keeper/msg_server_redeem_stake_test.go#L107-L113)]

dockernet/tests/integration_tests.bats
sampocs: good eye! I think this is a stale comment, resolved [[here](https://github.com/Stride-Labs/stride/pull/1012/commits/b928c91def434bceff0e953a2dc92da25860c13f)]

dockernet/tests/integration_tests.bats
sampocs: imho, I think this type of test case granularity is best suited in a unit test (more specifically the redeem stake unit test).

That said, we do run two redeem stakes in a row in these IG tests (1 from autopilot and 1 as a regular redeem), so it probably doesn't hurt to tweak it a bit to make sure they occur in the same epoch!

sampocs: done [[here](https://github.com/Stride-Labs/stride/pull/1012/commits/f63a6a91c1ef850f6c63bdbfb928e7e52ccdc2c5)]

x/stakeibc/types/message_liquid_stake.go
asalzmann: agreed, moved

x/autopilot/module_ibc.go
asalzmann: @sampocs it looks like this is the only comment that hasn't been addressed (let me know if that's not the case, and thank you for picking up the rest 🙏 )

should I update the integration test for this?

sampocs: > if this fails, do the tokens end up the receiver address on Stride?

If this fails, we get an ack error and the whole inbound transfer fails, sending the tokens back to the host zone.

I can update the integration test since you're hopping offline rn!

x/stakeibc/types/message_liquid_stake.go
sampocs: you may have forgotten to push up this commit?

sampocs: just moved, resolving

riley-stride (approved):
The new integration test looks great, thanks for adding that!

PR #1013, opened 2023-12-05
Unit tests for calibration ICQ callback

sampocs:
## Brief Changelog
* **Raised threshold to 5000**
* Changed evaluation of threshold so that 5000 is accepted instead of rejected
* Removed code that unmarshalled callback data (no longer needed)
* Updated unit tests
  * Simplified setup to only required state
  * Added more coverage around the shares to tokens calculation 
  * Added error case test coverage


x/stakeibc/keeper/icqcallbacks_callibrate_delegation.go
```diff
@@ -45,12 +44,6 @@ func CalibrateDelegationCallback(k Keeper, ctx sdk.Context, args []byte, query i
 	k.Logger(ctx).Info(utils.LogICQCallbackWithHostZone(chainId, ICQCallbackID_Calibrate, "Query response - Delegator: %s, Validator: %s, Shares: %v",
 		queriedDelegation.DelegatorAddress, queriedDelegation.ValidatorAddress, queriedDelegation.Shares))
 
-	// Unmarshal the callback data containing the previous delegation to the validator (from the time the query was submitted)
```
riley-stride, line 48: good catch

PR #1015, opened 2023-12-05
reward converter - Create/Update/Delete TradeRoute unit tests 

sampocs:
## Brief Changelog
* Added unit tests for create, update, and delete trade routes
* Update MockClientsAndConnections helper so that it has enough state to register ICAs
* Added helper function to mock out an ICA address (useful if you need to do more than one ICA)

### DIFF
--- a/app/apptesting/test_helpers.go
+++ b/app/apptesting/test_helpers.go
@@ -446,16 +446,42 @@ func (s *AppTestHelper) MockClientLatestHeight(height uint64) {
 
 // Helper function to mock out a client and connection to test
 // mapping from connection ID back to chain ID
+// This also mocks out the consensus state to enable testing registering interchain accounts
 func (s *AppTestHelper) MockClientAndConnection(chainId, clientId, connectionId string) {
+	clientHeight := clienttypes.Height{
+		RevisionHeight: uint64(s.Ctx.BlockHeight()),
+	}
 	clientState := tendermint.ClientState{
-		ChainId: chainId,
+		ChainId:        chainId,
+		LatestHeight:   clientHeight,
+		TrustingPeriod: time.Minute * 10,
 	}
 	s.App.IBCKeeper.ClientKeeper.SetClientState(s.Ctx, clientId, &clientState)
 
+	consensusState := tendermint.ConsensusState{
+		Timestamp: s.Ctx.BlockTime(),
+	}
+	s.App.IBCKeeper.ClientKeeper.SetClientConsensusState(s.Ctx, clientId, clientHeight, &consensusState)
+
 	connection := connectiontypes.ConnectionEnd{
 		ClientId: clientId,
+		Versions: []*connectiontypes.Version{connectiontypes.DefaultIBCVersion},
 	}
 	s.App.IBCKeeper.ConnectionKeeper.SetConnection(s.Ctx, connectionId, connection)
+}
+
+// Helper function to mock out an ICA address
+func (s *AppTestHelper) MockICAChannel(connectionId, channelId, owner, address string) {
+	// Create an open channel with the ICA port
+	portId, _ := icatypes.NewControllerPortID(owner)
+	channel := channeltypes.Channel{
+		State: channeltypes.OPEN,
+	}
+	s.App.IBCKeeper.ChannelKeeper.SetChannel(s.Ctx, portId, channelId, channel)
+
+	// Then set the address and make the channel active
+	s.App.ICAControllerKeeper.SetInterchainAccountAddress(s.Ctx, connectionId, portId, address)
+	s.App.ICAControllerKeeper.SetActiveChannelID(s.Ctx, connectionId, portId, channelId)
 }
 
 func (s *AppTestHelper) ConfirmUpgradeSucceededs(upgradeName string, upgradeHeight int64) {
--- a/x/stakeibc/keeper/msg_server_create_trade_route_test.go
+++ b/x/stakeibc/keeper/msg_server_create_trade_route_test.go
@@ -1,0 +1,127 @@
+package keeper_test
+
+import (
+	sdkmath "cosmossdk.io/math"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+
+	"github.com/Stride-Labs/stride/v16/x/stakeibc/types"
+)
+
+func (s *KeeperTestSuite) TestMsgCreateTradeRoute() {
+	rewardChainId := "reward-0"
+	tradeChainId := "trade-0"
+
+	hostConnectionId := "connection-0"
+	rewardConnectionId := "connection-1"
+	tradeConnectionId := "connection-2"
+
+	hostToRewardChannelId := "channel-100"
+	rewardToTradeChannelId := "channel-200"
+	tradeToHostChannelId := "channel-300"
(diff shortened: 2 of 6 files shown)


x/stakeibc/keeper/msg_server_create_trade_route_test.go
```diff
@@ -0,0 +1,231 @@
...
+	s.MockClientAndConnection(rewardChainId, "07-tendermint-0", rewardConnectionId)
+	s.MockClientAndConnection(tradeChainId, "07-tendermint-1", tradeConnectionId)
+
+	// Register an exisiting ICA account for the unwind ICA to test that
+	// existing accounts are re-used
+	// TODO [DYDX]: Replace with trade route owner
+	owner := types.FormatICAAccountOwner(rewardChainId, types.ICAAccountType_CONVERTER_UNWIND)
+	s.MockICAChannel(rewardConnectionId, "channel-0", owner, unwindAddress)
+
+	// Create a host zone with an exisiting withdrawal address
```
ethan-stride, line 49: The address exists but we never called MockICAChannel on the withdrawal address so we can't send messages to it yet

ethan-stride: Ok I understand better now, this function is just trying to setup the new trade route proposal and the expected outcome, so we don't need any of these connections to actually be real beyond generating the owner and ports as expected inside the setup.  The withdrawal account will be setup from the hostZone while the unwind and trade will be setup with these mock helpers.

x/stakeibc/keeper/msg_server_create_trade_route_test.go
```diff
@@ -0,0 +1,231 @@
...
+
+	return msg, expectedTradeRoute
+}
+
+// Helper function to create a trade route and check the created route matched expectations
+func (s *KeeperTestSuite) submitCreateTradeRouteAndValidate(msg types.MsgCreateTradeRoute, expectedRoute types.TradeRoute) {
+	_, err := s.GetMsgServer().CreateTradeRoute(sdk.WrapSDKContext(s.Ctx), &msg)
+	s.Require().NoError(err, "no error expected when creating trade route")
+
+	actualRoute, found := s.App.StakeibcKeeper.GetTradeRoute(s.Ctx, RewardDenom, HostDenom)
```
ethan-stride, line 130: Should you be loading these denoms off the expectedRoute rather than having them hard coded as globvars from the above setup?

x/stakeibc/keeper/msg_server_create_trade_route_test.go
```diff
@@ -0,0 +1,231 @@
...
+}
+
+// Tests creating a trade route where the ICA registration step fails
+func (s *KeeperTestSuite) TestCreateTradeRoute_Failure_UnableToRegisterICA() {
+	msg, expectedRoute := s.SetupTestCreateTradeRoute()
+
+	// Disable ICA middleware for the trade channel so the ICA fails
+	// TODO [DYDX]: Replace with new formatter
+	tradeAccount := expectedRoute.TradeAccount
+	tradeOwner := types.FormatICAAccountOwner(tradeAccount.ChainId, types.ICAAccountType_CONVERTER_TRADE)
```
ethan-stride, line 225: So this is purposefully the wrong ICAAccountOwner setup for the CONVERTER_TRADE type?

x/stakeibc/keeper/msg_server_create_trade_route_test.go
```diff
@@ -0,0 +1,231 @@
...
+	// Remove the host zone completely and check that that also fails
+	s.App.StakeibcKeeper.RemoveHostZone(s.Ctx, HostChainId)
+
+	_, err = s.GetMsgServer().CreateTradeRoute(sdk.WrapSDKContext(s.Ctx), &msg)
+	s.Require().ErrorContains(err, "host zone not found")
+}
+
+// Tests creating a trade route where the ICA channels cannot be created
+// because the ICA connections do not exist
+func (s *KeeperTestSuite) TestCreateTradeRoute_Failure_ConnectionNotFound() {
```
ethan-stride, line 201: We assume that the withdrawal connection is gotten from the HostZone and always is valid at the time we are making a tradeRoute

x/stakeibc/keeper/msg_server_update_trade_route_test.go
```diff
@@ -0,0 +1,86 @@
...
+	_, err := s.GetMsgServer().UpdateTradeRoute(sdk.WrapSDKContext(s.Ctx), &invalidMsg)
+	s.Require().ErrorContains(err, "invalid authority")
+
+	// Test that an error is thrown if the route doesn't exist
+	invalidMsg = msg
+	invalidMsg.RewardDenom = "invalid-reward-denom"
+
+	_, err = s.GetMsgServer().UpdateTradeRoute(sdk.WrapSDKContext(s.Ctx), &invalidMsg)
+	s.Require().ErrorContains(err, "trade route not found")
+}
```
ethan-stride, line 86: Since this is specifically a test for update, what about if the passed in values are out of bounds like Allowed swap loss rate being outside the 0-1 range etc.  Or minSwapAmount is negative. Or what if the maxSwapAmount is a string with a number, but the number is actually too large for the type?

ethan-stride (commented):
LGTM

app/apptesting/test_helpers.go
```diff
@@ -446,18 +446,44 @@ func (s *AppTestHelper) MockClientLatestHeight(height uint64) {
...
+
 	connection := connectiontypes.ConnectionEnd{
 		ClientId: clientId,
+		Versions: []*connectiontypes.Version{connectiontypes.DefaultIBCVersion},
 	}
 	s.App.IBCKeeper.ConnectionKeeper.SetConnection(s.Ctx, connectionId, connection)
 }
 
+// Helper function to mock out an ICA address
+func (s *AppTestHelper) MockICAChannel(connectionId, channelId, owner, address string) {
```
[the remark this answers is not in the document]
sampocs: It's a good question. I'm not sure all what's different, but I know for sure the CreateICAChannel one changes more state, including updating the "light client" / block time of each chain

My rec is to just use this when we need to test more than 1 account (cause I don't think CreateICAChannel supports more than one account at the moment)

x/stakeibc/keeper/msg_server_create_trade_route_test.go
sampocs: exactly!

x/stakeibc/keeper/msg_server_create_trade_route_test.go
sampocs: Either is fine imo. I think anytime we have only one trade route in a test, we can just use the global variables

x/stakeibc/keeper/msg_server_create_trade_route_test.go
sampocs: yeah the other hadn't been merged yet

x/stakeibc/keeper/msg_server_update_trade_route_test.go
sampocs: That's all covered in the validate basic unit test!

PR #1016, opened 2023-12-06
reward converter - pool price query unit test

sampocs:


x/stakeibc/keeper/reward_converter_test.go
```diff
@@ -344,3 +347,61 @@ func (s *KeeperTestSuite) TestSwapRewardTokens() {
 	err = s.App.StakeibcKeeper.SwapRewardTokens(s.Ctx, rewardAmount, route)
 	s.Require().ErrorContains(err, "epoch not found")
 }
+
+func (s *KeeperTestSuite) TestPoolPriceQuery() {
```
ethan-stride, line 351: So this is only the successful route, should we put "_Successful" on the end of the name?
What are the failure cases which would be worth checking?  
Non-legal connection id causes fail? Epoch tracker mismatch (would have caught the issue earlier with Hour vs Stride Epochs) Etc.

ethan-stride (commented):
LGTM
Should we be testing known common error cases for expected ICQ failures?

x/stakeibc/keeper/reward_converter_test.go
sampocs: I think there's only two error cases, epoch tracker and connection ID. I don't think it's possible to force the unmarshal to fail

I'm just going to add these to the bottom so we don't have to do make a new setup function

sampocs: done

x/stakeibc/keeper/keeper_test.go
```diff
@@ -67,16 +68,50 @@ func (s *KeeperTestSuite) MustGetHostZone(chainId string) types.HostZone {
...
+		EpochIdentifier:    epochType,
 		NextEpochStartTime: epochEndTime,
 	}
 	s.App.StakeibcKeeper.SetEpochTracker(s.Ctx, epochTracker)
 }
 
+// Validates the query object stored after an ICQ submission, using some default testing
+// values (e.g. HostChainId, stakeibc module name, etc.), and returning the query
+// NOTE: This assumes there was only one submission and grabs the first query from the store
+func (s *KeeperTestSuite) ValidateQuerySubmission(
```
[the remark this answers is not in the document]
sampocs: Sorry thought I had responded to this!

I tested the callback data outside of this because I couldn't get it to work with go generics [[ref](https://github.com/Stride-Labs/stride/pull/1016#discussion_r1419714144)]

x/stakeibc/keeper/reward_converter_test.go
```diff
@@ -344,3 +347,61 @@ func (s *KeeperTestSuite) TestSwapRewardTokens() {
...
+	expectedCallbackData := types.TradeRouteCallback{
+		RewardDenom: RewardDenom,
+		HostDenom:   HostDenom,
+	}
+
+	// Submit the pool price ICQ
+	err := s.App.StakeibcKeeper.PoolPriceQuery(s.Ctx, route)
+	s.Require().NoError(err, "no error expected when submitting pool price query")
+
+	// Confirm the query request key is the same regardless of which order the denom's are specified
```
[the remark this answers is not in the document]
sampocs: right, this was specific to the pool query - validating that the denom's are sorted

PR #1017, opened 2023-12-06
v14 cleanup

sampocs:
## Context
For upgrade v14, we added the field `DelegationChangesInProgress` initialized to 0. There was a tail risk that there would actually be an ICA in progress at the time that the upgrade happened, so we had temporarily swallowed an error case of the changes going below 0.

Now that the upgrade has succeeded and this case should no longer be possible, we can go back to throwing an error.

[[Relevant PR](https://github.com/Stride-Labs/stride/pull/910)]
[[Relevant Comment](https://github.com/Stride-Labs/stride/pull/910#discussion_r1306686032)]

## Brief Changelog
* Removed error swallow in delegation and undelegation callback

### DIFF
--- a/x/stakeibc/keeper/icacallbacks_delegate.go
+++ b/x/stakeibc/keeper/icacallbacks_delegate.go
@@ -1,7 +1,6 @@
 package keeper
 
 import (
-	"errors"
 	"fmt"
 
 	"github.com/spf13/cast"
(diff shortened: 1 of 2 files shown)


sampocs:
I don't really have a system - just happened to notice this! Fwiw, I'm not sure there are any other pressing TODOs, but if there are we can just make a GH issue!

PR #1018, opened 2023-12-06
sort unbonding prioritization by validator capacity 

sampocs:
## Context
Until unbonding from 32+ validators in a single batch is support, this sorts by capacity first to reduce the number of messages. 

## Changelog
* Removed condition to sort unbondings by balance ratio
* Added ConsolidateUnbondingMessages in the event that the number of messages is still greater than 32. This function will redistribute the remainder proportionally to each validator in the batch.

## Testing
* Tested with the unbonding simulator and a large unbond amount (3x the current mainnet amount):
  * Using mainnet code: 92 messages 
  * Using just the [sorting change](https://github.com/Stride-Labs/stride/pull/1018/files#diff-ba71e9ef6bf1ebcc5243683cbd44405947d8109d7ffb825c7c353d256f7e2179L163-L169): 34 messages
  * Using this branch: 32 messages


ethan-stride (commented):
Let's discuss assumptions around the incoming variables

x/stakeibc/keeper/unbonding_records.go
```diff
@@ -235,6 +245,93 @@ func (k Keeper) GetUnbondingICAMessages(
...
+				return finalUnbondings, fmt.Errorf("validator not found in initial unbonding plan")
+			}
+			unbondIncreaseProportion := delegationAfterFirstPass.Quo(totalDelegationsAfterFirstPass)
+			validatorUnbondIncrease = sdk.NewDecFromInt(totalExcessAmount).Mul(unbondIncreaseProportion).TruncateInt()
+
+			// Decrement excess
+			excessRemaining = excessRemaining.Sub(validatorUnbondIncrease)
+		} else {
+			// The last validator in the set should get any remainder from int truction
+			validatorUnbondIncrease = excessRemaining
```
ethan-stride, line 316: especially here we should check that validatorUnbondIncrease <= delegationAfterFirstPass on this last validator -- maybe we should even sort the capped set by validatorUnbondIncrease (increasing) so this final catch all would fall onto the validator with the most extra delegation in the set?

ethan-stride (commented):
After discussion on Friday, I think the algorithm is sound and makes sense.
Maybe a few renaming suggestions to make it clearer what the summed variables represent and maybe a few more safety checks that this unbonding is possible on a per validator level instead of the macro assumption that unbondings are always small compared to the total amount delegated in the capped set. 

In practice I think these are safe assumptions given the typically unequal delegations across validators and the assumption about the initialUnbondings coming in sorted, but there are some bad edge cases where this type of undelegation is not possible even in theory like equal delegation across 33 validators, batchsize = 32 and someone is trying to unbond everything for example

x/stakeibc/keeper/unbonding_records.go
```diff
@@ -310,10 +412,6 @@ func (k Keeper) UnbondFromHostZone(ctx sdk.Context, hostZone types.HostZone) err
 		return errorsmod.Wrap(sdkerrors.ErrInvalidRequest, "Target unbonded amount was 0 for each validator")
 	}
 
-	if len(msgs) > UndelegateICABatchSize {
```
riley-stride, line 313: Why not leave the sanity check?

x/stakeibc/keeper/unbonding_records.go
```diff
@@ -235,6 +245,93 @@ func (k Keeper) GetUnbondingICAMessages(
 	return msgs, unbondings, nil
 }
```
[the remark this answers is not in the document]
sampocs: yup that's all right!

x/stakeibc/keeper/unbonding_records.go
```diff
@@ -235,6 +245,93 @@ func (k Keeper) GetUnbondingICAMessages(
...
+		if initialUnbondAmount, ok := initialUnbondingsMap[capacity.ValidatorAddress]; ok {
+			delegationAfterFirstPass := sdk.NewDecFromInt(capacity.CurrentDelegation.Sub(initialUnbondAmount))
+
+			delegationsAfterFirstPass[capacity.ValidatorAddress] = delegationAfterFirstPass
+			totalDelegationsAfterFirstPass = totalDelegationsAfterFirstPass.Add(delegationAfterFirstPass)
+		}
+	}
+
+	// This should never happen, but it protects against a division by zero error below
+	if totalDelegationsAfterFirstPass.IsZero() {
```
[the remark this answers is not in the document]
sampocs: ah actually, I was misunderstand this when I wrote it (I previously thought it was impossible, but it's actually possible just unlikely)

This would occur if the 32 validators with the most capacity also happened to all be validators with 0 weight. Cause in that case, we would be unbonding everything from them

x/stakeibc/keeper/unbonding_records.go
sampocs: it was just moved upstream

x/stakeibc/keeper/unbonding_records.go
```diff
@@ -235,6 +245,93 @@ func (k Keeper) GetUnbondingICAMessages(
...
+		if i != len(unbondingsCapped)-1 {
+			// For all but the last validator, calculate their unbonding increase by
+			// splitting the excess proportionally in line with their remaining delegation
+			delegationAfterFirstPass, ok := delegationsAfterFirstPass[unbonding.Validator]
+			if !ok {
+				return finalUnbondings, fmt.Errorf("validator not found in initial unbonding plan")
+			}
+			unbondIncreaseProportion := delegationAfterFirstPass.Quo(totalDelegationsAfterFirstPass)
+			validatorUnbondIncrease = sdk.NewDecFromInt(totalExcessAmount).Mul(unbondIncreaseProportion).TruncateInt()
+
```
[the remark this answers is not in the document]
sampocs: agreed, will add these

x/stakeibc/keeper/unbonding_records.go
```diff
@@ -235,6 +245,93 @@ func (k Keeper) GetUnbondingICAMessages(
...
+	batchSize int,
+) (finalUnbondings []*types.SplitDelegation, err error) {
+	// Grab the first {batch_size} number of messages from the list
+	// This will consist of the validators with the most capacity
+	unbondingsCapped := initialUnbondings[:batchSize]
+
+	// Calculate the amount that was initially meant to be unbonded from that batch,
+	// and determine the remainder that needs to be redistributed
+	unbondAmountFromBatch := sdkmath.ZeroInt()
+	initialUnbondingsMap := map[string]sdkmath.Int{}
```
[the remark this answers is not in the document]
sampocs: Just pushed a new commit with some new names. Lemme know if it helps

For this particular variable, I renamed it to `initialUnbondAmountFromBatchByVal` - although, mentioning the batch is probably not necessary since this is just a map of the initial unbondings and we'll never lookup the vals outside of the batch

x/stakeibc/keeper/unbonding_records.go
```diff
@@ -235,6 +245,93 @@ func (k Keeper) GetUnbondingICAMessages(
...
+		unbondAmountFromBatch = unbondAmountFromBatch.Add(unbonding.Amount)
+		initialUnbondingsMap[unbonding.Validator] = unbonding.Amount
+	}
+	totalExcessAmount := totalUnbondAmount.Sub(unbondAmountFromBatch)
+
+	// Store the delegation of each validator that was expected *after* the originally
+	// planned unbonding went through
+	// e.g. If the validator had 10 before unbonding, and in the first pass, 3 was
+	//      supposed to be unbonded, their delegation after the first pass is 7
+	totalDelegationsAfterFirstPass := sdk.ZeroDec()
```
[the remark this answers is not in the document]
sampocs: See latest commits for a pass at improving this. Went with:

`totalRemainingDelegationsAcrossBatch` and `remainingDelegationsInBatchByVal`

x/stakeibc/keeper/unbonding_records.go
```diff
@@ -235,6 +245,93 @@ func (k Keeper) GetUnbondingICAMessages(
...
+	// the excess amongst the validators in the set
+	excessRemaining := totalExcessAmount
+	for i := range unbondingsCapped {
+		unbonding := unbondingsCapped[i]
+
+		var validatorUnbondIncrease sdkmath.Int
+		if i != len(unbondingsCapped)-1 {
+			// For all but the last validator, calculate their unbonding increase by
+			// splitting the excess proportionally in line with their remaining delegation
+			delegationAfterFirstPass, ok := delegationsAfterFirstPass[unbonding.Validator]
```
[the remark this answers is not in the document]
sampocs: agreed, I like remainingDelegation better!

x/stakeibc/keeper/unbonding_records.go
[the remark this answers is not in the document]
sampocs: These cases should all be covered by [this](https://github.com/Stride-Labs/stride/pull/1018/commits/c28539c1b1056747e2d1806e715b551864236c10#diff-ba71e9ef6bf1ebcc5243683cbd44405947d8109d7ffb825c7c353d256f7e2179R297-R301)

x/stakeibc/keeper/unbonding_records.go
sampocs: I think [this](https://github.com/Stride-Labs/stride/pull/1018/commits/c28539c1b1056747e2d1806e715b551864236c10#diff-ba71e9ef6bf1ebcc5243683cbd44405947d8109d7ffb825c7c353d256f7e2179R297-R301) check is sufficient, but added another check [here](https://github.com/Stride-Labs/stride/pull/1018/commits/e388120788759f76b792e4e1988a98f6623bc705) since it's free and gives a bit more peace of mind

x/stakeibc/keeper/unbonding_records.go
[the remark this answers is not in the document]
ethan-stride: The reason I was suggesting capped set be part somehow is that if we iterate over initialUnbondings or iterate over the capacities we would get validators outside the capped set.  In fact we do some of these types of iterations but with an if clause inside the loop which limits to validators inside the capped set.  We don't have to call it capped set if we don't want to but somehow make it clear it is a limited amount of the validators since that is one of the ways I got confused reading through the code the first time wondering if that was intended or a mistake to only have values for some but not all validators.

x/stakeibc/keeper/unbonding_records.go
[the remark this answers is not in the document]
ethan-stride: oh, this is an interesting case.  Even if not all 32 would be zero weight validators, there could be enough in the top 32 that the non-zero weight validators get all of their delegation unbonded... probably unlikely in practice without monster LSMs happening frequently.

But hold on, we wouldn't necessarily be unbonding everything from even the zero weight ones, it is just that for them capacity == currentDelegation because balancedDelegation = 0.  It would depend on the total unbonding amount across all the messages if it is greater than the delegations across all the zero weight ones but I get what you are saying.

ethan-stride:
LGTM
Looks good in that this is a temporary fix and the times we would fail with error (ie. respond we can't unbond this incoming request) are very unlikely situations in practice.  We are really scraping weird edge cases and most of the time this will work without any issue.  I think everything is sound and your modifications made the algorithm conceptually clearer.

x/stakeibc/keeper/unbonding_records.go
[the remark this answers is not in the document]
sampocs: does fromBatch clear it up though? Instead of capped?

x/stakeibc/keeper/unbonding_records_get_host_zone_unbondings_msgs_test.go
```diff
@@ -888,3 +912,140 @@ func (s *KeeperTestSuite) TestGetUnbondingICAMessages() {
...
+		initialUnbondAmount        int64
+		remainingDelegation        int64
+		expectedDelegationIncrease int64
+	}{
+		// Total Remaining Portion: 1000 + 500 + 250 + 250 = 2000
+		// Excess Portion = Remaining Delegation / Total Remaining Portion
+
+		// Excess Portion: 1000 / 2000 = 50%
+		// Delegation Increase: 50% * 101 = 50
+		{address: "val1", initialUnbondAmount: 500, remainingDelegation: 1000, expectedDelegationIncrease: 50},
```
riley-stride, line 932: Looks like we're testing with less than 32 vals. Do you think that's sufficient coverage if the batchSize is also low?

x/stakeibc/keeper/unbonding_records_get_host_zone_unbondings_msgs_test.go
```diff
@@ -888,3 +912,140 @@ func (s *KeeperTestSuite) TestGetUnbondingICAMessages() {
 		})
 	}
 }
+
+func (s *KeeperTestSuite) TestConsolidateUnbondingMessages_Success() {
+	batchSize := 4
```
riley-stride, line 917: Do we need tests for edge cases with o-weight vals?

x/stakeibc/keeper/unbonding_records_get_host_zone_unbondings_msgs_test.go
sampocs: Yeah since batchSIze is passed in as a param, I think it's functionally the same

x/stakeibc/keeper/unbonding_records_get_host_zone_unbondings_msgs_test.go
sampocs: Good question! It's actually covered by [this](https://github.com/Stride-Labs/stride/pull/1018/files#diff-42324da485a3df105dabdb479a72a8b7e82535250b0742a415e1c590081913ffR1023)

Zero weight validators will have their capacity equal to their current delegation ([example](https://github.com/Stride-Labs/stride/pull/1018/files#diff-42324da485a3df105dabdb479a72a8b7e82535250b0742a415e1c590081913ffR1031))

x/stakeibc/keeper/unbonding_records_get_host_zone_unbondings_msgs_test.go
riley-stride: I'm not confident the coverage is equivalent (I think there could be cases we'd hit with more vals, e.g. there could be much more spillover that needs to be distributed across the batchSize set) but I'm not very confident in them, couldn't point you to a specific case. As long as you did some checks with the unbonding simulator tool spreadsheet as well, I think I feel good about this. Sounds like you did some test with the unbonding simulator?

riley-stride (approved):
`ConsolidateUnbondingMessages` and `GetUnbondingICAMessages` are looking good, and thoroughly tested!

x/stakeibc/keeper/unbonding_records_get_host_zone_unbondings_msgs_test.go
sampocs: I would just think about it in terms of the different permutations of outcomes, which is generally agnostic to the batch size. 

From that framing, the only cases I could think of were:
* We are able to successfully re-distribute the excess  [[covered here](https://github.com/Stride-Labs/stride/pull/1018/files#diff-42324da485a3df105dabdb479a72a8b7e82535250b0742a415e1c590081913ffR916)]
* There is not enough existing stake to redistribute the access [[tested here](https://github.com/Stride-Labs/stride/pull/1018/files#diff-42324da485a3df105dabdb479a72a8b7e82535250b0742a415e1c590081913ffR1023)]
* There is no excess [covered by the `TestUnbondFromHostZone` cases]

Very open to adding more tests if you have a case in mind that doesn't fall into one of these!

sampocs: From riley on slack: 
> Yeah that makes sense. And I can't think of other cases. If I think of any, will follow up.

Marking this resolved!

x/stakeibc/keeper/unbonding_records.go
```diff
@@ -235,6 +245,111 @@ func (k Keeper) GetUnbondingICAMessages(
...
+	// planned unbonding went through
+	// e.g. If the validator had 10 before unbonding, and in the first pass, 3 was
+	//      supposed to be unbonded, their delegation after the first pass is 7
+	totalRemainingDelegationsAcrossBatch := sdk.ZeroDec()
+	remainingDelegationsInBatchByVal := map[string]sdk.Dec{}
+	for _, capacity := range unbondCapacities {
+		// Only add validators that were in the initial unbonding plan
+		// The delegation after the first pass is calculated by taking the "current delegation"
+		// (aka delegation before unbonding) and subtracting the unbond amount
+		if initialUnbondAmount, ok := initialUnbondAmountFromBatchByVal[capacity.ValidatorAddress]; ok {
```
shellvish, line 283: nit: should we change "ok" to "validatorInFinalUnbondingBatch" or something a bit more descriptive? Also recognize that it'll be more verbose, so I don't feel too strongly here

x/stakeibc/keeper/unbonding_records.go
```diff
@@ -235,6 +245,111 @@ func (k Keeper) GetUnbondingICAMessages(
...
+			remainingDelegation := sdk.NewDecFromInt(capacity.CurrentDelegation.Sub(initialUnbondAmount))
+
+			remainingDelegationsInBatchByVal[capacity.ValidatorAddress] = remainingDelegation
+			totalRemainingDelegationsAcrossBatch = totalRemainingDelegationsAcrossBatch.Add(remainingDelegation)
+		}
+	}
+
+	// This is to protect against a division by zero error, but this would technically be possible
+	// if the 32 validators with the most capacity were all 0 weight
```
shellvish, line 292: nit: would require validators with most capacity are at 0 weight AND we want to unbond the full amount from them in the initial pass (this is mostly the same thing, with the one exception being if the last validator in `unbondingsBatch` had some excess capacity)

x/stakeibc/keeper/unbonding_records.go
```diff
@@ -235,6 +245,111 @@ func (k Keeper) GetUnbondingICAMessages(
...
+			}
+			unbondIncreaseProportion := remainingDelegation.Quo(totalRemainingDelegationsAcrossBatch)
+			validatorUnbondIncrease = sdk.NewDecFromInt(totalExcessAmount).Mul(unbondIncreaseProportion).TruncateInt()
+
+			// Decrement excess
+			excessRemaining = excessRemaining.Sub(validatorUnbondIncrease)
+		} else {
+			// The last validator in the set should get any remainder from int truction
+			// First confirm the validator has sufficient remaining delegation to cover this
+			remainingDelegation, ok := remainingDelegationsInBatchByVal[unbonding.Validator]
```
shellvish, line 324: nit: could we move this above the `if i != len(unbondingsBatch) - 1` statement, to avoid duplicating code across the `if` and `else`?

shellvish (approved):
Thanks a lot for making this! I think this will make the unbonding algorithm significantly more resilient going forward.

The one edge case I'm thinking through is if the validator's with the highest capacities are NOT the validators with the highest delegated amounts. In this case, the algorithm might return a failure if we're unable to unbond from the highest capacity validators, when we would have instead been able to unbond from the highest delegated validators. 

That being said, I think this is a completely fine tradeoff. In general, these two sets should have very high overlap (given the capacity is calculated AFTER the unbonding processes, so validators with highest delegation should have the highest capacity as well), and in cases where they differ, the capacity is the "conceptually cleaner" way to undelegate. It should be extraordinarily rare that this algorithm would fail. 

Also, just want to shoutout the detail in the comments, makes reading and understanding this algorithm so much easier, and the code quality is very high (especially given the complexity of the code). Tests also look excellent.

x/stakeibc/keeper/unbonding_records.go
sampocs: yes, great call!

x/stakeibc/keeper/unbonding_records.go
sampocs: good clarification! updated [[here](https://github.com/Stride-Labs/stride/pull/1018/commits/fb122c549a7df2ef85fdbbb8b2798bfb3f9b8fd6)]

x/stakeibc/keeper/unbonding_records.go
sampocs: done [[here](https://github.com/Stride-Labs/stride/pull/1018/commits/fb122c549a7df2ef85fdbbb8b2798bfb3f9b8fd6)]

sampocs:
> The one edge case I'm thinking through is if the validator's with the highest capacities are NOT the validators with the highest delegated amounts. In this case, the algorithm might return a failure if we're unable to unbond from the highest capacity validators, when we would have instead been able to unbond from the highest delegated validators.

Yeah very good callout! My mental model for this is that for non-LSM chains, this algorithm should behave no different than just splitting messages purely based on weights. And for LSM chains, we target the validators with the most LSM stake first. 

So this failure case occurs if the validators with the most LSM stake also all happen to be 0 weight - which in practice, it should be the opposite! 

Also, each time we rebalance, this gets "reset"

@shellvish

x/stakeibc/keeper/unbonding_records.go
shellvish: 🤦 

of course haha, my bad! let's keep it as is

x/stakeibc/keeper/unbonding_records.go
shellvish: awesome, thank you!

x/stakeibc/keeper/unbonding_records.go
shellvish: awesome, thank you!

PR #1019, opened 2023-12-07
reward converter - fix timeouts

sampocs:
We need to timeout the swap at the end of the _hour_ (instead of stride epoch). And we need to timeout the withdrawal balance query at the epoch halfway mark since that's when the first transfer times out

### DIFF
--- a/x/stakeibc/keeper/reward_converter.go
+++ b/x/stakeibc/keeper/reward_converter.go
@@ -254,9 +254,9 @@ func (k Keeper) SwapRewardTokens(ctx sdk.Context, rewardAmount sdkmath.Int, rout
 		"Preparing MsgSwapExactAmountIn of %+v from the trade account", msg.TokenIn))
 
 	// Timeout the swap at the end of the epoch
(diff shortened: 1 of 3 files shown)


x/stakeibc/keeper/reward_converter.go
```diff
@@ -254,9 +254,9 @@ func (k Keeper) SwapRewardTokens(ctx sdk.Context, rewardAmount sdkmath.Int, rout
 		"Preparing MsgSwapExactAmountIn of %+v from the trade account", msg.TokenIn))
 
 	// Timeout the swap at the end of the epoch
-	strideEpochTracker, found := k.GetEpochTracker(ctx, epochstypes.STRIDE_EPOCH)
+	strideEpochTracker, found := k.GetEpochTracker(ctx, epochstypes.HOUR_EPOCH)
```
ethan-stride, line 257: Good catch, I wasn't even thinking about that the faster call cadence to the swap would cause these timeouts to overlap unless we used the hour cadence.

ethan-stride (commented):
LGTM, no open questions looks good to merge

PR #1023, opened 2023-12-09
reward converter - restore trade route ICA

sampocs:
Added support for restoring a trade route ICA. The original restore function fails because a trade route ICA is not necessarily associated with a host zone


x/stakeibc/client/cli/tx_restore_interchain_account.go
```diff
@@ -15,24 +13,26 @@ import (
...
+Specify the chain ID and account owner - where the owner is the alias for the ICA account
+
+For host zone ICA accounts, the owner is of the form {chainId}.{accountType}
+ex:
+>>> strided tx restore-interchain-account cosmoshub-4 connection-0 cosmoshub-4.DELEGATION 
+
+For trade route ICA accounts, the owner is of the form:
+    {chainId}.{rewardDenom}-{hostDenom}.{accountType}
+ex:
+>>> strided tx restore-interchain-account dydx-mainnet-1 connection-1 dydx-mainnet-1.uusdc.udydx.CONVERTER_TRADE
```
ethan-stride, line 29: dash not dot to separate the denoms

ethan-stride (commented):
LGTM

ethan-stride (approved):

x/stakeibc/client/cli/tx_restore_interchain_account.go
sampocs: nice catch!

PR #1028, opened 2023-12-14
whitelisted all gov proposals

sampocs:
## Context
Whitelisting all gov proposals because:
* It's a burden to maintain the list
* There have been multiple proposals that have been deleted on mainnet because of this
* This was mostly intended to block param changes to the CCV module - which are still possible in other ways

## Brief Changelog
* Returned true in `IsWhitelist` functions to whitelist all gov proposals
* Removed antehandler


app/app.go
```diff
@@ -748,7 +748,7 @@ func NewStrideApp(
 		capability.NewAppModule(appCodec, *app.CapabilityKeeper, false),
 		feegrantmodule.NewAppModule(appCodec, app.AccountKeeper, app.BankKeeper, app.FeeGrantKeeper, app.interfaceRegistry),
 		crisis.NewAppModule(app.CrisisKeeper, skipGenesisInvariants, app.GetSubspace(crisistypes.ModuleName)),
-		ccvgov.NewAppModule(appCodec, app.GovKeeper, app.AccountKeeper, app.BankKeeper, IsProposalWhitelisted, app.GetSubspace(govtypes.ModuleName), func(_ string) bool { return true }),
+		ccvgov.NewAppModule(appCodec, app.GovKeeper, app.AccountKeeper, app.BankKeeper, IsProposalWhitelisted, app.GetSubspace(govtypes.ModuleName), IsModuleWhiteList),
```
asalzmann, line 751: what's your thinking on using passing in a fn that returns true vs. having the `IsProposalWhitelisted` in another file that return true? the former approach seems slightly simpler to me, but I don't feel strongly!

asalzmann (approved):
lgtm, thank you for doing this!

cc @jtremback

app/app.go
sampocs: I think either is fine! 

My thinking was - when we were tracing through this together, it took us awhile to realize this function was passed in here cause we couldn't find it via global search. So I figured this gives us a bit more traceability / is a little more explicit!

asalzmann: Makes sense to me

PR #1030, opened 2023-12-15
autopilot redeem unit tests

sampocs:
## Context
Fixed, cleaned up, and added coverage to autopilot redeem unit tests.

## Brief Changelog
* Fixed build errors in unit tests
* Slimmed down mocked out state to only what's necessary
* Added coverage to all error cases (by testing inner TryRedeemStake function instead of OnRecvPacket)
* Added clarifying comments
* Renamed variables in redeem stake call for clarify

### DIFF
--- a/x/autopilot/keeper/redeem_stake.go
+++ b/x/autopilot/keeper/redeem_stake.go
@@ -4,9 +4,8 @@ import (
 	"fmt"
 
 	errorsmod "cosmossdk.io/errors"
-	"github.com/armon/go-metrics"
+	sdkmath "cosmossdk.io/math"
 	sdk "github.com/cosmos/cosmos-sdk/types"
-	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
 	transfertypes "github.com/cosmos/ibc-go/v7/modules/apps/transfer/types"
 	channeltypes "github.com/cosmos/ibc-go/v7/modules/core/04-channel/types"
 
@@ -26,9 +25,14 @@ func (k Keeper) TryRedeemStake(
 		return fmt.Errorf("packet forwarding param is not active")
 	}
 
-	// In this case, we can't process a liquid staking transaction, because we're dealing IBC tokens from other chains
+	// At this point in the stack, the denom's in the packet data appear as they existed on the sender zone,
+	//   but as a denom trace instead of a hash
+	// Meaning, for native stTokens, the port and channel on the host zone are part of the denom
+	//   (e.g. transfer/{channel-on-hub}/stuatom)
+	// Only stride native stTokens can be redeemed, so we confirm that the denom's prefix matches
+	//   the packet's "source" channel (i.e. the channel on the host zone)
 	if !transfertypes.ReceiverChainIsSource(packet.GetSourcePort(), packet.GetSourceChannel(), newData.Denom) {
-		return fmt.Errorf("the ibc tokens are not supported for redeem stake")
+		return fmt.Errorf("the ibc token %s is not supported for redeem stake", newData.Denom)
 	}
 
 	voucherPrefix := transfertypes.GetDenomPrefix(packet.GetSourcePort(), packet.GetSourceChannel())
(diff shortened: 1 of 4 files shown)


x/autopilot/keeper/redeem_stake.go
```diff
@@ -18,74 +17,64 @@ import (
...
+	redemptionReceiver := autopilotMetadata.IbcReceiver
 
-	if err := token.Validate(); err != nil {
-		return err
-	}
-
-	strideAddress, err := sdk.AccAddressFromBech32(newData.Receiver)
-	if err != nil {
-		return errorsmod.Wrapf(sdkerrors.ErrInvalidAddress, "invalid receiver (%s) in autopilot memo", strideAddress)
-	}
```
sampocs, line 60: * Removed token.Validate because `newData.Denom` isn't an actual denom, it's a denom trace path - so this check feels a bit misleading. Plus, if the denom doesn't match our host zone, we error later
* Removed address conversion because downstream we use it as a string instead of address, and it's validated in msg.ValidateBasic

x/autopilot/keeper/redeem_stake.go
```diff
@@ -18,74 +17,64 @@ import (
...
 	if !stakeibctypes.IsStAssetDenom(stAssetDenom) {
 		return fmt.Errorf("not a liquid staking token")
 	}
 
 	hostZoneDenom := stakeibctypes.HostZoneDenomFromStAssetDenom(stAssetDenom)
 
-	amount, ok := sdk.NewIntFromString(newData.Amount)
-	if !ok {
+	amount, ok := sdk.NewIntFromString(transferPacketData.Amount)
+	if !ok || amount.IsNegative() {
```
asalzmann, line 47: what's the scenario where `eval(ok is true and amount < 0) is true`?

x/autopilot/keeper/redeem_stake.go
asalzmann: > it's validated in msg.ValidateBasic

where does this check happen?

asalzmann (commented):
lgtm!

x/autopilot/module_ibc.go
```diff
@@ -116,6 +116,8 @@ func (im IBCModule) OnChanCloseConfirm(
 // OnRecvPacket implements the IBCModule interface. A successful acknowledgement
 // is returned if the packet data is successfully decoded and the receive application
 // logic returns without error.
+// TODO: Move this to the keeper so there's more transparency into errors
```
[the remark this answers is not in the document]
sampocs: neither haha, it's just advice for a later cleanup - shouldn't block this upgrade

x/autopilot/keeper/redeem_stake.go
sampocs: [here](https://github.com/Stride-Labs/stride/pull/1030/files#diff-6badb27c023a93b231bdecf85015f77643e5defa71a492747225acaf41d810f5R70), which runs [this](https://github.com/Stride-Labs/stride/blob/main/x/stakeibc/types/message_redeem_stake.go#L47-L50)

x/autopilot/keeper/redeem_stake.go
sampocs: if amount was, say, "-1000", it would be successfully parsed into an int (i.e. `ok == true`), but it would be negative

asalzmann: shouldn't the underlying ibc transfer module filter packets with negative amounts out? 

if it's just a sanity check, including it seems fine (it might make sense to check `amount.LTE(0)`)

sampocs: I don't think it matters much either way cause we validate that amount again in the redeem stake validate basic

I think the only reason I put it in was because it was the only meaningful check that was done in the coin.Validate function that I removed. 

I'll remove the isPositive in the autopilot-redeem branch

PR #1035, opened 2023-12-21
removed sender user redemption record

sampocs:
## Context
In order to allow multiple redemptions per epoch, we need to key the URR on the _receiver_ address instead of the _sender_ address. Otherwise, someone could steal or DDOS a user's redemption. 

## Brief Changelog
* Removed sender field from URR
* Changed URR key to use receiver
* Changed claim msg to take the receiver address
* Fixed references/unit tests


shellvish (approved):
This looks amazing, thank you for doing this! 

The one question from my end: what happens the epoch this upgrade happens? I _think_ unbondings prior to the upgrade will use the old format, and unbondings after the upgrade will use the new format. When those tokens are finished unbonding and need to be claimed, the claimer needs to make sure they pass the right format in, but everything should work as expected. Would be good to double check this intuition with you though! 

This will be a huge quality-of-life improvement 😄

x/records/migrations/v2/convert.go
```diff
@@ -43,7 +43,6 @@ func convertToNewEpochUnbondingRecord(oldEpochUnbondingRecord oldrecordstypes.Ep
 func convertToNewUserRedemptionRecord(oldRedemptionRecord oldrecordstypes.UserRedemptionRecord) recordstypes.UserRedemptionRecord {
 	return recordstypes.UserRedemptionRecord{
 		Id:             oldRedemptionRecord.Id,
-		Sender:         oldRedemptionRecord.Sender,
```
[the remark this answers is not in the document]
sampocs: yeah the old migration code doesn't really matter haha, it's more just a source of documentation

I removed it here cause otherwise it wouldn't be possible to compile because the type no longer has the sender field

As a rule of thumb, I think we should try to avoid changing these as much as possible, but I don't think small things like this matter much!

sampocs:
@shellvish that's my understanding as well!

riley-stride (approved):
Looks great. No comments on the code itself. 

Do we think we need to test the DOS / stealing scenario with this change?

sampocs:
@riley-stride Yeah good call! Are you thinking just test on the latest commit what would happen if someone else tried to redeem for a user? Or were you thinking we'd want to compare to an older commit?

riley-stride:
> @riley-stride Yeah good call! Are you thinking just test on the latest commit what would happen if someone else tried to redeem for a user? Or were you thinking we'd want to compare to an older commit?

Ideally replicate the attack at an old commit, then verify it's no longer possible here. But that might be overkill. I think testing it on the latest commit is probably enough.

x/records/migrations/v2/convert.go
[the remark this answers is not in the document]
sampocs: @shellvish can I mark this resolved?

sampocs:
@riley-stride this branch actually doesn't have the autopilot redeem implemented. I'll add a TODO to test that on one of the other branches.

I did test claiming from two different stride addresses to the same gaia address and confirmed the amount on the original record was incremented

proto/stride/stakeibc/tx.proto
```diff
@@ -131,10 +131,10 @@ message MsgRegisterHostZoneResponse {}
 
 message MsgClaimUndelegatedTokens {
   string creator = 1;
-  // UserUnbondingRecords are keyed on {chain_id}.{epoch}.{sender}
+  // UserUnbondingRecords are keyed on {chain_id}.{epoch}.{receiver}
   string host_zone_id = 2;
   uint64 epoch = 3;
-  string sender = 4;
+  string receiver = 4;
```
asalzmann, line 137: Why use a new field / reserve the old one above in UserRedemptionRecord, but not here?

x/records/migrations/v2/convert.go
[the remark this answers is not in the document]
asalzmann: maybe just comment it out (nit)

x/records/migrations/v2/convert_test.go
```diff
@@ -121,7 +121,6 @@ func TestConvertUserRedemptionRecord(t *testing.T) {
 	}
 	expectedNewUserRedemptionRecord := recordstypes.UserRedemptionRecord{
 		Id:             id,
-		Sender:         sender,
```
asalzmann, line 124: if you end up commenting out the field in `x/records/migrations/v2/convert.go` (nit, not required), should also comment this one out (for consistency)

x/stakeibc/keeper/msg_server_claim_undelegated_tokens.go
```diff
@@ -121,6 +121,7 @@ func (k Keeper) GetRedemptionTransferMsg(ctx sdk.Context, userRedemptionRecord *
 	})
 
 	// Give claims a 10 minute timeout
+	// TODO: Remove reference to epoch time (make timeout relative to block time)
```
asalzmann, line 124: does this TODO need to be executed?

x/stakeibc/types/message_claim_undelegated_tokens.go
```diff
@@ -47,11 +45,6 @@ func (msg *MsgClaimUndelegatedTokens) ValidateBasic() error {
 	if err != nil {
 		return errorsmod.Wrapf(sdkerrors.ErrInvalidAddress, "invalid creator address (%s)", err)
 	}
-	// sender must be a valid stride address
```
asalzmann, line 50: is there any reason to check the receiver address validity here? don't think there is (since it's verified in `RedeemStake`), but just double-checking

asalzmann (approved):
lgtm!

x/records/migrations/v2/convert.go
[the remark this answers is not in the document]
sampocs: good idea, done!

x/records/migrations/v2/convert_test.go
sampocs: done

x/stakeibc/keeper/msg_server_claim_undelegated_tokens.go
sampocs: I don't think it should block the upgrade, but it's a change we should make at some point

proto/stride/stakeibc/tx.proto
sampocs: I didn't think it mattered for messages since we don't read them from state. Probably low cost to use reserve though

sampocs: updated

x/stakeibc/keeper/msg_server_claim_undelegated_tokens.go
asalzmann: got it - we could use TODO [optimization] or TODO [cleanup] to make this more clear

PR #1038, opened 2023-12-27
autopilot hashed sender

sampocs:
## Context
During liquid stake and forward, the autopilot "receiver" of the inbound transfer becomes the "sender" of the transfer back to the host. However, downstream applications shouldn't trust this new "sender" so we need to use a generated address instead. 

To be clear, using the original sender would not introduce an attack vector on Stride. However, it could introduce an attack vector on a different zone, _if_ they were to trust the sender. The hashed sender is used to make the assumption more explicit that new zones should not trust the address.

This bug appeared in PFM (if more context is needed):
* [Issue](https://github.com/cosmos/ibc-apps/issues/68)
* [PR fix](https://github.com/cosmos/ibc-apps/pull/71)

## Design Considerations
There wasn't an immediately obvious way to implement this. The complexity arises in that the address used for the inbound transfer doesn't always line up with the address used in the autopilot action. Additionally, if a bank send is required after the transfer (in the event that we transferred to an intermediate recipient), some scenarios would require us to build the IBC denom hash ourselves (not impossible, but adds a lot of unnecessary code).

There didn't seem to be a clean way to refactor the callback such that we use a different receiver based on the action type (i.e. use "hashed" for LS&Forward, but "original" for Claim). As such, I think the only two options were as follows: 

1. _Send inbound transfer to original receiver_
  * ✅  **Claim**: No changes
  * ✅  **Liquid Stake**: No changes
  * ❌  **Liquid Stake and Forward**: Requires us to bank send to hashed receiver (but this is straightforward since the token is not an IBC denom)
2. _Send inbound transfer to hashed receiver_
  * ❌❌ **Claim**: Would require a bank send to the original receiver (we'd need to copy over all the IBC hashing code from ibc-go in order to handle both native and non-native denoms)
  * ❌  **Liquid Stake**: Would require a bank send to the original receiver (but this is straightforward since the token is not an IBC denom)
  * ✅ **Liquid Stake and Forward**: No changes
3. _Decide recipient based on action_
  * ✅  **Claim**: No changes
  * ✅  **Liquid Stake**: No changes
  * ✅ **Claim:** No changes
  * ❌ Requires some somewhat sloppy refactoring since the change must be upstream of the transfer

~I do think using the hashed receiver address as the inbound recipient and actor for each of the autopilot actions, does feel slightly more correct. However, I don't think it's a significant enough argument to justify the additional changes that would be required. Additionally, going with option 1 demands no changes to the existing autopilot functions that have been live on mainnet.~

After discussing offline, we decided to do option 3 and to simplify the autopilot schema so the refactor isn't as messy. This gives us the benefit of not having to worry about doing an extra bank send, but also not duplicating code in multiple places (see [here](https://github.com/Stride-Labs/stride/pull/1038#issuecomment-1879371007)).

Option 3 is tracked in #1046 

## Brief Changelog
* Added `GenerateHashedAddress` to generate the hashed address from the channel Id a previous sender (body of function taken from PFM)
* Renamed `PacketForwardMetadata` to `AutopilotMetadata`
* Added a bank send to the hashed address in LS and forward
* Set the hashed address as the outbound sender during LS and forward
* Restricted packets to being either autopilot or PFM (but not both)
* Cleaned up the variable names to avoid confusion with all the data structures (open to better names though!)
* Moved structs out of parser.go


riley-stride:
Just talked this one through with Sam and think I agree that (1) is a good approach.

My original intuition would have been to opt for approach (3): on the inbound IBC transfer, use a different receiver based on the action type (i.e. use "hashed" for LS&Forward, but "original" for LS and for Claim). Was thinking that at the stage of the IBC middleware stack where the transfer occurs, we could check the memo and decide whether to transfer to "hashed" or "original" at that stage. But it sounds like that might require some refactoring.

And (1) is easy to think about because it puts all the LS&Forward logic in `liquidstake.go` functions, rather than intermingling it in the IBC middleware stack.

x/autopilot/keeper/airdrop.go
```diff
@@ -39,11 +38,11 @@ func (k Keeper) TryUpdateAirdropClaim(
...
 
 	// grab relevant addresses
-	senderStrideAddress := utils.ConvertAddressToStrideAddress(data.Sender)
+	senderStrideAddress := utils.ConvertAddressToStrideAddress(transferMetadata.Sender)
 	if senderStrideAddress == "" {
-		return errorsmod.Wrapf(sdkerrors.ErrInvalidAddress, fmt.Sprintf("invalid sender address (%s)", data.Sender))
+		return errorsmod.Wrapf(sdkerrors.ErrInvalidAddress, fmt.Sprintf("invalid sender address (%s)", transferMetadata.Sender))
 	}
-	newStrideAddress := packetMetadata.StrideAddress
+	newStrideAddress := transferMetadata.Receiver
```
riley-stride, line 45: To double check these changes from `packetMetadata` -> `transferMetadata` would you mind writing out the mapping between the two's fields?

x/autopilot/keeper/liquidstake.go
```diff
@@ -16,103 +15,141 @@ import (
...
+	amount, ok := sdk.NewIntFromString(transferMetadata.Amount)
 	if !ok {
 		return errors.New("not a parsable amount field")
 	}
 
-	// Note: newData.denom is base denom e.g. uatom - not ibc/xxx
-	var token = sdk.NewCoin(newData.Denom, amount)
+	// In this case, we can't process a liquid staking transaction, because we're dealing with native tokens (e.g. STRD, stATOM)
+	if transfertypes.ReceiverChainIsSource(packet.GetSourcePort(), packet.GetSourceChannel(), transferMetadata.Denom) {
+		return errors.New("the native token is not supported for liquid staking")
```
riley-stride, line 39: Do we have unit tests covering the changes to this function, for example this line?

x/autopilot/keeper/liquidstake.go
```diff
@@ -16,103 +15,141 @@ import (
...
+	channelId := autopilotMetadata.TransferChannel
 	if channelId == "" {
 		channelId = hostZone.TransferChannelId
 	}
+
+	// Generate a hashed address for the sender to the next hop,
+	// to prevent impersonation at downstream zones
+	// Note: The channel ID here is different than the one used in PFM
+	// (we use the outbound channelID, they use the inbound channelID)
+	// DOUBLE CHECK ME that it shouldn't matter
```
riley-stride, line 120: Why do they include the channel in the hash in the first place? Not clear you need one at all. In fact, for our use case, there should only be one acceptable channel from host to Stride, so (check me) the channel doesn't add any information to the hash (it will always be the same for a given address).

Including the outbound channelID _does_ differentiate between LS-from-addrX&forward-to-chainA and LS-from-addrX&forward-to-chainB. Doesn't seem like we care whether there's a collision there but still might be nice to avoid it.

x/autopilot/module_ibc.go
```diff
@@ -7,7 +7,6 @@ import (
```
riley-stride, line 7: Not as familiar with `module_ibc.go` so can't give it a rubber stamp, though the changes _look_ to be just changes to names/types. Still, would recommend @asalzmann give it a very thorough review.

riley-stride (approved):

x/autopilot/types/autopilot.go
```diff
@@ -0,0 +1,50 @@
...
+
+// GenerateHashedSender generates a new  address for a packet, by hashing
+// the channel and original sender.
+// This makes the address deterministic and can used to identify the sender
+// from the preivous hop
+// Additionally, this prevents a forwarded packet from impersonating a different account
+// when moving to the next hop (i.e. receiver of one hop, becomes sender of next)
+//
+// This function was borrowed from PFM
+func GenerateHashedAddress(channelId, originalSender string) (string, error) {
```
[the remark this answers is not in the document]
sampocs: Potentially! I don't have a strong opinion either way

Generally, I like to eliminate dependencies when the code is short and simple like this (and it guarantees it wont be modified without us knowing), but we already have to add the pfm dependency so there's not much more overhead

x/autopilot/keeper/airdrop.go
sampocs: `transferMetadata` is the `transfertypes.FungibleTokenData` and `packetMetadata` _was_ the autopilot+airdrop specific struct that had has custom fields needed for the autopilot action.

But... the custom fields for the airdrop only consisted of `StrideAddress`, and when we build the `packetMetadata`, the `StrideAddress` is pulled directly from `transferMetadata.Receiver` (i.e. it's duplicated). So we can just remove the `packetMetadata` struct entirely from this function and use the receiver instead

x/autopilot/keeper/liquidstake.go
sampocs: In [this](https://github.com/Stride-Labs/stride/pull/1041) PR!

x/autopilot/module_ibc.go
sampocs: yeah it should be mostly just variable renaming! agreed though it'd be good to get Aidan's review

shellvish (commented):
This looks great, thank you so much for doing this! Very meticulous, and a good solution.

I left a couple of small clarifying questions, but should hopefully be quick! All signed off otherwise

x/autopilot/keeper/liquidstake.go
```diff
@@ -16,103 +15,141 @@ import (
...
 		msg,
 	)
 	if err != nil {
 		return errorsmod.Wrapf(err, err.Error())
 	}
 
-	if packetMetadata.IbcReceiver == "" {
+	// If the IBCReceiver is empty, there is no forwarding step
+	if autopilotMetadata.IbcReceiver == "" {
 		return nil
```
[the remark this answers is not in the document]
sampocs: It's verified when we parse the payload [here](https://github.com/Stride-Labs/stride/blob/main/x/autopilot/types/parser.go#L82-L84)!

x/autopilot/keeper/liquidstake.go
```diff
@@ -16,103 +15,141 @@ import (
...
-		// TODO: does this reintroduce the bug in PFM where senders can be spoofed?
-		// If so, should we instead call PFM directly to forward the packet?
-		// Or should we obfuscate the sender, making it a random address?
-		Sender:           sender,
-		Receiver:         packetMetadata.IbcReceiver,
+		SourcePort:       transfertypes.PortID,
+		SourceChannel:    channelId,
+		Token:            stToken,
+		Sender:           hashedAddress,
+		Receiver:         autopilotMetadata.IbcReceiver,
```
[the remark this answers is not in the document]
sampocs: I was wondering this as well, but we don't know the prefix at the time (since they could be sending it anywhere), so I don't think we can validate it. 

But if the address is invalid, they transfer will just fail and the tokens will fallback to the stride receiver so I don't think it's a concern if an invalid address is passed!

x/autopilot/module_ibc.go
riley-stride: Leaving this one open as a reminder to Aidan to scrutinize this section. 

@asalzmann or @sampocs, feel free to resolve this once Aidan reviews.

x/autopilot/types/autopilot.go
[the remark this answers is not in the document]
riley-stride: > it guarantees it wont be modified without us knowing
this is a really strong point imo, I'm in favor of leaving it as is

x/autopilot/keeper/liquidstake.go
riley-stride: Leaving this open until @asalzmann chimes in. Please feel free to close thereafter.

sampocs:
@asalzmann @riley-stride 

I now remember why I thought option (3) would be messy. it's because we call the same TryLiquidStake function regardless of whether we're doing just LS, or LS&Forward. Which means we have to have a clunky switch statement to check whether we want to use the hashed receiver or original receiver.

I implemented option (3) [here](https://github.com/Stride-Labs/stride/pull/1046) to make things more concrete. (Funny enough option (1) and (3) are the **exact** same number of LOC haha).

---

I personally think option (1) is a bit cleaner than (3) because we only have action-specific stuff in one place instead of two. And more importantly, we don't duplicate logic between the two [ex: [here](https://github.com/Stride-Labs/stride/pull/1038/files#diff-72d80985ee3cc3b2c9a66d97d351c9530307f87a845191907fc40207946ee82cR89) and [here](https://github.com/Stride-Labs/stride/pull/1046/files#diff-37f5036a7893e665fdf8a4402b96e107dcf7e4781d27c311a57821944593f8daR180)].

That said, I'm still open to option (3) if that's your preference!

x/autopilot/keeper/liquidstake.go
```diff
@@ -16,103 +15,141 @@ import (
...
+	ctx sdk.Context,
+	amount sdkmath.Int,
+	transferMetadata transfertypes.FungibleTokenPacketData,
+	autopilotMetadata types.StakeibcPacketMetadata,
+) error {
 	msg := &stakeibctypes.MsgLiquidStake{
-		Creator:   addr.String(),
-		Amount:    token.Amount,
-		HostDenom: token.Denom,
+		Creator:   transferMetadata.Receiver,
```
asalzmann, line 70: is this address still verified to be a valid stride address?

x/autopilot/keeper/liquidstake.go
asalzmann: By default, I'd opt to use the inbound channel id (given that's what PFM does). 

Actually, it's not entirely clear to me why they use the channel id in the has, so I'd be hesitant to make a decision on this before understanding that.

My understanding is as follows
* The bug in PFM was that anyone could forward via arbitrary addresses, leading to untrusted senders
* To resolve this, the high-level idea was to obfuscate senders by hashing them
* To implement this solution, the hash was hash(channel, sender)

Shouldn't hashing the sender alone obfuscate the sender for downstream chains?

@agouin would you be able to weigh in on why PFM _also_ includes the channel in the hash?

x/autopilot/keeper/liquidstake.go
```diff
@@ -16,103 +15,141 @@ import (
...
+	hashedSender, err := sdk.AccAddressFromBech32(hashedAddress)
+	if err != nil {
+		return err
+	}
+	if err := k.bankKeeper.SendCoins(ctx, originalReceiver, hashedSender, sdk.NewCoins(stToken)); err != nil {
+		return err
+	}
+
+	// Use the default transfer timeout of 10 minutes
+	timeoutTimestamp := uint64(ctx.BlockTime().UnixNano()) + transfertypes.DefaultRelativePacketTimeoutTimestamp
```
asalzmann, line 140: how are you thinking about setting the timeout?

my default assumption would have been that longer timeouts (on the order of hours) are better

asalzmann (commented):
Thanks for taking this PR!

x/autopilot/module_ibc.go
sampocs: There was a separate [commit](https://github.com/Stride-Labs/stride/pull/1038/commits/99db7240ed37186bac6fe5635ff3735d70da467b) with just the renaming. Sorry forgot to call it out in the PR description

Fwiw, I always try to logically separate commits so if you ever find that a PR seems a bit too large, I would recommend going commit by commit (I usually do try to mention it in the PR but forgot in this case)

x/autopilot/keeper/liquidstake.go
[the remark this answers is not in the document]
sampocs: Just realized the linked code is actually the host zone address. Did you mean the stride or host zone address?

We validate the stride address, but not the host zone one. But I don't think normal IBC transfers validate the destination address (it will just ack fail)

x/autopilot/keeper/liquidstake.go
sampocs: It's verified when we parse the payload [here](https://github.com/Stride-Labs/stride/blob/main/x/autopilot/types/parser.go#L82-L84)!

x/autopilot/keeper/liquidstake.go
sampocs: My thinking was two fold:
* (1) Users are used to a 10minute timeout for transfers. If I was waiting longer than that and my tokens were in limbo, it would make me a bit uneasy
* (2) Timeouts on transfers are very rare. I don't think any of our epochly transfers has ever timed out and those have a 30minute timeout

Not opposed to making it longer, but that was my thought process!

x/autopilot/types/parser.go
```diff
@@ -74,18 +56,25 @@ func (m ClaimPacketMetadata) Validate() error {
...
 	// assume packet forwarding was no used and pass back nil so that autopilot is ignored
 	var raw RawPacketMetadata
 	if err := json.Unmarshal([]byte(metadata), &raw); err != nil {
 		return nil, nil
 	}
 
-	// If no forwarding logic was used for autopilot, return the metadata with each disabled
+	// Packets cannot be used for both autopilot and PFM at the same time
+	// If both fields were provided, reject the packet
+	if raw.Autopilot != nil && raw.Forward != nil {
```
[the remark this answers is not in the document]
sampocs: > Unwind a packet to Stride then use autopilot

I believe in this case, the memo would start as:
```json
{
  "forward": {
    "channel": "umee-to-osmo-channel",

    "next": {
      "forward": {
        "channel": "osmo-to-stride-channel",

        "next": {
          "autopilot": "..."
        }
```

And by the time it got to stride, it would be just
```json
{
   "autopilot": "..."
}
```

---

> I think there's also a slightly different case where Stride first is a hop on the PFM path, then the token lands back on Stride 

I can't think of any examples either. Imo, I think it's fine to reject this case. If you're attempting to use autopilot, I think we can assume the PFM'ing should only tough stride in one place, at the last hop.

x/autopilot/keeper/liquidstake.go
asalzmann: That makes sense! For what it's worth, I've only seen transfers time out in periods of high activity on Osmosis. Two that come to mind - (1) Terra collapse (2) Celestia launch. In both cases, it took me about ~2 hours to get an IBC transfer processed.

In a bull market, we could experience events of this magnitude monthly.

If we end up retrying forwards on timed out packets, it doesn't matter too much what the timeout is. If we limit the retries, I'd bias to either a longer timeout or allowing for more retries

x/autopilot/types/parser.go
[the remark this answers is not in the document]
asalzmann: Sounds good and makes sense to me! Thanks for confirming

x/autopilot/keeper/liquidstake.go
sampocs: I also experienced the timeouts with the Celestia launch, which is partially what motivated this sentiment. (I found myself a bit anxious while I was waiting haha)

Do you think either approach introduces more complexity on the frontend based on how long we cache a users session?

x/autopilot/keeper/liquidstake.go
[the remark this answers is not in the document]
shellvish: I just meant the stride address, in the case where we're not forwarding the tokens! 

Sorry if this is a dumb question, but do you know where we validate the Stride address? I'm pretty sure we do this as well, but just can't find the code off-hand.

x/autopilot/keeper/liquidstake.go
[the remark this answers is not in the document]
shellvish: Ah yeah that makes sense! I was thinking we might be able to get the prefix from the channel-id, but just realized the light client doesn't store prefixes. 

That logic makes sense to me, it doesn't sound like a concern then! Will mark this resolved 🙂

x/autopilot/keeper/liquidstake.go
sampocs: Chatted offline, landed on 3 hour timeouts

x/autopilot/keeper/liquidstake.go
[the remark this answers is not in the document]
sampocs: yeah we verify it in the [[link](https://github.com/Stride-Labs/stride/blob/main/x/autopilot/types/parser.go#L82-L84)] above!

sampocs:
Chatted on line - decided on option 3 and will refactor the autopilot schema to cleanup the switch statement

x/autopilot/keeper/liquidstake.go
sampocs: updated to 3 hours [[here](https://github.com/Stride-Labs/stride/pull/1038/commits/2bf291f924a1587cbb8092237f4602e9d8e3830f)]

x/autopilot/keeper/liquidstake.go
```diff
@@ -136,8 +141,8 @@ func (k Keeper) IBCTransferStToken(
 		return err
 	}
 
-	// Use the default transfer timeout of 10 minutes
-	timeoutTimestamp := uint64(ctx.BlockTime().UnixNano()) + transfertypes.DefaultRelativePacketTimeoutTimestamp
+	// Use a conservative timeout for the transfer
```
riley-stride, line 144: pico nit: conservative -> long

(to me, conservative means safe so is a bit unclear in the context of a timeout; can mean short or long depending on what you interpret as safe)

riley-stride (commented):
timeout looks good to me. 

should we add our rationale for it into the comments? so that if we ever want to change it in the future that context isn't lost.

x/autopilot/keeper/liquidstake.go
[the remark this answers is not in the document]
shellvish: Ah yes! Sounds good, resolving!

x/autopilot/keeper/liquidstake.go
sampocs: good call, updated!

sampocs:
> should we add our rationale for it into the comments? so that if we ever want to change it in the future that context isn't lost.

Great idea! Updated [here](https://github.com/Stride-Labs/stride/pull/1038/commits/a814f92d5ea7081b87b25fa669e5b4c9a4978961)

x/autopilot/module_ibc.go
```diff
@@ -127,53 +126,55 @@ func (im IBCModule) OnRecvPacket(
...
 		return channeltypes.NewErrorAcknowledgement(err)
 	}
 
-	// If the parsed metadata is nil, that means there is no forwarding logic
+	// If the parsed metadata is nil, that means there is no autopilot forwarding logic
 	// Pass the packet down to the next middleware
-	if packetForwardMetadata == nil {
+	// PFM packets will also go down this path
+	if autopilotMetadata == nil {
 		return im.app.OnRecvPacket(ctx, packet, relayer)
```
asalzmann, line 169: Why do we have `return im.app.OnRecvPacket` twice  - here and a few lines below? 

Recommend we consolidate the two if statements and make the reasons for triggering `return im.app.OnRecvPacket(ctx, packet, relayer)` more clear 

I think the reasons are 
(1) if the memo is nil and the receiver is an address, it's a normal IBC packet (so short-circuit autopilot)
(2) if the memo is _not_ empty, but can't be parsed by autopilot, pass it down the stack

sampocs: good observation!

sampocs: although, I think the reasoning was so we didn't mess around with autopilot related functionality when it's not an autopilot packet.

Did you have a specific refactor in mind? If not, I can take a pass at it

asalzmann: Fiddled with it - I think it's actually clearest as written. Don't see a straightforward way to simplify. Sorry

x/autopilot/keeper/liquidstake.go
asalzmann: Riley: Hey team we're shipping an upgrade to autopilot that hashes channel and sender (like PFM), but are considering removing the channel from the hash inputs. Tagged you guys in our PR here with more context: https://github.com/Stride-Labs/stride/pull/1038#discussion_r1443767344
Do you know why PFM includes the channel in the hash?

Aidan: My impression of the issue was that a previous version of PFM passed along untrusted senders, and the solution was to obfuscate the sender by hashing it. Doesn't hashing just the sender obfuscate it? Why do you also need the transfer channel?

Andrew: Just the sender should be enough afaict. We followed the same pattern as IBC hooks. I didn't see a problem with including the channel so I included it also.

It adds more uniqueness but the bech32 prefix of the source sender is included in the hash anyways

So it's mainly just extra collision prevention


**Based on this, would probably use the incoming channel id to stay consistent with PFM. 

But if it's a lot of extra work, it's not so important.**

sampocs: Based on that, I think I'd vote to leave it how we have it in this PR. Avoids having to pass an extra parameter across multiple functions and gives the same (weak) collision resistance as the inbound channel ID

asalzmann: sgtm

riley-stride: sgtm too!

x/autopilot/keeper/liquidstake.go
sampocs: @riley-stride am I good to resolve?

riley-stride: resolving

x/autopilot/keeper/liquidstake.go
sampocs: actually nvm, now that we're doing option 3, it's actually the same amount of work to use the same channel as PFM (and it's already implemented that way in the option 3 PR)

PR #1041, opened 2024-01-03
autopilot liquid stake and forward unit tests

sampocs:
## Context
Fixed, cleaned up and added coverage to autopilot liquid stake and forward unit tests.

## Brief Changelog
* Slimmed down mocked state to only what's necessary
* Added coverage to all error cases (by testing inner TryLiquidStake function instead of OnRecvPacket)
* Improved error messages


asalzmann (approved):
these changes all lgtm - thanks for significantly cleaning up x/autopilot/keeper/liquidstake_test.go !

PR #1045, opened 2024-01-05
send disable tokenization tx for delegation account on hub

ethan-stride:
Runs the disable tokenization ICA tx for the delegation ICA on the hub
Should prevent fast and large withdrawal attempts using tokenization


proto/stride/stakeibc/ica_account.proto
```diff
@@ -20,3 +22,13 @@ message ICAAccount {
...
+
+// MsgDisableTokenizeShares prevents LSM tokenization of shares for a given
+// address
+message MsgDisableTokenizeShares {
+  option (gogoproto.equal) = false;
+  option (gogoproto.goproto_getters) = false;
+
+  string delegator_address = 1
+      [ (gogoproto.moretags) = "yaml:\"delegator_address\"" ];
+}
```
sampocs, line 34: move this to `proto/cosmos/.../lsm_tx.proto`

app/upgrades/v17/upgrades.go
```diff
@@ -101,10 +103,36 @@ func CreateUpgradeHandler(
 			return vm, errorsmod.Wrapf(err, "unable to add rate limits to Osmosis")
 		}
 
+		ctx.Logger().Info("Disabling tokenization on the hub...")
+		DisableTokenization(ctx, stakeibcKeeper, "cosmoshub-4")
```
sampocs, line 107: add `GaiaChainId` to constants at top

app/upgrades/v17/upgrades.go
```diff
@@ -101,10 +103,36 @@ func CreateUpgradeHandler(
...
+
 		return mm.RunMigrations(ctx, configurator, vm)
 	}
 }
 
+// Sends the ICA message which disables LSM style tokenization of shares from the delegation
+// account for this chain as a security to prevent possibility of large/fast withdrawls
+func DisableTokenization(ctx sdk.Context, k stakeibckeeper.Keeper, chainId string) error {
+	for _, hostZone := range k.GetAllHostZone(ctx) {
+		if hostZone.ChainId == chainId {
```
sampocs, line 117:
```
hostZone, found := k.GetHostZone(ctx, chainId)
```

app/upgrades/v17/upgrades.go
```diff
@@ -101,10 +103,36 @@ func CreateUpgradeHandler(
...
+			// Build the msg for the disable tokenization ICA tx
+			var msgs []proto.Message
+			msgs = append(msgs, &stakeibctypes.MsgDisableTokenizeShares{
+				DelegatorAddress: hostZone.DelegationIcaAddress,
+			})
+				
+			// Send the ICA tx to disable tokenization
+			timeoutTimestamp := uint64(time.Now().Add(24 * time.Hour).Nanosecond())
```
sampocs, line 125:
```suggestion
			timeoutTimestamp := uint64(ctx.BlockTime().Add(24 * time.Hour).Nanosecond())
```

sampocs (commented):
looks good so far! Couple small comments and then needs the unit tests

app/upgrades/v17/upgrades.go
ethan-stride: So we do have access to the block time and the epoch info when running the upgrade handler, it is just unwise to depend on the next epoch as the timeout point because this could be run at any point through the epoch and could be close to the end. Makes sense

ethan-stride:
LGTM thanks for figuring out the unit test

riley-stride:
Has this been tested on Dockernet?

riley-stride (approved):
needs to be tested on dockernet. otherwise lgtm

PR #1046, opened 2024-01-05
autopilot hashed sender option (3)

sampocs:
Implemented option 3 from the discussion in #1038 where, for LS and forwards, we update the receiver before passing it down the stack


x/autopilot/module_ibc.go
```diff
@@ -169,9 +169,31 @@ func (im IBCModule) OnRecvPacket(
...
-	tokenPacketData.Receiver = autopilotMetadata.Receiver
+	// Update the reciever in the packet data so that we can pass the packet down the stack
+	// (since the "receiver" may have technically been a full JSON memo)
+	switch routingInfo := autopilotMetadata.RoutingInfo.(type) {
+	case types.StakeibcPacketMetadata:
+		// For autopilot liquid stake and forward, we'll use a hashed receiver address instead
+		// that will also be used as the sender in the forwarding step
+		// This is to prevent impersonation at downstream zones
+		// We can identify the forwarding step by whether there's a non-empty IBC receiver field
+		if routingInfo.Action == types.LiquidStake && routingInfo.IbcReceiver != "" {
```
riley-stride, line 180: Is `routingInfo.IbcReceiver != ""` the best way to identify that a packet is a LS&forward packet?

riley-stride (commented):
Looking good. Seeing it implemented, I prefer this approach! Wdyt?

x/autopilot/module_ibc.go
sampocs: this is temporary for now - will replace with a switch on the action after we change the schema

x/autopilot/module_ibc.go
```diff
@@ -169,9 +169,31 @@ func (im IBCModule) OnRecvPacket(
 
 	//// At this point, we are officially dealing with an autopilot packet
 
-	// Modify the packet data by replacing the JSON metadata field with a receiver address
-	// to allow the packet to continue down the stack
-	tokenPacketData.Receiver = autopilotMetadata.Receiver
+	// Update the reciever in the packet data so that we can pass the packet down the stack
+	// (since the "receiver" may have technically been a full JSON memo)
+	switch routingInfo := autopilotMetadata.RoutingInfo.(type) {
```
asalzmann, line 174: Wdyt of simplifying like so?

```
// Set default receiver
tokenPacketData.Receiver = autopilotMetadata.Receiver

// Override receiver for autopilot liquid stake with non-empty IBC receiver
if routingInfo, ok := autopilotMetadata.RoutingInfo.(types.StakeibcPacketMetadata); ok &&
   routingInfo.Action == types.LiquidStake && routingInfo.IbcReceiver != "" {

    var err error
    tokenPacketData.Receiver, err = types.GenerateHashedAddress(packet.DestinationChannel, tokenPacketData.Sender)
    if err != nil {
    }
}
```

sampocs: Done [[here](https://github.com/Stride-Labs/stride/pull/1046/commits/4a8269211d5f90cfc34c3ecc109c5ff1aeb75473)]

asalzmann: lgtm!

asalzmann (approved):

PR #1047, opened 2024-01-06
autopilot fallback address with callbacks

sampocs:
## Context
Took a first pass at implementing the fallback address using ibc callbacks instead of the keepers. **This PR is far from complete**, I'm just opening it so that you all can have a more concrete representation of what this solution would look like.

## Potential Issues
**Since autopilot is in a different stack, there's a significant amount of refactoring that would be needed to complete the callback wiring.** 

Most notably, as it currently sits in this PR, the callback will technically be [executed](https://github.com/Stride-Labs/stride/blob/main/x/records/keeper/ibc.go#L26-L31) via the records module. However:
* (a) The records callback runs before the packet is sent down the stack (meaning the tokens wont be available to send to the fallback address), and
* (b) There's a comment in the log linked above that says the callback will only work for transfers from the stakeibc module address (and I don't fully understand what it means but can dig in more later)

## Remaining Work
* Correctly execute the autopilot callback
  * Initial reaction is to sub out the records module from our stack with the callbacks module (like how we have it in stakeibc). That way callbacks can route to the relevant module (records vs autopilot).
    * However, the transfer module has to sit at the bottom of the stack, and the callbacks module is currently implemented such that it sits at the bottom of the stack, so it might not be trivial to do this.
  * Otherwise, another option would be to `CallRegisteredICACallback` from the autopilot OnAck/OnTimeout's (instead of from records). 
    * However, this would require adding a guardrail to each module to make sure it doesn't run the callback from the other module.
* Once the above is solved, we still need to make sure the packet is sent down the packet first. Unclear what changes are required for this atm.
* Finally, would need to clean this up, test, and add unit tests


sampocs:
closing this PR as we decided that this was too complex of an undertaking for the current timeline.

We'll revisit in the next few months what the best design for our middleware stack is.

PR #1049, opened 2024-01-09
Updates User Redemption Records dynamically before unbonding

shellvish:
WIP - Still need to add unit tests.

High level, this change makes it so redemption rates get set at the time that unbondings are initiated, not at the time of redemption. 

I would focus on a few main things:
- Does it ever break any assumptions if the NativeTokenAmount on a HostZone is set to 0 before unbondings finish?
- Do we handle failure cases (e.g. callback errors or times out) correctly? 
- Do we migrate our current data structures over properly?


app/upgrades/v17/upgrades.go
```diff
@@ -303,3 +309,61 @@ func AddRateLimitToOsmosis(ctx sdk.Context, k ratelimitkeeper.Keeper) error {
...
+			nativeTokenAmountDec := sdk.NewDecFromInt(hostZoneUnbonding.NativeTokenAmount)
+			stTokenAmountDec := sdk.NewDecFromInt(hostZoneUnbonding.StTokenAmount)
+			// this estimated rate is the amount of stTokens that would be received for 1 native token
+			// e.g. if the rate is 0.5, then 1 native token would be worth 0.5 stTokens
+			estimatedStTokenConversionRate := stTokenAmountDec.Quo(nativeTokenAmountDec)
+
+			// store if the unbonding has not been initiated
+			unbondingNotInitiated := hostZoneUnbonding.Status == recordtypes.HostZoneUnbonding_UNBONDING_QUEUE
+
+			// Loop through User Redemption Records and insert an estimated stTokenAmount
```
riley-stride, line 340: Do we want update records that have the following statuses all in the same way?

1. `HostZoneUnbonding_UNBONDING_QUEUE`
2. `HostZoneUnbonding_UNBONDING_IN_PROGRESS` <= is there a risk we are going to catch any of these in flight?
3. `HostZoneUnbonding_EXIT_TRANSFER_QUEUE`
4. `HostZoneUnbonding_EXIT_TRANSFER_IN_PROGRESS`

app/upgrades/v17/upgrades.go
```diff
@@ -303,3 +309,61 @@ func AddRateLimitToOsmosis(ctx sdk.Context, k ratelimitkeeper.Keeper) error {
...
+			for _, userRedemptionRecordId := range hostZoneUnbonding.GetUserRedemptionRecords() {
+				userRedemptionRecord, found := k.RecordsKeeper.GetUserRedemptionRecord(ctx, userRedemptionRecordId)
+				if !found {
+					// this would happen if the user has already claimed the unbonding, but given the status check above, this should never happen
+					k.Logger(ctx).Error(fmt.Sprintf("user redemption record %s not found", userRedemptionRecordId))
+					continue
+				}
+
+				userRedemptionRecord.StTokenAmount = estimatedStTokenConversionRate.Mul(sdkmath.LegacyDec(userRedemptionRecord.Amount)).RoundInt()
```
riley-stride, line 349: conversion math lgtm

riley-stride (commented):
Looking good, thanks for taking point on this one! Added some questions and suggestions.

x/stakeibc/keeper/unbonding_records.go
```diff
@@ -74,6 +74,14 @@ func (k Keeper) CreateEpochUnbondingRecord(ctx sdk.Context, epochNumber uint64)
 	return true
 }
 
+// Helper function to evaluate if a host zone unbonding record still needs to be initiated
+func (k Keeper) ShouldInitiateHostZoneUnbondingRecord(ctx sdk.Context, hostZoneRecord *recordstypes.HostZoneUnbonding) bool {
+	if hostZoneRecord.Status == recordstypes.HostZoneUnbonding_UNBONDING_QUEUE && hostZoneRecord.NativeTokenAmount.GT(sdkmath.ZeroInt()) {
```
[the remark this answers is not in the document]
shellvish: 🤦 

Yes, you're 100% right. Great catch, fixing!

x/stakeibc/keeper/msg_server_redeem_stake.go
```diff
@@ -89,16 +89,17 @@ func (k msgServer) RedeemStake(goCtx context.Context, msg *types.MsgRedeemStake)
 		k.Logger(ctx).Info(fmt.Sprintf("UserRedemptionRecord found for %s", redemptionId))
 		// Add the unbonded amount to the UserRedemptionRecord
 		// The record is set below
-		userRedemptionRecord.Amount = userRedemptionRecord.Amount.Add(nativeAmount)
+		userRedemptionRecord.StTokenAmount = userRedemptionRecord.StTokenAmount.Add(msg.Amount)
```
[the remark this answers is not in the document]
shellvish: Yeah I was wondering this too, I agree let's remove it. I'm going to delete inCoin from this file (I don't believe it's used outside of logs)

x/stakeibc/keeper/msg_server_redeem_stake.go
```diff
@@ -89,16 +89,17 @@ func (k msgServer) RedeemStake(goCtx context.Context, msg *types.MsgRedeemStake)
...
 		userRedemptionRecord = recordstypes.UserRedemptionRecord{
-			Id:          redemptionId,
-			Receiver:    msg.Receiver,
-			Amount:      nativeAmount,
-			Denom:       hostZone.HostDenom,
-			HostZoneId:  hostZone.ChainId,
-			EpochNumber: epochTracker.EpochNumber,
+			Id:            redemptionId,
+			Receiver:      msg.Receiver,
+			Amount:        sdk.ZeroInt(),
```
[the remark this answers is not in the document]
shellvish: Definitely, totally agree! Made a separate PR for it just to avoid cluttering this PR: https://github.com/Stride-Labs/stride/pull/1051

app/upgrades/v17/upgrades.go
```diff
@@ -303,3 +309,61 @@ func AddRateLimitToOsmosis(ctx sdk.Context, k ratelimitkeeper.Keeper) error {
...
+	for _, epochUnbondingRecord := range k.RecordsKeeper.GetAllEpochUnbondingRecord(ctx) {
+		for _, hostZoneUnbonding := range epochUnbondingRecord.GetHostZoneUnbondings() {
+			// if the tokens have unbonded, we don't want to modify the record
+			// this is because we won't be able to estimate a redemption rate
+			if hostZoneUnbonding.Status == recordtypes.HostZoneUnbonding_CLAIMABLE {
+				continue
+			}
+			// similarly, if there aren't any tokens to unbond, we don't want to modify the record
+			// as we won't be able to estimate a redemption rate
+			if hostZoneUnbonding.NativeTokenAmount == sdkmath.ZeroInt() {
```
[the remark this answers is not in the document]
shellvish: I think the only case where this is true but _not_ the above if-statement is if no tokens were redeemed in that day-epoch

app/upgrades/v17/upgrades.go
shellvish: Thanks for verifying!

x/stakeibc/keeper/unbonding_records.go
```diff
@@ -96,6 +111,74 @@ func (k Keeper) GetTotalUnbondAmountAndRecordsIds(ctx sdk.Context, chainId strin
...
+		}
+		userRedemptionRecord.Amount = nativeAmount
+		totalNativeAmount = totalNativeAmount.Add(nativeAmount)
+		k.RecordsKeeper.SetUserRedemptionRecord(ctx, userRedemptionRecord)
+	}
+	// update the host zone unbonding record
+	hostZoneUnbondingRecord.NativeTokenAmount = totalNativeAmount
+	updatedEpochUnbondingRecord, success := k.RecordsKeeper.AddHostZoneToEpochUnbondingRecord(ctx, epochNumber, chainId, hostZoneUnbondingRecord)
+	if !success {
+		// TODO QUESTION: what should we do on failure? do we revert the userRedemptionRecord changes above?
```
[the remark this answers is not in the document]
shellvish: Yeah that makes sense! The reason I threw an error in the first location is because we need that function call to succeed (otherwise we don't know how many tokens to unbond), but we don't strictly need it in the other function call (it's only used for housekeeping niceness). 

That being said, I lean towards throwing an error regardless. I put the tradeoffs in [here](https://github.com/Stride-Labs/stride/pull/1049#discussion_r1446777459), my recommendation is we get one-notch more confident that the function won't fail, and then we choose to always revert.

x/stakeibc/keeper/msg_server_redeem_stake.go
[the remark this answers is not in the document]
riley-stride: Fixes [here](https://github.com/Stride-Labs/stride/pull/1049/commits/69a9b7be69dae7a6be10b525be6211656a5544db) lgtm

x/stakeibc/keeper/msg_server_redeem_stake.go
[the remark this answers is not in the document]
riley-stride: Just reviewed that PR. Resolving this comment.

x/stakeibc/keeper/unbonding_records.go
```diff
@@ -96,6 +111,74 @@ func (k Keeper) GetTotalUnbondAmountAndRecordsIds(ctx sdk.Context, chainId strin
...
+		k.Logger(ctx).Info(fmt.Sprintf("Updating native token amount on host zone unbondings on EpochUnbondingRecord %d to zero", epochUnbondingRecordId))
+		// fetch the host zone unbonding
+		hostZoneUnbonding, found := k.RecordsKeeper.GetHostZoneUnbondingByChainId(ctx, epochUnbondingRecordId, chainId)
+		if !found {
+			errMsg := fmt.Sprintf("Error fetching host zone unbonding record for epoch: %d, host zone: %s", epochUnbondingRecordId, chainId)
+			k.Logger(ctx).Error(errMsg)
+			return errorsmod.Wrapf(types.ErrHostZoneNotFound, errMsg)
+		}
+		success := k.UpdateNativeTokensForHostZoneUnbondingRecord(ctx, epochUnbondingRecordId, hostZoneUnbonding, true)
+		if !success {
```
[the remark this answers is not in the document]
riley-stride: > 1. Currently, we don't actually need the output of UpdateNativeTokensForHostZoneUnbondingRecord to succeed, it's mainly housekeeping to keep the records neat. If this failed for whatever reason, it would be nice to not halt the chain / break critical functionality

If it succeeded up to and including [this line](https://github.com/Stride-Labs/stride/pull/1049/files/a71b9b47e72f6454a585e88c5cee8a7b24ec6553#diff-5dd4bc8c0f6254ddad2534b66e244b9576d32326538b71c0051ea58eacc6f1e6R355), then `AddHostZoneToEpochUnbondingRecord` failed and we didn't revert, would we end up in a bad state? Afaict the `totalNativeAmount` would be incremented and written to the URR, but the corresponding HZUR would not be updated, so the two would be out of sync.

app/upgrades/v17/upgrades.go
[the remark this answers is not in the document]
riley-stride: makes sense. and makes sense to continue in that case. resolving.

x/stakeibc/keeper/unbonding_records.go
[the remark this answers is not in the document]
riley-stride: Added a comments below at the [link](https://github.com/Stride-Labs/stride/pull/1049#discussion_r1446777459). (Low confidence) I think the records could get out of sync if we don't revert. Agree we should get one-notch more confident that the function won't fail, and then we choose to always revert.

app/upgrades/v17/upgrades.go
riley-stride: 1. Yep, makes sense. 
2. Thinking about this a bit more, not only will nothing bad happen here, in fact we _need_ to migrate these or their undelegate callbacks will fail because in `UndelegateCallback` we invoke `SetNativeTokensToZeroInUnbondingRecords` which references their `stTokenAmount` attribute. If they have no `stTokenAmount` that callback will return err. 
3. Agreed, we should migrate it. And feels safe: I can't find a reference to stTokenAmount in any code post-UndelegateCallback. Imo this is the same as 4 but different than 2 (2 _needs_ the migration to avoid erroring).
4. Yeah I think we should migrate it. Imo this is the same as 3 but different than 2 (2 _needs_ the migration to avoid erroring).

x/stakeibc/keeper/unbonding_records.go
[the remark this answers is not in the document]
riley-stride: I think your fix [here](https://github.com/Stride-Labs/stride/pull/1049/commits/69a9b7be69dae7a6be10b525be6211656a5544db) works, since `NativeTokenAmount` will (once set) equal `stTokenAmount` times `RR`. So checking `stTokenAmount` for equality to 0 has the same effect as checking whether the value `NativeTokenAmount` will eventually be set to equals 0.

x/stakeibc/keeper/unbonding_records.go
```diff
@@ -74,6 +74,14 @@ func (k Keeper) CreateEpochUnbondingRecord(ctx sdk.Context, epochNumber uint64)
 	return true
 }
 
+// Helper function to evaluate if a host zone unbonding record still needs to be initiated
+func (k Keeper) ShouldInitiateHostZoneUnbondingRecord(ctx sdk.Context, hostZoneRecord *recordstypes.HostZoneUnbonding) bool {
+	if hostZoneRecord.Status == recordstypes.HostZoneUnbonding_UNBONDING_QUEUE && hostZoneRecord.StTokenAmount.GT(sdkmath.ZeroInt()) {
+		return true
+	}
+	return false
```
sampocs, line 82:
```suggestion
	return hostZoneRecord.Status == recordstypes.HostZoneUnbonding_UNBONDING_QUEUE && hostZoneRecord.StTokenAmount.GT(sdkmath.ZeroInt()) 
```

x/stakeibc/keeper/unbonding_records.go
```diff
@@ -96,6 +111,74 @@ func (k Keeper) GetTotalUnbondAmountAndRecordsIds(ctx sdk.Context, chainId strin
 	return totalUnbonded, unbondingRecordIds
 }
 
+// Update Native Amounts in the Host Zone Unbonding Record (and associated User Redemption Records) to reflect the current redemption rate
```
sampocs, line 114: A couple nit comments in this PR to clean things up a bit and keep things standardized:
* Use `.Attribute` instead of `.GetAttribute()`
* Add some vertical whitespace between logical groups of logic (makes the review much easier! - similar to new line breaks in a paragraph)
* We don't have to do that `errMsg` variable format anymore, and no need to log each error cause their logged upstream, we can just return the wrapped error

x/stakeibc/keeper/unbonding_records.go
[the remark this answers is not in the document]
sampocs: Even if we throw an error, state doesn't get reverted since this is in the begin blocker. We should use a cache context wrapper around this function and throw an error

x/stakeibc/keeper/unbonding_records.go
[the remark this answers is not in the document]
sampocs: Mentioned in the other thread as well - state does not get reverted in the begin/end blocker even if we return an error. We should wrap this in a temporary context so we can discard state changes if an error is thrown. I believe this is already implemented in an old PR

x/stakeibc/keeper/unbonding_records.go
```diff
@@ -74,6 +74,14 @@ func (k Keeper) CreateEpochUnbondingRecord(ctx sdk.Context, epochNumber uint64)
 	return true
 }
 
+// Helper function to evaluate if a host zone unbonding record still needs to be initiated
+func (k Keeper) ShouldInitiateHostZoneUnbondingRecord(ctx sdk.Context, hostZoneRecord *recordstypes.HostZoneUnbonding) bool {
```
sampocs, line 78: thoughts on renaming to `ShouldHostZoneRecordUnbond`?

x/stakeibc/keeper/unbonding_records.go
```diff
@@ -87,7 +95,14 @@ func (k Keeper) GetTotalUnbondAmountAndRecordsIds(ctx sdk.Context, chainId strin
...
-		if hostZoneRecord.Status == recordstypes.HostZoneUnbonding_UNBONDING_QUEUE && hostZoneRecord.NativeTokenAmount.GT(sdkmath.ZeroInt()) {
+		if k.ShouldInitiateHostZoneUnbondingRecord(ctx, hostZoneRecord) {
+			// Dynamically calculate the unbonding amount based on the current redemption rate
+			success := k.UpdateNativeTokensForHostZoneUnbondingRecord(ctx, epochUnbonding.EpochNumber, hostZoneRecord, false)
+			if !success {
+				k.Logger(ctx).Error(utils.LogWithHostZone(chainId, "Failed to update native tokens for host zone unbonding record in epoch %d", epochUnbonding.EpochNumber))
+				continue
+			}
+
 			totalUnbonded = totalUnbonded.Add(hostZoneRecord.NativeTokenAmount)
```
sampocs, line 106: are you sure this summation will work? I think the native amount might always be 0 here

sampocs (commented):
sorry kinda running out of steam rn, but just dropping this partial review for now and will finish the review first thing in the morning!

x/stakeibc/keeper/unbonding_records.go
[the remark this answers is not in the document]
shellvish: Ah yeah that's a great point Riley! Agreed, I think these will get out-of-sync the way it's currently written

Sam, do you have an example of the temporary context? I can implement it here, I think that's likely the best path forward

app/upgrades/v17/upgrades.go
shellvish: Ah yeah great call re:(2)! Let's keep this as is then, seems like the migration is good for all of the cases

riley-stride: Agreed let's keep it as is

riley-stride: Leaving this open so @sampocs and @asalzmann can think through each case with this context as well. Feel free to resolve after.

x/stakeibc/keeper/unbonding_records.go
riley-stride: Imo the proposed change is more clear.

x/stakeibc/keeper/unbonding_records.go
riley-stride: Good eye. Agreed I don't think this will work. @shellvish @sampocs should we run this PR through the dockernet integration tests to make sure the high level flow works? Would be nice to have that guarantee as we review.

x/stakeibc/keeper/unbonding_records.go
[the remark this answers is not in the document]
sampocs: https://github.com/Stride-Labs/stride/pull/861/files#diff-ba71e9ef6bf1ebcc5243683cbd44405947d8109d7ffb825c7c353d256f7e2179R355-R358

x/stakeibc/keeper/unbonding_records.go
sampocs: we should try and clean this up a bit. It's a bit dangerous to update pointers inside functions and then make assumptions in the calling function

In general, I think we should either:
* Return the struct from the function
* Write to state in the updating function, and read that value back in the calling function 

I'll take a peak at what makes the most sense here

sampocs:
Considering our unbonding flow is already pretty tricky, I think we should optimize for readability over performance.

One paradigm that I think helps remove implicit assumptions is having a "main" like function (UnbondFromHost) and then having the sub functions be small/focused and explicit about whether they are just doing calculations or are updating state.

That said, I think we should disentangle this `SetNativeAmount` functions from `GetTotalUnbondAmountAndRecordsIds`. So in the main we have something like:
```
// set the native amount for each EUR
// sum the native amount and get each EUR
```

This requires we do another loop, but then we don't have to modify the existing `GetTotalUnbondAmountAndRecordsIds` function at all, we can keep it focused on the one task, and we don't have to rely on pointers/global side effects. 

If you're all in agreement with this, I can open a PR to this one

x/stakeibc/keeper/msg_server_restore_interchain_account.go
```diff
@@ -90,8 +90,15 @@ func (k msgServer) RestoreInterchainAccount(goCtx context.Context, msg *types.Ms
...
+		// Set UNBONDING_IN_PROGRESS records NativeTokenAmounts to 0
+		err := k.SetNativeTokensToZeroInUnbondingRecords(ctx, hostZone.ChainId, epochNumberForPendingUnbondingRecords)
+		if err != nil {
+			errMsg := fmt.Sprintf("unable to set native token amounts to 0 for chainId: %s and epochUnbondingRecordIds: %v, err: %s")
+			k.Logger(ctx).Error(errMsg)
+			// TODO Question: should we return an error here, or just proceed?
+		}
 		// Revert UNBONDING_IN_PROGRESS records to UNBONDING_QUEUE
-		err := k.RecordsKeeper.SetHostZoneUnbondings(ctx, hostZone.ChainId, epochNumberForPendingUnbondingRecords, recordtypes.HostZoneUnbonding_UNBONDING_QUEUE)
+		err = k.RecordsKeeper.SetHostZoneUnbondings(ctx, hostZone.ChainId, epochNumberForPendingUnbondingRecords, recordtypes.HostZoneUnbonding_UNBONDING_QUEUE)
```
asalzmann, line 101: Should we rename `SetHostZoneUnbondings` -> `SetHostZoneUnbondingsStatus` for clarity?

x/stakeibc/keeper/msg_server_restore_interchain_account.go
```diff
@@ -90,8 +90,15 @@ func (k msgServer) RestoreInterchainAccount(goCtx context.Context, msg *types.Ms
 				epochNumberForPendingTransferRecords = append(epochNumberForPendingTransferRecords, epochUnbondingRecord.EpochNumber)
 			}
 		}
+		// Set UNBONDING_IN_PROGRESS records NativeTokenAmounts to 0
+		err := k.SetNativeTokensToZeroInUnbondingRecords(ctx, hostZone.ChainId, epochNumberForPendingUnbondingRecords)
+		if err != nil {
+			errMsg := fmt.Sprintf("unable to set native token amounts to 0 for chainId: %s and epochUnbondingRecordIds: %v, err: %s")
+			k.Logger(ctx).Error(errMsg)
+			// TODO Question: should we return an error here, or just proceed?
```
asalzmann, line 98: imo we should return an error - we don't want to re-open the ICA channel if we're in a partially recovered state.

x/stakeibc/keeper/msg_server_restore_interchain_account.go
asalzmann: Also, would it make sense to consolidate `SetNativeTokensToZeroInUnbondingRecords` and `SetHostZoneUnbondings`?

x/stakeibc/keeper/unbonding_records.go
```diff
@@ -96,6 +111,74 @@ func (k Keeper) GetTotalUnbondAmountAndRecordsIds(ctx sdk.Context, chainId strin
...
+	// loop through user redemption records for the host zone unbonding record, update each one
+	for _, userRedemptionRecordId := range hostZoneUnbondingRecord.GetUserRedemptionRecords() {
+		userRedemptionRecord, found := k.RecordsKeeper.GetUserRedemptionRecord(ctx, userRedemptionRecordId)
+		if !found {
+			k.Logger(ctx).Info(utils.LogWithHostZone(chainId, "No user redemption record found for id %s", userRedemptionRecordId))
+			continue
+		}
+		// update the amount, based on the current redemption rate (or set to zero if desired)
+		nativeAmount := sdkmath.ZeroInt()
+		if !setNativeAmountsToZero {
```
asalzmann, line 140: imo we shouldn't mix this into the function - we should have 
```
UpdateNativeTokensForHostZoneUnbondingRecord
ResetNativeTokensForHostZoneUnbondingRecord
```
and these could optionally call a sub function

the current setup of `SetNativeTokensToZeroInUnbondingRecords` calling `UpdateNativeTokensForHostZoneUnbondingRecord` is confusing

x/stakeibc/keeper/unbonding_records.go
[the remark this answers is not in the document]
asalzmann: it's not necessarily in the begin blocker - re-opening ICAs is just a tx

x/stakeibc/keeper/unbonding_records.go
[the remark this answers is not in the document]
asalzmann: same comment as above - this is called outside of the begin blocker (acks, tx to re-open ICAs)

asalzmann (commented):
didn't have time to review the migration logic yet (but planning on doing that last, since it's more straightforward than implementing the functions themselves)


* Do we handle failure cases (e.g. callback errors or times out) correctly?

I think so

* Do we migrate our current data structures over properly?

Haven't checked yet

* Does it ever break any assumptions if the NativeTokenAmount on a HostZone is set to 0 before unbondings finish

Can you clarify this question? Do you mean HostZoneUnbonding?

If so, I think we clear out EURs when all HZUs have `native_token_amount == 0`

sampocs:
closing in favor of #1053

PR #1051, opened 2024-01-10
Rename "amount" to "nativeTokenAmount" in UserRedemptionRecord

shellvish:
Just renames a variable in the protos!

All I did to verify this worked was checking that the build passed. After unit tests are done, will also check unit tests. Please let me know if there are other ways to verify this worked as expected!


proto/stride/records/records.proto
```diff
@@ -9,7 +9,7 @@ option go_package = "github.com/Stride-Labs/stride/v16/x/records/types";
 message UserRedemptionRecord {
   string id = 1; // {chain_id}.{epoch}.{receiver}
   string receiver = 3;
-  string amount = 4 [
+  string native_token_amount = 4 [
```
riley-stride, line 12: Can we rename a field like this? It's possible we need to depracate the field and create a new one with a new index, I'm not sure. I bet @sampocs will know.

riley-stride (approved):
lgtm! 

only q is about changing the proto field name, do we need to depracate the field and create a new one to change its name?

proto/stride/records/records.proto
sampocs: no need to deprecate or create a new index, just a renaming is fine!

shellvish: It's a good question though! We should definitely double check this in localstride (along with all of the unbonding records)

sampocs: For sure! Fwiw, we did this on our host zone struct in the lsm upgrade - renamed all ica address fields and did not do a migration

riley-stride: Great. Resolving.

PR #1053, opened 2024-01-10
dynamically update user redemption records during unbonding

sampocs:
## Context
(See #1049 for more context)

Refactored the dynamic redemption rate during unbonding PR.

Motivation for refactoring:
* Further modularize functions to give more fine grained test coverage
* Explicitly separate steps that modify state vs those that are just doing calculations
* Put the new dynamic unbonding code as upstream as possible (so it's less invasive to the current unbonding code), while still keeping the native calculation and the unbonding ICAs atomic


x/stakeibc/keeper/unbonding_records.go
```diff
@@ -74,18 +74,12 @@ func (k Keeper) CreateEpochUnbondingRecord(ctx sdk.Context, epochNumber uint64)
...
-		return true
-	}
-	return false
-}
-
-// Gets the total unbonded amount for a host zone by looping through the epoch unbonding records
-// Also returns the epoch unbonding record ids
-func (k Keeper) GetTotalUnbondAmountAndRecordsIds(ctx sdk.Context, chainId string) (totalUnbonded sdkmath.Int, unbondingRecordIds []uint64) {
-	totalUnbonded = sdk.ZeroInt()
+// Returns the list of all epoch unbonding record IDs and host zone unbonding records that should unbond this epoch
```
sampocs, line 77: Gonna change that to return a map when I’m back

x/stakeibc/keeper/unbonding_records.go
```diff
@@ -86,14 +90,85 @@ func (k Keeper) GetTotalUnbondAmountAndRecordsIds(ctx sdk.Context, chainId strin
...
+	totalNativeAmount = sdkmath.ZeroInt()
+	for _, userRedemptionRecordId := range redemptionRecordIds {
+		userRedemptionRecord, found := k.RecordsKeeper.GetUserRedemptionRecord(ctx, userRedemptionRecordId)
+		if !found {
+			k.Logger(ctx).Error(utils.LogWithHostZone(chainId, "No user redemption record found for id %s", userRedemptionRecordId))
+			continue
+		}
+
+		// A zero redemption rate is used to indicate that the native amount should be set to zero
+		// If the redemption rate is non-zero, calculate the number of native tokens
```
shellvish, line 129: I _think_ redemption rate should always be positive now in this function, should we explicitly verify that though (e.g. if rr <= 0, throw an error)?

shellvish (commented):
This looks great, thank you for doing this! So much cleaner and easier to reason about with this new paradigm 🙂 

Left a couple of small suggestions, but should be quick to address!

x/stakeibc/keeper/unbonding_records.go
sampocs: chatted offline, no need to throw an error since it's covered by our invariants, but will remove this comment

x/stakeibc/keeper/unbonding_records.go
```diff
@@ -86,14 +90,85 @@ func (k Keeper) GetTotalUnbondAmountAndRecordsIds(ctx sdk.Context, chainId strin
...
+	// Loop and set the native amount for each record, keeping track fo the total
+	totalNativeAmount = sdkmath.ZeroInt()
+	for _, userRedemptionRecordId := range redemptionRecordIds {
+		userRedemptionRecord, found := k.RecordsKeeper.GetUserRedemptionRecord(ctx, userRedemptionRecordId)
+		if !found {
+			k.Logger(ctx).Error(utils.LogWithHostZone(chainId, "No user redemption record found for id %s", userRedemptionRecordId))
+			continue
+		}
+
+		// A zero redemption rate is used to indicate that the native amount should be set to zero
```
asalzmann: why?

app/upgrades/v17/upgrades.go
```diff
@@ -303,3 +309,61 @@ func AddRateLimitToOsmosis(ctx sdk.Context, k ratelimitkeeper.Keeper) error {
...
+				userRedemptionRecord.StTokenAmount = estimatedStTokenConversionRate.Mul(sdkmath.LegacyDec(userRedemptionRecord.Amount)).RoundInt()
+
+				if unbondingNotInitiated {
+					userRedemptionRecord.Amount = sdkmath.ZeroInt()
+				}
+
+				k.RecordsKeeper.SetUserRedemptionRecord(ctx, userRedemptionRecord)
+			}
+
+			// if the unbonding has not been initiated, we want to set nativeTokenAmount to 0 for the whole HostZoneUnbonding
```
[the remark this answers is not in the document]
sampocs: good call! just updated this as well as some other stuff that no longer seems necessary!

asalzmann (commented):
this is looking clean! thank you for refactoring

will do a final review once it's merged into the underlying branch

x/stakeibc/keeper/unbonding_records.go
sampocs: sorry stale comment, removed it a second ago

riley-stride:
do we still need to add the cache context?

x/stakeibc/keeper/msg_server_redeem_stake.go
```diff
@@ -89,16 +82,18 @@ func (k msgServer) RedeemStake(goCtx context.Context, msg *types.MsgRedeemStake)
...
 		userRedemptionRecord = recordstypes.UserRedemptionRecord{
-			Id:          redemptionId,
-			Receiver:    msg.Receiver,
-			Amount:      nativeAmount,
-			Denom:       hostZone.HostDenom,
-			HostZoneId:  hostZone.ChainId,
-			EpochNumber: epochTracker.EpochNumber,
+			Id:            redemptionId,
+			Receiver:      msg.Receiver,
+			Amount:        nativeAmount,
```
asalzmann, line 92: not important for this PR - at some point we should rename all "amount"s to either native amount or stToken amount 

in this file, Amount is native amount (here) and stToken amount (above on MsgRedeemStake)

very confusing

x/stakeibc/keeper/unbonding_records.go
```diff
@@ -500,7 +583,10 @@ func (k Keeper) InitiateAllHostZoneUnbondings(ctx sdk.Context, dayNumber uint64)
 		}
 
 		// Get host zone unbonding message by summing up the unbonding records
-		if err := k.UnbondFromHostZone(ctx, hostZone); err != nil {
+		err := utils.ApplyFuncIfNoError(ctx, func(ctx sdk.Context) error {
```
asalzmann, line 586: I think this is ok, but double checking because it's different than what we discussed live - this makes (1) updating the records and (2) sending the MsgUndelegate ICAs atomic, right?

asalzmann: nit: should we rename to `UpdateRecordsAndSendUnbondingICAs`?

x/stakeibc/keeper/unbonding_records.go
```diff
@@ -86,14 +90,84 @@ func (k Keeper) GetTotalUnbondAmountAndRecordsIds(ctx sdk.Context, chainId strin
...
+
+		// Set the native amount on the record
+		userRedemptionRecord.Amount = nativeAmount
+		k.RecordsKeeper.SetUserRedemptionRecord(ctx, userRedemptionRecord)
+	}
+	return totalNativeAmount
+}
+
+// Sets the native token amount unbonded on the host zone unbonding record and the associated user redemption records
+func (k Keeper) RefreshHostZoneUnbondingNativeTokenAmounts(
```
asalzmann, line 140:
```suggestion
func (k Keeper) RefreshHostZoneUnbondingNativeTokenAmount(
```

Since there's just one variable we're refreshing on the HZU

x/stakeibc/keeper/unbonding_records.go
```diff
@@ -86,14 +90,84 @@ func (k Keeper) GetTotalUnbondAmountAndRecordsIds(ctx sdk.Context, chainId strin
...
+
+// Given a list of redemption record IDs and a redemption rate, sets the native token amount on each record and returns the total
+// The native amount is calculated using the redemption rate. If a zero redemption rate is passed, set the amount to zero
+func (k Keeper) RefreshUserRedemptionRecordNativeAmounts(
+	ctx sdk.Context,
+	chainId string,
+	redemptionRecordIds []string,
+	redemptionRate sdk.Dec,
+) (totalNativeAmount sdkmath.Int) {
+	// Loop and set the native amount for each record, keeping track fo the total
```
asalzmann, line 119:
```suggestion
	// Loop and set the native amount for each record, keeping track of the total
```

x/stakeibc/keeper/unbonding_records.go
```diff
@@ -86,14 +90,84 @@ func (k Keeper) GetTotalUnbondAmountAndRecordsIds(ctx sdk.Context, chainId strin
...
+	}
+	return totalUnbonded
+}
+
+// Given a list of redemption record IDs and a redemption rate, sets the native token amount on each record and returns the total
+// The native amount is calculated using the redemption rate. If a zero redemption rate is passed, set the amount to zero
+func (k Keeper) RefreshUserRedemptionRecordNativeAmounts(
+	ctx sdk.Context,
+	chainId string,
+	redemptionRecordIds []string,
```
asalzmann, line 116:
```suggestion
	userRedemptionRecordIds []string,
```

nit

x/stakeibc/keeper/unbonding_records.go
```diff
@@ -86,14 +90,84 @@ func (k Keeper) GetTotalUnbondAmountAndRecordsIds(ctx sdk.Context, chainId strin
...
+// Also returns the epoch unbonding record ids
+func (k Keeper) GetTotalUnbondAmount(ctx sdk.Context, hostZoneUnbondingRecords map[uint64]recordstypes.HostZoneUnbonding) (totalUnbonded sdkmath.Int) {
+	totalUnbonded = sdk.ZeroInt()
+	for _, hostZoneRecord := range hostZoneUnbondingRecords {
+		totalUnbonded = totalUnbonded.Add(hostZoneRecord.NativeTokenAmount)
+	}
+	return totalUnbonded
+}
+
+// Given a list of redemption record IDs and a redemption rate, sets the native token amount on each record and returns the total
```
asalzmann, line 111:
```suggestion
// Given a list of user redemption record IDs and a redemption rate, sets the native token amount on each record and returns the sum of all native token amounts across all user redemption records  
```

nit

x/records/types/records.go
```diff
@@ -0,0 +1,8 @@
+package types
+
+import sdkmath "cosmossdk.io/math"
+
+// Helper function to evaluate if a host zone unbonding record still needs to be initiated
+func (r HostZoneUnbonding) ShouldInitiateUnbonding() bool {
+	return r.Status == HostZoneUnbonding_UNBONDING_QUEUE && r.NativeTokenAmount.GT(sdkmath.ZeroInt())
```
riley-stride, line 7: nit for clarity
```suggestion
    notYetUnbonding = r.Status == HostZoneUnbonding_UNBONDING_QUEUE
    hasAtLeastOneRecord = r.NativeTokenAmount.GT(sdkmath.ZeroInt())
	return notYetUnbonding && hasAtLeastOneRecord
```

x/stakeibc/keeper/unbonding_records.go
```diff
@@ -86,14 +90,84 @@ func (k Keeper) GetTotalUnbondAmountAndRecordsIds(ctx sdk.Context, chainId strin
...
+// The native amount is calculated using the redemption rate. If a zero redemption rate is passed, set the amount to zero
+func (k Keeper) RefreshUserRedemptionRecordNativeAmounts(
+	ctx sdk.Context,
+	chainId string,
+	redemptionRecordIds []string,
+	redemptionRate sdk.Dec,
+) (totalNativeAmount sdkmath.Int) {
+	// Loop and set the native amount for each record, keeping track fo the total
+	totalNativeAmount = sdkmath.ZeroInt()
+	for _, userRedemptionRecordId := range redemptionRecordIds {
```
asalzmann, line 121: This is the first time we're looping all URRs (afaik) - don't think it's an issue, but flagging is it could potentially cause performance issues if there are too many URRs.

One simple solution is to enforce a min redemption amount.

Wdyt @riley-stride @sampocs ?

x/stakeibc/keeper/unbonding_records.go
```diff
@@ -365,8 +439,17 @@ func (k Keeper) UnbondFromHostZone(ctx sdk.Context, hostZone types.HostZone) err
...
+	epochUnbondingRecordIds, epochNumberToHostZoneUnbondingMap := k.GetQueuedHostZoneUnbondingRecords(ctx, hostZone.ChainId)
+
+	// Update the native unbond amount on all relevant records
+	// The native amount is calculated from the stTokens
+	if err := k.RefreshUnbondingNativeTokenAmounts(ctx, epochNumberToHostZoneUnbondingMap); err != nil {
+		return err
+	}
+
+	// Sum the total native unbond amount across all records
+	totalUnbondAmount := k.GetTotalUnbondAmount(ctx, epochNumberToHostZoneUnbondingMap)
```
riley-stride, line 452: Afaiu `GetTotalUnbondAmount` returns the total amount to be unbonded. If so, should we slightly rename and change the comment above the fn def for clarity, maybe `GetTotalAmountToBeUnbonded`?

Afaiu `GetTotalUnbondAmount` only works if passed a a _filtered_ map of `HostZoneUnbonding`s (here, the filtering is performed by `GetQueuedHostZoneUnbondingRecords`. Should we make this filtering requirement more clear?

x/stakeibc/keeper/unbonding_records.go
```diff
@@ -86,14 +90,84 @@ func (k Keeper) GetTotalUnbondAmountAndRecordsIds(ctx sdk.Context, chainId strin
...
-			totalUnbonded = totalUnbonded.Add(hostZoneRecord.NativeTokenAmount)
-			unbondingRecordIds = append(unbondingRecordIds, epochUnbonding.EpochNumber)
-			k.Logger(ctx).Info(utils.LogWithHostZone(chainId, "  %v%s included in total unbonding", hostZoneRecord.NativeTokenAmount, hostZoneRecord.Denom))
+		if hostZoneRecord.ShouldInitiateUnbonding() {
+			epochNumbers = append(epochNumbers, epochUnbonding.EpochNumber)
+			hostZoneUnbondingMap[epochUnbonding.EpochNumber] = *hostZoneRecord
 		}
 	}
-	return totalUnbonded, unbondingRecordIds
+	return epochNumbers, hostZoneUnbondingMap
```
asalzmann, line 98: small nit, I think `epochNumbers` are just the keys of `hostZoneUnbondingsMap`

would it make sense to just return `hostZoneUnbondingsMap`?

x/stakeibc/keeper/unbonding_records.go
```diff
@@ -74,10 +74,14 @@ func (k Keeper) CreateEpochUnbondingRecord(ctx sdk.Context, epochNumber uint64)
...
-// Also returns the epoch unbonding record ids
-func (k Keeper) GetTotalUnbondAmountAndRecordsIds(ctx sdk.Context, chainId string) (totalUnbonded sdkmath.Int, unbondingRecordIds []uint64) {
-	totalUnbonded = sdk.ZeroInt()
+// Returns all the host zone unbonding records that should unbond this epoch
+// Records are returned as a mapping of epoch unbonding record ID to host zone unbonding record
+// Records ready to be unbonded are identified by status UNBONDING_QUEUE and a non-zero native amount
+func (k Keeper) GetQueuedHostZoneUnbondingRecords(
+	ctx sdk.Context,
+	chainId string,
+) (epochNumbers []uint64, hostZoneUnbondingMap map[uint64]recordstypes.HostZoneUnbonding) {
```
asalzmann, line 83:
```suggestion
) (epochNumbers []uint64, epochToHostZoneUnbondingMap map[uint64]recordstypes.HostZoneUnbonding) {
```
nit

riley-stride (commented):
This refactor is beautiful, much easier to reason about. Added just two very light comments. The cache context lgtm.

asalzmann (commented):
In particular, reviewed 
* Set native_token_amount on redemption - OK
* Set native_token_amount before sending ICAs - OK
* Only execute (2) if (1) succeeds - OK (cache context)
* If (2) fails, don't revert native_token_amount - different from what we discussed, but I think that's fine

Documenting some assumptions
* Native token amount is set on the HZU / all associated URRs upon redemption, then refreshed just before sending the unbonding ICA 
* Both the refresh step _and_ the send ICA step happen atomically
* Native token amount will either be (1) an amount corresponding to the RR at the time of unbonding (2) an amount corresponding to the RR at the time of a successful unbonding 

Looking really good! I'm signed off, pending a few nits

x/stakeibc/keeper/msg_server_redeem_stake.go
riley-stride: I made the same comment on Vishals' original PR so he opened one to fix this, maybe we can do the same here (shuld be quick) https://github.com/Stride-Labs/stride/pull/1051

asalzmann: We should keep in mind that we'll need to update the FE as well if we change the names on these protos

sampocs: agreed! was planning on doing this in a separate PR along with a couple other renaming's mentioned in todos

x/stakeibc/keeper/unbonding_records.go
riley-stride: Good flag. 

The min redemption amount makes sense, is the easiest fix I think.

Maybe 1 unit of native token.

x/records/types/records.go
sampocs: great suggestion! updated

x/stakeibc/keeper/unbonding_records.go
sampocs: good call, just updated

x/stakeibc/keeper/unbonding_records.go
sampocs: Ah sorry - must have misunderstood. I though we wanted both (1) and (2) to be atomic, with the thinking being that there's no need to do (1) outside of the context of (2)?

riley-stride: Wasn't making them atomic the plan we discussed live?

> Only execute (2) if (1) succeeds

sampocs: re: naming, my thinking was we have a few core operations in stakeibc (delegation, unbonding, redemptionSweep, etc.). So I thought it made sense to name them using just that high level category as the name instead of getting too specific (e.g. `UnbondFromHostZone`, `DelegateToHostZone`, etc.). Makes it a little easier to locate the main function imo instead of conflating it with helpers

That said, open to something else, but that was my thinking!

asalzmann: > Wasn't making them atomic the plan we discussed live?

I think we discussed not rolling back setting native token amounts if sending ICAs failed, but might have misunderstood. Either way, not important and I'm signed off on this approach 

> re: naming...

I bias to explicit clarity in naming variables and functions (at the expense of making them verbose), but don't feel strongly

x/stakeibc/keeper/unbonding_records.go
sampocs: done

x/stakeibc/keeper/unbonding_records.go
sampocs: done

x/stakeibc/keeper/unbonding_records.go
sampocs: done

x/stakeibc/keeper/unbonding_records.go
sampocs: done

x/stakeibc/keeper/unbonding_records.go
sampocs: done

x/stakeibc/keeper/unbonding_records.go
sampocs: I did that originally, but we eventually need them as a list for the callback data. So it seemed a bit cleaner to do the above, rather than have to do something like the following later:
```go
epochNumbers := []uint64{}
for epochNumber, _ := range hostZoneUnbondingMap {
  epochNumbers = append(epochNumbers, epochNumber)
}
```

If only there was a .keys like function haha

asalzmann: ah yeah that makes sense!

asalzmann (approved):

riley-stride (approved):

x/stakeibc/keeper/unbonding_records.go
sampocs: let's table naming for now, still planning on doing the file org refactor soon we can re-evaluate then

riley-stride:
Manual testing to inspect the native_token_amount incrementing:

x/stakeibc/keeper/unbonding_records_get_host_zone_unbondings_msgs_test.go
```diff
@@ -594,6 +594,139 @@ func (s *KeeperTestSuite) TestGetTotalUnbondAmount() {
...
+		actualRecord, found := s.App.RecordsKeeper.GetUserRedemptionRecord(s.Ctx, expectedRecord.Id)
+		s.Require().True(found, "record %s should have been found", expectedRecord.Id)
+		s.Require().Equal(expectedRecord.Amount.Int64(), actualRecord.Amount.Int64(),
+			"record %s native amount", expectedRecord.Id)
+	}
+}
+
+// Tests RefreshUnbondingNativeTokenAmounts which indirectly tests
+// RefreshHostZoneUnbondingNativeTokenAmount and RefreshUserRedemptionRecordNativeAmounts
+func (s *KeeperTestSuite) TestRefreshUnbondingNativeTokenAmounts() {
```
riley-stride, line 640: Do we need to test the case where there is a record value 0?

riley-stride (approved):
Unit tests look quite solid. Covering all the functions and edge cases I can think of. If we wanted to beef them up we could test `RefreshHostZoneUnbondingNativeTokenAmount` and `RefreshUserRedemptionRecordNativeAmounts` more directly, but I think `RefreshHostZoneUnbondingNativeTokenAmount` already has good coverage of these indirectly, as you noted.

x/stakeibc/keeper/unbonding_records_get_host_zone_unbondings_msgs_test.go
sampocs: I don't think so cause it will still just do the same multiplication

PR #1054, opened 2024-01-10
add pfm testing scripts

riley-stride:
1. a->b->c pfm works. to verify:
- open two terminal windows as described below
    - terminal 1 monitors the sender on A: `while true; do gaiadl q bank balances $(gaiadl keys show gval1 -a)  && echo "--------------" && sleep 1; done`
    - terminal 2 monitors the receiver on C: `while true; do osmosisdl q bank balances osmo1uk4ze0x4nvh4fk0xm4jdud58eqn4yxhrqyeqwq  && echo "--------------" && sleep 1; done`
- run `sh dockernet/scripts/pfm/pfm_abc_default.sh`
- watch the terminal windows to see C's balance increase after A's decreases.

2. pfm'ing + autopilot coin A thru chain A will work after upgrade but pfm'ing + autopilot coin A thru B won't work 
  -there's a script that sends from atom from CH -> stride then from Stride -> osmosis, then tries to autopilot LS that ATOM from Osmosis. If fails.
  -there's a script `pfm_to_autopilot_ls.sh` that tries to PFM (a -> b -> autopilot-LS-on-a: stride -> osmosis -> stride (autopilot LS). It doesn't work right now. Because the autopilot step doesn't work.


asalzmann (approved):
lgtm!

asalzmann:
I think since `pfm` was merged you can change the base branch and merge right to `main`

PR #1056, opened 2024-01-11
unittest MigrateUnbondingRecords

asalzmann:
**Summary**

Fix two issues in the migration logic
(1) divide by zero wasn't being caught
(2) LegacyDec shouldn't be used (it takes an int like 250 and makes it 0.00...00250)

**Test plan**
Add unittests


riley-stride (approved):
Tests look great. Verified you're covering 
- all the error cases in `MigrateUnbondingRecords` (except the seemingly impossible "user redemption record not found" error on ln 342)
- all the record types 
- zero values for at least one record type. I believe (check me) that 0 is not possible for records in status other than UNBONDING_QUEUE. If that's the case, we have full coverage of the cases where `NativeTokenAmount` could be 0.

Open question — I see a bunch of other uses of `sdkmath.ZeroInt()` across the codebase. Should those be removed, or only in the case where it's being used to check whether a Dec is 0-value?

riley-stride:
> Open question — I see a bunch of other uses of `sdkmath.ZeroInt()` across the codebase. Should those be removed, or only in the case where it's being used to check whether a Dec is 0-value?

Sam and I connected about this. Sounds like the only case that breaks is comparing a value as follows: `X == sdkmath.ZeroInt()` (it compares the pointers rather than values, so even if X is `0` this equality will return false).

I searched through the rest of the codebase and can verify we don't use this elsewhere. 

Good to go on this PR!

PR #1058, opened 2024-01-11
rename amount to native_token_amount

shellvish:
Renames "amount" on the UserRedemptionRecords to "nativeTokenAmount" 

Confirmed that binary builds, but please double check I didn't inadvertently change any wrong variables!


x/records/keeper/epoch_unbonding_record.go
```diff
@@ -134,7 +134,7 @@ func (k Keeper) SetHostZoneUnbondingRecord(ctx sdk.Context, epochNumber uint64,
 
 // Updates the status for a given host zone across relevant epoch unbonding record IDs
 // TODO [cleanup]: Rename to SetHostZoneUnbondingStatus
```
sampocs, line 136: remove this TODO

sampocs (approved):
one small comment, otherwise lgtm!

x/records/keeper/epoch_unbonding_record.go
sampocs: done

PR #1060, opened 2024-01-11
fix bug in changelog script

sampocs:
## Context
Need to set the base branch to main, otherwise we need a tag before we can create the changelog


shellvish (approved):
Nice, thanks for doing this!

PR #1074, opened 2024-01-12
reset delegation change in progress to 0 when a channel is restored

sampocs:
Closes: #XXX

## Context 
The `DelegationChangesInProgress` field on the validator struct is a mechanism to prevent ICA/ICQ overlap. When ICAs are submitted, the `DelegationChangesInProgress` field is incremented, and when the ack comes back, it is decremented.

In order to prevent ICA/ICQ overlap, when a slash ICQ is relayed, it [[checks](https://github.com/Stride-Labs/stride/blob/main/x/stakeibc/keeper/icqcallbacks_delegator_shares.go#L147-L148)] if there are any active delegation changes in progress, and if so, the query is retried (since the results can’t be trusted)

However, when an ICA packet times out, the channel is closed and the pending packets are never relayed. Since we never receive the ack for these packets, the `DelegationChangesInProgress` field never gets decremented. 

This PR sets the `DelegationChangesInProgress` field to 0 when the ICA channel is restored.

## Brief Changelog
* Set the `DelegationChangesInProgress` to 0 when a channel is restored and added unit test

### DIFF
--- a/x/stakeibc/keeper/msg_server_restore_interchain_account.go
+++ b/x/stakeibc/keeper/msg_server_restore_interchain_account.go
@@ -53,6 +53,13 @@ func (k msgServer) RestoreInterchainAccount(goCtx context.Context, msg *types.Ms
 			return nil, types.ErrHostZoneNotFound.Wrapf("delegation ICA supplied, but no associated host zone")
 		}
 
+		// Since any ICAs along the original channel will never get relayed,
+		// we have to reset the delegation_changes_in_progress field on each validator
+		for _, validator := range hostZone.Validators {
+			validator.DelegationChangesInProgress = 0
+		}
+		k.SetHostZone(ctx, hostZone)
+
 		// revert DELEGATION_IN_PROGRESS records for the closed ICA channel (so that they can be staked)
 		depositRecords := k.RecordsKeeper.GetAllDepositRecord(ctx)
 		for _, depositRecord := range depositRecords {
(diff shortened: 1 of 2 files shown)


x/stakeibc/keeper/msg_server_restore_interchain_account.go
```diff
@@ -53,6 +53,13 @@ func (k msgServer) RestoreInterchainAccount(goCtx context.Context, msg *types.Ms
 			return nil, types.ErrHostZoneNotFound.Wrapf("delegation ICA supplied, but no associated host zone")
 		}
 
+		// Since any ICAs along the original channel will never get relayed,
+		// we have to reset the delegation_changes_in_progress field on each validator
```
riley-stride, line 57: This makes sense if we're restoring the delegation account channel. 

But if we're restoring another channel like the withdrawalAccount, do we want to set these `DelegationChangesInProgress` to 0?

riley-stride (commented):
Just one question

x/stakeibc/keeper/msg_server_restore_interchain_account.go
sampocs: I think the delegation channel is the only relevant one cause the only ICAs that change our delegations go through that channel (but please double check me)

riley-stride: Sorry, I don't understand. 

Afaiu we have multiple ICA channels to multiple ICAs on the host zone. If any of them close, we need to restore them.

The only way to restore them is via this function.

But we ONLY want to reset the DelegationChangesInProgress to 0 if we're restoring the Delegation ICA channel.
However this fn resets DelegationChangesInProgress to 0 for any channel restoration (E.g. withdrawal channel)

sampocs: this is actually nested under a check that only runs for the delegation channel! [ref]

Easy to miss with the GH diffs haha

riley-stride (approved):

x/stakeibc/keeper/msg_server_restore_interchain_account.go
sampocs: great call out though - appreciate the scrutiny!

PR #1075, opened 2024-01-13
Query redemption records for multiple addresses

ethan-stride:
This allows the QueryAddressUnbondings RPC to either pass in a single address or a comma separated list of multiple addresses.  The proto was not changed on purpose so that the grpc endpoint wouldn't change, and queries which were working before with a single input address will continue to work just fine.  The input should be robust to whether the user includes spaces or not but it is expected to be a string of addresses separated by commas with no spaces normally.

One nit change, initializing the addressUnbondings variable on ln33 to an empty array, before it was given a type but could have been nil if the address provided had no redemption records.

Noticed this keeper function never had a unit test before so added grpc_query_address_unbondings_test.go which tests:
- Single address input with no redemption records found
- Single address input with a single redemption record found
- Single address input multiple redemption records across different epochs
- Multiple addresses input which finds record for only one address but not the others
- Multiple addresses input (even on different hostZones) which finds user records for each, more than one for some
- Invalid query or empty address given is an expected error case

### DIFF
--- a/x/stakeibc/keeper/grpc_query_address_unbondings.go
+++ b/x/stakeibc/keeper/grpc_query_address_unbondings.go
@@ -19,7 +19,7 @@ import (
 const nanosecondsInDay = 86400000000000
 
 func (k Keeper) AddressUnbondings(c context.Context, req *types.QueryAddressUnbondings) (*types.QueryAddressUnbondingsResponse, error) {
-	// The function queries all the unbondings associated with a Stride address.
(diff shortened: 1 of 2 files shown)


shellvish (approved):
This looks great, thanks a lot Ethan! Verified locally that this works with mainnet state.

Pushed up a very small optimization (to stop the inner loop when we find a match), otherwise looks fantsatic.

PR #1084, opened 2024-01-19
V18 Upgrade Migrations

shellvish:
Migrates User Redemption Records to new format in v18 upgrade

Note: we should test this twice with LocalStride. First, with a _slightly_ stale cache, so we can more accurately gauge what will happen at upgrade time. Second, with the most recent mainnet state (as part of the standard localstride flow). 

TODO: Add Unit Tests


x/stakeibc/keeper/unbonding_records_get_host_zone_unbondings_msgs_test.go
```diff
@@ -351,6 +365,42 @@ func (s *KeeperTestSuite) TestUnbondFromHostZone_Successful_UnbondTotalGreaterTh
 	s.CheckUnbondingMessages(tc, expectedUnbondings)
 }
 
+func (s *KeeperTestSuite) TestUnbondFromHostZone_Successful_RefreshedNativeAmount() {
```
riley-stride, line 368: Should we do more unit testing? Feels like this adds a bit of coverage but doesn't cover everything

app/upgrades/v18/upgrades.go
```diff
@@ -0,0 +1,216 @@
...
+		"phoenix-1":    sdk.MustNewDecFromStr("1.175092793361528752"),
+		"sommelier-3":  sdk.MustNewDecFromStr("1.025191976926387110"),
+		"stargaze-1":   sdk.MustNewDecFromStr("1.426572776580499529"),
+		"umee-1":       sdk.MustNewDecFromStr("1.125555216159213353"),
+	}
+
+	// Get Amount Unbonded for each HostZone for Unbonding Records Migration
+	TotalAmountUnbonded = map[string]map[string]sdkmath.Int{
+		"comdex-1": {
+			"1705863627": NewIntFromString("1640812802"),
```
riley-stride, line 57: Will these need refreshing before this upgrade goes live?

app/upgrades/v18/upgrades.go
```diff
@@ -0,0 +1,216 @@
...
+			}
+			// if we can't find unbonding amount, then we use an estimated redemption rate
+			if amountUnbonded.IsZero() {
+				initialRedemptionRate := InitialRedemptionRates[hostZoneUnbonding.HostZoneId]
+				hostZone, found := k.GetHostZone(ctx, hostZoneUnbonding.HostZoneId)
+				if !found {
+					return errorsmod.Wrapf(stakeibctypes.ErrHostZoneNotFound, "unable to find host zone with chain-id %s", hostZoneUnbonding.HostZoneId)
+				}
+				currentRedemptionRate := hostZone.RedemptionRate
+				blendedRedemptionRate := initialRedemptionRate.Add(currentRedemptionRate).Quo(sdk.NewDecFromInt(sdk.NewInt(2)))
```
riley-stride, line 196: Just sanity checking — how does order of operations work here? Add should precede Mul, right?

Might be good to add a sanity check below that makes sure blended is between initial and current, just to be safe. Or could add more thorough unit tests.

app/upgrades/v18/upgrades.go
```diff
@@ -0,0 +1,216 @@
...
+				unbondTime := strconv.FormatUint(hostZoneUnbonding.UnbondingTime, 10)
+				for unbondTimePrefixInCache, amountUnbondedInCache := range allAmountsUnbondedForHost {
+					if strings.HasPrefix(unbondTime, unbondTimePrefixInCache) {
+						amountUnbonded = amountUnbondedInCache
+						break
+					}
+				}
+			}
+			// if we can't find unbonding amount, then we use an estimated redemption rate
+			if amountUnbonded.IsZero() {
```
riley-stride, line 189: Afaict this will run for every host zone, is that expected? I don't see any INJ-specific logic in this PR.

app/upgrades/v18/upgrades.go
```diff
@@ -0,0 +1,216 @@
...
+			"1706468412": NewIntFromString("508229473185"),
+			"1706727610": NewIntFromString("1033532361858"),
+		},
+		"umee-1": {
+			"1705690815": NewIntFromString("510084698639"),
+		},
+	}
+)
+
+// CreateUpgradeHandler creates an SDK upgrade handler for v17
```
riley-stride, line 121: nit
```suggestion
// CreateUpgradeHandler creates an SDK upgrade handler for v18
```

riley-stride (changes requested):
The logic looks sound. My main comment is that we should add more unit testing as this is quite a sensitive change.

asalzmann (commented):
Overall looks good - need to review in detail again tomorrow. +1 to Riley's comment - should we add more detailed unittests?

app/upgrades/v18/upgrades.go
```diff
@@ -0,0 +1,216 @@
...
+		// loop over host zone unbonding records
+		for _, hostZoneUnbonding := range epochUnbondingRecord.GetHostZoneUnbondings() {
+			recordStatus := hostZoneUnbonding.Status
+			if recordStatus != recordtypes.HostZoneUnbonding_EXIT_TRANSFER_QUEUE {
+				continue // can ignore this record, as it's not unbonding
+			}
+			amountUnbonded := sdkmath.ZeroInt()
+			// check if the unbond amount is stored in cache above
+			allAmountsUnbondedForHost, found := TotalAmountUnbonded[hostZoneUnbonding.HostZoneId]
+			if !found {
```
sampocs, line 180: is this supposed to be `found` not `!found`?

app/upgrades/v18/upgrades.go
```diff
@@ -0,0 +1,216 @@
...
+				// check if this unbond time is in the cache above
+				unbondTime := strconv.FormatUint(hostZoneUnbonding.UnbondingTime, 10)
+				for unbondTimePrefixInCache, amountUnbondedInCache := range allAmountsUnbondedForHost {
+					if strings.HasPrefix(unbondTime, unbondTimePrefixInCache) {
```
sampocs, line 184: what's the rationale for using `HasPrefix` instead of just checking equality?

app/upgrades/v18/upgrades.go
```diff
@@ -0,0 +1,216 @@
...
+	RedemptionRateOuterMaxAdjustment = sdk.MustNewDecFromStr("0.10")
+
+	// Osmosis will have a slighly larger buffer with the redemption rate
+	// since their yield is less predictable
+	OsmosisChainId              = "osmosis-1"
+	OsmosisRedemptionRateBuffer = sdk.MustNewDecFromStr("0.02")
+
+	// Get Initial Redemption Rates for Unbonding Records Migration
+	InitialRedemptionRates = map[string]sdk.Dec{
+		"comdex-1":     sdk.MustNewDecFromStr("1.199107579864670291"),
```
sampocs, line 42: did you determine these by doing division on the records or by doing a historical query on the host zone?

app/upgrades/v18/upgrades.go
```diff
@@ -0,0 +1,216 @@
...
+			impliedRedemptionRate := sdk.NewDecFromInt(amountUnbonded).Quo(sdk.NewDecFromInt(hostZoneUnbonding.StTokenAmount))
+			// now update all userRedemptionRecords
+			for _, userRedemptionRecordId := range hostZoneUnbonding.GetUserRedemptionRecords() {
+				userRedemptionRecord, found := rk.GetUserRedemptionRecord(ctx, userRedemptionRecordId)
+				if !found {
+					return errorsmod.Wrapf(recordtypes.ErrHostUnbondingRecordNotFound, "unable to find user redemption record with id %s", userRedemptionRecordId)
+				}
+				userRedemptionRecord.NativeTokenAmount = sdk.NewDecFromInt(userRedemptionRecord.StTokenAmount).Mul(impliedRedemptionRate).TruncateInt()
+				rk.SetUserRedemptionRecord(ctx, userRedemptionRecord)
+			}
```
sampocs: do we have to do a catch all here on the last record to make sure there's no precision error?

app/upgrades/v18/upgrades.go
sampocs: or is this supposed to be as refreshed as possible and we'll replace it with the RR's at the time the prop goes up?

sampocs (commented):
thanks for taking this - it looked like a tedious one!

x/stakeibc/keeper/unbonding_records_get_host_zone_unbondings_msgs_test.go
shellvish: 100%, this is just adding coverage to the existing code, but we should definitely test the migrations very thoroughly

app/upgrades/v18/upgrades.go
shellvish: It's a great question! I think Add goes before Quo, although we should test it explicitly with a unit test

app/upgrades/v18/upgrades.go
shellvish: It's meant to run for every host zone!

app/upgrades/v18/upgrades.go
```diff
@@ -0,0 +1,216 @@
...
+	upgradetypes "github.com/cosmos/cosmos-sdk/x/upgrade/types"
+
+	recordskeeper "github.com/Stride-Labs/stride/v17/x/records/keeper"
+	recordtypes "github.com/Stride-Labs/stride/v17/x/records/types"
+	stakeibckeeper "github.com/Stride-Labs/stride/v17/x/stakeibc/keeper"
+	stakeibctypes "github.com/Stride-Labs/stride/v17/x/stakeibc/types"
+)
+
+func NewIntFromString(val string) sdkmath.Int {
+	out_int, ok := sdkmath.NewIntFromString(val)
```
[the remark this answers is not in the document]
shellvish: I think the only reason is the INJ/EVMOS values overflow if they're regular ints, so string felt a bit easier. But no preference on my end on what casting method we use, as long as it can handle the big numbers!

app/upgrades/v18/upgrades.go
shellvish: The latter! We'll refresh these with the live RRs at the time of the prop (we have a script that will refresh all of the constants at the top of this file)

app/upgrades/v18/upgrades.go
shellvish: 🤦  Yes definitely, that was a mistake on my part

app/upgrades/v18/upgrades.go
shellvish: When we pull the unbond times from the ground truth on the host zone, they report it to slightly lower precision than we store it on the unbonding records. There might be a way to pull the more precise value from the host zone struct though

app/upgrades/v18/upgrades.go
shellvish: Ah yes that's a great point, we should update it to that!

app/upgrades/v18/upgrades.go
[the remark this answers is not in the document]
sampocs: ah that makes sense!

app/upgrades/v18/upgrades.go
```diff
@@ -0,0 +1,116 @@
...
+			// now update all userRedemptionRecords by using the redemption rate to set the native token amount
+			for _, userRedemptionRecordId := range hostZoneUnbonding.UserRedemptionRecords {
+				userRedemptionRecord, found := rk.GetUserRedemptionRecord(ctx, userRedemptionRecordId)
+				if !found {
+					return errorsmod.Wrapf(recordtypes.ErrHostUnbondingRecordNotFound, "unable to find user redemption record with id %s", userRedemptionRecordId)
+				}
+				userRedemptionRecord.NativeTokenAmount = sdk.NewDecFromInt(userRedemptionRecord.StTokenAmount).Mul(recordRedemptionRate).TruncateInt()
+				rk.SetUserRedemptionRecord(ctx, userRedemptionRecord)
+			}
+
```
shellvish, line 110: should we also update the `NativeTokenAmount` on the host zone struct? I don't think it's strictly necessary, although feels a bit easier to reason about all of the records lining up!

shellvish (commented):
Looks good to me Sam! One small comment on updating the HZU but otherwise beautiful

app/upgrades/v18/upgrades.go
sampocs: omg, good catch - not sure how I missed that!

sampocs: just pushed up the fix. I also think there was  bug where there was a premature return after calling SetHZUnbonding

app/upgrades/v18/constants.go
```diff
@@ -0,0 +1,252 @@
...
+			500: sdk.MustNewDecFromStr("1.4161992241945822"),
+			501: sdk.MustNewDecFromStr("1.4161992241945822"),
+			503: sdk.MustNewDecFromStr("1.4161992241945822"),
+			504: sdk.MustNewDecFromStr("1.4161992241945822"),
+			505: sdk.MustNewDecFromStr("1.417724248601981"),
+			507: sdk.MustNewDecFromStr("1.417724248601981"),
+			508: sdk.MustNewDecFromStr("1.417724248601981"),
+		},
+		"phoenix-1": {
+			480: sdk.MustNewDecFromStr("1.1617706976314004"),
```
asalzmann, line 57: these seem slightly off @shellvish

app/upgrades/v18/upgrades.go
```diff
@@ -0,0 +1,116 @@
...
+	}
+}
+
+// Modify HostZoneUnbonding and UserRedemptionRecords NativeTokenAmount to reflect new data structs
+func UpdateUnbondingRecords(
+	ctx sdk.Context,
+	sk stakeibckeeper.Keeper,
+	rk recordskeeper.Keeper,
+	redemptionRatesBeforeProp map[string]map[uint64]sdk.Dec,
+	redemptionRatesDuringProp map[string]sdk.Dec,
```
asalzmann, line 67:
```suggestion
	redemptionRatesAtTimeOfProp map[string]sdk.Dec,
```

asalzmann (commented):
This is looking good, should we also unittest test the upgrade itself? Can also happen in a different PR / someone else could grab it!

sampocs:
@asalzmann sorry still wrapping up the unit test - kept getting side tracked!

app/upgrades/v18/upgrades_test.go
```diff
@@ -49,12 +49,155 @@ type UpgradeTestSuite struct {
...
+	})
+	s.App.RecordsKeeper.SetUserRedemptionRecord(s.Ctx, recordtypes.UserRedemptionRecord{
+		Id:                "B",
+		EpochNumber:       epochNumberAfter,
+		HostZoneId:        chainId,
+		StTokenAmount:     stTokenAmount,
+		NativeTokenAmount: initialNativeAmount2,
+	})
+
+	// Create epoch unbonding records
```
riley-stride, line 120: Should we test records in other statuses?

riley-stride (commented):
Unit test looks beautiful!

app/upgrades/v18/upgrades_test.go
sampocs: was considering that too, but didn't want the test to get too bloated. 

Happy to add it in the morn if you think it makes sense!

riley-stride: Maybe just one other status to make sure the record filtering works

sampocs: done

riley-stride (approved):
re-reviewed with fresh eyes: i'm signed off.

app/upgrades/v18/constants.go
```diff
@@ -0,0 +1,167 @@
...
+	// since their yield is less predictable
+	OsmosisChainId              = "osmosis-1"
+	OsmosisRedemptionRateBuffer = sdk.MustNewDecFromStr("0.02")
+
+	// Terra chain ID for delegation changes in progress
+	TerraChainId = "phoenix-1"
+
+	// Get Initial Redemption Rates for Unbonding Records Migration
+	RedemptionRatesAtTimeOfProp = map[string]sdk.Dec{
+		"comdex-1":     sdk.MustNewDecFromStr("1.204789775596912578"),
```
riley-stride: 1.204883527965105396 from https://stride-fleet.main.stridenet.co/api/Stride-Labs/stride/stakeibc/host_zone

PR #1091, opened 2024-01-26
feature fee address

sampocs:
## Context 
There's been some discussion on the best way to create the fee module account.

## How we got here
* The code on main was working on dockernet, but threw an error when initializing the chain for the testnet
* It's not clear why it didn't work on both, but it appears that the genesis validation was being run independent from InitGenesis
* When testing the `.GetModuleAccount` approach from informal - we were forgetting to register the module with maccPerms, which is why it was failing

## Conclusion
* The `.GetModuleAccount` approach does work if we add it to maccPerms
* There's really no need to store this on the host zone if it's a normal module account - we can query it through the auth store
* There's also no need to validate the address in genesis since it's a module account

## Brief Changelog
* Removed fee address from host zone
* Added module account using `.GetModuleAccount` approach
* Added module name to maccPerms 
* Updated LS and distribute logic to use new account
* Fixed unit tests
* Fixed dockernet scripts

## Testing
* Covered by unit test
* Tested distribution to STRD stakers in dockernet


x/staketia/keeper/genesis.go
```diff
@@ -1,30 +1,21 @@
...
-	}
-	genState.HostZone.FeeAddress = feeAddress.String()
-
 	// Validate that all required fields are specified
 	if err := genState.Validate(); err != nil {
 		panic(err)
 	}
 
+	// Create fee module account
+	k.accountKeeper.GetModuleAccount(ctx, types.FeeAddress)
```
asalzmann, line 17: Is this required?

asalzmann (approved):
lgtm, this seems cleaner!

x/staketia/keeper/genesis.go
sampocs: yeah this what sets it actually!

asalzmann: Huh, weird naming that a getter sets it

PR #1103, opened 2024-02-14
Add cosmwasm 

sampocs:
## Context
Added permissioned cosmwasm to Stride for hyperlane

## Brief Changelog
* Added wasm module
* Added upgrade handler and restricted permissions (see next section)
* Updated dockernet image to include wasm bindings
* Removed the cosmwasm protos as they are no longer needed as the types can now be imported
* Switched msg types for icaoracle (used wasm types)

## Context on Permissions
* You can restrict (a) who can upload and (b) the _default_ instantiation permissions. Instantiation permissions can also be specified when uploading
* Regardless of how the permissions are configured, governance can always upload or instantiate
* Setting "Nobody" as the access type means it's governance only
* For uploading, I think we should restrict it to governance only
* For instantiating, I think we should explicitly set an admin address during uploading to make the instantiation process easier. Since we're doing this, we can set the _default_ permissions to governance only.

## Open Questions
* Are we in agreement on the permissions?
* Do we need IBC hooks? assuming no
* Do we need stargate? depends on hyperlane but assuming no
* Do we need custom wasm options?
* Are default wasm config settings fine? (i.e. gas limit, mem cache size, etc.)

## Testing
This was tested with the ICA oracle contract

#### Testing w/o Upgrade
First this was tested without the upgrade and with open permissions
* Started dockernet
```bash
make start-docker 
```
* Uploaded contract
```bash
strided tx wasm store ./artifacts/ica_oracle.wasm --gas-prices 0.1ustrd --gas auto --gas-adjustment 1.3 --from val1 -y
```
* Instantiated contract
```bash
init_msg='{"admin_address": "stride1uk4ze0x4nvh4fk0xm4jdud58eqn4yxhrt52vv7", "transfer_channel_id": "channel-0"}'
strided tx wasm instantiate 1 "$init_msg" --from val1 --label "ica-oracle" --no-admin --gas auto --gas-adjustment 1.3 -y
```
* Confirmed query worked (indicating that the contract was instantiated)
```
>>> stridedl q wasm contract-state smart stride14hj2tavq8fpesdwxxcu44rty3hh90vhujrvcmstl4zr3txmfvw9szh70ur '{"all_latest_metrics": {}}'
data:
  metrics: []
```

#### Testing Permissioned Upload after Upgrade
This was then tested via running an upgrade and confirming the permissions were set correctly
* Started dockernet with the old binary and ran through the upgrade
```bash
make UPGRADE_OLD_VERSION=v18.0.0 upgrade-build-old-binary
make UPGRADE_OLD_VERSION=v18.0.0 UPGRADE_NAME=v19 start-docker 
make UPGRADE_NAME=v17 submit-upgrade-immediately
```
* Attempted to upload the contract using the code above and confirmed it failed with error `cannot create code: unauthorized`
* Submitted proposal to upload contract
```bash
strided tx wasm submit-proposal wasm-store ./artifacts/ica_oracle.wasm \
   --title "upload oracle contract" --summary "upload oracle contract" \
   --authority stride10d07y265gmmuvt4z0w9aw880jnsr700jefnezl \
    --from val1 -y --gas 20000000 --deposit 10000000ustrd

```
* Confirmed contract was uploaded after the vote passed
```bash
>>> strided q wasm list-code
code_infos:
- code_id: "1"
  creator: stride10d07y265gmmuvt4z0w9aw880jnsr700jefnezl
  data_hash: 5EE10302357FF0F8531FE00029CD35CFE5A2C521C6A818CEE06CD61A3DF0CB42
  instantiate_permission:
    addresses: []
    permission: Nobody
pagination:
  next_key: null
  total: "0"
```

#### Testing Permissioned Instantiation after Upgrade 
Continuing off the previous example:
* Attempted to instantiate the contract using the code above and confirmed it failed with error `cannot instantiate: unauthorized`
* Submitted a proposal to instantiate through goverance
```
init_msg='{"admin_address": "stride1uk4ze0x4nvh4fk0xm4jdud58eqn4yxhrt52vv7", "transfer_channel_id": "channel-0"}'
strided tx wasm submit-proposal instantiate-contract 1 "$init_msg" \
    --authority stride10d07y265gmmuvt4z0w9aw880jnsr700jefnezl \
    --from val1 -y --gas 20000000 --deposit 10000000ustrd \
    --label "ica-oracle" --title "instantiate oracle" --summary "instantiate oracle" --no-admin
```
* Confirmed contract was instantiated 
```
>>> stridedl q wasm contract-state smart stride14hj2tavq8fpesdwxxcu44rty3hh90vhujrvcmstl4zr3txmfvw9szh70ur '{"all_latest_metrics": {}}'
data:
  metrics: []
```


go.mod
```diff
@@ -43,6 +45,7 @@ require (
 	github.com/99designs/go-keychain v0.0.0-20191008050251-8e49817e8af4 // indirect
 	github.com/99designs/keyring v1.2.1 // indirect
 	github.com/ChainSafe/go-schnorrkel v0.0.0-20200405005733-88cbf1b4c40d // indirect
+	github.com/CosmWasm/wasmvm v1.5.2 // indirect
```
p0mvn, line 48: Could this be wrong?

See this compatibility table:
https://github.com/CosmWasm/wasmd?tab=readme-ov-file#for-node-developers

sampocs: Good thought! But I believe the versions should be compatible. Stargaze and Juno both use this combo

I think the way to read that table is that wasmd `v0.45.0` uses wasmvm `v1.5.0` in it's own go.mod, _but_ wasmd `v0.45.0` should be compatible with any `v1.5.x`

Fwiw, I tried a few combinations of versions (I think I tried v1.5.0, v1.5.1 and v1.5.2) and got the same error with all of them 😞

app/app.go
```diff
@@ -666,6 +643,40 @@ func NewStrideApp(
...
+		&app.IBCKeeper.PortKeeper,
+		scopedWasmKeeper,
+		app.TransferKeeper,
+		app.MsgServiceRouter(),
+		app.GRPCQueryRouter(),
+		wasmDir,
+		wasmConfig,
+		wasmCapabilities,
+		authtypes.NewModuleAddress(govtypes.ModuleName).String(),
+		wasmOpts...,
```
sampocs, line 675: @p0mvn do you know if we need to include the [stargate wasm options](https://github.com/osmosis-labs/osmosis/blob/main/app/keepers/keepers.go#L483-L484)? 

For context, we want a minimum installation that would allow deploying hyperlane.

My assumption is that we would only need this if the hyperlane contracts needed to query from the chain layer? But very low confidence

piobab: IMO you are right. We use `stargate` queries on Osmosis for chain modules such as pool manager, twap, downtimedetector.

sampocs:
per open questions:

> Are we in agreement on the permissions?

Discussed offline, in agreement. Need to remember to set the instantiate address when uploading.

> Do we need IBC hooks? Do we need stargate? Do we need custom wasm options?

Successful hyperlane deployment confirms we do not need any of these

> Are default wasm config settings fine? (i.e. gas limit, mem cache size, etc.)

I believe these are just validator config settings so easy to change later if needed. And I assume the defaults are fine since this is a light-weight deployment

app/upgrades/v19/upgrades.go
```diff
@@ -0,0 +1,46 @@
...
+		ctx.Logger().Info("Running module migrations...")
+		newVm, err := mm.RunMigrations(ctx, configurator, vm)
+		if err != nil {
+			return newVm, err
+		}
+
+		// Update wasm params so that contracts can only be uploaded through governance
+		wasmParams := wasmKeeper.GetParams(ctx)
+		wasmParams.CodeUploadAccess = wasmtypes.AccessConfig{
+			Permission: wasmtypes.AccessTypeNobody,
```
ethan-stride, line 36: So this means when we deploy to the real main net we will deploy hyperlane via gov props?

app/app.go
```diff
@@ -666,6 +643,39 @@ func NewStrideApp(
 		app.AccountKeeper, app.BankKeeper, app.DistrKeeper, app.StakingKeeper,
 	)
 
+	// Add wasm keeper
+	wasmCapabilities := "iterator,staking,stargate,cosmwasm_1_1,cosmwasm_1_2,cosmwasm_1_3,cosmwasm_1_4"
```
ethan-stride, line 647: In the go mod it looks like it is using wasm 1.5.2 so does this mean we should add wasmCapabilities cosmwasm_1_5 as well?

ethan-stride (approved):
LGTM
As discussed none of us are experts for many of these settings so I am just calling out anything which looks potentially unneeded at this point.  I was comparing against the osmosis cosmwasm setup but they probably have additional usecases we don't want so we might be able to lock it down even more.

app/app.go
sampocs: I don't think there is a 1_5

https://github.com/CosmWasm/wasmd?tab=readme-ov-file#for-contract-developers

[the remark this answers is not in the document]
sampocs: yeah i meant the wasm options and query interface. I'm not sure what these capabilities are, but I think it's safest to just leave it

app/app.go
```diff
@@ -821,6 +831,7 @@ func NewStrideApp(
 		claim.NewAppModule(appCodec, app.ClaimKeeper),
 		// technically, app.GetSubspace(packetforwardtypes.ModuleName) will never be run https://github.com/cosmos/ibc-apps/issues/146#issuecomment-1839242144
 		packetforward.NewAppModule(app.PacketForwardKeeper, app.GetSubspace(packetforwardtypes.ModuleName)),
+		wasm.NewAppModule(appCodec, &app.WasmKeeper, app.StakingKeeper, app.AccountKeeper, app.BankKeeper, app.BaseApp.MsgServiceRouter(), app.GetSubspace(wasmtypes.ModuleName)),
```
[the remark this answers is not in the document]
sampocs: It's the store that stores parameters

app/upgrades/v19/upgrades.go
sampocs: yeah that's the current plan. I think there's a chance we may want to switch it to an address but we can handle that in a separate PR if needed

PR #1099, opened 2024-02-04
chore(dev-deps): upgrade ledger dependencies

keupsonite:
Please see:

- https://github.com/cosmos/cosmos-sdk/issues/18319
- https://github.com/cosmos/cosmos-sdk/issues/18423#issuecomment-1804129041

PR #1106, opened 2024-02-23
ledger nano s fix

riley-stride:
nano s signing issues, had to modify go.mod to use zondax/ledger-go v0.14.3 instead of the v0.14.1 then rebuild, then was able to sign


keupsonite:
Just in case, a PR already cover this https://github.com/Stride-Labs/stride/pull/1099

sampocs:
closing in favor of #1099

PR #1112, opened 2024-03-01
stakeibc re-org part 1 (keepers)

sampocs:
## Context
Stakeibc has grown quite large over the years. This PR does a file re-org cleanup to better organize functions and remove ignite tech debt. This should make it easier to onboard new devs to the codebase. 

**Note to reviewer**: There were no logical changes made in this PR. It only copy pasted functions to different files. When reviewing, I'd recommend scrolling each commit and making sure all functions are accounted for during the move.

## Brief Changelog
* Renamed validator_section.go → rebalance.go
* Renamed unbonding_record.go → unbonding.go
* Created validator.go
* Moved function into host_zone.go
* Moved functions into epoch_tracker.go
* Created records.go


ethan-stride (approved):
LGTM 

Every function which was deleted I searched for and found where it was moved
Same thing for the tests

There were 3 net deletions which were not moved but you claim they were dead unused code
I don't think there were net logical changes but the naming of the files is better

I wish I could give better design feedback but I would need to parse how everything is calling to each other in order to do that, this was a very mechanical check.  Hopefully the linter/type checker and the unit tests are enough to ensure soundness on this change.

PR #1114, opened 2024-03-01
Upgrade to ICS v4.0.0 and sdk v0.47.10 

sampocs:
## Context
Upgraded ICS to v4.0.0 and SDK to v0.47.10. Per the [upgrade guide](https://github.com/cosmos/interchain-security/blob/release/v4.0.x/UPGRADING.md#v40x), this upgrade requires the inclusion of a custom migration function.

## Brief Changelog
* Upgraded to sdk v0.47.10 (no changes required outside of go.mod)
* Upgraded to ICS v4.0.0
  * Required a change to import paths from v3 → v4
  * Required a change of the genesis state which now has the provider state as a nested struct [[ref](https://github.com/cosmos/interchain-security/pull/1324/files#diff-62edbb49aa296149eefa80f374cf36a4ff5e7ee4285146120e71cbdfd4eed130)]
  * Updated unit test setup as needed

## Testing
* I've asked the informal team if they have a unit test for the ICS migration function. This will also be tested on testnet.


shellvish (approved):
I don't have a great understanding of what ICS v4 entails, but this PR overall looks good to me! 

Would also just want to double check in dockernet testing and localstride that this produces the expected output

PR #1120, opened 2024-03-04
stakeibc file re-org part 5 (msg_server)

sampocs:
## Context
Stakeibc has grown quite large over the years. This PR does a file re-org cleanup to better organize functions and remove ignite tech debt. This should make it easier to onboard new devs to the codebase. 

**Note to reviewer**: There were no logical changes made in this PR. It only copy pasted functions to different files. When reviewing, I'd recommend scrolling each commit and making sure all functions are accounted for during the move.

## Brief Changelog
Consolidates msg_server functions into one file. This is just a starting point. In later PRs, any longer message (specifically those that have many cases of test coverage) will be split into their own files (e.g. liquidstake.go).


ethan-stride (approved):
LGTM
Verified all functions and constants were simply moved, anything which was created was deleted from somewhere else.

PR #1123, opened 2024-03-05
switched to open sourced rate limit module

sampocs:
## Context
Switched over the the rate limit module that's moved into it's own separate [repo](https://github.com/Stride-Labs/ibc-rate-limiting).

Noteable changes to the version in the isolated repo is it removed the legacy gov handlers and epochs dependency.

## Brief Changelog
* Wired different keeper into app.go
* Updated import statements
* Removed legacy gov proposals types and rate limit epoch hooks
  * The open sourced module uses the non-legacy gov tx format and removes the epoch dependency with a lighter weight implementation
* Removed `x/ratelimit` and `proto/stride/ratelimit`
* Added migration to new module in upgrade handler


ethan-stride (approved):
LGTM

PR #1124, opened 2024-03-05
parameterize max messages per ICA

sampocs:
## Context
To optionally enable delegations to the whole validator set, this PR parameterizes the number of messages per ICA.

## Brief Changelog
* Added parameter `MaxIcaMessagesPerTx` 
* Replaced the hard coded "batch size" with the parameter
* Sorted the unbonding messages by unbalance ratio (reversion of [this](https://github.com/Stride-Labs/stride/commit/8f37be3e23179a92b48a27c04c227c97024de33a) commit). The sort was changed a few upgrades ago to ensure that the number of validators was under the limit, but that is no longer necessary
* Added new parameter in upgrade handler

## Assumptions made (please check me!)
* ~We can change the sort now that we're delegating to more than 32 validators (in fact, if we don't add it back in, we won't be able to unbond from the full set)~ EDIT: Leaving the sort as is in case we run into issues over 32 vals
* `ConsolidateUnbondingMessages` should be left in place in case we want to set the parameter back to 32
* There isn't anything we need to do for delegations since we can control it by adding/removing validators
* Confirm all our parameters are currently set to the default on mainnet


x/stakeibc/keeper/unbonding.go
```diff
@@ -206,7 +202,13 @@ func SortUnbondingCapacityByPriority(validatorUnbondCapacity []ValidatorUnbondCa
 		validatorA := validatorUnbondCapacity[i]
 		validatorB := validatorUnbondCapacity[j]
 
-		// TODO: Once more than 32 validators are supported, change back to using balance ratio first
+		balanceRatioValA, _ := validatorA.GetBalanceRatio()
+		balanceRatioValB, _ := validatorB.GetBalanceRatio()
+
+		// Sort by the balance ratio first - in ascending order - so the more unbalanced validators appear first
```
ethan-stride, line 208: If the default value is literally 100 and we can reach all validators now, why is sorting at all important?  There is no cutoff for the order here to matter right?

proto/stride/stakeibc/params.proto
```diff
@@ -27,6 +27,7 @@ message Params {
   uint64 ibc_transfer_timeout_nanos = 16;
   uint64 validator_slash_query_threshold = 19;
   uint64 validator_weight_cap = 20;
+  uint64 max_ica_messages_per_tx = 21;
```
ethan-stride, line 30: Is it per tx or per block?  Before I thought that it was about the block size how many would fit simultaneously.  Can't more than one tx happen in a block but from the chains point of view these would still be essentially happening at the same time?

ethan-stride (approved):
LGTM

x/stakeibc/keeper/unbonding.go
sampocs: disregard this change - I've since reverted it

But to answer your questions, it's important in the case that we have to fall back down to 32 validators, and it makes the unbondings more "fair" instead of just targeting the top vals

proto/stride/stakeibc/params.proto
sampocs: the limit (we believe) is a gas limit for a single tx. Each ICA is one tx that's built up of many messages - the more messages, the more gas it uses

asalzmann (changes requested):
I don't think this will work - some host zones have lower gas limits. I thought we were going to do this by host zone, not across them all

asalzmann:
https://docs.google.com/spreadsheets/d/1eMvlfZTLTt9_TElU2l5nhSv9LvKJdt4dtVEwZ2SSg5w/edit#gid=0
Screenshot 2024-03-07 at 12 13 11 PM

min across all chains is Evmos with 40M, not sure that's enough to support a tx that stakes/unstakes to 100 vals (on Celestia a tx with ~70-80 takes ~25M gas, but it's been a bit unpredictable so would be nice to leave a buffer)

sampocs:
> I don't think this will work - some host zones have lower gas limits. I thought we were going to do this by host zone, not across them all

I thought you said to do it as a global param and we would control the list of validators for non-hub/osmo

@asalzmann

asalzmann:
Vishal and I chatted
* Main relevant deadline for delegating to more than 32 vals is when Celestia adds ICAs (so we can continue delegating to the whole set) - likely April
* In theory, we could set the limit higher today (e.g. 100 vals) and keep track of how many validators we delegate to by host zone, but seems complex / like a lot of work
* Adding this as a param vs. having it hard coded doesn't add much value

=> going to close this PR in favor of either setting this by host zone or adding a proper fix (unbundling the staking/unstaking tx so they can be spread across multiple txs) in a later upgrade

PR #1125, opened 2024-03-05
set wasm permissions

sampocs:
## Context
The hyperlane deployment scripts are setup to upload and instantiate via an account instead through governance. This PR sets an admin address to use throughout the deployment phase. Afterwards, the permissions will be updated again to be through governance. 

## Brief Changelog
* Set admin address with upload permissions
* Pulled out code into it's own function and added unit test


app/upgrades/v19/upgrades_test.go
```diff
@@ -0,0 +1,40 @@
...
+
+	// Check state after upgrade
+	s.CheckWasmPerms()
+}
+
+func (s *UpgradeTestSuite) CheckWasmPerms() {
+	wasmParams := s.App.WasmKeeper.GetParams(s.Ctx)
+	s.Require().Equal(wasmtypes.AccessTypeAnyOfAddresses, wasmParams.CodeUploadAccess.Permission, "upload permission")
+	s.Require().Equal(v19.WasmAdmin, wasmParams.CodeUploadAccess.Addresses[0], "upload address")
+	s.Require().Equal(wasmtypes.AccessTypeNobody, wasmParams.InstantiateDefaultPermission, "instantiate permission")
```
ethan-stride, line 39: What is the purpose of this check?  This shouldn't be relevant once we explicitly whitelist an admin address

ethan-stride (approved):
LGTM

app/upgrades/v19/upgrades_test.go
ethan-stride: Specifically line 39

sampocs: instantiate permissions just define the "default", meaning the perms if you don't specify anything when uploading

But when we upload, we'll specify the admin address as the only one who can instantiate

PR #1126, opened 2024-03-05
Add x/stakedym

shellvish:
Adds x/stakedym module, tested on local->mainnet flow (staking, unstaking, compounding).

Please note, this does not test granular authz permissions, those should be tested separately.

Also note, we will need to update addresses in `x/stakedym/types/dymension.go`.

Combed through each file, and made sure any reference to tia could be changed without issue. Also high-level traced through logic and think everything should function smoothly.


x/stakedym/types/dymension.go
```diff
@@ -0,0 +1,20 @@
+// #nosec G101
+package types
```
sampocs, line 2: verified everything but the addresses. Do you have a doc I can use for the addresses?

x/stakedym/types/msgs_test.go
```diff
@@ -1,950 +0,0 @@
-package types_test
```
sampocs: did you mean to delete this file?

sampocs (commented):

Also can you add the store key?

x/stakedym/types/dymension.go
```diff
@@ -0,0 +1,20 @@
...
+
+const (
+	DymensionChainId                   = "dymension_1100-1"
+	StrideToDymensionTransferChannelId = "channel-197"
+	DymensionNativeTokenDenom          = "adym"
+	DymensionNativeTokenIBCDenom       = "ibc/FF6C2E86490C1C4FBBD24F55032831D2415B9D7882F85C3CC9C2401D79362BEA" // #nosec G101
```
riley-stride, line 8:
```
base ❯ dymd q ibc-transfer denom-trace FF6C2E86490C1C4FBBD24F55032831D2415B9D7882F85C3CC9C2401D79362BEA
denom_trace:
  base_denom: ustrd
  path: transfer/channel-0
```

x/stakedym/types/msgs_test.go
shellvish: No, good catch, adding back in!

asalzmann (approved):

PR #1127, opened 2024-03-06
added sttia rate limit

sampocs:
## Context 
Added 10% stTIA rate limit to celestia and osmosis


app/upgrades/v19/upgrades.go
```diff
@@ -65,3 +76,28 @@ func MigrateRateLimitModule(ctx sdk.Context, k ratelimitkeeper.Keeper) {
 	hourEpoch.EpochStartHeight = ctx.BlockHeight()
 	k.SetHourEpoch(ctx, hourEpoch)
 }
+
+// Add a 10% rate limit for stTIA to Celestia and Osmosis
+func AddStTiaRateLimit(ctx sdk.Context, k ratelimitkeeper.Keeper) error {
+	addRateLimitMsgTemplate := ratelimittypes.MsgAddRateLimit{
```
ethan-stride, line 82: I thought that this was done via proposal which had to be approved via governance... I didn't think we could directly do this via keeper functions

app/upgrades/v19/upgrades.go
```diff
@@ -65,3 +76,28 @@ func MigrateRateLimitModule(ctx sdk.Context, k ratelimitkeeper.Keeper) {
...
+		MaxPercentRecv: sdk.NewInt(10),
+		DurationHours:  RateLimitDurationHours,
+	}
+
+	addCelestiaMsg := addRateLimitMsgTemplate
+	addCelestiaMsg.ChannelId = CelestiaTransferChannelId
+
+	addOsmosisMsg := addRateLimitMsgTemplate
+	addOsmosisMsg.ChannelId = OsmosisTransferChannelId
+
```
ethan-stride, line 94: Now that we have incentivized stTia on so many different chains should we add similar messages for all the chains we are incentivizing?  Seems like right now people can move a lot of volume through other chains and we wouldn't notice.  Even if we give those other chains looser different bounds because the capital might only be from a handful of people at the moment seems like a good check

ethan-stride (approved):
LGTM

app/upgrades/v19/upgrades.go
sampocs: We’re planning to switch to a global rate limit soon that should cover all these cases.

in the short term, we could probably also add neutron since that pool has a lot of liquidity but I think we can wait on the others

app/upgrades/v19/upgrades.go
sampocs: Anything is possible in the upgrade handler!

ethan-stride (approved):
LGTM

PR #1135, opened 2024-03-15
add community pool rebate feature

sampocs:
## Context 
This feature provides the option for large community pool liquid stakes to be eligible for a rebate.

This PR will act as the base PR for this effort. Initial commits in this PR are only:
* Proto changes to add the rebate to the host zone struct
* Added helper on host zone to check if there is a rebate
* Fixed misc lint errors (figured they were small enough that I'd just bundle them into this PR)

Remaining work for this feature will be in branched PRs, including:
* Rebate calculation and fund community pool ICA
* Tx to register a rebate
* Txs to grant/revoke authz permissions
* Disabling of ICA-enabled trading


ethan-stride (approved):
LGTM

riley-stride:
looking good so far

proto/stride/stakeibc/host_zone.proto
```diff
@@ -7,6 +7,19 @@ import "cosmos_proto/cosmos.proto";
...
 option go_package = "github.com/Stride-Labs/stride/v19/x/stakeibc/types";
 
+// CommunityPoolRebate stores the size of the community pool liquid stake
+// and the rebate percentage
+message CommunityPoolRebate {
+  string rebate_percentage = 1 [
+    (gogoproto.customtype) = "github.com/cosmos/cosmos-sdk/types.Dec",
+    (gogoproto.nullable) = false
+  ];
+  string liquid_stake_amount = 2 [
```
riley-stride, line 17: nit: I think `liquid_staked_amount` (past tense) would be more clear. I found the present tense confusing [here](https://github.com/Stride-Labs/stride/pull/1136/files#diff-8850ef830f8cbe0e6ca0be787be1036bc793215c526fb43491a812ca4f0bb079R108), made me think we were referring to tokens in the process of being liquid staked.

proto/stride/stakeibc/host_zone.proto
```diff
@@ -7,6 +7,19 @@ import "cosmos_proto/cosmos.proto";
 
 option go_package = "github.com/Stride-Labs/stride/v19/x/stakeibc/types";
 
+// CommunityPoolRebate stores the size of the community pool liquid stake
+// and the rebate percentage
+message CommunityPoolRebate {
+  string rebate_percentage = 1 [
```
ethan-stride, line 13: As I have continued to think about the whole design, I think it would be much simpler to not have this additional struct, but just have the rebate_percentage which will be 0 by default on the hostzone struct.  

The liquid stake amount should not remain fixed because the amount of stDYDX in the comm pool won't remain fixed -- we should epochly ICQ to measure the current amount in the comm pool and do computations on the callback of that ICQ.  This should be simpler than having to update this structure by hand using txs when they vote to stake more or choose to sell off part of their holdings and it eliminates this whole message type.

ethan-stride: Nevermind, ignore this comment.  Discussion on slack concluded this might be a fast follow but not needed in this first pass.

proto/stride/stakeibc/host_zone.proto
sampocs: totally agree! going to leave this comment open and address it once everything's merged into this so I don't have to deal with a bunch of annoying merge conflicts

proto/stride/stakeibc/host_zone.proto
```diff
@@ -7,6 +7,19 @@ import "cosmos_proto/cosmos.proto";
 
 option go_package = "github.com/Stride-Labs/stride/v19/x/stakeibc/types";
 
+// CommunityPoolRebate stores the size of the community pool liquid stake
+// and the rebate percentage
+message CommunityPoolRebate {
```
asalzmann, line 12: Might not have time, but it seems like we could generalize this by
* calling the struct `Rebate`
* storing the address on the struct
* making `community_pool_rebate` on the host zone (whatever it would be renamed to) a repeated field

Then this could support arbitrary DAOs, or chains with multiple community pools. Not needed for DYDX

x/stakeibc/keeper/redemption_sweep.go
```diff
@@ -49,11 +49,6 @@ func (k Keeper) SweepAllUnbondedTokensForHostZone(ctx sdk.Context, hostZone type
 		if inTransferQueue && validUnbondingTime {
 			k.Logger(ctx).Info(utils.LogWithHostZone(hostZone.ChainId, "  %v%s included in sweep", hostZoneUnbonding.NativeTokenAmount, hostZoneUnbonding.Denom))
 
-			if err != nil {
```
asalzmann, line 52: why remove this?

sampocs: the `err` variable in this statement is actually defined way up above. This was caught by the linter btw

My guess is that we used to have something like
```go
if inTransferQueue && validUnbondingTime {
    k.Logger(...)
    
    err := something...  <---
    if err != nil {
        ...
```

and we removed the line with the arrow, but not the corresponding if statement below

proto/stride/stakeibc/host_zone.proto
sampocs: discussed this this morning - decided that it makes more sense to focus on just 1 pool for now since a second pool would require a different ICA message and DYDX only needs the main pool to be funded atm

proto/stride/stakeibc/host_zone.proto
sampocs: @riley-stride going to handle this in the PR that changes the % calculations to use stDYDX instead of DYDX.

Do you have a name preference? Maybe `liquid_staked_sttoken_amount` ?

riley-stride: `liquid_staked_sttoken_amount` sgtm!

sampocs: sweet - going to mark this resolved since it will be in the other PR

PR #1136, opened 2024-03-15
rebate calculation and fund community pool ICA

sampocs:
## Context
This PR adds the code to send the rebated funds to the community pool.

The high level flow is:
* Rewards land in the withdrawal account (as USDC)
* We query the USDC balance epochly
* In the callback, we
  * Check if there's a rebate for the host zone
  * If so, we split out the rebate amount and send it to the community pool
  * The remaining amount will sent to Osmosis so it can get swapped for the native token
* Next epoch, we query the DYDX balance
* In the callback, we
  * Check if there's a rebate for the host zone
  * If so, calculate what the rewards were _before_ the rebate was issued
  * Then use that reward amount to calculate the fee amount
  * Send the fee to the fee account, and reinvest the remainder

The rebate amount was calculated as follows (see code docstring for more details and an example):
```
rebateAmount = rewardsAmount * strideFeePercentage * (communityPoolLiquidStakeSize / TVL) * rebatePercentage
```

**Note to reviewer:** I'd recommend reviewing commit by commit

## Brief Changelog
* Added `CheckForCommunityPoolRebate` to check if there's a rebate for the given host zone, and, if so, calculate the rebate amount
  * Note: I refer to the non-rebate amount as the "reinvest" amount, which is slightly misleading so lmk if you have a preference for a different name (maybe `tradeAmount`?)
* Parameterized the fund community pool ICA function so that it can be submitted from the withdrawal ICA
* Wired up the above two functions in to the withdrawal reward balance callback
* Added/extended unit tests for the above


x/stakeibc/keeper/reward_converter.go
```diff
@@ -51,6 +57,67 @@ type ForwardMetadata struct {
...
+	// It shouldn't be possible to have 0 delegations (since there are rewards and there was a community pool stake)
+	// This will also prevent a division by 0 error
+	if hostZone.TotalDelegations.IsZero() {
+		return feeInfo, errorsmod.Wrapf(types.ErrDivisionByZero,
+			"unable to calculate rebate amount for %s since total delegations are 0", chainId)
+	}
+
+	// The rebate amount is determined by the contribution of the community pool stake towards the total TVL,
+	// multiplied by the rebate fee percentage
+	contributePercentage := sdk.NewDecFromInt(rebateInfo.LiquidStakeAmount).Quo(sdk.NewDecFromInt(hostZone.TotalDelegations))
```
riley-stride, line 108: Sanity checking the denominator

> Should we count newly-deposited-not-yet-delegated TVL in the denominator? 

No bc it's not contributing to rewards for this epoch 

> Should we count newly-unbonding TVL whose rewards from this epoch are in the withdrawal account?

No because it would be very challenging to parcel that out (would need to measure what share of the epoch it was earning rewards for and exclude it from this calc) - much easier and very similar to take this simpler approach. The net effect if there are unbondings during the epoch is that the comm pool gets a slightly higher rebate that it otherwise would, which seems fine.

x/stakeibc/keeper/reward_converter.go
```diff
@@ -51,6 +57,67 @@ type ForwardMetadata struct {
...
+	rebateAmount := sdk.NewDecFromInt(totalFeesAmount).Mul(contributePercentage).Mul(rebateInfo.RebatePercentage).TruncateInt()
+	strideFeeAmount := totalFeesAmount.Sub(rebateAmount)
```
riley-stride, line 110: Should we add some sanity checks in here e.g.
- strideFeeAmount > 0 (should it just be a uint?)
- 0 <= contributePercentage <= 1
- rebateAmount >= 0

riley-stride, line 110: Should we add some sanity checks in here e.g.
- strideFeeAmount > 0 (should it just be a uint?)
- 0 <= contributePercentage <= 1
- rebateAmount >= 0

x/stakeibc/keeper/icqcallbacks_withdrawal_balance.go
```diff
@@ -62,32 +58,15 @@ func WithdrawalBalanceCallback(k Keeper, ctx sdk.Context, args []byte, query icq
 		return errorsmod.Wrapf(types.ErrICAAccountNotFound, "no fee account found for %s", chainId)
 	}
 
-	// Determine the stride commission rate to the relevant portion can be sent to the fee account
-	params := k.GetParams(ctx)
-	strideCommissionInt, err := cast.ToInt64E(params.StrideCommission)
+	// Split the withdrawal amount into a rebate, stride fee, and reinvest portion
```
riley-stride, line 61: It seems like some of the sanity checks from `WithdrawalBalanceCallback` didn't get ported over to `CheckForCommunityPoolRebate` - should we port those over? e.g. sanity checks on commission (0 <= x <= 1)

x/stakeibc/keeper/icqcallbacks_withdrawal_balance.go
```diff
@@ -62,32 +58,15 @@ func WithdrawalBalanceCallback(k Keeper, ctx sdk.Context, args []byte, query icq
...
+	feeInfo, err := k.CheckForCommunityPoolRebate(ctx, chainId, withdrawalBalanceAmount)
```
riley-stride, line 62: I don't understand the difference between `WithdrawalBalanceCallback` and `WithdrawalRewardBalanceCallback` - why do we discard `feeInfo.RebateAmount` in `WithdrawalBalanceCallback`?

If it's discarded because this callback is only used for non-DYDX zones, might be safe to verify it's zero rather than ignoring it here.

x/stakeibc/keeper/icqcallbacks_withdrawal_balance_test.go
```diff
@@ -228,6 +228,22 @@ func (s *KeeperTestSuite) TestWithdrawalBalanceCallback_NoFeeAccount() {
 	s.Require().EqualError(err, "no fee account found for GAIA: ICA acccount not found on host zone")
 }
 
+func (s *KeeperTestSuite) TestWithdrawalBalanceCallback_FailedToCheckForRebate() {
```
riley-stride, line 231: Note to self @riley-stride -  review this again once you understand the difference between `WithdrawalBalanceCallback` and `WithdrawalRewardBalanceCallback`

x/stakeibc/keeper/icqcallbacks_withdrawal_reward_balance_test.go
```diff
@@ -102,6 +110,25 @@ func (s *KeeperTestSuite) TestWithdrawalRewardBalanceCallback_SuccessfulNoTransf
 	})
 }
 
+// Verify that if the amount returned by the ICQ response is less than the min_swap_amount, no transfer happens
+func (s *KeeperTestSuite) TestWithdrawalRewardBalanceCallback_SuccessfulWithRebate() {
```
riley-stride, line 114: Nit: we should add tests for the rebate = 0 and rebate = 1 edge cases

x/stakeibc/keeper/reward_converter_test.go
```diff
@@ -33,6 +33,184 @@ type TransferRewardHostToTradeTestCase struct {
...
+			rewardAmount:             sdkmath.NewInt(1000),
+			strideFee:                10,
+			rebatePercentage:         sdk.MustNewDecFromStr("0.79"),
+
+			expectedRebateAmount:    sdkmath.NewInt(7),
+			expectedStrideFeeAmount: sdkmath.NewInt(93),
+			expectedReinvestAmount:  sdkmath.NewInt(900),
+		},
+		{
+			// No rebate - all should be reinvested
```
riley-stride, line 155: nit: add case where entire TVL is comm pool TVL

riley-stride (changes requested):
Looking like a great start! 

Added some questions and suggestions. 

Main big picture question: what's the difference between `WithdrawalBalanceCallback` and `WithdrawalRewardBalanceCallback` — why is the rebate handled differently across the two?

x/stakeibc/keeper/community_pool.go
```diff
@@ -250,7 +255,7 @@ func (k Keeper) FundCommunityPool(ctx sdk.Context, hostZone types.HostZone, toke
 	_, err := k.SubmitTxs(ctx,
 		hostZone.ConnectionId,
 		msgs,
-		types.ICAAccountType_COMMUNITY_POOL_RETURN,
+		senderAccountType,
```
ethan-stride, line 258: Why do you need to make this parametrized?  The account which sends the funds back into the pool is known to be this ICA and the type is known as well right?

x/stakeibc/keeper/icqcallbacks_withdrawal_balance.go
ethan-stride: WithdrawalBalance is the ICQ which is part of the normal staking flow.  In this case it is looking for DYDX to turn into stDYDX.  WithdrawalRewardBalance is looking for reward tokens in the withdrawal account which need to be swapped first before being returned to the withdrawal account.  In this case looking for USDC in withdrawal.

WithdrawalBalanceCallback gets the amount of DYDX in withdrawal and does the liquid stake.  WithdrawalRewardBalanceCallback gets the amount of USDC and transfers it to the trade ICA.  Eventually it will come back to the withdrawal account having been swapped as DYDX and then the normal further staking flow can happen from there.  The reward one only applies to DYDX at the moment because it is the only one with trade routes yet but the one without reward is part of the core triggers for liquid stake reinvestment functioning

ethan-stride: The Rebate amount is used when the USDC comes in originally because the rebate in USDC is sent back at that time.  This function is detecting funds which have already been converted to DYDX -- the withdrawalBalanceAmount is in DYDX -- because this is when stride takes its fee.  RebateAmount is ignored in this function but it is used in the other one which is measuring the USDC which has come in as revenue.

ethan-stride (commented):
Still reading it again in a second pass but mostly looks good to me.

x/stakeibc/keeper/icqcallbacks_withdrawal_balance.go
riley-stride: That's very helpful, thanks @ethan-stride! Makes sense that these are separate. 

Could I recommend we rename them for clarity?

Here's one option that makes sense to me, but open to others:
*  `WithdrawalBalanceCallback` -> `WithdrawalNativeBalanceCallback`
* `WithdrawalRewardBalanceCallback` -> WithdrawalNonNativeRewardBalanceCallback`

riley-stride:
Check me - I think this is what’s happening here but not 100% sure:

**Example:**
10 CP Liquid Stake, 100 TVL => 10% contribution
1000 rewards, 10% stride fee => 100 total fees

**Desired outcome:**
100 total fees * 10% contribution * 50% rebate
=> 5 rebate, 95 stride fee, 900 reinvest

Currently what happens for the DYDX zone (assume for simplicity DYDX price is $1) is:
* `WithdrawalNonNativeDenomRewardBalanceCallback` will send 5 to the CP, reinvest 995
* `WithdrawalNativeDenomRewardBalanceCallback` will get 995 and take 10% of that as a fee which is 99.5

However it should instead do:
* `WithdrawalNonNativeDenomRewardBalanceCallback` will send 5 to the CP, reinvest 995
* `WithdrawalNativeDenomRewardBalanceCallback` will get 995. It should take 95 as a fee, which it can do in the following way:
    * OriginalRewardAmount = 995/(1-rebateAmount) [note: rebate amount calc’d [here](https://github.com/Stride-Labs/stride/pull/1136/files#diff-8850ef830f8cbe0e6ca0be787be1036bc793215c526fb43491a812ca4f0bb079R109%5D)
    * StrideFee = calc strideFeeAmount using OriginalRewardAmount as `CheckForCommunityPoolRebate
`’s param rewardAmount

To implement this new logic, could either branch inside `CheckForCommunityPoolRebate` or might be cleanest to write a replacement function to be called from `WithdrawalNativeDenomRewardBalanceCallback` in the case feeinfo.RebateAmount > 0.

x/stakeibc/keeper/icqcallbacks_withdrawal_balance.go
ethan-stride: We had been using the word "Reward" denom to specifically mean revenue tokens which need to be swapped as compared to the "HostDenom" which is what most chains put into the withdrawal account directly because they inflate the token native to the hostzone.  This terminology is used through the traderoute and hostzone datastructs

So I would suggest `WithdrawalBalanceCallback` -> `WithdrawalHostDenomBalanceCallback` if we want to change it at all but definitely don't include the word reward in that one.  It is part of the route which is most core to all liquid staking stride logic so I am fine with it being the very short name it has now.  And then for the other one, rewards are implicitly in another denom than the host denom so no need to include NonNative

riley-stride: I find it confusing because the SDK uses "rewards" as the standard name for all staking rewards, which are typically in the native denom. E.g. claim rewards usually means claim native token rewards.

ethan-stride: We should probably do a refactor and name what we had called "rewards" to "revenue" instead.  Then USDC would be the revenue denom, and all other chains would get their rewards in the host denom

x/stakeibc/keeper/community_pool.go
```diff
@@ -226,7 +226,12 @@ func (k Keeper) RedeemCommunityPoolTokens(ctx sdk.Context, hostZone types.HostZo
 
 // Using tokens in the CommunityPoolReturnIcaAddress, trigger ICA tx to fund community pool
 // Note: The denom of the passed in token has to be the denom which exists on the hostZone not Stride
-func (k Keeper) FundCommunityPool(ctx sdk.Context, hostZone types.HostZone, token sdk.Coin) error {
+func (k Keeper) FundCommunityPool(
+	ctx sdk.Context,
+	hostZone types.HostZone,
+	token sdk.Coin,
+	senderAccountType types.ICAAccountType,
+) error {
```
ethan-stride, line 234: !
You need to also add the address of the ICA which is sending these tokens to community pool.  This is hardcoded on line 235/240 in the MsgFundCommunityPool as the Despositor field for the way that it was before but you forgot to update it -- if you intend to call this from the withdrawal address you need to pass in the address for withdrawal ICA to that Msg type as Depositor.

x/stakeibc/keeper/icqcallbacks_withdrawal_reward_balance.go
```diff
@@ -54,14 +58,34 @@ func WithdrawalRewardBalanceCallback(k Keeper, ctx sdk.Context, args []byte, que
...
+	// Split the withdrawal amount into a rebate, stride fee, and reinvest portion
+	feeInfo, err := k.CheckForCommunityPoolRebate(ctx, chainId, withdrawalRewardBalanceAmount)
+	if err != nil {
+		return errorsmod.Wrapf(err, "unable to check for community pool rebate")
+	}
+
+	// If there's a rebate portion, fund the community pool with that amount
+	if feeInfo.RebateAmount.GT(sdkmath.ZeroInt()) {
+		rebateToken := sdk.NewCoin(tradeRouteCallback.RewardDenom, feeInfo.RebateAmount)
+		if err := k.FundCommunityPool(ctx, hostZone, rebateToken, types.ICAAccountType_WITHDRAWAL); err != nil {
```
ethan-stride, line 70: Reminder, also pass in the Withdrawal ICA address as Depositor on the actual MsgFundCommunityPool

x/stakeibc/keeper/icqcallbacks_withdrawal_reward_balance.go
```diff
@@ -54,14 +58,34 @@ func WithdrawalRewardBalanceCallback(k Keeper, ctx sdk.Context, args []byte, que
...
+			return errorsmod.Wrapf(err, "unable to submit fund community pool ICA")
+		}
+
+		k.Logger(ctx).Info(utils.LogICQCallbackWithHostZone(chainId, ICQCallbackID_WithdrawalRewardBalance,
+			"Sending rebate tokens %v %s to community pool",
+			feeInfo.RebateAmount, tradeRoute.RewardDenomOnHostZone))
```
ethan-stride, line 76: tradeRouteCallback.RewardDenom is actually tradeRoute.RewardDenomOnRewardZone not on host zone which would be an ibc hash in this log statement

x/stakeibc/keeper/icqcallbacks_withdrawal_reward_balance.go
```diff
@@ -54,14 +58,34 @@ func WithdrawalRewardBalanceCallback(k Keeper, ctx sdk.Context, args []byte, que
...
+	// We transfer both the amount to be reinvested, and the amount for the stride fee
+	tradeAmount := feeInfo.ReinvestAmount.Add(feeInfo.StrideFeeAmount)
+	if err := k.TransferRewardTokensHostToTrade(ctx, tradeAmount, tradeRoute); err != nil {
 		return errorsmod.Wrapf(err, "initiating transfer of reward tokens to trade ICA failed")
 	}
 
 	k.Logger(ctx).Info(utils.LogICQCallbackWithHostZone(chainId, ICQCallbackID_WithdrawalRewardBalance,
 		"Sending discovered reward tokens %v %s from hostZone to tradeZone",
-		withdrawalRewardBalanceAmount, tradeRoute.RewardDenomOnHostZone))
+		feeInfo.ReinvestAmount, tradeRoute.RewardDenomOnHostZone))
```
ethan-stride, line 88: I think this should be `tradeAmount` instead of `feeInfo.ReinvestAmount`
and `tradeRoute.RewardDenomOnRewardZone` instead of `tradeRoute.RewardDenomOnHostZone` which is ibc

x/stakeibc/keeper/icqcallbacks_withdrawal_balance.go
ethan-stride, line 62: This is a logic error to call this here unmodified because it will compute and remove the rebateFee again...
The amount being passed in here is `withdrawalBalanceAmount` which measures the DYDX in the withdrawal -- ie, the output of the swap.  The USDC rebate was already sent back before the swap.  This call will set aside some proportion of the DYDX amount now for the rebateFee even though we don't do anything with it.


Concrete numerical example: CP has all of the stDYDX tokens and rebate of 50% on them. Stride fee is 10%. We get 100 USDC reward, compute at that point rebateFee = 5 USDC, strideFee = 5 USDC, reinvest = 90 USDC all correct so far.  Then we send off 95 USDC to be swapped and it comes back as 95 DYDX assuming a 1:1 price for simplicity.  Now you get to this call back and withdrawalBalanceAmount is 95 representing the 95 DYDX post swap.  And it tries to compute rebateFee = 4.75 DYDX, stride fee = 4.75 DYDX and reinvest = 85.5 DYDX when it should be stride fee = 5 DYDX reinvest = 90 DYDX.   

This assumes that 100 would be passed in so it can do the same computation as it did before, except the rebate was already taken out so instead it gets 95.  

Here is one way you could solve it: add an additional param flag which says if you are doing the rebate call or the stride fee call.  If doing the rebate call, the function would be as is currently.  If doing the stride fee call, you would scale the input value up as though the rebate hadn't already been removed and then do the computation the same way from that point on.  You would do this by computing the rebateFee, figuring out what percent of the total it is as rebateFeePct, then multiplying the current input by 1.0/(1.0 - rebateFeePct)  to get the original input.

So for example we compute the rebateFee is 4.75 DYDX of the incoming value of 95 DYDX, then we find that rebateFeePct = 5%, so we multiply the input of 95 DYDX by 1.0 / (1.0 - 0.05) to get 100 DYDX as the massaged input.  From that point on the computation is as normal so you would get rebateFee = 5 DYDX, stride fee = 5 DYDX and reinvest = 90 DYDX and you just ignore the rebate fee for this later call since it is intended for stride.

ethan-stride: In fact you don't have to compute the actual values you can find just the scale factor needed:

```
contributePercentage = ....

if strideFeeComputation {
    rebatePctOfTotal = strideFeePct * contributePercentage * rebateInfo.RebatePercentage
    scaleFactor = 1.0 / (1.0 - rebatePctOfTotal)
    rewardAmount *= scaleFactor
}

[Rest of computation as normal]
```

ethan-stride (changes requested):
There are a few errors for certain, and a few which I have questions on.  
- For sure the second call to CheckForCommunityPoolRebate is not doing what you think it is doing because the amount coming in has been reduced
- For sure the call to FundCommunityPool when performing the rebate from the Withdrawal account directly needs to include the withdrawal ICA address as the Depositor

- I think but I am not sure that to find the contributePercentage we want to measure both numerator and denominator in stDYDX instead of DYDX but I'm not certain so let's discuss.

x/stakeibc/keeper/reward_converter.go
sampocs: All great suggestions!

> RebateAmount + StrideFeeAmount + ReinvestAmount = rewardAmount

Just added

> strideFeeAmount > 0 (should it just be a uint?)

This is enforced by the param validation. It has to be a sdk.Int to allow for the calculation. 

> 0 <= contributePercentage <= 1

Just added a check that liquid staked amount <= TVL (essentially the same thing)

> rebateAmount >= 0

Already enforced in validate basic

---

Also, fwiw, my philosophy on sanity checks like this in general is that:
* All else equal, less code is better than more code (invariants aren't free, any new code has the potential for a bug)
* Sanity checking simple math doesn't add value (e.g. saying x = y - 10, and then validating that x + y == 10 is pointless)
* Futile sanity checks bloat the code and make it less readable 

That said, I think your suggested checks add enough value to break out of the categories above and I'm in agreement with adding them! Just wanted to explain my personal framework!

sampocs: duplicate of above, responded there

x/stakeibc/keeper/icqcallbacks_withdrawal_balance.go
sampocs: commission is validated by the params. Imo the other check falls into my category above of unnecessarily validating simple math (i.e. reinvest = balance - fee, then validating fee + balance = reinvest). Wdyt?

x/stakeibc/keeper/icqcallbacks_withdrawal_balance.go
sampocs: agreed on all the points above. I see how reward denom is a bit confusing but, as Ethan mentioned, it is common terminology across the whole trade route feature so I'd prefer to not do a large refactor across everything right now

My vote would either be to rename `WithdrawalBalanceCallback` to `WithdrawalHostDenomBalanceCallback` as Ethan suggested, or to just leave it as is as the short name `WithdrawalBalanceCallback` since it's a fundamental part of the core liquid staking flow

---
> confusing because the SDK uses "reward" as the standard name for all staking rewards, which are typically in the native denom

Fwiw, it seems like the confusion is more a byproduct of dydx collecting rewards in a USDC, and less a byproduct of our naming. I would still consider the USDC "staking rewards"

x/stakeibc/keeper/icqcallbacks_withdrawal_reward_balance_test.go
sampocs: added more test cases to `CheckForCommunityPoolRebate`

x/stakeibc/keeper/reward_converter_test.go
sampocs: added

x/stakeibc/keeper/icqcallbacks_withdrawal_balance_test.go
sampocs: I see Ethan already addressed it, but difference is that `WithdrawalBalanceCallback` queries for native tokens, and `WithdrawalRewardBalanceCallback` queries for USDC (or any other revenue token that's been registered with the trade route)

The reason their handled differently is because in `WithdrawalRewardBalanceCallback`, we need the amount to trade (reinvestAmount + strideFee) and the rebateAmount.

Whereas in `WithdrawalBalanceCallback`, we're post-trade and post-rebate, so we only need strideFee and reinvestAmount.

x/stakeibc/keeper/community_pool.go
sampocs: yeah for the rebate we fund from the withdrawal account. For the CP liquid stake, we fund from the community pool return account

sampocs:
@riley-stride 

> I think there's a slight bug with CheckForCommunityPoolRebate but not 100% sure - pls check me!

You're 100% right! Great catch! I'll take a pass at resolving

sampocs:
@riley-stride @ethan-stride - I'm not sure the proposed suggestions will work for the fee calculation in `WithdrawalBalanceCallback`

I think it only works in the case where all the DYDX tokens in the withdrawal account came from USDC. But I believe DYDX also gives out a small amount of native DYDX rewards as well - in which case, there would be no way to discern the two.

I'm thinking it might just make the most sense to make this stateful and write out the rebate amount at the time that it's issued. What do you guys think?

x/stakeibc/keeper/reward_converter.go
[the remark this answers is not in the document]
riley-stride: Good catch!

I think I agree with measuring stDYDX instead of DYDX based on your logic:

> Over time LiquidStakeAmount will remain fixed while TotalDelegations grows from interest reinvestment happening so won't this ratio naturally drop over time even if no actions by outside holders are happening

Seems easy enough to make the change too!

x/stakeibc/keeper/icqcallbacks_withdrawal_balance_test.go
riley-stride: Ty!

Leaving open to until we address [this](https://github.com/Stride-Labs/stride/pull/1136#issuecomment-2002047681)

x/stakeibc/keeper/icqcallbacks_withdrawal_balance.go
riley-stride: Of the two options you listed, if it's not too invasive would prefer to rename to `WithdrawalHostDenomBalanceCallback` - these functions were pretty confusing to me so I can only imagine how lost someone who doesn't understand the codebase could be

x/stakeibc/keeper/reward_converter.go
riley-stride: This all looks good except I am not sure this check is going to work. Opened up a separate comment about it, so we can resolve this longer thread. 

> 0 <= contributePercentage <= 1
> Just added a check that liquid staked amount <= TVL (essentially the same thing)

x/stakeibc/keeper/reward_converter.go
riley-stride: Leaving open for other reviewers to read through since it's very sensitive logic

sampocs:
@riley-stride @ethan-stride Just took a pass at the [correction](https://github.com/Stride-Labs/stride/pull/1136/commits/a66c0cb1a52031ecbea66fab44f95a7f32f554b4). I think you guys were both proposing the same thing in slightly different ways, but let me if this aligns with what you were thinking or if you had a different approach in mind

Going to hold off on making/correcting unit tests for these until I get your feedback to avoid any wasted work

x/stakeibc/keeper/reward_converter.go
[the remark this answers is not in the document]
sampocs: agreed, this change makes sense to me. Might wanna save this for a separate PR though cause it will be a pretty invasive change

x/stakeibc/keeper/community_pool.go
sampocs: done (handled it by switching on the account type)

x/stakeibc/keeper/icqcallbacks_withdrawal_reward_balance.go
sampocs: handled it by switching on the account type

x/stakeibc/keeper/icqcallbacks_withdrawal_reward_balance.go
sampocs: are you suggesting to use `RewardDenomOnRewardZone` so that the logs don't have an ibc hash? If so I agree with you, but just double checking cause `RewardDenomOnHostZone` is the actual denom of the transfer

x/stakeibc/keeper/icqcallbacks_withdrawal_reward_balance.go
sampocs: done

x/stakeibc/keeper/icqcallbacks_withdrawal_balance.go
sampocs: took a stab at it in the latest commit

x/stakeibc/keeper/reward_converter.go
```diff
@@ -103,12 +103,23 @@ func (k Keeper) CheckForCommunityPoolRebate(
 			"unable to calculate rebate amount for %s since total delegations are 0", chainId)
 	}
 
+	// It also shouldn't be possible for the liquid stake amount to be greater than the full TVL
+	if rebateInfo.LiquidStakeAmount.GT(hostZone.TotalDelegations) {
```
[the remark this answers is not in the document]
sampocs: yeah good catch - although I think we'd want to avoid registering the rebate until it's been delegated (otherwise the rewards don't technically belong to the CP yet)

Also, if we switch to using stDYDX for the liquid staked amount, this should no longer be able to error in normal conditions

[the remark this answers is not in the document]
riley-stride: > we'd want to avoid registering the rebate until it's been delegated (otherwise the rewards don't technically belong to the CP yet)

ah yeah that makes sense 

> Also, if we switch to using stDYDX for the liquid staked amount, this should no longer be able to error in normal conditions

because we'd compare liquidStakedAmount to stTokenSupply instead of to TotalDelegations?

x/stakeibc/keeper/reward_converter.go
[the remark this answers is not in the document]
riley-stride: agreed, separate PR

x/stakeibc/keeper/reward_converter.go
```diff
@@ -57,75 +57,95 @@ type FeeInfo struct {
...
-		StrideFeeAmount: strideFeeAmount,
-		ReinvestAmount:  reinvestAmount,
+// Given the reward balance from an ICQ, calculates the portions intended as a stride fee vs for reinvestment
+//
+// If the chain doesn't have a rebate in place, the split is decided entirely from the stride commission percent
+// However, if the chain does have a rebate, we need to factor that into the calculation
+//
+// For instance, if 100 rewards were collected and 5 were sent as a rebate, then the stride fee should be based
+// on the original 100 rewards instead of the remaining 95
+func (k Keeper) CalculateRewardsFeeSplit(
```
riley-stride, line 119: If I understand correctly, `CalculateRewardsFeeSplit` is used when calculating the split on the DYDX rewards in WithdrawalBalanceCallback, and `CalculateRewardsRebateSplit` is used to calculate the split on USDC rewards. Finding the names a bit confusing - had to trace the code for calls to this and `CalculateRewardsRebateSplit` to understand which was which. Possible I'm missing some detail that makes them clear - one idea would be to name them according to the functions they're called in e.g. `CalculateRewardsSplitHostDenom` or similar.

x/stakeibc/keeper/reward_converter.go
```diff
@@ -57,75 +57,95 @@ type FeeInfo struct {
 // and the normal staking and distribution flow will continue from there.
 //////////////////////////////////////////////////////////////////////////////////////////////////////////////////
 
-// Breaks down the accumulated rewards into the portions intended for reinvestment, stride commission, or a rebate
-// Returns the rebate amount, the stride fee amount, and the reinvestment amount
+// Breaks down the accumulated rewards into the portions intended for a rebate vs the remainder
```
riley-stride, line 60: suggest: Break down the split of non-host-denom rewards into the portions intended for a rebate vs the remainder that's used for fees and reinvestment

e.g. if DYDX staking has earned 100 USDC in rewards, and DYDX community pool has a 2.5% rebate `CalculateRewardsRebateSplit` will return:
* rebateAmount = 2.5
* reinvestAmount = 97.5

x/stakeibc/keeper/reward_converter.go
```diff
@@ -57,75 +57,95 @@ type FeeInfo struct {
...
-		return feeInfo, errorsmod.Wrapf(types.ErrFeeSplitInvariantFailed, "fee info compontents do not sum to rewards")
-	}
+	return rebateAmount, remainingAmount, nil
+}
 
-	feeInfo = FeeInfo{
-		RebateAmount:    rebateAmount,
-		StrideFeeAmount: strideFeeAmount,
-		ReinvestAmount:  reinvestAmount,
+// Given the reward balance from an ICQ, calculates the portions intended as a stride fee vs for reinvestment
```
riley-stride, line 112: suggestion for clarity:


Given the balance of **host-denom** rewards from an ICQ, calculate the portions intended as a stride fee vs for reinvestment

*note: `CalculateRewardsFeeSplit` is run on **host-denom** rewards. In the case of DYDX this is run **_after_** rewards are pushed through the trade route.*

x/stakeibc/keeper/reward_converter.go
```diff
@@ -57,75 +57,95 @@ type FeeInfo struct {
...
+
+	// If there's no rebate, the fee split just uses the commission
+	if !chainHasRebate {
+		strideFeeAmount = totalFeePercentage.Mul(sdk.NewDecFromInt(rewardsAmount)).TruncateInt()
+	} else {
+		// Otherwise, the rebate must be considered in the fee split
+		// The rebate portion the portion of TVL contributed to by the liquid stake * the rebate percentage
+		// The stride fee poriton is the remaining percentage
+		contributePercentage := sdk.NewDecFromInt(rebateInfo.LiquidStakeAmount).Quo(sdk.NewDecFromInt(hostZone.TotalDelegations))
+		rebatePercentage := contributePercentage.Mul(rebateInfo.RebatePercentage)
```
riley-stride, line 140: nit: confusing that this var and rebateInfo.RebatePercentage have the same name but represent different concepts

suggest renaming this

maybe `RebatePctOfAllZoneRewards` or similar

x/stakeibc/keeper/reward_converter.go
```diff
@@ -57,75 +57,95 @@ type FeeInfo struct {
...
+	strideFee := sdk.NewIntFromUint64(k.GetParams(ctx).StrideCommission)
+	totalFeePercentage := sdk.NewDecFromInt(strideFee).Quo(sdk.NewDec(100))
+	totalFeesAmount := sdk.NewDecFromInt(rewardsAmount).Mul(totalFeePercentage)
+
+	// Check if the chain has a rebate
+	rebateInfo, chainHasRebate := hostZone.SafelyGetCommunityPoolRebate()
+
+	// If there's no rebate, the fee split just uses the commission
+	if !chainHasRebate {
+		strideFeeAmount = totalFeePercentage.Mul(sdk.NewDecFromInt(rewardsAmount)).TruncateInt()
```
riley-stride, line 134: Isn't this already calculated above as `totalFeesAmount`?

x/stakeibc/keeper/reward_converter.go
```diff
@@ -57,75 +57,95 @@ type FeeInfo struct {
...
+	} else {
+		// Otherwise, the rebate must be considered in the fee split
+		// The rebate portion the portion of TVL contributed to by the liquid stake * the rebate percentage
+		// The stride fee poriton is the remaining percentage
+		contributePercentage := sdk.NewDecFromInt(rebateInfo.LiquidStakeAmount).Quo(sdk.NewDecFromInt(hostZone.TotalDelegations))
+		rebatePercentage := contributePercentage.Mul(rebateInfo.RebatePercentage)
+		strideFeePercentage := sdk.OneDec().Sub(rebatePercentage)
+		strideFeeAmount = totalFeesAmount.Mul(strideFeePercentage).TruncateInt()
```
riley-stride, line 142: I think what you're currently doing here is calc'ing the % owed as rebate, then subtracting that from the % of total stride fees, to get the portion of fees that stride gets (not rebated). 

Which makes sense. 

_But_ it seems that to get _token amounts_, you use the input to this fn `rewardsAmount`. 

The problem is that we've already removed the rebate amount from `rewardsAmount` upstream in `CalculateRewardsRebateSplit`.

To fix, we just need to recover the amount that was input to `CalculateRewardsRebateSplit`, and apply all the calcs you have here to _that_ value.

I think you can get it with `rewardsAmount.Quo(1-rebatePercentage)` per [here](https://github.com/Stride-Labs/stride/pull/1136#issuecomment-2002047681).

x/stakeibc/keeper/icqcallbacks_withdrawal_balance.go
sampocs: can I address this in a separate PR since it will require a file renaming? I can add it to the canvas

x/stakeibc/keeper/reward_converter.go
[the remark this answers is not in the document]
sampocs: yeah exactly

x/stakeibc/keeper/reward_converter.go
sampocs: done

x/stakeibc/keeper/reward_converter.go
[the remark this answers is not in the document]
riley-stride: Makes sense. Resolving.

x/stakeibc/keeper/icqcallbacks_withdrawal_balance.go
riley-stride: Sg! Resolving

x/stakeibc/keeper/reward_converter.go
sampocs: done

x/stakeibc/keeper/reward_converter.go
sampocs: done

x/stakeibc/keeper/reward_converter.go
sampocs: haha yes, good eye! done

x/stakeibc/keeper/reward_converter.go
sampocs: :face_palm: 

sorry it's getting pretty sloppy with all this back and forth. will resolve

x/stakeibc/keeper/reward_converter.go
```diff
@@ -51,6 +57,119 @@ type ForwardMetadata struct {
...
+// that's used for fees and reinvestment
+// For instance, in the case of dYdX, this is used on the USDC that has not been pushed through the trade route
+// yet, and has not yet been converted to DYDX
+//
+// The rebate percentage is determined by: (% of total TVL contributed by commuity pool) * (rebate percentage)
+//
+// E.g. Community pool liquid staked 1M, TVL is 10M, rebate is 50%
+// Total rewards this epoch are 1000, and the stride fee is 10%
+// => Then the rebate is 1000 rewards * 10% stride fee * (1M / 10M) * 50% rebate = 5
+func (k Keeper) CalculateRewardsSplitRewardDenom(
```
riley-stride, line 70: Sorry to be pedantic about this: `RewardDenom` is confusing to me, afaiu we'd only call this function on chains who have rewards in non-host denom (non-native denoms). Why not just call it that? RewardDenom is confusing because the denom of rewards is usually the native denom, but this function should never be run on native denoms.

x/stakeibc/keeper/reward_converter.go
riley-stride: `CalculateRewardsSplitHostDenom` lgtm, though `CalculateRewardsSplitRewardDenom` still confusing to me. Opened a new comment about that so will resolve this.

x/stakeibc/keeper/reward_converter.go
riley-stride: lgtm

x/stakeibc/keeper/reward_converter.go
riley-stride: lgtm

x/stakeibc/keeper/reward_converter.go
riley-stride: lgtm

x/stakeibc/keeper/reward_converter.go
riley-stride: I have the same comment for the naming of `CalculateRewardsSplitRewardDenom`

x/stakeibc/keeper/icqcallbacks_withdrawal_balance.go
```diff
@@ -62,27 +58,10 @@ func WithdrawalBalanceCallback(k Keeper, ctx sdk.Context, args []byte, query icq
 		return errorsmod.Wrapf(types.ErrICAAccountNotFound, "no fee account found for %s", chainId)
 	}
 
-	// Determine the stride commission rate to the relevant portion can be sent to the fee account
-	params := k.GetParams(ctx)
-	strideCommissionInt, err := cast.ToInt64E(params.StrideCommission)
+	// Split the withdrawal amount into a rebate, stride fee, and reinvest portion
```
asalzmann, line 61: This is just stride fee + reinvest right?

asalzmann (commented):
Overall lgtm!
* is the change to use stDYDX to calculate the comm pool share going to be in a different PR?
* it seems like we have some logic to support a native token fee split, but it's not actually used. should we remove it?

x/stakeibc/keeper/reward_converter.go
sampocs: as mentioned before the RewardDenom/HostDenom split is used across the codebase for the trade route. 

I'm open to renaming it, but I'd prefer to avoid a refactor of all the trade route stuff and I think having inconsistent terminology across the codebase is the worst thing we can do. 

What about `CalculateRewardsSplitBeforeRebate` and `CalculateRewardSplitAfterRebate`?

x/stakeibc/keeper/reward_converter.go
riley-stride: Resolving as I'm doing a new review.

x/stakeibc/keeper/icqcallbacks_community_pool_ica_balance.go
```diff
@@ -75,7 +75,7 @@ func CommunityPoolIcaBalanceCallback(k Keeper, ctx sdk.Context, args []byte, que
 	} else if icaType == types.ICAAccountType_COMMUNITY_POOL_RETURN {
 		// Send ICA msg to FundCommunityPool with token found in return ICA
 		err := utils.ApplyFuncIfNoError(ctx, func(c sdk.Context) error {
-			return k.FundCommunityPool(ctx, hostZone, token)
+			return k.FundCommunityPool(ctx, hostZone, token, types.ICAAccountType_COMMUNITY_POOL_RETURN)
```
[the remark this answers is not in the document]
sampocs: done

x/stakeibc/keeper/reward_converter.go
```diff
@@ -51,6 +57,119 @@ type ForwardMetadata struct {
...
+// This is used on the host zone's native denom, either from traditional staking rewards in most case, or
+// from rewards that have been traded in order to convert them into the native denom
+// For instance, with dYdX, this is applied to the DYDX tokens that were converted from USDC
+//
+// If the chain doesn't have a rebate in place, the split is decided entirely from the stride commission percent
+// However, if the chain does have a rebate, we need to factor that into the calculation
+//
+// For instance, if 100 rewards were collected and 5 were sent as a rebate, then the stride fee should be based
+// on the original 100 rewards instead of the remaining 95
+func (k Keeper) CalculateRewardsSplitHostDenom(
```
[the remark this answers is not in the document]
riley-stride: afaiu this isn't for rewards received as DYDX, it's for DYDX obtained after swapping on Osmosis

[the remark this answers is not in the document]
sampocs: this is for the rewards _after_ they've been converted to DYDX. Sorry I should have updated the description after all the recent changes.

The flow is:
* Query USDC balance
* Call `CalculateRewardsSplitRewardDenom` → get rebate amount and remaining amount
* Send rebate to CP, convert remaining amount to DYDX
* Query DYDX balance
* Call `CalculateRewardsSplitHostDenom` → get stride fee and reinvest amount
* Send feeAmount to fee account, reinvestAmount to delegation account

x/stakeibc/keeper/reward_converter.go
riley-stride: `CalculateRewardsSplitBeforeRebate` and `CalculateRewardSplitAfterRebate` is much more clear, in favor!

Agreed that a large renaming now is tricky, sorry had forgotten about that detail.

x/stakeibc/keeper/reward_converter.go
```diff
@@ -51,6 +57,119 @@ type ForwardMetadata struct {
...
 //////////////////////////////////////////////////////////////////////////////////////////////////////////////////
 
+// Breaks down the split of non-native-denom rewards into the portions intended for a rebate vs the remainder
+// that's used for fees and reinvestment
+// For instance, in the case of dYdX, this is used on the USDC that has not been pushed through the trade route
+// yet, and has not yet been converted to DYDX
+//
+// The rebate percentage is determined by: (% of total TVL contributed by commuity pool) * (rebate percentage)
+//
+// E.g. Community pool liquid staked 1M, TVL is 10M, rebate is 50%
```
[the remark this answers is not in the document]
sampocs: done

x/stakeibc/keeper/reward_converter.go
```diff
@@ -51,6 +57,119 @@ type ForwardMetadata struct {
...
+
+	// It shouldn't be possible to have 0 delegations (since there are rewards and there was a community pool stake)
+	// This will also prevent a division by 0 error
+	if hostZone.TotalDelegations.IsZero() {
+		return sdkmath.ZeroInt(), sdkmath.ZeroInt(), errorsmod.Wrapf(types.ErrDivisionByZero,
+			"unable to calculate rebate amount for %s since total delegations are 0", chainId)
+	}
+
+	// It also shouldn't be possible for the liquid stake amount to be greater than the full TVL
+	if rebateInfo.LiquidStakeAmount.GT(hostZone.TotalDelegations) {
```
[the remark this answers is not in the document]
sampocs: yup!

x/stakeibc/keeper/reward_converter.go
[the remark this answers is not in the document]
asalzmann: I see - the shared naming convention confused me at first glance. Is the idea of using one naming convention (`CalculateRewardsSplitXXXDenom`) that both denoms _could_ have a split of stride fee + reinvest/trade + rebate?

Don't have a strong view on what we call these, if we want more clarity on what they do today could use `CalculateRebateAmount` and `CalculateFeeAmount`

x/stakeibc/keeper/reward_converter.go
sampocs: pushed!

x/stakeibc/keeper/reward_converter.go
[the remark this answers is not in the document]
sampocs: just pushed an update to rename them to `CalculateRewardsSplitBeforeRebate` and `CalculateRewardsSplitAfterRebate`

Also open to `CalculateRebateAmount` / `CalculateFeeAmount` as well, although I think I had had something similar at the start and it received similar confusion haha

x/stakeibc/keeper/icqcallbacks_withdrawal_reward_balance.go
ethan-stride: Yes I was suggesting it for readability but also the variable from the callback tradeRouteCallback.RewardDenom what goes into that it the `RewardDenomOnRewardZone` as well

ethan-stride (approved):
LGTM

@sampocs good job with these changes.  I now agree with Riley that the math in this one looks conceptually correct to me even if some extra small amount of the rewards come in as DYDX should be close enough.

I think the remaining large logic thing to do is change to tracking the contribute fraction with stDYDX instead of DYDX although I know the changes to do that are pretty mechanical.  Approving now

riley-stride (approved):
if the grouping is reverted I'm signed off

ofc still needs ample integration testing etc

x/stakeibc/keeper/reward_converter.go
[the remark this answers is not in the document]
asalzmann: Makes sense, fine to keep imo

asalzmann (approved):
signed off

x/stakeibc/keeper/reward_converter.go
```diff
@@ -53,6 +59,124 @@ type ForwardMetadata struct {
 // and the normal staking and distribution flow will continue from there.
 //////////////////////////////////////////////////////////////////////////////////////////////////////////////////
 
+// Breaks down the split of non-native-denom rewards into the portions intended for a rebate vs the remainder
```
shellvish, line 62: These comments are very helpful!

x/stakeibc/keeper/reward_converter.go
```diff
@@ -53,6 +59,124 @@ type ForwardMetadata struct {
...
+		// It also shouldn't be possible for the liquid stake amount to be greater than the full TVL
+		if rebateInfo.LiquidStakeAmount.GT(hostZone.TotalDelegations) {
+			return sdkmath.ZeroInt(), sdkmath.ZeroInt(), errorsmod.Wrapf(types.ErrFeeSplitInvariantFailed,
+				"community pool liquid staked amount greater than total delegations")
+		}
+
+		// Otherwise, the rebate must be considered in the fee split
+		// The rebate portion is the portion of TVL contributed to by the liquid stake * the rebate percentage
+		// The stride fee poriton is the remaining percentage
+		contributionRate := sdk.NewDecFromInt(rebateInfo.LiquidStakeAmount).Quo(sdk.NewDecFromInt(hostZone.TotalDelegations))
```
shellvish, line 164: Small thing to note: the `TotalDelegations` here is being fetched in a different epoch than when `CalculateRewardsSplitBeforeRebate` is called. Generally, the two values should be very close together, and I don't think there are any bad edge cases, but just verifying that lines up with your intuition too

x/stakeibc/keeper/reward_converter.go
```diff
@@ -53,6 +59,124 @@ type ForwardMetadata struct {
...
+		effectiveRebateRate := totalFeeRate.Mul(contributionRate).Mul(rebateInfo.RebatePercentage)
+		effectiveStrideFeeRate := totalFeeRate.Sub(effectiveRebateRate)
+
+		// Before calculating the fee, we have to scale up the rewards amount to the amount before the rebate
+		originalRewardScalingFactor := sdk.OneDec().Sub(effectiveRebateRate)
+		originalRewardsAmount := sdk.NewDecFromInt(rewardsAmount).Quo(originalRewardScalingFactor).TruncateDec()
+		strideFeeAmount = originalRewardsAmount.Mul(effectiveStrideFeeRate).Ceil().TruncateInt() // ceiling since rebate was truncated
```
shellvish, line 171: Verified this math makes sense to me. Thanks for writing this out so clearly!

shellvish (approved):
This looks great to me! Only one comment around a potential edge case. I think it shouldn't cause any issues, but want to verify that lines up with your view too

Thanks a lot for doing this Sam!

x/stakeibc/keeper/reward_converter.go
sampocs: yeah my thinking was that it's not perfect, but the values are close enough together and it prevents us from having to make this stateful. 

Also, fwiw, this was changed in the other PR to be the stTokens and total supply (similar story there though)

PR #1137, opened 2024-03-15
disable swap

sampocs:
Small one here but definitely double check me that this is all we have to do to disable the swaps.

### DIFF
--- a/x/stakeibc/keeper/hooks.go
+++ b/x/stakeibc/keeper/hooks.go
@@ -92,9 +92,7 @@ func (k Keeper) BeforeEpochStart(ctx sdk.Context, epochInfo epochstypes.EpochInf
 	// At the hour epoch, query the swap price on each trade route and initiate the token swap
 	if epochInfo.Identifier == epochstypes.HOUR_EPOCH {
 		k.UpdateAllSwapPrices(ctx)
-		k.SwapAllRewardTokens(ctx)
 	}
-
 }
 
 func (k Keeper) AfterEpochEnd(ctx sdk.Context, epochInfo epochstypes.EpochInfo) {}


ethan-stride:
This will indeed stop the chain from doing swaps and I do think it is all that is needed.  The 3 phase design we ended on means the swaps were their own phase and their callbacks don't exist.  The third phase is still the return which exists in the TransferAllRewardTokens keeper function.

But you could change the swap route types and all the associated tx routes to update trade routes if we are sure we won't want to go back since now they no longer need info for how to do the swaps like which pool and slippage params etc.  If we think we might want to enable them again, then you can leave them be.

ethan-stride (approved):
LGTM

ethan-stride:
Side note: because the return path does an ICQ only looking for DYDX tokens in the TradeAccount ICA it would ignore other tokens in there... so the operator could actually do multiple swaps if they thought it would give a better price.  Eg. Swap USDC for OSMO then swap the OSMO for DYDX

x/stakeibc/keeper/hooks.go
```diff
@@ -92,9 +92,7 @@ func (k Keeper) BeforeEpochStart(ctx sdk.Context, epochInfo epochstypes.EpochInf
 	// At the hour epoch, query the swap price on each trade route and initiate the token swap
 	if epochInfo.Identifier == epochstypes.HOUR_EPOCH {
 		k.UpdateAllSwapPrices(ctx)
```
riley-stride, line 94: should we be removing this too? the authz grantee could just query the price offchain right before swapping rather than using on-chain metric that is more stale

riley-stride (changes requested):
should we remove the price query too? the authz grantee could just query the price offchain right before swapping rather than using on-chain metric that is more stale

x/stakeibc/keeper/hooks.go
sampocs: I don't feel strongly either way.

My thinking was I wanted to make the changes as minimal as possible, including not modifying the trade route structs (so we don't need to do a migration). And if we're going to leave the price on the struct, it felt like we may as well keep it updated. 

That said, I also don't love leaving in dead code so can see an argument for removing it

riley-stride: You know what - maybe we should just leave it. It's working now and will act as a backup price - you never know why we may need it. Given the current upgrade timeline I agree with you philosophy of `make the changes as minimal as possible`

riley-stride: To sum up: don't feel that strongly either way, slightly leaning towards keeping this as is. 

Leaving this open for comments from @shellvish @asalzmann @ethan-stride.

riley-stride: Ethan's comment from Slack: _We don’t need it any longer and it does fire once an hour which is a lot. I would only keep it in if we thought that after the community pool comes through we might go back to the auto trade on chain approach._

sampocs: yeah the more I think about it, I think I actually lean towards pulling it out. I think a lot of the trade route struct will no longer be relevant (and we can migrate it later to reflect that)

It's pretty simple to remove and leaves one less place for something to break

riley-stride: If it's easy to pull out with no migrations then let's do it! I got the impression from above that it might require a migration?

sampocs: oh sorry to clarify the options are:
* (1) Leave query and struct as is
* (2) Turn off the query but _keep_ the price and last updated time on the struct
* (3) Turn off the query and _remove_ the price and last updated time from the struct (requires migration)

My argument originally was that #2 felt a bit odd to turn off the updates but leave a stale value on the struct. But on second thought, I think I'd prefer #2 and then we can do the migration in a later upgrade to clean it up

riley-stride: Thanks for writing these out - number 2 feels right!

asalzmann: #2 makes sense - why does #3 require a migration? I'd have assumed we'd just mark the proto field as deprecated, which doesn't require a migration afaik.

sampocs: that's true, probably wouldn't require a migration

asalzmann: If no migration is required, 3 seems preferable to 2, although I realize we're getting away from the principle of making minimal changes

sampocs: I think 3 would still require us to update all references to those fields (otherwise we'll get link errors)

I agree that's the best long term solution but I think I'd prefer to push the cleanup to a separate upgrade to make sure we can meet our timeline

sampocs: @riley-stride just turned the price query in the latest commit

riley-stride: Do we make any assumptions about the price being non-nil? If you're confident we don't, this lgtm and I'm signed off. Approving now.

sampocs: It wont be nil - it'll just be stale.

Also, a nil price would just lead to us disabling the on-chain swap (which we already are doing!)

PR #1139, opened 2024-03-15
fixing init_chain script to not clobber new params

ethan-stride:
There are new params when starting the chain which were being overwritten
experimental_max_gossip_connections_to_persistent_peers
experimental_max_gossip_connections_to_non_persistent_peers
Because in the init_chain.sh script we had a regex looking for the variable persistent_peers and it was writing values to these
With the wrong type being written to these params, nodes after the first one would fail to start
This was causing dockernet to only start one validator when expecting 3 and then no blocks could reach consensus

Added a word boundary to that intended variable so these new variables can be left alone.

Dockernet can now start
All unit tests in `make test-unit` are passing again
All integration tests in `make test-integration-docker` are passing again

### DIFF
--- a/dockernet/src/init_chain.sh
+++ b/dockernet/src/init_chain.sh
@@ -362,7 +362,7 @@ for (( i=2; i <= $NUM_NODES; i++ )); do
     genesis_json="${STATE}/${node_name}/config/genesis.json"
 
     # add the main node as a persistent peer
-    sed -i -E "s|persistent_peers = .*|persistent_peers = \"${MAIN_NODE_ID}\"|g" $config_toml
+    sed -i -E "s|\bpersistent_peers = .*|persistent_peers = \"${MAIN_NODE_ID}\"|g" $config_toml
     # copy the main node's genesis to the peer nodes to ensure they all have the same genesis
     cp $MAIN_GENESIS $genesis_json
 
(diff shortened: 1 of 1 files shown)


dockernet/src/init_chain.sh
```diff
@@ -362,7 +362,7 @@ for (( i=2; i <= $NUM_NODES; i++ )); do
     genesis_json="${STATE}/${node_name}/config/genesis.json"
 
     # add the main node as a persistent peer
-    sed -i -E "s|persistent_peers = .*|persistent_peers = \"${MAIN_NODE_ID}\"|g" $config_toml
+    sed -i -E "s|\bpersistent_peers = .*|persistent_peers = \"${MAIN_NODE_ID}\"|g" $config_toml
```
sampocs, line 365: what does the `\b` do / is it the same as doing `^persistent_peers`?

sampocs (approved):
lgtm, thanks for fixing!

dockernet/src/init_chain.sh
ethan-stride: Not exactly the same but in this case either would work. \b is word boundary so it can't be part of another word.  Applies to new lines, spaces, etc.  The carat only applies to new lines.  In this case the one we want to update is on a new line.  I'll update it to carat for simplicity

PR #1142, opened 2024-03-16
added toggle trade controller tx

sampocs:
## Context
Added tx to grant or revoke trade permissions to an account on behalf of the trade ICA. The transaction does this via an ICA with an authz grant/revoke message.

Reviewer - please double check the authz message contents look correct.


x/stakeibc/types/message_toggle_trade_controller.go
```diff
@@ -0,0 +1,64 @@
...
+	}
+	return []sdk.AccAddress{creator}
+}
+
+func (msg *MsgToggleTradeController) GetSignBytes() []byte {
+	bz := ModuleCdc.MustMarshalJSON(msg)
+	return sdk.MustSortJSON(bz)
+}
+
+func (msg *MsgToggleTradeController) ValidateBasic() error {
```
riley-stride, line 48: should we add another check on the enum here or are the checks in `tx_toggle_trade_controller` enough?

x/stakeibc/keeper/msg_server.go
```diff
@@ -1103,3 +1106,66 @@ func (k msgServer) ResumeHostZone(goCtx context.Context, msg *types.MsgResumeHos
...
+) (*types.MsgToggleTradeControllerResponse, error) {
+	ctx := sdk.UnwrapSDKContext(goCtx)
+
+	// Fetch the trade ICA which will be the granter
+	tradeRoute, found := k.GetTradeRouteFromTradeChainId(ctx, msg.ChainId)
+	if !found {
+		return nil, types.ErrTradeRouteNotFound.Wrapf("trade route not found for chain ID %s", msg.ChainId)
+	}
+
+	swapMsgTypeUrl := "/" + proto.MessageName(&types.SwapAmountInRoute{})
```
riley-stride, line 1124: could you explain how this works?

x/stakeibc/keeper/trade_route_test.go
```diff
@@ -96,3 +96,32 @@ func (s *KeeperTestSuite) TestGetAllTradeRoutes() {
...
+			RewardDenomOnRewardZone: rewardDenom,
+			HostDenomOnHostZone:     hostDenom,
+			TradeAccount: types.ICAAccount{
+				ChainId: chainId,
+			},
+		})
+	}
+
+	// Search for each of them by chain ID
+	for i := 1; i <= 3; i++ {
```
riley-stride, line 117: also search for chainids that do _not_ have trade routes!

x/stakeibc/types/message_toggle_trade_controller_test.go
```diff
@@ -0,0 +1,94 @@
...
+				Address:          "",
+			},
+			err: "address must be specified",
+		},
+		{
+			name: "invalid chain ID",
+			msg: types.MsgToggleTradeController{
+				Creator:          validAdminAddress,
+				ChainId:          "",
+				PermissionChange: validPermissionChange,
```
riley-stride, line 72: is it possible to test invalidPermissionChange too, or not because that's handled at the tx level?

riley-stride (commented):
Looking like a great start! Few small comments, will see if I can fill in the authz grant.

x/stakeibc/keeper/msg_server.go
```diff
@@ -1103,3 +1106,66 @@ func (k msgServer) ResumeHostZone(goCtx context.Context, msg *types.MsgResumeHos
...
+	// Build the authz message, depending on whether permissions are granted or revoked
+	// The granter will always be the trade ICA and the grantee will be the address
+	// specified in the message
+	var authzMsg []proto.Message
+	switch msg.PermissionChange {
+	case types.AuthzPermissionChange_GRANT:
+		authzMsg = []proto.Message{&authz.MsgGrant{
+			Granter: tradeRoute.TradeAccount.Address,
+			Grantee: msg.Address,
+			Grant:   authz.Grant{
```
ethan-stride, line 1135: Was this supposed to be the msgTypeUrl which you made?

x/stakeibc/keeper/msg_server.go
ethan-stride, line 1124: Shouldn't this be MsgSwapExactAmountIn?  The SwapAmountInRoute is a type which goes into one of the fields of this message but the actual tx they would be running is the MsgSwapExactAmountIn

x/stakeibc/types/message_toggle_trade_controller.go
```diff
@@ -0,0 +1,64 @@
...
+	if err != nil {
+		return errorsmod.Wrapf(sdkerrors.ErrInvalidAddress, "invalid creator address (%s)", err)
+	}
+	if err := utils.ValidateAdminAddress(msg.Creator); err != nil {
+		return err
+	}
+	if msg.ChainId == "" {
+		return errors.New("chain ID must be specified")
+	}
+	if msg.Address == "" {
```
ethan-stride, line 59: Should we also do a check with utils.ValidateAdminAddress on the Address passed in to be gaining permissions?  This would only allow us to use a limited but trusted set of addresses whereas right now we can run this many times and grant many addresses ability to trade even though it has to be an admin who is doing the granting.  Making the set limited to a chain upgrade might give more trust?

x/stakeibc/keeper/msg_server.go
```diff
@@ -1103,3 +1106,66 @@ func (k msgServer) ResumeHostZone(goCtx context.Context, msg *types.MsgResumeHos
...
+		ctx,
+		tradeRoute.TradeAccount.ConnectionId,
+		tradeRouteAccountOwner,
+		authzMsg,
+		timeoutTimestamp,
+	)
+	if err != nil {
+		return nil, err
+	}
+
```
ethan-stride, line 1169: When you run this, the memory of which permissions which addresses currently have is on the other chain where the ICA exists, but you could store an array of which addresses currrently are granted or revoked.  Then instead of the PR which just removes the hook for the Trade to happen on chain, it could check if there are any operators granted permissions and it would only do the epochly on chain trades if there are no current operators with permissions.  Sort of a fall back fail safe and allows us to turn on or off operator mode very easily.

ethan-stride (approved):
LGTM I think.  I am not super certain on the msg url which you built but if that is right then the PR looks right

x/stakeibc/types/message_toggle_trade_controller.go
sampocs: good call, added

x/stakeibc/keeper/msg_server.go
sampocs: It get's the type URL of a message type (e.g. `/osmosis.gamm.v1beta1.MsgSwapExactAmountIn`)

x/stakeibc/keeper/trade_route_test.go
sampocs: good call! meant to add this and then totally forgot

x/stakeibc/types/message_toggle_trade_controller_test.go
sampocs: i think it is - just added a case for it

x/stakeibc/keeper/msg_server.go
sampocs: It takes an `Authorization` object - which has the type url as a component. Just pushed the update

x/stakeibc/keeper/msg_server.go
sampocs: great eye! you're 100% right - just updated

x/stakeibc/types/message_toggle_trade_controller.go
sampocs: hmm it's a good question. I think it'd be nice to have some flexibility here though. Hypothetically if the account was compromised, we wouldn't be able to update it to a new account until we did a chain upgrade

x/stakeibc/keeper/msg_server.go
sampocs: that's an interesting idea! that would be a nice fallback, but I'm not sure having a fallback is valuable enough to justify making this process stateful. Curious to hear other people's thoughts as well

x/stakeibc/keeper/trade_route_test.go
riley-stride: leaving open until resolved

x/stakeibc/types/message_toggle_trade_controller.go
riley-stride: Isn't the address passed in here supposed to be the account to which we're granting authz permissions on Osmosis? That would not pass a utils.ValidateAdminAddress check. Maybe you're proposing limiting Address to be in a separate list

x/stakeibc/keeper/trade_route_test.go
sampocs: it was updated [here](https://github.com/Stride-Labs/stride/pull/1142/commits/31a7205b1038d1a40371d249a34537266ad518f1#diff-769192289f97a856eec27684fae4cf94c5d8a5990362349831ca7b01c4f7e61dR128-R130)

x/stakeibc/keeper/msg_server.go
riley-stride: ok I understand now. do you think we should unit test? assembling the msg feels kinda complicated

x/stakeibc/keeper/reward_converter.go
```diff
@@ -51,6 +53,40 @@ type ForwardMetadata struct {
 // and the normal staking and distribution flow will continue from there.
 //////////////////////////////////////////////////////////////////////////////////////////////////////////////////
 
+// Builds an authz MsgGrant or MsgRevoke to grant an account trade capabilties on behalf of the trade ICA
+func (k Keeper) BuildTradeAuthzMsg(
```
riley-stride, line 57: looks very clean in its own function!

x/stakeibc/keeper/msg_server.go
sampocs: did you see the [unit tests](https://github.com/Stride-Labs/stride/pull/1142/commits/33b8ca1acff2193b84f0d5f57ec312b36a50264c#diff-1629187c35b24ec74868f7e895ab32337777eaf56723e246d834f27e31e7d75eR41) that are already in there?

x/stakeibc/types/message_toggle_trade_controller.go
sampocs: My understanding of the suggestion was that this would be a separate list of osmo addresses

x/stakeibc/keeper/msg_server.go
riley-stride: i missed those - they look good ty!

x/stakeibc/types/message_toggle_trade_controller.go
riley-stride: Makes sense. I lean toward maintaining some flexibility - admin is the main point of trust

riley-stride (approved):
Re-reviewed and resolved comments - looking good!

x/stakeibc/keeper/msg_server.go
```diff
@@ -1103,3 +1104,47 @@ func (k msgServer) ResumeHostZone(goCtx context.Context, msg *types.MsgResumeHos
...
+	// Build the ICA channel owner from the trade route
+	tradeRouteAccountOwner := types.FormatTradeRouteICAOwnerFromRouteId(
+		msg.ChainId,
+		tradeRoute.GetRouteId(),
+		types.ICAAccountType_CONVERTER_TRADE,
+	)
+
+	// Submit the ICA tx from the trade ICA account
+	// Timeout the ICA at 1 hour
```
shellvish, line 1136: extremely nit: should we time this out at 1 hour, or on the epoch? Conceptually, this feels completely disjoint from the rest of the stakeibc logic, but I believe we time out most of the other stakeibc ICA messages at the `stride_epoch`

x/stakeibc/keeper/msg_server.go
shellvish: I agree, I think the additional stateful complexity isn't worth the auto-fall-back on-chain! Especially given the tight timeline, I'd bias to just shipping the simple thing and we can add the additional complexity later

x/stakeibc/keeper/reward_converter.go
```diff
@@ -51,6 +53,40 @@ type ForwardMetadata struct {
...
+	ctx sdk.Context,
+	tradeRoute types.TradeRoute,
+	permissionChange types.AuthzPermissionChange,
+	grantee string,
+) (authzMsg []proto.Message, err error) {
+	swapMsgTypeUrl := "/" + proto.MessageName(&types.MsgSwapExactAmountIn{})
+
+	switch permissionChange {
+	case types.AuthzPermissionChange_GRANT:
+		authorization := authz.NewGenericAuthorization(swapMsgTypeUrl)
```
shellvish, line 67: nit: is it possible to limit the authorization to only swap from token A to token B? imo this would give a small amount of additional protection / make the system easier to reason about, but not critical if it can't be done!

x/stakeibc/types/message_toggle_trade_controller.go
shellvish: Agreed, it's a good suggestion but I also lean slightly towards letting the admin set any address as a trade address. Gives us a bit more flexibility in a nightmare scenario, and I think the "restricted trade list" only grants protection in a scenario where the admin is compromised but the trade routers aren't. This is relatively unlikely imo.

shellvish (approved):
Looks great, thanks a lot for doing this Sam! Very clean, easy to follow code.

x/stakeibc/types/message_toggle_trade_controller.go
ethan-stride: I think that people who have weighed in are all in agreement just to use the admin to turn on or off operators as is without putting any additional constraints on the operator addresses themselves.  Resolving.

x/stakeibc/keeper/msg_server.go
sampocs: I don't think we should time it out at the stride epoch, cause then we have to think about the timing of when we send the message (e.g. if we submit the message at the epoch boundary, we could automatically get a timeout)

imo, we should only use the stride epoch timeout for things that are triggered in our epochly hooks and can't have any overlap across epochs

x/stakeibc/keeper/reward_converter.go
sampocs: it's a good question! I'm not sure. @riley-stride do you know?

x/stakeibc/types/codec.go
```diff
@@ -26,7 +26,7 @@ func RegisterCodec(cdc *codec.LegacyAmino) {
 	cdc.RegisterConcrete(&MsgCalibrateDelegation{}, "stakeibc/CalibrateDelegation", nil)
 	cdc.RegisterConcrete(&MsgUpdateInnerRedemptionRateBounds{}, "stakeibc/UpdateInnerRedemptionRateBounds", nil)
 	cdc.RegisterConcrete(&MsgResumeHostZone{}, "stakeibc/ResumeHostZone", nil)
-	// this line is used by starport scaffolding # 2
+	cdc.RegisterConcrete(&MsgToggleTradeController{}, "stakeibc/ToggleTradeController", nil)
```
asalzmann, line 29: same comment as other PR, let's test with ledger. added to checklist

proto/stride/stakeibc/tx.proto
```diff
@@ -314,3 +316,26 @@ message MsgUpdateTradeRoute {
...
+  option (cosmos.msg.v1.signer) = "creator";
+  option (amino.name) = "stride/x/stakeibc/MsgToggleTradeController";
+
+  // Message signer (admin only)
+  string creator = 1 [ (cosmos_proto.scalar) = "cosmos.AddressString" ];
+  // Chain ID of the trade account
+  string chain_id = 2;
+  // Permission change (either grant or revoke)
+  AuthzPermissionChange permission_change = 3;
+  // Address of controller
```
asalzmann, line 338: Should we call this operator instead of controller, to keep it consistent with stTIA/stDYM?

x/stakeibc/keeper/trade_route.go
```diff
@@ -51,3 +51,13 @@ func (k Keeper) GetAllTradeRoutes(ctx sdk.Context) (list []types.TradeRoute) {
 
 	return
 }
+
+// Searches for a trade route by the trade account chain ID
+func (k Keeper) GetTradeRouteFromTradeChainId(ctx sdk.Context, chainId string) (tradeRoute types.TradeRoute, found bool) {
```
asalzmann, line 56:
```suggestion
func (k Keeper) GetTradeRouteFromTradeAccountChainId(ctx sdk.Context, chainId string) (tradeRoute types.TradeRoute, found bool) {
```

asalzmann (approved):
lgtm!

x/stakeibc/keeper/msg_server.go
shellvish: This makes sense to me! Resolving

x/stakeibc/keeper/reward_converter.go
shellvish: Got it, good to know. We should mark this as a fast follow, it doesn't seem terribly hard to do, and it sounds like the Osmosis team is conceptually on-board.

proto/stride/stakeibc/tx.proto
sampocs: done

x/stakeibc/keeper/trade_route.go
sampocs: done

PR #1145, opened 2024-03-19
change reward calculation to use stDYDX instead of DYDX

sampocs:
## Context
When calculating the rebate, we need to determine the portion of TVL that's made up from the community pool liquid stake. We were originally using the number of native tokens and the total delegation field on the host zone to build this ratio, but these two numbers naturally diverge as rewards are reinvested. Instead we can use number of stTokens which does not change during reinvestment.

## Brief Changelog
* Updated tx and host zone protos to have `RebateRate` instead of `RebatePercentage` and `LiquidStakedStTokenAmount` instead of `LiquidStakedAmount`
* Updated reward calculation to build the "contribution ratio" from the stToken supply instead of total delegations


x/stakeibc/keeper/icqcallbacks_withdrawal_reward_balance_test.go
```diff
@@ -115,14 +116,18 @@ func (s *KeeperTestSuite) TestWithdrawalRewardBalanceCallback_SuccessfulWithReba
...
 	hostZone.TotalDelegations = sdkmath.NewInt(10_000_000)
 	hostZone.CommunityPoolRebate = &types.CommunityPoolRebate{
-		RebatePercentage:  sdk.MustNewDecFromStr("0.5"),
-		LiquidStakeAmount: sdkmath.NewInt(1_000_000),
+		RebateRate:                sdk.MustNewDecFromStr("0.5"),
+		LiquidStakedStTokenAmount: stTokenSupply,
 	}
 	s.App.StakeibcKeeper.SetHostZone(s.Ctx, hostZone)
 
+	// Mint stTokens so there the supply matches the liquid staked amount
```
riley-stride, line 128:
```suggestion
	// Mint stTokens so that the supply matches the liquid staked amount
```

x/stakeibc/keeper/msg_server.go
```diff
@@ -1119,12 +1119,12 @@ func (k msgServer) SetCommunityPoolRebate(
 
 	// If a zero rebate is specified, set the rebate to nil
```
riley-stride, line 1120:
```suggestion
	// If a zero rebate rate or zero LiquidStakedStTokenAmount is specified, set the rebate to nil
```

riley-stride (approved):
looks great - I'm signed off on this one!

proto/stride/stakeibc/host_zone.proto
```diff
@@ -8,13 +8,15 @@ import "cosmos_proto/cosmos.proto";
...
 message CommunityPoolRebate {
-  string rebate_percentage = 1 [
+  // Rebate percentage as a decimal (e.g. 0.2 for 20%)
+  string rebate_rate = 1 [
     (gogoproto.customtype) = "github.com/cosmos/cosmos-sdk/types.Dec",
     (gogoproto.nullable) = false
   ];
-  string liquid_stake_amount = 2 [
+  // Number of stTokens received from the community pool liquid stake
+  string liquid_staked_st_token_amount = 2 [
```
shellvish, line 19: I appreciate the clarity on the variable naming!

x/stakeibc/keeper/reward_converter.go
```diff
@@ -85,24 +85,28 @@ func (k Keeper) CalculateRewardsSplitBeforeRebate(
...
 	// This will also prevent a division by 0 error
-	if hostZone.TotalDelegations.IsZero() {
+	if stTokenSupply.IsZero() {
 		return sdkmath.ZeroInt(), sdkmath.ZeroInt(), errorsmod.Wrapf(types.ErrDivisionByZero,
 			"unable to calculate rebate amount for %s since total delegations are 0", hostZone.ChainId)
 	}
 
 	// It also shouldn't be possible for the liquid stake amount to be greater than the full TVL
-	if rebateInfo.LiquidStakeAmount.GT(hostZone.TotalDelegations) {
+	if rebateInfo.LiquidStakedStTokenAmount.GT(stTokenSupply) {
```
shellvish, line 100: Just curious, what happens in this case? Should we add any protection to setting `rebateInfo` to make sure this case doesn't trigger?

x/stakeibc/keeper/reward_converter.go
```diff
@@ -145,24 +149,28 @@ func (k Keeper) CalculateRewardsSplitAfterRebate(
 	if !chainHasRebate {
 		strideFeeAmount = sdk.NewDecFromInt(rewardsAmount).Mul(totalFeeRate).TruncateInt()
 	} else {
-		// It shouldn't be possible to have 0 delegations (since there are rewards and there was a community pool stake)
+		// Get supply of stTokens to determine the portion of TVL that the community pool liquid stake makes up
+		stDenom := utils.StAssetDenomFromHostZoneDenom(hostZone.HostDenom)
```
shellvish, line 153: Do you think it's worth consolidating some of the code in this function and `CalculateRewardsSplitBeforeRebate` into a helper, to reduce code duplication? It could make things slightly easier to follow. These are relatively small functions though, so would understand if you'd rather keep as is

shellvish (approved):
Looks great, very easy to follow! Thanks for doing this.

x/stakeibc/keeper/reward_converter.go
sampocs: yeah I took a pass it it in the other PR but the initial reaction from people was that it actually made the code a little harder to follow / it was nice having all the math in one place

I may take another stab at it, but I think I _slightly_ lean towards keeping it as it is for now

x/stakeibc/keeper/reward_converter.go
sampocs: realistically I don't think it should ever trigger! We could add a check when registering a rebate that the LS amount is less than the current supply, but the supply can change so it doesn't perfectly eliminate this case

sampocs: just added the check during registration

shellvish: That makes sense to me, thanks for adding it! 

Out of curiosity, do you know what would happen if it did trigger? Is there a way to handle it gracefully (e.g. just not pay a rebate)?

x/stakeibc/keeper/reward_converter.go
shellvish: That makes sense, would definitely believe it would make it harder to follow. Let's keep as is, resolving!

asalzmann (approved):
lgtm!

ethan-stride (approved):
LGTM

x/stakeibc/keeper/reward_converter.go
sampocs: We could just admin update the rebate struct if it did trigger (and then next epoch the rebate would be ignored)

But imo, we should throw and error here instead of silently failing since it's unexpected!

shellvish: Makes sense to me, marking as resolved!

PR #1147, opened 2024-03-20
generalized community pool destination address

sampocs:
## Context
In order to support sending tokens to a community pool treasury address (which is not the default community pool), this PR adds an optional address field to the host zone struct. If the field is empty, tokens are sent to the default community pool via `MsgFundCommunityPool. If the field is not empty and an address is specified, tokens are bank sent to that address.

**I would recommend reviewing commit by commit**.

## Brief Changelog
* Added `community_pool_treasury_address` to host zone struct
* Added docstrings to the host zone struct (needed to add it for this new field so figured I'd add em all)
* Created a separate `BuildFundCommunityPoolMsg` to build the underlying ICA message, but not send it, so that we can add a unit test that inspects the message contents
* Added a branching statement to handle both the bank send case and the MsgFundCommunityPool case
* Added unit test


x/stakeibc/keeper/community_pool.go
```diff
@@ -245,12 +245,22 @@ func (k Keeper) BuildFundCommunityPoolMsg(
...
 
-	msgs := []proto.Message{&disttypes.MsgFundCommunityPool{
-		Amount:    sdk.NewCoins(token),
-		Depositor: sender,
-	}}
+	// If the community pool treasury address is specified, bank send there
+	if hostZone.CommunityPoolTreasuryAddress != "" {
+		fundMsg = []proto.Message{&banktypes.MsgSend{
+			FromAddress: sender,
+			ToAddress:   hostZone.CommunityPoolTreasuryAddress,
```
riley-stride, line 252: potential error case: we could mistakenly set `hostZone.CommunityPoolTreasuryAddress` to the address of the actual community pool, which is not whitelisted for bank sends. any way to prevent this case, or should we just be vigilant about setting it properly?

x/stakeibc/keeper/community_pool_test.go
```diff
@@ -438,3 +439,95 @@ func (s *KeeperTestSuite) TestRedeemCommunityPoolTokens_Failure_NotEnoughDelegat
...
+	testCases := []struct {
+		name             string
+		senderAccoutType types.ICAAccountType
+		sendToTreasury   bool
+		expectedSender   string
+		expectedReceiver string
+		expectedError    string
+	}{
+		{
+			name:             "community pool return ICA to main community pool",
```
riley-stride, line 461: are withdrawalica and community pool return ica the only possible senders here?

riley-stride (commented):
looking good!

x/stakeibc/keeper/community_pool.go
sampocs: I think we should just be vigilant about setting it properly.

If it wasn't whitelisted, I think the ICA would just fail

x/stakeibc/keeper/community_pool_test.go
sampocs: yup, at least with the current implementation (easy to extend it later if needed, but need custom code for each address)

riley-stride (approved):

asalzmann (approved):
lgtm!

proto/stride/stakeibc/host_zone.proto
```diff
@@ -20,71 +20,114 @@ message CommunityPoolRebate {
   ];
 }
 
+// Core data structure to track liquid staking zones
```
shellvish, line 25: Honestly shocking we've gone this long without commenting this file

Thank you for commenting it so thoroughly!

x/stakeibc/keeper/community_pool.go
```diff
@@ -224,33 +224,57 @@ func (k Keeper) RedeemCommunityPoolTokens(ctx sdk.Context, hostZone types.HostZo
...
 	}
 
-	var msgs []proto.Message
-	msgs = append(msgs, &disttypes.MsgFundCommunityPool{
-		Amount:    fundCoins,
-		Depositor: depositor,
-	})
+	// If the community pool treasury address is specified, bank send there
+	if hostZone.CommunityPoolTreasuryAddress != "" {
+		fundMsg = []proto.Message{&banktypes.MsgSend{
```
ethan-stride, line 250: So I assume that we have verified that the Treasury module account they are using can accept normal bank sends then.  And the only addresses which can send are the withdrawal address or the community pool return ICA both of which actually exist on the DYDX chain so bank sends will work.  I think this looks good

ethan-stride (approved):
LGTM
Assuming that we have already verified that the Treasury module account they are using is able to accept normal bank sends.

x/stakeibc/keeper/msg_server.go
```diff
@@ -112,6 +112,14 @@ func (k msgServer) RegisterHostZone(goCtx context.Context, msg *types.MsgRegiste
 		return nil, errorsmod.Wrapf(err, "unable to create community pool redeem account for host zone %s", chainId)
 	}
 
+	// Validate the community pool treasury address if it's non-empty
+	if msg.CommunityPoolTreasuryAddress != "" {
+		_, err := utils.AccAddressFromBech32(msg.CommunityPoolTreasuryAddress, msg.Bech32Prefix)
```
shellvish, line 117: Just verifying, this function doesn't error on modular accounts, ICA accounts, or CW accounts? I think that's the case, but just want to double-check!

shellvish (approved):
This looks great to me, thanks a lot for doing this! 

Just flagging, I think we want to do two additional testing steps: 
1. In Dockernet, try funding the community treasury 
2. Run this flow on local->mainnet

Neither should be particularly hard, just writing down so we don't forget!

x/stakeibc/keeper/msg_server.go
sampocs: yup that's right! it works for all of them

PR #1148, opened 2024-03-20
v20 upgrade handler scaffolding

riley-stride:
Adds the upgrade handler shell for v20. Upgrade logic to be done in separate PRs


app/upgrades.go
```diff
@@ -263,6 +264,17 @@ func (app *StrideApp) setupUpgradeHandlers(appOpts servertypes.AppOptions) {
 		),
 	)
 
+	// v20 upgrade handler
+	app.UpgradeKeeper.SetUpgradeHandler(
+		v20.UpgradeName,
+		v20.CreateUpgradeHandler(
+			app.mm,
+			app.configurator,
+			app.appCodec,
```
sampocs, line 273: I don't think we need the appCodec

app/upgrades/v20/upgrades.go
```diff
@@ -0,0 +1,57 @@
...
+	cdc codec.Codec,
+	stakeIbcKeeper stakeibckeeper.Keeper,
+) upgradetypes.UpgradeHandler {
+	return func(ctx sdk.Context, _ upgradetypes.Plan, vm module.VersionMap) (module.VersionMap, error) {
+		ctx.Logger().Info("Starting upgrade v20...")
+
+		ctx.Logger().Info("Adding DYDX Community Pool Treasury Address...")
+		if err := SetDydxCommunityPoolTreasuryAddress(ctx, stakeIbcKeeper); err != nil {
+			return nil, errorsmod.Wrapf(err, "unable to set dydx community pool treasury address")
+		}
```
sampocs, line 34: per the slack convo, can you scratch this from this PR so that this PR is just the upgrade handler shell?

app/upgrades/v20/upgrades.go
```diff
@@ -0,0 +1,57 @@
...
+
+		return nil, nil
```
sampocs, line 36:
```suggestion
		return mm.RunMigrations(ctx, configurator, vm)
```

sampocs (commented):
Thanks for taking this! Can you pull out the dydx stuff so we can merge this to main?

app/upgrades/v20/upgrades.go
riley-stride: done

app/upgrades/v20/upgrades.go
riley-stride: done

sampocs (approved):
lgtm!

PR #1153, opened 2024-03-21
Upgraded ICS to 4.0

sampocs:
## Context 
Upgraded ICS to v4

## Brief Changelog
* Changed ICS version and import paths
* Added migrations to upgrade handler
* Fixed unit tests


app/upgrades/v20/upgrades.go
```diff
@@ -40,3 +58,24 @@ func SetDydxCommunityPoolTreasuryAddress(ctx sdk.Context, stakeIbcKeeper stakeib
...
+			return err
+		}
+		ck.DeleteOutstandingDowntime(ctx, consAddr)
+	}
+	return nil
+}
+
+// Migrates ICS Params to add the new RetryDelayParam
+func MigrateICSParams(ctx sdk.Context, ck ccvconsumerkeeper.Keeper) {
+	params := ck.GetConsumerParams(ctx)
```
asalzmann, line 78: Nice - so this didn't work pre-SDK 47, and now it does?

asalzmann (approved):
LGTM

app/upgrades/v20/upgrades.go
sampocs: that's my guess. Although, I think it might be up to the specific module. For instance, Stride as a chain is on SDK 47 but I don't think stakeibc ever upgraded the param store to the new format (low confidence though)

Either way, we can figure out if this works right away through the dockernet IG tests

PR #1157, opened 2024-03-21
Modify ICS Params in v20 Migration

shellvish:


sampocs (approved):
lgtm, thanks for the quick fix!

app/upgrades/v20/upgrades.go
```diff
@@ -84,7 +86,33 @@ func MigrateICSOutstandingDowntime(ctx sdk.Context, ck ccvconsumerkeeper.Keeper)
 
 // Migrates ICS Params to add the new RetryDelayParam
```
sampocs, line 87: can you just add a comment here that the reason we set from the default is because you can't unmarshal old params?

app/upgrades/v20/upgrades.go
```diff
@@ -84,7 +86,33 @@ func MigrateICSOutstandingDowntime(ctx sdk.Context, ck ccvconsumerkeeper.Keeper)
...
+	params.BlocksPerDistributionTransmission = 1000
+	params.DistributionTransmissionChannel = "channel-0"
+	params.ProviderFeePoolAddrStr = "cosmos1ap0mh6xzfn8943urr84q6ae7zfnar48am2erhd"
+	params.CcvTimeoutPeriod = time.Second * 2419200
+	params.TransferTimeoutPeriod = time.Second * 3600
+	params.ConsumerRedistributionFraction = "0.85"
+	params.HistoricalEntries = 10000
+	params.UnbondingPeriod = time.Second * 1209600
+	params.SoftOptOutThreshold = "0.05"
+	params.RewardDenoms = []string{"stucmdx",
```
asalzmann, line 103: should we add in the new reward denoms? stutia stadym

asalzmann (approved):

app/upgrades/v20/upgrades.go
shellvish: It's a good question, would defer to you! Do you have a sense of what the Hub devs would prefer?

app/upgrades/v20/upgrades.go
shellvish: Definitely, added!

asalzmann (approved):

PR #1158, opened 2024-03-21
added wasm to cosmovisor upgrade image

sampocs:
## Context
In order to run the dockernet upgrade, we need wasm on the docker image that builds the old binary


shellvish (approved):
Thanks for taking this quick fix!

asalzmann (approved):

PR #1163, opened 2024-03-25
added help info to toggle command

sampocs:
## Context
The chain-id arg was a bit confusing as to whether it was supposed to be dydx or osmosis. Added long help option to clarify


shellvish (approved):
Incredible, thank you for doing this!

PR #1165, opened 2024-03-26
added v21 upgrade handler

sampocs:
## Context
Added v21 upgrade handler which has no custom logic


shellvish (approved):
A tricky one, thanks for doing this!

ethan-stride (approved):
LGTM

PR #1166, opened 2024-03-26
added function to unregister host zone

sampocs:
## Context
Added helper function (not exposed via txs) to unregister a host zone. This will be used only on our testnet

## Brief Changelog
* Added function to undo the commands during host zone registration. This includes:
  * Removing the host zones struct
  * Removing the module accounts
  * Removing the deposit and epoch unbonding records
  * Removing the whitelisted addresses pairs

## Testing
To test, I added a condition to the register host zone function (for testing purposes only) that would un-register the host zone if the function was ran with a host that already exists. Then I tested with the following:
* Updated start_relayers to register a second set of clients/connections/channels for GAIA
* Started dockernet
```bash
make start-docker
```
* Confirmed the host zone was registered
```bash
build/strided --home dockernet/state/stride1 q stakeibc list-host-zone
```
* Called the register-host-zone command - which since the host already exists, should un-register it
```bash
build/strided --home dockernet/state/stride1 tx stakeibc register-host-zone \
    connection-0 uatom cosmos \
    ibc/27394FB092D2ECCD56123C74F36E4C1F926001CEADA9CA97EA622B25F41E5EB2 \
    channel-0 1 true \
    --from admin -y --gas 4000000
```
* Confirmed the host zone was removed
```bash
build/strided --home dockernet/state/stride1 q stakeibc list-host-zone
```
* Ran the registration command again to re-register with a different connection ID
```bash
build/strided --home dockernet/state/stride1 tx stakeibc register-host-zone \
    connection-1 uatom cosmos \
    ibc/27394FB092D2ECCD56123C74F36E4C1F926001CEADA9CA97EA622B25F41E5EB2 \
    channel-1 1 true \
    --from admin -y --gas 4000000
```
* Finally, confirmed the host zone was registered again
```bash
build/strided --home dockernet/state/stride1 q stakeibc list-host-zone
```


shellvish (commented):
This looks good to me, although how is this function meant to be called on testnet? Is it through governance?

x/stakeibc/keeper/host_zone.go
```diff
@@ -112,6 +113,55 @@ func (k Keeper) GetAllHostZone(ctx sdk.Context) (list []types.HostZone) {
...
+	depositAddress := types.NewHostZoneDepositAddress(chainId)
+	communityPoolStakeAddress := types.NewHostZoneModuleAddress(chainId, CommunityPoolStakeHoldingAddressKey)
+	communityPoolRedeemAddress := types.NewHostZoneModuleAddress(chainId, CommunityPoolRedeemHoldingAddressKey)
+
+	k.AccountKeeper.RemoveAccount(ctx, k.AccountKeeper.GetAccount(ctx, depositAddress))
+	k.AccountKeeper.RemoveAccount(ctx, k.AccountKeeper.GetAccount(ctx, communityPoolStakeAddress))
+	k.AccountKeeper.RemoveAccount(ctx, k.AccountKeeper.GetAccount(ctx, communityPoolRedeemAddress))
+
+	// Remove all deposit records for the host zone
+	for _, depositRecord := range k.RecordsKeeper.GetAllDepositRecord(ctx) {
```
ethan-stride, line 133: I understand that this function will often be called when we are actually keeping the data in a different way and doing an upgrade or something, but in case this remove is called on a hostzone without replacement, do we want to error or at least log if there are are records still because this means we have control of funds and we are losing the records which track them?

ethan-stride (approved):
LGTM

Do we want to add this behind an admin gated tx?

Should we do similar ones for staketia and stakedym which have different fields and record types?

x/stakeibc/keeper/host_zone.go
ethan-stride: I see that this is only expected to be called on the testnet, but how do we ensure that is the case?

sampocs: we ensure it's only called on testnet by not creating a tx for it

sampocs:
@shellvish 
> This looks good to me, although how is this function meant to be called on testnet? Is it through governance?

Just through the upgrade handler! Considering it's unlikely to be used outside of the testnet, I don't think we should make a tx for it

shellvish:
> Just through the upgrade handler! Considering it's unlikely to be used outside of the testnet, I don't think we should make a tx for it

That's perfect @sampocs! Just want to make sure there's no worry of this accidentally being called where it shouldn't be. I'm signed off on this!

shellvish (approved):
Approved based on comments!

PR #1173, opened 2024-04-12
added ibc-hooks

sampocs:
## Context
Added IBCHooks by roughly following the guide [here](https://github.com/cosmos/ibc-apps/blob/main/modules/ibc-hooks/README.md#installation). The guide seemed to be in accurate in places, so I filled in the gaps by referencing osmosis/stargaze/neutron, and by leveraging my own understanding of the app.go wiring.

**Note to reviewer**: Please double check the ICS4Wrapper wiring.

Should we also restrict packets to be one of (a) autopilot, (b) pfm or (c) wasm?

## Brief Changelog
* Upgraded ibc to v7.4.0
* Added IBCHooks module
* Added store key

### High Level Approach
1. Deployed a test contract to stride on dockernet (ica oracle contract)
2. Sent a transfer with a hook to add a metric to the oracle
3. Confirmed that the state changed accordingly in the contract

### Takeaways
I had to increase the relayer gas adjustment in order to get the tx relayed (since the OnRecv now has more computation). **We'll want to update our relayers accordingly!**

### Steps
* Started dockernet normally
```bash
>>> make start-docker
```
* Uploaded the ICA oracle contract to stride
```bash
>>> strided tx wasm store ../ica-oracle/artifacts/ica_oracle.wasm --from val1 -y --gas 5000000
```
* Determined the address that would execute the contract by creating a temporary unit test in the ibc-hooks repo and using their function that builds the "intermediate sender" address
```go
senderBech32, err := ibchookskeeper.DeriveIntermediateSender(
	"channel-0",
	"cosmos1uk4ze0x4nvh4fk0xm4jdud58eqn4yxhrgl2scj", // gaia validator 1
	"stride",
)
fmt.Println(senderBech32)

Address: stride1038hhcqj3syag68jrzhj387uy0qx88u3ta537e4nk8xn8pwuxt5sl0xrr9
```
* Instantiated the contract with the admin address as `stride1038hhcqj3syag68jrzhj387uy0qx88u3ta537e4nk8xn8pwuxt5sl0xrr9` (the address that will be generated from ibc-hooks) since adding metrics to the oracle is permissioned
```bash
>>> init_msg='{"admin_address": "stride1038hhcqj3syag68jrzhj387uy0qx88u3ta537e4nk8xn8pwuxt5sl0xrr9", "transfer_channel_id": "channel-0"}'
>>> strided tx wasm instantiate 1 "$init_msg" --from val1 --label "ica-oracle" --no-admin --gas auto --gas-adjustment 1.3 -y

Contract Address: stride14hj2tavq8fpesdwxxcu44rty3hh90vhujrvcmstl4zr3txmfvw9szh70ur
```
* Queried the oracle and confirmed there were no metrics
```bash
>>> strided q wasm contract-state smart stride14hj2tavq8fpesdwxxcu44rty3hh90vhujrvcmstl4zr3txmfvw9szh70ur '{"all_latest_metrics": {}}'

data:
  metrics: []
```
* Sent a transfer from the hub to Stride with a wasm ibc hook to write a metric to the oracle
```bash
>>> attributes=$(echo '{"sttoken_denom": "stuatom"}' | base64)
>>> add_metric_message='{"post_metric": {"key": "stuatom_redemption_rate", "value": "1.1", "metric_type": "redemption_rate", "update_time": 100, "block_height": 1000, "attributes": "'$attributes'"}}'
>>> memo='{"wasm": {"contract": "stride14hj2tavq8fpesdwxxcu44rty3hh90vhujrvcmstl4zr3txmfvw9szh70ur", "msg": '"$add_metric_message"'}}'
>>> gaiad tx ibc-transfer transfer transfer channel-0 stride14hj2tavq8fpesdwxxcu44rty3hh90vhujrvcmstl4zr3txmfvw9szh70ur 1000uatom --from gval1 -y --memo "$memo" -y
```
* Confirmed the metric was added
```bash
>>> strided q wasm contract-state smart stride14hj2tavq8fpesdwxxcu44rty3hh90vhujrvcmstl4zr3txmfvw9szh70ur '{"all_latest_metrics": {}}'

data:
  metrics:
  - attributes: eyJzdHRva2VuX2Rlbm9tIjogInN0dWF0b20ifQo=
    block_height: 1000
    key: stuatom_redemption_rate
    metric_type: redemption_rate
    update_time: 100
    value: "1.1"
```


app/app.go
```diff
@@ -498,7 +523,7 @@ func NewStrideApp(
 		app.IBCKeeper.ChannelKeeper,
 		app.DistrKeeper,
 		app.BankKeeper,
-		app.RatelimitKeeper, // ICS4Wrapper
+		app.HooksICS4Wrapper, // ICS4Wrapper
```
asalzmann, line 526: Why this change?

app/app.go
```diff
@@ -781,19 +801,21 @@ func NewStrideApp(
 	// - staketia
 	// - stakedym
 	// - ratelimit
+	// - ibchooks
```
asalzmann, line 804: re: ordering - I think memo data should trigger one of autopilot / ibchooks / pfm, but should not trigger 2 or 3 of them. Is this also your understanding?

If so, I don't think the order of those 3 modules in the stack matters

sampocs: I think today we don't have any use cases of all 3 at once. In the future, I could see us using autopilot + ibchooks together. And then I think there should never be a use case for (autopilot + pfm) or (ibchooks + pfm).

The only current restriction in the code is that packets can't be autopilot and pfm. Feels safest rn to just extend it to only 1 of the 3

app/app.go
sampocs: Traffic up the stack is routed via the ICS4Wrappers that are wired on each keeper. The stack was previously 
```
channel
...
ratelimit
pfm
transfer
```
where the ICS4 routing was:
* transfer wires pfm 
* pfm wires rate limit
* rate limit wires channel

---

Now the stack is
```
channel
...
ratelimit
ibchooks [NEW]
pfm
transfer
```
So the routing needs to be:
* transfer wires pfm 
* pfm wires IBC hooks [NEW]
* IBC hooks wires rate limit [NEW]
* rate limit wires channel

asalzmann: Ah yeah, this always confuses me. So the order of outbound packets depends on the Keeper definitions?

app/app.go
sampocs: [done](https://github.com/Stride-Labs/stride/pull/1173/commits/f00f7c1bf1594f6e19cd479a5951bc10b1efbb7d)

app/app.go
sampocs: yeah exactly. The part that I could use a second set of eyes on is usually, you just pass the keeper as the ICS4Wrapper, but hooks instead uses this `HooksICS4Wrapper` struct instead. 

I think they're functionally the same but would be good to have someone independently verify

app/app.go
asalzmann: Nice!

asalzmann (approved):
LGTM

x/autopilot/types/parser.go
```diff
@@ -66,10 +66,21 @@ func ParseAutopilotMetadata(metadata string) (*AutopilotMetadata, error) {
 		return nil, nil
 	}
 
-	// Packets cannot be used for both autopilot and PFM at the same time
-	// If both fields were provided, reject the packet
-	if raw.Autopilot != nil && raw.Forward != nil {
-		return nil, errorsmod.Wrapf(ErrInvalidPacketMetadata, "autopilot and pfm cannot both be used in the same packet")
+	// Packets cannot be used for more than one of autopilot, pfm, or wasmhooks at the same time
```
shellvish, line 69: This is smart! It might be worth adding an integration test to explicitly check for this (def no need to do it as part of this PR, but just generally it might be worth verifying this invariant given it's a potentially very impactful assumption we're making in the codebase now)

shellvish (approved):
This looks great to me! 

One caveat though that the wiring in `app.go` is a bit arcane, would just want to make sure we test this the migration thoroughly on localstride + testnet + locally. Might be worthwhile to make a list of tests that we want to do, and then we can verify on all of the various environments. 

More than happy to help with that of course!

app/app.go
```diff
@@ -517,8 +542,36 @@ func NewStrideApp(
 	// Set the TransferKeeper reference in the PacketForwardKeeper
 	app.PacketForwardKeeper.SetTransferKeeper(app.TransferKeeper)
 
-	transferModule := transfer.NewAppModule(app.TransferKeeper)
-	transferIBCModule := transfer.NewIBCModule(app.TransferKeeper)
+	// Add wasm keeper (must be after IBCKeeper and TransferKeeper)
+	wasmCapabilities := "iterator,staking,stargate,cosmwasm_1_1,cosmwasm_1_2,cosmwasm_1_3,cosmwasm_1_4"
+	wasmDir := filepath.Join(homePath, "wasm")
```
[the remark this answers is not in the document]
sampocs: Yeah I think it’s just a standard config file! Similar to config.toml/app.toml

app/app.go
```diff
@@ -781,19 +801,21 @@ func NewStrideApp(
...
 	// - stakedym
 	// - ratelimit
+	// - ibchooks
 	// - pfm
 	// - transfer
 	// - base app
 	// Note: Traffic up the stack does not pass through records or autopilot,
 	// as defined via the ICS4Wrappers of each keeper
-	var transferStack porttypes.IBCModule = transferIBCModule
+	var transferStack porttypes.IBCModule = transfer.NewIBCModule(app.TransferKeeper)
```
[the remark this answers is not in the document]
sampocs: Just seemed better to define it near where it’s used instead of way up above. Especially since most stuff above is really just keeper definitions

x/autopilot/types/parser.go
sampocs: There’s unit tests for it in this PR!

app/app.go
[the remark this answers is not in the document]
shellvish: Makes sense!

x/autopilot/types/parser.go
shellvish: I'm on-board with keeping it as just unit-tests!

app/app.go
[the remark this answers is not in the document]
shellvish: Makes sense, resolving

app/app.go
sampocs: @asalzmann am I good to mark this resolved?

asalzmann: Yes

x/autopilot/types/autopilot.go
```diff
@@ -8,17 +8,18 @@ import (
...
+// As a result, only the key is needed, cause the actual parsing of the PFM/wasm
+// packet will occur in the respective module
 type RawPacketMetadata struct {
 	Autopilot *struct {
 		Receiver string                  `json:"receiver"`
 		Stakeibc *StakeibcPacketMetadata `json:"stakeibc,omitempty"`
 		Claim    *ClaimPacketMetadata    `json:"claim,omitempty"`
 	} `json:"autopilot"`
 	Forward *interface{} `json:"forward"`
+	Wasm    *interface{} `json:"wasm"`
```
ethan-stride, line 22: The main thing I was confused about before was that normal packets could be picked up and interpreted by a wasm contract but the diagram you made in excalidraw of how all the middlewares interact helped a lot.  I was confused because I thought before that only tx specifically aimed at cosmwasm modules would be seen by wasm but now I see how it all fits together and they use different parts of the data so that it is possible for these cosm middlewares and another middleware like autopilot can act on the same packets even using data from the same memo.  Unit tests also made it clearer

ethan-stride (commented):
LGTM

PR #1175, opened 2024-04-12
added stToken burn to unregister host zone function

sampocs:
## Context
See PR #1166 for context on the unregister host zone function. TLDR is it's only for our testnet.

This PR extends the function to also remove redemption records and burn any outstanding stTokens. 

## Testing 
* Updated `RegisterHostZone` to "unregister" if the host zone exists. This allows me to test the function by first running the register tx back to back (with the first execution adding it, and the second one removing it)
```go
	// get zone
	_, found = k.GetHostZone(ctx, chainId)
	if found {
		if err := k.UnregisterHostZone(ctx, chainId); err != nil {
			return nil, err
		}
		return &types.MsgRegisterHostZoneResponse{}, nil

```
* Started dockernet and confirmed the host zone was registered
* Liquid staked and transferred tokens off stride
```bash
gaiadl tx ibc-transfer transfer transfer channel-0 stride1uk4ze0x4nvh4fk0xm4jdud58eqn4yxhrt52vv7 10000uatom --from gval1 -y
stridedl tx stakeibc liquid-stake 10000 uatom --from val1 -y
stridedl tx ibc-transfer transfer transfer channel-0 cosmos1uk4ze0x4nvh4fk0xm4jdud58eqn4yxhrgl2scj 1000stuatom --from val1 -y
```
* Confirmed stToken balances and supply before the host is unregistered
```bash
stridedl q bank balances $(stridedl keys show val1 -a) --denom stuatom
stridedl q bank balances $(stridedl q ibc-transfer escrow-address transfer channel-0) --denom stuatom
stridedl q bank total --denom=stuatom
stridedl q ibc-transfer total-escrow stuatom
```
* Call the register host command which will **unregister**
```bash
stridedl tx stakeibc register-host-zone \
    connection-0 uatom cosmos \
    ibc/27394FB092D2ECCD56123C74F36E4C1F926001CEADA9CA97EA622B25F41E5EB2 \
    channel-0 1 true \
    --from admin -y --gas 4000000
```
* Confirmed all balances were set back to 0
```bash
stridedl q bank balances $(stridedl keys show val1 -a) --denom stuatom
stridedl q bank balances $(stridedl q ibc-transfer escrow-address transfer channel-0) --denom stuatom
stridedl q bank total --denom=stuatom
stridedl q ibc-transfer total-escrow stuatom
```
* Exec into the relayer and create a new channel
```
rly tx link stride-gaia --override
```
* Run the registration command again to re-register with the new connection ID and channel ID
```
stridedl tx stakeibc register-host-zone \
    connection-1 uatom cosmos \
    ibc/25418646C017D377ADF3202FF1E43590D0DAE3346E594E8D78176A139A928F88 \
    channel-7 1 true \
    --from admin -y --gas 4000000
```
* Liquid stake again and confirm the delegation goes through and there's no chain halt

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 <!-- GH ACTIONS TEMPLATE - INSERT NEW VERSION HERE -->
 
+## [v21.0.0](https://github.com/Stride-Labs/stride/releases/tag/v21.0.0) - 2024-03-27
+
+### On-Chain changes
+1. added help info to toggle command ([#1163](https://github.com/Stride-Labs/stride/pull/1163))
+2. added v21 upgrade handler ([#1165](https://github.com/Stride-Labs/stride/pull/1165))
+3. updated rebate to use native denom instead of reward denom ([#1162](https://github.com/Stride-Labs/stride/pull/1162))
+4. added function to unregister host zone ([#1166](https://github.com/Stride-Labs/stride/pull/1166))
+5. v21 Import Paths ([#1167](https://github.com/Stride-Labs/stride/pull/1167))
+
+
 ## [v20.0.0](https://github.com/Stride-Labs/stride/releases/tag/v20.0.0) - 2024-03-21
 
 ### On-Chain changes
--- a/app/upgrades.go
+++ b/app/upgrades.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 
 	wasmtypes "github.com/CosmWasm/wasmd/x/wasm/types"
+	ratelimittypes "github.com/Stride-Labs/ibc-rate-limiting/ratelimit/types"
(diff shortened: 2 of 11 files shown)


ethan-stride (approved):
LGTM
Does this work for the other hostZones like stTia and stDym which are non-standard?  The ones with operator accounts instead of ICAs?

sampocs:
it does not cause those are in a different module

ethan-stride:
> it does not cause those are in a different module

Sorry yes, I knew that this currently wouldn't affect those.  I should have asked don't we need this same type of logic for resetting those ones in the other module and specifically will this accounts keeper method to burn all the stTokens work with the operator accounts on the testnet the same way?

sampocs:
oh I see - good question! i don't think we do atm. Hopefully those host zones become normal ICA host zones by the time we ever have to use this again. If not, we could always make the custom version when needed

PR #1176, opened 2024-04-12
Fix validator weight check

sampocs:
## Context
The validator weight check was be run after each validator was added. It should instead be done once after all validators were added - otherwise, it can error prematurely. 

## Brief Changelog
* Pulled out `CheckValidatorWeightsBelowCap` into relevant msg_server functions (with and without gov)
* Updated `CheckValidatorWeightsBelowCap` to take a chainId as an arg and look up the host zone + validators (instead of taking validators as an arg)
* Fixed the unit test

## Testing
* Created the following [add-validator-propoal-file](https://github.com/Stride-Labs/stride/files/14953726/add-vals.json). This has the same weights that were used for stSAGA
* Started dockernet **on main branch**, placing an early exit in `register_host.sh` before the validators were added normally
* Added validators using the attached file, confirmed it errored
```bash
>>> strided tx stakeibc add-validators GAIA add-vals.json --from admin -y --gas 1000000

raw_log: 'failed to execute message; message index: 0: validator cosmosvaloper14kn0kk33szpwus9nh8n87fjel8djx0y070ymmj
  exceeds weight cap, has 11.39913057478667% of the total weight when the cap is 10%:
```
* Restarted the process on this PR's branch, added validators with the same file, and confirmed it succeeded
* Changed the first validators weight to 1200 and repeated the process - confirming that registration failed

### DIFF
--- a/x/stakeibc/keeper/gov.go
+++ b/x/stakeibc/keeper/gov.go
@@ -13,6 +13,11 @@ func (k Keeper) AddValidatorsProposal(ctx sdk.Context, msg *types.AddValidatorsP
 		}
 	}
 
+	// Confirm none of the validator's exceed the weight cap
+	if err := k.CheckValidatorWeightsBelowCap(ctx, msg.HostZone); err != nil {
+		return err
+	}
+
 	return nil
 }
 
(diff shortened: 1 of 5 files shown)


shellvish (approved):
This looks great to me, thank you for doing it! 

Only question, should we also enforce the ValidatorWeightCap is respected when validators are deleted? My bias is yes, but could reasonably see it going either way

sampocs:
hmm very good question! I'd lean yes as well. Seems like it's a tradeoff with having some flexibility - but that flexibility doesn't seem that valuable

sampocs:
@shellvish I actually just realized that we can only remove validators if they have 0 weight - so the weight distribution shouldn't change when one is removed. I don't think we need the check

PR #1177, opened 2024-04-12
parameterize max messages per ICA tx on host zone

sampocs:
## Context
See #1124 for more context. This PR implements the same batching technique, but instead the parameter is included on the host zone instead of as a global param.

**Note to reviewer**: I would recommend reviewing this PR commit by commit

## Brief Changelog
* Added `MaxMessagesPerIcaTx` parameter to host zone type
* Added `MaxMessagesPerIcaTx` to `MsgRegisterHostZone` type
* Added optional flag to specify the parameter when registering a host zone
* Set the parameter on the host zone during registration - defaulting to 32 if not provided
* Used parameter in unbonding for the undelegate batch size
* **Check me**: no need to use the parameter during delegations since that can be controlled with the weights
* Added unit test

## Testing
* Started dockernet as normal, confirmed the parameter was set to 32
* Updated `register_host.sh` to pass `100` as the max instead and confirmed it was set on the host zone


proto/stride/stakeibc/host_zone.proto
```diff
@@ -124,6 +124,9 @@ message HostZone {
     (gogoproto.customtype) = "github.com/cosmos/cosmos-sdk/types.Dec",
     (gogoproto.nullable) = false
   ];
+  // The max number of messages that can be sent in a delegation
+  // or undelegation ICA tx
+  uint64 max_messages_per_ica_tx = 36;
```
ethan-stride, line 129: Why would this number vary per hostZone?  It totally makes sense on the tx proto but what factors would cause it to be different per hostZone and if there are such factors how will we experimentally figure it out for each one?  Can it be automatically measured and if so can we set this param empirically instead of through tx setting up hostZones

ethan-stride (approved):
LGTM

proto/stride/stakeibc/host_zone.proto
sampocs: it's based on the gas limit on each host zone - which is up to the chain's governance so it can change at any time. And yeah initially, I think we'll have to set it based on experimentation

sampocs:
I just realized, we may also want a gov transaction to change this value

proto/stride/stakeibc/host_zone.proto
ethan-stride: Hmm interesting, maybe the two values have to change together?  If governance changes the gas value but this setting remains too high will that cause tx with too many messages to be attempted and they will fail because of this?  If so then the governance to change the gas fees should update this value at the same time always

sampocs: To clarify - it's the governance on the host zone that decides the gas limit. So the hub could put up a proposal to reduce the gas limit, and, as you said, it would cause these ICAs to start failing. And then stride would have to put up a prop to change the max messages to take the new limit into account

sampocs:
I assume this wont be the last host-zone parameter we'll add. I'm thinking the ideal long term solution would be to have a nested params struct on the host zone, with a single tx in charge of overriding all params.

If I can get a thumbs up from someone that they agree, then in this PR I'll keep the host zone struct as is, but add the `UpdateHostZoneParams` tx that will only update this MaxIcaMessages param for now

x/stakeibc/keeper/unbonding.go
```diff
@@ -485,8 +482,8 @@ func (k Keeper) UnbondFromHostZone(ctx sdk.Context, hostZone types.HostZone) err
 	}
 
 	// Send the messages in batches so the gas limit isn't exceedeed
```
asalzmann, line 484: This comment is inaccurate right? Since we error if the number of undelegation messages is greater than the batch size

app/upgrades/v22/upgrades.go
```diff
@@ -4,20 +4,49 @@ import (
...
 
-const (
+var (
 	UpgradeName = "v22"
+
+	CosmosHubChainId = "cosmoshub-4"
+	OsmosisChainId   = "osmosis-1"
+	DydxChainId      = "dydx-mainnet-1"
+
+	MaxMessagesPerIcaByHost = map[string]uint64{
```
asalzmann, line 18: DYDX/Osmosis have higher blockly gas limits than CH (infinite and 300M), so think these limits should be ok for both

For CH, the gas to stake/unstake is somewhat unpredictable because it scales with additional ICS tokens.

Trying to track down pStake/QS ICA accounts to see how many validators they're delegating to

asalzmann (commented):
LGTM - just need to decide on number of validators for Cosmos Hub

x/stakeibc/keeper/unbonding.go
sampocs: good eye! updated

x/stakeibc/keeper/msg_server.go
```diff
@@ -277,6 +285,23 @@ func (k msgServer) RegisterHostZone(goCtx context.Context, msg *types.MsgRegiste
 	return &types.MsgRegisterHostZoneResponse{}, nil
 }
 
+func (ms msgServer) UpdateHostZoneParams(goCtx context.Context, msg *types.MsgUpdateHostZoneParams) (*types.MsgUpdateHostZoneParamsResponse, error) {
```
shellvish, line 288: Is this added as a function so governance can call this function? If so, is there any additional wiring needed (e.g. do we need to register anything with the gov module)?

proto/stride/stakeibc/tx.proto
```diff
@@ -366,3 +369,18 @@ message MsgToggleTradeController {
   string address = 4;
 }
 message MsgToggleTradeControllerResponse {}
+
+// Updates host zone params
+message MsgUpdateHostZoneParams {
```
shellvish, line 374: Re:your comment, I'm on-board with adding a gov-callable function that can update host zone params!

x/stakeibc/keeper/msg_server.go
```diff
@@ -277,6 +285,23 @@ func (k msgServer) RegisterHostZone(goCtx context.Context, msg *types.MsgRegiste
...
+	if ms.authority != msg.Authority {
+		return nil, errorsmod.Wrapf(govtypes.ErrInvalidSigner, "invalid authority; expected %s, got %s", ms.authority, msg.Authority)
+	}
+
+	hostZone, found := ms.Keeper.GetHostZone(ctx, msg.ChainId)
+	if !found {
+		return nil, types.ErrHostZoneNotFound.Wrapf("host zone %s not found", msg.ChainId)
+	}
+
+	hostZone.MaxMessagesPerIcaTx = msg.MaxMessagesPerIcaTx
```
shellvish, line 299: Should we do any validation on `msg.MaxMessagesPerIcaTx`? Possible it's handled elsewhere already!

x/stakeibc/keeper/msg_server.go
```diff
@@ -277,6 +285,23 @@ func (k msgServer) RegisterHostZone(goCtx context.Context, msg *types.MsgRegiste
 	return &types.MsgRegisterHostZoneResponse{}, nil
 }
 
+func (ms msgServer) UpdateHostZoneParams(goCtx context.Context, msg *types.MsgUpdateHostZoneParams) (*types.MsgUpdateHostZoneParamsResponse, error) {
+	ctx := sdk.UnwrapSDKContext(goCtx)
+	if ms.authority != msg.Authority {
```
shellvish, line 290: Sorry, this might be a silly question, what does this line do? Is `ms.authority` the gov module?

x/stakeibc/keeper/unbonding.go
```diff
@@ -485,8 +482,11 @@ func (k Keeper) UnbondFromHostZone(ctx sdk.Context, hostZone types.HostZone) err
...
-	for start := 0; start < len(msgs); start += UndelegateICABatchSize {
-		end := start + UndelegateICABatchSize
+	// NOTE: With the current implementation, an error is thrown upstream if there are
+	// more messages than undelegateBatchSize - so there should only be one iteration in this loop
+	// This is because the undelegation callback is not currently setup to handle multiple batches
+	for start := 0; start < len(msgs); start += undelegateBatchSize {
```
shellvish, line 488: I know we check upstream, but do you think it's at all worth it to add a redundant check here to verify this loop only fires once? This is just one of the places that's the most risky if anything unexpected happened

x/stakeibc/types/message_update_host_zone_params.go
```diff
@@ -0,0 +1,44 @@
...
+	addr, _ := sdk.AccAddressFromBech32(msg.Authority)
+	return []sdk.AccAddress{addr}
+}
+func (msg *MsgUpdateHostZoneParams) ValidateBasic() error {
+	if _, err := sdk.AccAddressFromBech32(msg.Authority); err != nil {
+		return errorsmod.Wrap(err, "invalid authority address")
+	}
+	if msg.ChainId == "" {
+		return errors.New("chain ID must be specified")
+	}
```
shellvish, line 42: Should we add "MaxICAMessages" is positive?

shellvish (commented):
This looks great! A few small questions, but overall looks fantastic. Thanks for doing this!

x/stakeibc/keeper/msg_server.go
sampocs: yup it's a gov function! the only part that's needed is the `authority` address check

x/stakeibc/keeper/msg_server.go
sampocs: good call - forgot the default coalesce was in the main register function. Just updated!

x/stakeibc/keeper/msg_server.go
sampocs: not a silly question! authority is the gov module account - so this is how gov transactions are handled in sdk 47+

You can run `submit-proposal` with any arbitrary sdk message, the gov module will route it to the msg_server with the module address as the "signer" and then, if you want to restrict the tx to gov-only, you have to add this clause to confirm that the signer is the gov module address!

x/stakeibc/keeper/unbonding.go
sampocs: imo, I don't think it's necessary. At that point, I think we'd just remove the loop, but sounded like we were on board with laying off this code for now to make the change less invasive

x/stakeibc/types/message_update_host_zone_params.go
sampocs: it's a uint so has to be positive!

x/stakeibc/keeper/msg_server.go
asalzmann: What prevents users from calling this function (by setting msg.Authority)?

asalzmann (commented):
Change to add the gov function looks reasonable, although I don't fully follow how permissions work for this function.

What stops users from calling the function?

x/stakeibc/keeper/msg_server.go
shellvish: Gotcha, perfect then! We might have already done this, but should make sure in Dockernet (or localstride or another env) that Governance can call this function, but other accounts can't

x/stakeibc/keeper/msg_server.go
shellvish: Nice, resolving!

x/stakeibc/types/message_update_host_zone_params.go
shellvish: Ah good call, resolving!

x/stakeibc/keeper/unbonding.go
shellvish: Makes sense, I also didn't realize the check is run so close to here (line 468). On-board with this, resolving

x/stakeibc/keeper/msg_server.go
sampocs: yup, already tested in dockernet!

x/stakeibc/keeper/msg_server.go
sampocs: they can't set `msg.Authority` cause that would require them signing a tx from the gov module account

app/upgrades/v22/upgrades.go
sampocs: 111 for hub
56 for dydx 
86 for osmo

PR #1179, opened 2024-04-13
remove GetConnectionIdFromICAPort, use connection ID from host zone

sampocs:
## Context
There are a couple issues with how the connection ID is handled that prevents the re-registering of a host zone after it's been unregistered. This is primarily relevant for testnet, but also indicates a flaw in the current logic.

Namely the issues are:
* The function `GetConnectionIdFromICAPort` returns the first connection ID that matches the port ID - but there can be more than one connection with a given port ID
* The connection ID for the Delegation and SetWithdrawalAddress ICA's is pulled from the above function instead of from the connection ID on the host zone
* The OnChanOpenAck callback uses the above erroneous function which prevents registering ICAs on a new connection (since it will always match to the old expired connection)

## Brief Changelog
* Removed `GetConnectionIdFromICAPort`
* Used `ConnnectionId` from host zone in Delegation and SetWithdrawalAddress ICAs
* Used `GetConnectionFromChannel` in place of the removed function in the `OnChanOpenAck` callback


riley-stride:
Nice to get rid of those blocks of unnecessary code!

Do we need changes to ICAs that are submitted via submittxs to ICAs that are not the delegate acct? E.g. the ICA to the withdrawal acct that sweep rewards (submitted in FeeBalanceCallback) 

As long as we address that question and integration test this I'm signed off

riley-stride (approved):

sampocs:
> Do we need changes to ICAs that are submitted via submittxs to ICAs that are not the delegate acct? E.g. the ICA to the withdrawal acct that sweep rewards (submitted in FeeBalanceCallback)

@riley-stride sorry can you say more on this question / point to a specific example? 

I think all other references just grab the connection ID from the host zone (e.g. [fee balance callback](https://github.com/Stride-Labs/stride/blob/main/x/stakeibc/keeper/icqcallbacks_fee_balance.go#L88))

riley-stride:
> > Do we need changes to ICAs that are submitted via submittxs to ICAs that are not the delegate acct? E.g. the ICA to the withdrawal acct that sweep rewards (submitted in FeeBalanceCallback)
> 
> @riley-stride sorry can you say more on this question / point to a specific example?
> 
> I think all other references just grab the connection ID from the host zone (e.g. [fee balance callback](https://github.com/Stride-Labs/stride/blob/main/x/stakeibc/keeper/icqcallbacks_fee_balance.go#L88))

Resolved in Slack with 
"There’s only one connection ID per host zone - and each ICA channel belongs to that same connection. So the delegation and withdrawal ICA accounts/channels, share the same connection". 

I'm signed off.

asalzmann:
I thought each ICA had its own connection?

sampocs:
Each ICA has it's own _channel_ - and all channels are on the same connection. 

[example](https://stride-api.polkachu.com/ibc/core/channel/v1/connections/connection-0/channels)

x/stakeibc/keeper/ibc.go
```diff
@@ -17,11 +17,10 @@ import (
 )
 
 func (k Keeper) OnChanOpenAck(ctx sdk.Context, portID, channelID string) error {
-	// Lookup the ICA address and chainId from the port and connection
-	controllerConnectionId, found := k.GetConnectionIdFromICAPortId(ctx, portID)
-	if !found {
-		k.Logger(ctx).Info(fmt.Sprintf("portId %s has no associated ICA account", portID))
-		return nil
+	// Lookup connection ID, counterparty chain ID, and ICA address from the channel ID
+	controllerConnectionId, _, err := k.IBCKeeper.ChannelKeeper.GetChannelConnection(ctx, portID, channelID)
```
asalzmann, line 21: > The OnChanOpenAck callback uses the above erroneous function which prevents registering ICAs on a new connection (since it will always match to the old expired connection)

Was the host zone connection updated?

asalzmann (commented):
> The function GetConnectionIdFromICAPort returns the first connection ID that matches the port ID - but there can be more than one connection with a given port ID

What's the scenario where there are two connections associated with an ICA port ID? Is it due to migrating the connection on an active host zone?

x/stakeibc/keeper/delegation.go
```diff
@@ -10,26 +10,12 @@ import (
...
 
-	icatypes "github.com/cosmos/ibc-go/v7/modules/apps/27-interchain-accounts/types"
-
 	"github.com/Stride-Labs/stride/v21/utils"
 	recordstypes "github.com/Stride-Labs/stride/v21/x/records/types"
 	"github.com/Stride-Labs/stride/v21/x/stakeibc/types"
 )
 
 func (k Keeper) DelegateOnHost(ctx sdk.Context, hostZone types.HostZone, amt sdk.Coin, depositRecord recordstypes.DepositRecord) error {
-	// TODO: Remove this block and use connection-id from host zone
```
shellvish, line 21: This change feels very reasonable, but do you know why we did it this way at all? I think `connection-id`'s always been on the host_zone struct?

shellvish (approved):
This looks great to me! This is a great catch, it's quite an edge case. 

I'm surprised we use the `port_id->connection_id` lookup when the `connection_id` is so readily available, but on-board with changing it now.

x/stakeibc/keeper/delegation.go
sampocs: I think it's just a bit of tech debt! this was probably some of the first code written haha

My guess at what happens is:
*  We first wrote the `GetConnectionIdFromICAPort` function specifically for OnChanOpenAck 
    * maybe because weren't familiar with the IBC functions / didn't know you could map from channel ID to connection ID?
    *  I _think_ I used to believe that channel IDs were **not** globally unique and that they were only unique to each connection. Possible others thought this as well
* Then when were were writing functions like `DelegateOnHost`, `GetConnectionIdFromICAPort` seemed like the easiest helper to get what we wanted

Just speculating! But I've had it on my radar to fix this for awhile (hence the TODOs) and this was the first time I realized the lookup was actually incorrect!

sampocs:
> What's the scenario where there are two connections associated with an ICA port ID? Is it due to migrating the connection on an active host zone?

Yeah the use case would be if a light client expired for a host and we had to update the connection. Hope to god we never have to worry about this on mainnet, but it already happened on testnet with dydx unfortunately :(

x/stakeibc/keeper/ibc.go
sampocs: yeah the specific use casa here was for dydx on testnet. The LC expired so we had to make new clients/connections

But even when the new connection ID was passed to register host zone, the lookup in OnChanOpenAck was grabbing the old ICAs on the old channels

x/stakeibc/keeper/keeper.go
```diff
@@ -108,17 +108,6 @@ func (k Keeper) GetAuthority() string {
 	return k.authority
 }
 
-// Searches all interchain accounts and finds the connection ID that corresponds with a given port ID
-func (k Keeper) GetConnectionIdFromICAPortId(ctx sdk.Context, portId string) (connectionId string, found bool) {
```
ethan-stride, line 112: For ICA controllers I see how the port is unique but what about for transfer channels where the port is always 'transfer'
How was this reverse lookup function working before?

ethan-stride (commented):
LGTM
Much simplified, these Mocks are going to help testing in the future because in unit tests I often had to do the exact type of reverse lookups you now have built in.

PR #1182, opened 2024-04-15
added event for redeem stake

sampocs:
## Context
Added events to redeem stake in stakeibc to match the event format that's used for stakedym/staketia. This will significantly simplify the indexing work that numia has to do.


shellvish (approved):
This looks great, thank you for doing this!

riley-stride (approved):

PR #1184, opened 2024-04-15
fixed ibc hooks store key

sampocs:
## Context
Contrary to most new modules, the store key for ibc hooks is not the module name


asalzmann (approved):
lgtm! nice catch

PR #1189, opened 2024-05-08
removed trade route dead code

sampocs:
## Context
Now that trades are executed off-chain, the ICA swapping code can be removed.

## Brief Changelog
* Deprecated `TradeConfig`
* Deprecated `trade_config` field in `TradeRoute` and added `min_transfer_amount` field (only field from `TradeConfig` that's still used
* Deprecated fields in `CreateTradeRoute` and `UpdateTradeRoute` txs
* Updated references of `MinSwapAmount` to `MinTransferAmount`
* Updated txs to not set deprecated fields
* Removed unused hook entry points 
  * `SwapAllRewardTokens` 
  * `UpdateAllSwapPrices`
* Removed unused hook business logic
  *`SwapRewardTokens`
  * `TradeRewardBalanceQuery` (initiates query of USDC)
  * `PoolPriceQuery`
* Removed unused ICQ callbacks 
  * `PoolPriceCallback` 
  * `TradeRewardBalanceCallback` (query of USDC in trade account)
* Fixed unit tests


proto/stride/stakeibc/trade_route.proto
```diff
@@ -84,7 +89,18 @@ message TradeRoute {
   // zone, back to the host zone. This is the channel ID on the trade zone side
   string trade_to_host_channel_id = 11;
 
+  // Minimum amount of reward token that must be accumulated before
+  // the tokens are transferred to the trade ICA
+  string min_transfer_amount = 13 [
```
ethan-stride, line 94: I'm confused why this wasn't here before? Were we using a variable which was on the trade_config like min_swap amount and so now this transfer specific functionality needs a new variable to track it?

ethan-stride (approved):
LGTM
So many lines of code removed. Wow.  Much simpler design now

proto/stride/stakeibc/trade_route.proto
sampocs: Yeah exactly. We just had a MinSwap and we used it both for the transfer and the actual swap

but since this deprecates the config, I just figured we’d pull it out and rename it to it’s actual use

sampocs: Yeah i dont think any need for a minimum in the other direction

PR #1195, opened 2024-05-20
batched undelegations

sampocs:
## Context 
Base PR for the batched undelegation work. This PR consists of only proto changes, but subsequent PR's will be merged into this before main.

### New Fields on HZU
Now that the StToken and NativeToken amounts are decremented in different places, I think it makes more sense to add explicit fields for each purpose. In this vein, I added the following:

`NativeTokensToUnbond`
  * purpose is so that if we have failed batches, we know how much we have left that needs to be unbonded
  * set right before the ICA is submitted
  * decremented in the callback

`StTokensToBurn`
  * purpose is to track tokens that need to be burned 
  * set right before ICA is submitted
  * decremented in callback

`ClaimableNativeTokens`
  * purpose is to track tokens that have been claimed
  * set right before redemption sweep ICA is submitted
  * decremented during each user claim

`UndelegationTxsInProgress`
* purpose is to prevent submitting a retry until we know all txs for a given epoch have been completed
* otherwise, it would could attempt to unbond with funds that were already unbonded (happens if the ack comes back after the epoch)
* This is incremented when each ICA is submitted and decremented in the callback

### New Status Field
Additionally, we need to differentiate EUR's that are in a retry state vs those that are new unbondings. For this, I added the new status field `UNBONDING_RETRY_QUEUE`

### Updated Callback Data
Finally, the callback data was updated to also store the number of stTokens. Given the fact that there's some precision error when breaking down a total stToken amount across validators, I think there are two options:
* Option 1: Only store native amount in callback, handle precision error by checking if the stAmount is below some epsilon, and if, so, rounding it down to 0
* Option 2: When building the ICA messages, capture the stToken remainder on the last validator and store it in the callback data

I opted for option 2 as it seems less error prone.


riley-stride (approved):
looks good, ty for the detailed description

ethan-stride (approved):
LGTM, thank you for explaining new fields.
As the github alerts point out, this change to the protos means that this PR alone has some inconsistent types now -- like in icacallback_undelegate on line 48 it still refers to the old field called SplitDelegations instead of the new SplitUndelegation but I assume that these are fixed in the later PRs which actually update the callbacks

proto/stride/records/records.proto
```diff
@@ -57,9 +57,13 @@ message HostZoneUnbonding {
   enum Status {
     // tokens bonded on delegate account
     UNBONDING_QUEUE = 0;
+    // unbonding ICA has been submitted
```
shellvish, line 60: Thank you for putting comments in this, long overdue :slightly_smiling_face:

proto/stride/records/records.proto
```diff
@@ -72,6 +76,19 @@ message HostZoneUnbonding {
...
   ];
+  string st_tokens_to_burn = 8 [
+    (gogoproto.customtype) = "github.com/cosmos/cosmos-sdk/types.Int",
+    (gogoproto.nullable) = false
+  ];
+  string native_tokens_to_unbond = 9 [
+    (gogoproto.customtype) = "github.com/cosmos/cosmos-sdk/types.Int",
+    (gogoproto.nullable) = false
+  ];
+  string claimable_native_tokens = 10 [
```
shellvish, line 87: Do you think we need to store this, vs just using the last unbonding time in `unbonding_time` (i.e. assume all native tokens aren't claimable until the last unbonding finishes)? 

Also possible this is described in more detail in another PR, in which case please ignore!

proto/stride/records/records.proto
```diff
@@ -72,6 +76,19 @@ message HostZoneUnbonding {
...
+    (gogoproto.customtype) = "github.com/cosmos/cosmos-sdk/types.Int",
+    (gogoproto.nullable) = false
+  ];
+  uint64 undelegation_txs_in_progress = 11;
```
shellvish, line 91: Just curious, do we need this variable if we're already storing `native_tokens_to_unbond` and `claimable_native_tokens` / `unbonding_time`? 

Even if it's not used, seems reasonable to me to store it for inspectability!

shellvish (approved):
Looks great, thanks a lot for doing this! Will drop more questions if they come up as I go through the other PRs, but this looks great

proto/stride/records/records.proto
sampocs: Good question! We definitely don't need it, but I figured that now that we have different rules around decrementing these amounts, it's better to just be explicit (at the cost of adding an additional field).

Feels much more likely to footgun ourselves if we start having different implicit rules around when a field should be set or when it's used as a counter

proto/stride/records/records.proto
sampocs: Yeah so this was for an edge case that came up kind of late. The idea is if one of the undelegation batches lands on the host zone in time, but we don't get the ack back until after the epoch, we need to make sure not to resubmit the  undelegation for that batch. 

So, although I don't love adding this field, I think adding a semaphore like this is the easiest way to build a locking mechanism to make sure everything comes back before we retry. Open to other ideas if you have em though!

sampocs: Example:
* EUR with 1000 tokens
* Submit two batches, first for 400, then for 600
* First batch lands and acks, we decrement the EUR to 600
* Next batch lands but ack is still pending
* Next epoch kicks off, we need to make sure we don't submit an undelegation for 600 again

PR #1197, opened 2024-05-20
batched undelegations - batch helper

sampocs:
Split out the batching code into it's own function so it can have a dedicated unit test


x/stakeibc/keeper/unbonding.go
```diff
@@ -309,6 +309,61 @@ func (k Keeper) GetUnbondingICAMessages(
...
+			SplitUndelegations:      unbondingsBatch,
+			EpochUnbondingRecordIds: epochUnbondingRecordIds,
+		}
+		callbackArgsBz, err := proto.Marshal(&undelegateCallback)
+		if err != nil {
+			return numTxsSubmitted, errorsmod.Wrap(err, "unable to marshal undelegate callback args")
+		}
+
+		// Submit the undelegation ICA
+		if _, err := k.SubmitTxsDayEpoch(
```
ethan-stride, line 343: Now that we are batching these together a particular undelegation ICA can contain msgs from different records, all of these will succeed or fail together because a level or two up the whole call to UnbondFromHostZone is inside of a utils.ApplyFuncIfNoError -- so if there is any hostZoneRecord which has a persistent problem unbonding, this batching will now mean all will be blocked correct?  There is no way that part of a hostZoneRecord can succeed in unbonding while other messages from the same hostZoneRecord fail, all will succeed or fail together?

x/stakeibc/keeper/unbonding.go
```diff
@@ -309,6 +309,61 @@ func (k Keeper) GetUnbondingICAMessages(
...
+			ICACallbackID_Undelegate,
+			callbackArgsBz,
+		); err != nil {
+			return numTxsSubmitted, errorsmod.Wrapf(err, "unable to submit unbonding ICA for %s", hostZone.ChainId)
+		}
+		numTxsSubmitted += 1
+
+		// flag the delegation change in progress on each validator
+		for _, unbonding := range unbondingsBatch {
+			if err := k.IncrementValidatorDelegationChangesInProgress(&hostZone, unbonding.Validator); err != nil {
```
ethan-stride, line 357: Thinking through this: because when building these messages we are combining the total amounts needed to be undelegated before we run the algorithm which chooses which validators to undelegate from -- it still should be the case that across the whole batch we are only sending one message to each validator at most correct?  There should not be a case where the same validator gets more than one of these unbonding messages?

x/stakeibc/keeper/unbonding.go
```diff
@@ -400,57 +455,28 @@ func (k Keeper) UnbondFromHostZone(ctx sdk.Context, hostZone types.HostZone) (er
...
+		undelegateBatchSize,
+	)
+	if err != nil {
 		return err
 	}
 
+	// Update the epoch unbonding record status and number of undelegation ICAs
+	for epochNumber, hostZoneUnbonding := range epochNumberToHostZoneUnbonding {
+		hostZoneUnbonding.Status = recordstypes.HostZoneUnbonding_UNBONDING_IN_PROGRESS
+		hostZoneUnbonding.UndelegationTxsInProgress += numTxsSubmitted
```
ethan-stride, line 473: Now that things are batched, this numTxsSubmitted isn't really specific to the this particular hostZoneUnbonding record but I see how it will function largely the same, I assume in the callback the plan is to decrement this number to ensure that all relevant Txs have finished and since all will succeed or fail together, I guess this is still reasonable to apply it here on each hostZoneUnbonding record but it did cause me to pause on reading it the first time.

ethan-stride (approved):
LGTM
After seeing all the callback logic, I want to do another pass because then the whole picture will be clearer but this looks good for now.  Awesome work!

x/stakeibc/keeper/unbonding.go
sampocs: so the ApplyFuncIfNoError is wrapped around the ICA _submission_ - so there's atomicity in that either all ICAs will fire or none of them will

But once they all fire, we no longer have atomicity with the responses. That's where certain batches can succeed and others can fail (hence why we have the retry logic and why those tokens are pooled with the new unbondings next epoch).

x/stakeibc/keeper/unbonding.go
sampocs: yup that's right!

x/stakeibc/keeper/unbonding.go
sampocs: yeah, this is where it's a bit messy to have messages overlap records. But the main reason for this is so that we know to only initiate a retry when there's no more outstanding messages

Otherwise you could have a scenario where the last batch lands on the host zone in time (so there's no timeout) but the ack doesn't come back until after the next epoch. In this case we would accidentally re-submit that failed batch

ethan-stride: Good looking out for that edge case.  In general I think these semaphores are the more robust way to handle the inherent race conditions without having to think through every case -- definitely a good pattern I think we had done during LSM the first time that I noticed it in the codebase.

x/stakeibc/keeper/unbonding.go
ethan-stride: This is the super key info.  As you had previously clarified on slack:

- Every epoch we send 20 batches of 5 undelegate messages, for a total of 100 messages
- This means each ICA tx contains 5 messages
- Submission for all batches/txs/messages is atomic
- Ack for each message within a tx/batch is atomic
- Ack for different txs/batches are not atomic

So importantly individual batches can fail or succeed separate from each other even though the whole batch itself will have the same fate.  I'm going to do another pass with this in mind since this is so crucial to the whole design.

x/stakeibc/keeper/unbonding.go
shellvish, line 357: Just checking, is it right that there's no non-critical reason this should fail?

x/stakeibc/keeper/unbonding_test.go
```diff
@@ -1149,6 +1149,83 @@ func (s *KeeperTestSuite) TestGetUnbondingICAMessages() {
 	}
 }
 
+func (s *KeeperTestSuite) TestBatchSubmitUndelegateICAMessages() {
```
shellvish, line 1152: Very comprehensive test! 

I don't think there are any meaningful error cases we need to check (the only errors I can think of is the HZ / connection being misconfigured), but just double checking that with you

shellvish (approved):
This looks great! Thanks a lot for splitting out this function, `UnbondFromHostZone` is very unwieldy as-is. This is outside the scope of this PR, but just making a mental note that it might be good for us to do the same in the future with some of the other logic in `UnbondFromHostZone`

x/stakeibc/keeper/unbonding.go
sampocs: that's right! it would only fail if the validator doesn't exist which shouldn't be possible

x/stakeibc/keeper/unbonding_test.go
sampocs: It's a good callout, it definitely wouldn't hurt to add a failure case test

sampocs: done!

PR #1198, opened 2024-05-21
batched undelegations - undelegation ICA callback

sampocs:
## Context
This PR updates the undelegation callback to support batching. The previous implementation of this callback would burn all the tokens associated with a record in the callback. This implementation updates the callback to only burn as many as were in the ICA batch.

This accounting update is done in a cascading fashion starting from the earliest epoch unbonding record. That way, if there's a failed unbonding, a future successful unbonding can cover the portion from the failed record so that it can move to the next status. 

## Logical Changelog
* Regardless of the ack status, decrements the `UndelegationTxsInProgress`
* In the case of a failure, the callback sets the record status to RETRY
* In the case of a successful ack, the unbonding record has the `StTokensToBurn` and `NativeTokensToUnbond` amounts decremented if the unbonding was successful, in a cascading fashion
* If the unbonding time from the message response is later than the time in the record, the record is updated 
* Reset `UndelegationChangesInProgress` on HZU to 0 after a timeout when the ICA channel is restored (since the timed out packet will never ack)

## Code Changelog
* Added function `MarkUndelegationAckReceived` to reset the `DelegationChangesInProgress` on the validators and the `UndelegationTxsInProgress` on the HZU
* Added function `CalculateTokensFromBatch` to sum up the total stTokens and native tokens from the callback data
* Updated `GetLastestCompletionTime` to return unix time instead of `time.Time`
* Updated `UpdateHostZoneUnbondings` to apply the logical changes above to the HZU's
* Renamed some of the helper functions since they're global in the keeper (e.g. `BurnTokens` → `BurnTokensAfterUndelegation`)
* Cleaned up existing unit tests and added test coverage


x/stakeibc/keeper/icacallbacks_undelegate.go
```diff
@@ -129,83 +161,103 @@ func (k Keeper) UpdateDelegationBalances(ctx sdk.Context, hostZone types.HostZon
...
+				"host zone unbonding not found for epoch %d and %s", epochNumber, chainId)
 		}
-		hostZoneUnbonding, found := k.RecordsKeeper.GetHostZoneUnbondingByChainId(ctx, epochUnbondingRecord.EpochNumber, chainId)
-		if !found {
-			errMsg := fmt.Sprintf("Host zone unbonding not found (%s) in epoch unbonding record: %d", chainId, epochNumber)
-			k.Logger(ctx).Error(errMsg)
-			return sdkmath.ZeroInt(), errorsmod.Wrapf(sdkerrors.ErrKeyNotFound, errMsg)
+
+		// Determine the amount to decrement from each record, flooring at 0
+		stTokensToBurn := sdkmath.MinInt(hostZoneUnbonding.StTokensToBurn, totalStTokensToBurn)
```
ethan-stride, line 220: Hold on, explain this logic for me in more detail please.
What if we have this situation and the middle batch2 is calledback first:  
- batch1 epochIds [1, 1, 1, 1, 1]
- batch2 epochids [1, 1, 2, 2, 2]
- batch3 epochids [2, 2, 2, 2, 2]
Then when processing epochid1 the totalStTokensToBurn would be the sum across the [1, 1, 2, 2, 2] and the hostZoneUnbonding.stTokensToBurn would be the sum across the [1, 1, 1, 1, 1] [1, 1] but for this you really only want the sum across [1, 1] the sum across [1, 1, 1, 1, 1] will be handled in a different callback...

x/stakeibc/keeper/icacallbacks_undelegate.go
```diff
@@ -129,83 +161,103 @@ func (k Keeper) UpdateDelegationBalances(ctx sdk.Context, hostZone types.HostZon
...
+		// Decrement the records
+		hostZoneUnbonding.StTokensToBurn = hostZoneUnbonding.StTokensToBurn.Sub(stTokensToBurn)
+		hostZoneUnbonding.NativeTokensToUnbond = hostZoneUnbonding.NativeTokensToUnbond.Sub(nativeTokensUnbonded)
+
+		// Update the totals for the next loop
+		totalStTokensToBurn = totalStTokensToBurn.Sub(stTokensToBurn)
+		totalNativeTokensUnbonded = totalNativeTokensUnbonded.Sub(nativeTokensUnbonded)
+
+		// If there are no more tokens to burn after this batch, iterate the record to the next status
+		if hostZoneUnbonding.StTokensToBurn.IsZero() && hostZoneUnbonding.NativeTokensToUnbond.IsZero() {
```
ethan-stride, line 232: Is this where you would need to take into account the off by one rounding errors instead of absolute comparison to zero?

x/stakeibc/keeper/icacallbacks_undelegate.go
ethan-stride: I think you might need to pass in the splitundelegations which were part of the batch to compute this accurately not just the total sums but I could be missing something.

ethan-stride (commented):
LGTM except for the one crucial question around computing the stTokensToBurn and nativeTokensUnbonded per unbonding record in the UpdateHostZoneUnbondingsAfterUndelegation function.  I could be missing something but it feels like you need to pass in the actual splitUndelegations which were part of this specific batch to get these numbers instead of just the totals.

x/stakeibc/keeper/icacallbacks_undelegate.go
```diff
@@ -66,12 +64,12 @@ func (k Keeper) UndelegateCallback(ctx sdk.Context, packet channeltypes.Packet,
...
 			icacallbackstypes.AckResponseStatus_FAILURE, packet))
 
-		// Reset unbondings record status
+		// Set the unbonding status to RETRY
 		if err := k.RecordsKeeper.SetHostZoneUnbondingStatus(
 			ctx,
 			chainId,
 			undelegateCallback.EpochUnbondingRecordIds,
-			recordstypes.HostZoneUnbonding_UNBONDING_QUEUE,
+			recordstypes.HostZoneUnbonding_UNBONDING_RETRY_QUEUE,
```
ethan-stride, line 72: Checking my understanding: In the case of this unbonding record going back into the retry queue there is no need to re-marshal and save the batch data because those particular split Undelegations and batching only mattered for this particular ICA submission.  When this record gets retried the whole validator assignment and batching flow will happen again from scratch and new splitUndelegations will be made for that future batching correct?

x/stakeibc/keeper/icacallbacks_undelegate.go
```diff
@@ -81,36 +79,70 @@ func (k Keeper) UndelegateCallback(ctx sdk.Context, packet channeltypes.Packet,
...
 	if err != nil {
 		k.Logger(ctx).Error(fmt.Sprintf("UndelegateCallback | %s", err.Error()))
 		return err
 	}
 
-	// Burn the stTokens
-	stTokenBurnAmount, err := k.UpdateHostZoneUnbondings(ctx, *latestCompletionTime, chainId, undelegateCallback)
-	if err != nil {
+	// Update the accounting on the host zone unbondings
+	if err := k.UpdateHostZoneUnbondingsAfterUndelegation(
```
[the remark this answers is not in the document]
sampocs: the whole callback function is already atomic - if an error is thrown, the whole tx that submitted the ack fails and everything is reverted

x/stakeibc/keeper/icacallbacks_undelegate.go
```diff
@@ -129,83 +161,103 @@ func (k Keeper) UpdateDelegationBalances(ctx sdk.Context, hostZone types.HostZon
...
+
+		// If there are no more tokens to burn after this batch, iterate the record to the next status
+		if hostZoneUnbonding.StTokensToBurn.IsZero() && hostZoneUnbonding.NativeTokensToUnbond.IsZero() {
+			hostZoneUnbonding.Status = recordstypes.HostZoneUnbonding_EXIT_TRANSFER_QUEUE
 		}
 
-		// Keep track of the stTokens that need to be burned
-		stTokenAmount := hostZoneUnbonding.StTokenAmount
-		stTokenBurnAmount = stTokenBurnAmount.Add(stTokenAmount)
+		// Update the unbonding time if the time from this batch is later than what's on the record
```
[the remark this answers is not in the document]
sampocs: yup that's right

x/stakeibc/keeper/icacallbacks_undelegate.go
```diff
@@ -81,36 +79,70 @@ func (k Keeper) UndelegateCallback(ctx sdk.Context, packet channeltypes.Packet,
...
 	}
 
-	// Burn the stTokens
-	stTokenBurnAmount, err := k.UpdateHostZoneUnbondings(ctx, *latestCompletionTime, chainId, undelegateCallback)
-	if err != nil {
+	// Update the accounting on the host zone unbondings
+	if err := k.UpdateHostZoneUnbondingsAfterUndelegation(
+		ctx,
+		chainId,
+		undelegateCallback.EpochUnbondingRecordIds,
```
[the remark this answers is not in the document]
sampocs: since the messages span multiple records, this will actually always be multiple. 

In a normal case, we'd have 4 epoch unbonding records here. You can see the excalidraw for a visual, but for example, we might have:

EUR1, EUR2, EUR3, EUR4 
val1, val2, val3, .... , val100

And a batch might just be val1 ... val5, but it will have EUR 1,2,3, and 4 in the callback data

x/stakeibc/keeper/icacallbacks_undelegate.go
sampocs: so instead, you would just have [1,2] in the callback data for each batch. The batch will always reference all EUR's that were included in the total sum of native tokens that needed to be unbonded

I'm not sure if that perfectly answer your question, but if you're still confused after we can hop on a quick call

x/stakeibc/keeper/icacallbacks_undelegate.go
sampocs: this is actually from your point in the other PR! since we have bended redemption rates across records, there's no guarantee that they will zero out at the same time, so we have to wait for both to be zero before we iterate the record. 

[Here's](https://github.com/Stride-Labs/stride/pull/1198/files#diff-b534601426c987137850288cf1ff0c31eec9ba0127489956e0399974b15dcf33R582-R584) a unit test that cover's this situation

x/stakeibc/keeper/icacallbacks_undelegate.go
sampocs: yup exactly - the callback data from this batch will be deleted as soon as the ack completes. And then new callback data will be created in the retry

x/stakeibc/keeper/icacallbacks_undelegate.go
```diff
@@ -81,36 +79,70 @@ func (k Keeper) UndelegateCallback(ctx sdk.Context, packet channeltypes.Packet,
...
-		return err
+	return nil
+}
+
+// Regardless of failure/success/timeout, indicate that this ICA has completed on each validator
+// on the host zone, and on the epoch unbonding record
+func (k Keeper) MarkUndelegationAckReceived(ctx sdk.Context, hostZone types.HostZone, undelegateCallback types.UndelegateCallback) error {
+	// Indicate that this ICA has completed on each validator
+	for _, splitDelegation := range undelegateCallback.SplitUndelegations {
+		if err := k.DecrementValidatorDelegationChangesInProgress(&hostZone, splitDelegation.Validator); err != nil {
```
riley-stride, line 125: To make sure I understand: this counter includes both Delegation and Undelegation changes - is that right?

x/stakeibc/keeper/icacallbacks_undelegate.go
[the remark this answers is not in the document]
riley-stride: > this will actually always be multiple

Could you point me to where this is explained in the Excalidraw? My understanding is that whether they come from one record vs multiple would depend on the sorting or capacity / unbalanced ratio.

x/stakeibc/keeper/icacallbacks_undelegate.go
riley-stride: I have the same question as Ethan.

>  The batch will always reference all EUR's that were included in the total sum of native tokens that needed to be unbonded

It seems like you're saying the CB from batch2 would process all 1s and all 2s but my understanding is that in this case we would only want to handle [1,1] and [2,2,2].

Are there unit tests for a case like this that I could read to help me understand?

riley-stride (commented):
Looking good. Only big picture question is the same as Ethan's.

x/stakeibc/keeper/icacallbacks_undelegate.go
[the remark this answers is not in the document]
sampocs: Screenshot 2024-05-23 at 11 39 08 AM

Setup: Two EUR's, 6 undelegation messages in batches of 2 per ICA

Notice here EUR1 and EUR2 are passed into the callback data _for every ICA_. And then in the callback it just cascades through each in order. So eventually EUR1 will be zero but we'll still loop over it, there will just be nothing to decrement

x/stakeibc/keeper/icacallbacks_undelegate.go
```diff
@@ -129,83 +161,112 @@ func (k Keeper) UpdateDelegationBalances(ctx sdk.Context, hostZone types.HostZon
...
 		}
-		hostZoneUnbonding, found := k.RecordsKeeper.GetHostZoneUnbondingByChainId(ctx, epochUnbondingRecord.EpochNumber, chainId)
-		if !found {
-			errMsg := fmt.Sprintf("Host zone unbonding not found (%s) in epoch unbonding record: %d", chainId, epochNumber)
-			k.Logger(ctx).Error(errMsg)
-			return sdkmath.ZeroInt(), errorsmod.Wrapf(sdkerrors.ErrKeyNotFound, errMsg)
+
+		// Determine the amount to decrement from the record, flooring at 0
+		// Also decrement the total for th next loop
+		nativeTokensUnbonded := sdkmath.MinInt(hostZoneUnbonding.NativeTokensToUnbond, totalNativeTokensUnbonded)
```
shellvish, line 221: Is the `MinInt` just here for redundancy? Could `totalNativeTokensUnbonded` ever be larger than `hostZoneUnbonding.NativeTokensToUnbond`?

x/stakeibc/keeper/icacallbacks_undelegate.go
shellvish: I could definitely be misunderstanding, but what's the case where `NativeTokensToUnbond > 0` and we don't want to iterate the record?

x/stakeibc/keeper/icacallbacks_undelegate.go
```diff
@@ -129,83 +161,112 @@ func (k Keeper) UpdateDelegationBalances(ctx sdk.Context, hostZone types.HostZon
...
-		// Update the bonded time
-		hostZoneUnbonding.UnbondingTime = cast.ToUint64(latestCompletionTime.UnixNano())
-		updatedEpochUnbondingRecord, success := k.RecordsKeeper.AddHostZoneToEpochUnbondingRecord(ctx, epochUnbondingRecord.EpochNumber, chainId, hostZoneUnbonding)
+		// If there are no more tokens to burn after this batch, iterate the record to the next status
+		if hostZoneUnbonding.StTokensToBurn.IsZero() && hostZoneUnbonding.NativeTokensToUnbond.IsZero() {
+			hostZoneUnbonding.Status = recordstypes.HostZoneUnbonding_EXIT_TRANSFER_QUEUE
+		}
+
+		// Update the unbonding time if the time from this batch is later than what's on the record
+		if unbondingTime > hostZoneUnbonding.UnbondingTime {
```
shellvish, line 246: Great catch here with this edge case!

shellvish (approved):
This looks great, thanks a ton for doing this Sam!

x/stakeibc/keeper/icacallbacks_undelegate.go
```diff
@@ -81,36 +79,70 @@ func (k Keeper) UndelegateCallback(ctx sdk.Context, packet channeltypes.Packet,
...
+	return nil
+}
+
+// Regardless of failure/success/timeout, indicate that this ICA has completed on each validator
+// on the host zone, and on the epoch unbonding record
+func (k Keeper) MarkUndelegationAckReceived(ctx sdk.Context, hostZone types.HostZone, undelegateCallback types.UndelegateCallback) error {
+	// Indicate that this ICA has completed on each validator
+	for _, splitDelegation := range undelegateCallback.SplitUndelegations {
+		if err := k.DecrementValidatorDelegationChangesInProgress(&hostZone, splitDelegation.Validator); err != nil {
+			return err
```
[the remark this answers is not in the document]
sampocs: any error will get logged upstream in the icacallback module!  Lemme know if you think I should add some more context to the message though

In general the pattern we have in stakeibc where we log and then return the error is kind of an antipattern from the early days where we were being extra cautious. Everything should be setup now so that you can just return an error and then it will get logged by the calling function!

x/stakeibc/keeper/icacallbacks_undelegate.go
sampocs: It will actually quite often be greater than hostZoneUnbonding.NativeTokensToUnbond because there's multiple records!

E.g. Two EUR, each with 1000 native
Total would be 2000
In the first loop, `MinInt` would set `nativeTokensUnbonded` to `min(1000, 2000) = 1000` and then only decrement the record by 1000

x/stakeibc/keeper/icacallbacks_undelegate.go
sampocs: After completing the first batch, the `NativeTokensToUnbond` will likely be still greater than 0, so we want to keep the record in `IN_PROGRESS` until enough batches complete that it decrements all the tokens on that record

x/stakeibc/keeper/icacallbacks_undelegate.go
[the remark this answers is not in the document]
riley-stride: Consider this resolved on my end

x/stakeibc/keeper/icacallbacks_undelegate.go
riley-stride: Thanks that helps. Consider this resolved from my end.

x/stakeibc/keeper/icacallbacks_undelegate.go
sampocs: yep that's right

x/stakeibc/keeper/icacallbacks_undelegate.go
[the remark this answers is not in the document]
shellvish: Ah gotcha, this makes sense! I'm onboard with not adding additional print statements then

x/stakeibc/keeper/icacallbacks_undelegate.go
shellvish: AH yes of course. Signed off on this then!

PR #1199, opened 2024-05-21
batch undelegations - redemption sweep

sampocs:
## Context
Updated redemption sweep to use `ClaimableNativeTokens` instead of `NativeTokens`

## Brief Changelog
* Updated redemption callback to set `ClaimableNativeTokens` at the same time that the status is updated to `Claimable`
* Updated the claim callback to decrement `ClaimableNativeTokens` 
* Fixed a bug in CleanupEpochUnbondingRecords so that it only removes records that are in status Claimable
* Updated unit tests

### DIFF
--- a/x/stakeibc/keeper/icacallbacks_claim.go
+++ b/x/stakeibc/keeper/icacallbacks_claim.go
@@ -100,7 +100,7 @@ func (k Keeper) DecrementHostZoneUnbonding(ctx sdk.Context, userRedemptionRecord
 	}
 
 	// decrement the hzu by the amount claimed
-	hostZoneUnbonding.NativeTokenAmount = hostZoneUnbonding.NativeTokenAmount.Sub(userRedemptionRecord.NativeTokenAmount)
+	hostZoneUnbonding.ClaimableNativeTokens = hostZoneUnbonding.ClaimableNativeTokens.Sub(userRedemptionRecord.NativeTokenAmount)
 
 	// save the updated hzu on the epoch unbonding record
 	epochUnbondingRecord, success := k.RecordsKeeper.AddHostZoneToEpochUnbondingRecord(ctx, callbackArgs.EpochNumber, callbackArgs.ChainId, hostZoneUnbonding)
--- a/x/stakeibc/keeper/records.go
+++ b/x/stakeibc/keeper/records.go
@@ -60,7 +60,7 @@ func (k Keeper) CreateEpochUnbondingRecord(ctx sdk.Context, epochNumber uint64) 
 }
 
 // Deletes any epoch unbonding records that have had all unbondings claimed
-func (k Keeper) CleanupEpochUnbondingRecords(ctx sdk.Context, epochNumber uint64) bool {
+func (k Keeper) CleanupEpochUnbondingRecords(ctx sdk.Context, epochNumber uint64) {
 	k.Logger(ctx).Info("Cleaning Claimed Epoch Unbonding Records...")
 
 	for _, epochUnbondingRecord := range k.RecordsKeeper.GetAllEpochUnbondingRecord(ctx) {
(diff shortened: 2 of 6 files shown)


x/stakeibc/keeper/records.go
```diff
@@ -70,18 +70,15 @@ func (k Keeper) CleanupEpochUnbondingRecords(ctx sdk.Context, epochNumber uint64
 		for _, hostZoneUnbonding := range hostZoneUnbondings {
 			// if an EpochUnbondingRecord has any HostZoneUnbonding with non-zero balances, we don't delete the EpochUnbondingRecord
 			// because it has outstanding tokens that need to be claimed
-			if !hostZoneUnbonding.NativeTokenAmount.Equal(sdkmath.ZeroInt()) {
+			notClaimable := hostZoneUnbonding.Status != recordstypes.HostZoneUnbonding_CLAIMABLE
+			hasUnclaimedTokens := !hostZoneUnbonding.ClaimableNativeTokens.Equal(sdkmath.ZeroInt())
```
ethan-stride, line 74: Do you have to take into account the off by one error again here or is it always in the direction to short the user, never to leave an extra token behind so the absolute compare to zero is fine?

x/stakeibc/keeper/icacallbacks_claim.go
```diff
@@ -100,7 +100,7 @@ func (k Keeper) DecrementHostZoneUnbonding(ctx sdk.Context, userRedemptionRecord
 	}
 
 	// decrement the hzu by the amount claimed
-	hostZoneUnbonding.NativeTokenAmount = hostZoneUnbonding.NativeTokenAmount.Sub(userRedemptionRecord.NativeTokenAmount)
+	hostZoneUnbonding.ClaimableNativeTokens = hostZoneUnbonding.ClaimableNativeTokens.Sub(userRedemptionRecord.NativeTokenAmount)
```
ethan-stride, line 103: So the big question I am still trying to grasp to see the whole picture is how this number gets accurately updated for a partial unbonding which has to be retried at different redemption rates when the the two parts succeed.  My current understanding is that the first partial success happens at RR1, then the hostZoneUnbonding record gets picked up again from the RETRY_QUEUE when we are now at RR2, because we haven't yet assigned which users get which tokens, I think the update user records would modify all of these numbers so that the expected native amounts would be in line with RR2 right?  So we end up short tokens overall because RR2 > RR1?

If that is correct, you can still fix this issue by recomputing the implied RR for the whole hostZoneUnbonding record at the time when it finishes all of its parts successfully if you know the total stTokens it was tracking and you know the total native stake which got unbonded at the different times, then you proportionally hand out those realized unbonded tokens proportional to the stTokens in the connected user records.

If I'm off in my understanding can you walk through how this per user native token amount would be correct at this point for a partial retry at different Redemption Rates?

ethan-stride (commented):
Looks good but one big question remains for me which is really about the pre-computation of the per user native token amounts when there is a retry.  It is perhaps more related to the other PRs but this is where that value is used.  I might be missing some core piece and this is already fine but I think the current thing would end up short tokens?  Please show me where my understanding is wrong

Awesome work on this whole project overall though, insanely complicated when you look at all the pieces together and impressively fast turn around.

x/stakeibc/keeper/records.go
sampocs: To clarify the off by one error - it actually happens entirely on the host zone and it's materialized in a difference between our internal representation of the host zone and the ground truth.

To walk through an example:
* Let's say we intend to unbond 100uatom
* We have 100uatom stored on our EURs 
* We submit an undelegation for 100uatom
* The hub receives the message for 100uatom and processes it
* In our ack, we decrement whatever we need to by 100uatom
* 21 days later, only 99uatom land in our delegation ICA account after the unbonding finishes
* On stride, we still think we unbonded 100uatom <- here's where the discrepancy is

So this is resolved by two things:
1. We send a little extra atom to our delegation account when we register a host zone so that when we try and "sweep" 100uatom after the unbonding, it's able to grab the missing 1uatom from our initial funding
2. We submit slash/calibration queries that will update the total delegation on the host zone and validators to whatever the true value is (as determined by the query)

x/stakeibc/keeper/icacallbacks_claim.go
sampocs: To your first questions: the host zone unbonding is not updated with the latest redemption rate when it retries - it just uses whatever the RR was when the unbonding was first attempted

But this whole PR is downsteam of all that. At this point we can assume that we've unbonded all tokens for the record, _and_ that the unbonding has finished. So we can just use set the `ClaimableNativeTokens` to the records original `NativeTokenAmount` (that was set in the "refresh" function before the unbonding)

For additional context, we previously decremented `NativeTokenAmount` to track claims and then when an unbonding record finally had 0 native tokens, we would remove the record (since all users claimed). This is functionally the exact same except the field has been renamed to make it more explicit

x/stakeibc/keeper/records.go
riley-stride: If it's truly always off by 1utoken and unavoidable as part of the SDK, funding the delegation account with a little extra at the start seems like a good solution - covers us for 1 million tries!

x/stakeibc/keeper/icacallbacks_claim.go
sampocs: discussed offline, will revisit after the audit

PR #1200, opened 2024-05-23
batch undelegations - sort by unbalance ratio

sampocs:
## Context
Changed the sorting back to being by unbalance ratio now that we can support multiple batches.

This is how the sorting was originally done as it's more fair. We had changed it temporarily to avoid the 32 message constraint.


shellvish:
Looks great! For what it's worth, I think the main advantage of using the UnbalancedRatio is it helps simplify the final delegations Stride has (e.g. if any validators become zero weight, we immediately undelegate from them). It also feels a bit more fair to me too, I'm on board to reverting to this algorithm!

PR #1203, opened 2024-05-26
batch undelegations - moved st burn calculation

sampocs:
## Context
In each undelegate callback, we have to know the relevant number of stTokens to burn _for that batch_. This requires us to calculate the relevant stToken amounts that correspond with each validator undelegation message. This PR moved the calculation out of the message submission and into the callback.

See this [comment](https://github.com/Stride-Labs/stride/pull/1196#discussion_r1609541509) for the main motivation.

TLDR: With the original implementation, you can end up with a case where the `NativeTokensToUnbond` goes to zero on the record, but the `StTokensToBurn` is non-zero, due to there being a blended redemption rate across new and retry HZU records.

While there is value in doing the full stToken calculation up front in the undelegation submission (only need to do the calculation once, gracefully handles precision error, displayed more prominently, etc.), the fact that we could have HZU records in states that are hard to reason about wound up being the trump card.

This new approach will no longer store the stToken amounts for each validator in the callback data, but will instead calculate it in each callback using the RR of a specific HZU record that is being adjusted. See #1204 for implementation. 

## Brief Changelog
* Removed stToken field from callback data
* Removed validator-level stToken burn calculation in ICA submission
* Updated unit tests


x/stakeibc/keeper/unbonding.go
```diff
@@ -67,17 +67,13 @@ func (k Keeper) GetQueuedHostZoneUnbondingRecords(
...
-func (k Keeper) GetTotalUnbondAmount(
-	hostZoneUnbondingRecords map[uint64]recordstypes.HostZoneUnbonding,
-) (totalNativeAmount sdkmath.Int, totalStAmount sdkmath.Int) {
-	totalNativeAmount = sdk.ZeroInt()
-	totalStAmount = sdk.ZeroInt()
+// that should be unbonded
+func (k Keeper) GetTotalUnbondAmount(hostZoneUnbondingRecords map[uint64]recordstypes.HostZoneUnbonding) sdkmath.Int {
+	totalNativeAmount := sdk.ZeroInt()
 	for _, hostZoneRecord := range hostZoneUnbondingRecords {
 		totalNativeAmount = totalNativeAmount.Add(hostZoneRecord.NativeTokensToUnbond)
```
ethan-stride, line 74: Just to confirm, if there are partially handled records in the incoming hostZoneUnbondingRecords, this field will be only the remaining amount from previous failures -- it will already have been reduced in the callbacks which succeeded by the successfully handled amount and in order to have the RETRY_QUEUE status, all callbacks have to have resolved succeed or fail but there can't be any still outstanding... correct?

x/stakeibc/keeper/unbonding.go
```diff
@@ -393,14 +372,13 @@ func (k Keeper) UnbondFromHostZone(ctx sdk.Context, hostZone types.HostZone) (er
 		return err
 	}
 
-	// Sum the total number of native tokens that from the records above that are ready to unbond
-	totalNativeUnbondAmount, totalStBurnAmount := k.GetTotalUnbondAmount(epochNumberToHostZoneUnbonding)
+	// Sum the total number of native tokens from the records above that are ready to unbond
+	totalUnbondAmount := k.GetTotalUnbondAmount(epochNumberToHostZoneUnbonding)
```
ethan-stride, line 376: The name epochNumberToHostZoneUnbonding confused me because it sounded singular -- could we do epochNumbersToHostZoneUnbondings for clarity?

ethan-stride: Still could be useful to call it totalNativeUnbondAmount just for clarity IMO

ethan-stride (commented):
LGTM
Much simplified overall

x/stakeibc/keeper/unbonding.go
sampocs: yeah first part is right about the amount potentially being partial. RETRY_QUEUE is set as soon as a single batch fails

x/stakeibc/keeper/unbonding.go
sampocs: good call, updated

riley-stride (commented):
Lgtm

shellvish:
LGTM, I don't think I fully understand this part `the fact that we could have HZU records in states that are hard to reason about wound up being the trump card`, but fully believe it's true! A bit unfortunate that we couldn't keep doing the full stToken calculation up front, but this approach looks great

PR #1204, opened 2024-05-26
batched undelegations - move st token burn calculation callback

sampocs:
## Context
Counterpart to #1203 

## Brief Changelog
* Used the total native and total st amounts on the HZU record to get an implied redemption rate
* Uses the implied redemption rate to figure out how many tokens to burn for each batch
* Once the HZU has 0 native remaining, stToken amount is also dropped to 0 (to handle any precision error from the RR equation)
* Updated unit tests


x/stakeibc/keeper/icacallbacks_undelegate.go
```diff
@@ -204,31 +202,42 @@ func (k Keeper) UpdateHostZoneUnbondingsAfterUndelegation(
...
 
-		// If there are no more tokens to burn after this batch, iterate the record to the next status
+		// Calculate the relative stToken portion using the implied RR from the record
+		// If the native amount has already been decremented to 0, just use the full stToken remainder
+		// from the record to prevent any precision error
+		var stTokensToBurn sdkmath.Int
+		if hostZoneUnbonding.NativeTokensToUnbond.IsZero() {
+			stTokensToBurn = hostZoneUnbonding.StTokensToBurn
+		} else {
+			impliedRedemptionRate := sdk.NewDecFromInt(hostZoneUnbonding.NativeTokenAmount).Quo(sdk.NewDecFromInt(hostZoneUnbonding.StTokenAmount))
```
ethan-stride, line 232: NativeTokenAmount never changes and is the original full amount before any unbondings while NativeTokensToUnbond  is the tracker on the record which is expected to change right?  Zero check is good and probably short circuits this almost all of the time.

x/stakeibc/keeper/icacallbacks_undelegate.go
```diff
@@ -204,31 +202,42 @@ func (k Keeper) UpdateHostZoneUnbondingsAfterUndelegation(
...
+			stTokensToBurn = sdk.NewDecFromInt(nativeTokensUnbonded).Quo(impliedRedemptionRate).TruncateInt()
```
ethan-stride, line 233: Whereas normally I worry about Int Truncates leaving us with unpredictable roundings, but beautifully the other case here should make it not a problem -- there will be remaining stTokens only in the case now when there is remaining native, and when this record gets finished off and native is zero, the other case sets the stTokensToBurn to exactly the stTokensToBurn so any previous rounding issues wouldn't matter anyway. Nice

ethan-stride (commented):
LGTM
I think this was simpler than I was expecting because of all the talk we had about we are really only moving the complexity to the callback -- I know that now we compute the stTokensToBurn here instead of before hand but it feels a lot more natural to do it separately for each EUR and it feels like not that many lines more in the callbacks for decent savings in the other part of the code.  Overall I like this solution a lot.

x/stakeibc/keeper/icacallbacks_undelegate.go
sampocs: yup, that's right

riley-stride (commented):
Are we sure we catch all cases in which Unbondings could fail? I'm concerned that if we don't catch them all, the calculated impliedRedemptionRate could be off, leading to incorrect token burns. I assume we do catch all such cases, just double checking!

sampocs:
@riley-stride can you say more on that?

x/stakeibc/keeper/icacallbacks_undelegate.go
```diff
@@ -204,31 +202,42 @@ func (k Keeper) UpdateHostZoneUnbondingsAfterUndelegation(
...
+		}
+
+		// Decrement st amount on the record and increment the tota
```
riley-stride, line 236: nit: total

riley-stride:
> @riley-stride can you say more on that?

Please disregard. I misunderstood the callback logic. Just re-read and it makes sense. I'm signed off!

x/stakeibc/keeper/icacallbacks_undelegate.go
sampocs: https://github.com/Stride-Labs/stride/pull/1223/commits/19e3617521bb1f7ad446d950ba5eedb363cc76a5

PR #1205, opened 2024-05-26
batch unbondings - cleanup

sampocs:
## Brief Changelog
* Updated `AddHostZoneToEpochUnbonding` to return an error instead of bool
* Called `SetHostZoneUnbonding` instead of `AddHostZoneToEpochUnbonding` + `SetEpochUnbondingRecord`
* Cleaned up errors

### DIFF
--- a/x/records/keeper/epoch_unbonding_record.go
+++ b/x/records/keeper/epoch_unbonding_record.go
@@ -9,8 +9,6 @@ import (
 
 	errorsmod "cosmossdk.io/errors"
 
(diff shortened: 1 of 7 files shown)


ethan-stride (commented):
LGTM
What is the philosophy between the two functions which take in a hostZoneUnbonding why one passes by pointer and the other doesn't?  Should they both pass it in by pointer?

sampocs:
Good observation, imo I think none of them should use pointers as they can lead to unnecessary side effects. I'll update

PR #1211, opened 2024-06-03
Consolidated stakeibc cli commands

sampocs:
## Context
Removing some ignite baggage, consolidating the cli functions down to one file. I'd recommend reviewing commit by commit to easily check the diffs


assafmo (approved):
Yeah that's much less cluttered!

PR #1213, opened 2024-06-04
staketia migration - disable staketia LS

sampocs:
During the stTIA migration, liquid stakes are disabled from staketia and will be handled in stakeibc. This PR removes the `LiquidStake` functionality from staketia


x/staketia/keeper/msg_server.go
```diff
@@ -25,12 +26,7 @@ var _ types.MsgServer = msgServer{}
 
 // User transaction to liquid stake native tokens into stTokens
 func (k msgServer) LiquidStake(goCtx context.Context, msg *types.MsgLiquidStake) (*types.MsgLiquidStakeResponse, error) {
-	ctx := sdk.UnwrapSDKContext(goCtx)
-	stToken, err := k.Keeper.LiquidStake(ctx, msg.Staker, msg.NativeAmount)
-	if err != nil {
-		return nil, err
-	}
-	return &types.MsgLiquidStakeResponse{StToken: stToken}, nil
+	return nil, errors.New("Liquid staking is no longer enabled in staketia")
```
shellvish, line 29: _very_ nit, should this message say "use stakeibc instead"?

shellvish (commented):
This looks great, thank you for doing this!

riley-stride:
Should we also pull LiquidStake out of:
- msgs.go
- msgs_test.go
A few other files, I see some mentions we could prob clean up when searching "LiquidStake" in x/staketia

riley-stride:
Otherwise this lgtm!

x/staketia/keeper/msg_server.go
sampocs: done!

sampocs:
@riley-stride we definitely could remove them from msgs but I was trying to keep the changes to a minimum and I don't think there's any harm in keeping them there since we'll eventually deprecate the whole module!

riley-stride:
> @riley-stride we definitely could remove them from msgs but I was trying to keep the changes to a minimum and I don't think there's any harm in keeping them there since we'll eventually deprecate the whole module!

Makes sense. Good point about deprecating the module. I'm signed off here.

PR #1214, opened 2024-06-04
staketia migration - staketia accounting

sampocs:
## Context
This PR handles the main accounting changes that will take place in the staketia module during the migration. At a high level, the main changes are:
* Staketia now reads the redemption rate from stakeibc
* The `DelegatedBalance` field on the host zone is now renamed to `RemainingDelegatedBalance` and is only meant to track the delegations left in the multisig account. 

## Context on Accounting
In staketia, the `DelegatedBalance` field previously was changed in the following stages:
* `LiquidStake` - no change
* `ConfirmDelegation` - incremented
* `RedeemStake` - no change
* `ConfirmUndelegation` - decremented

And in stakeibc, we similarly have the following with the `TotalDelegations` field on host zone
* `LiquidStake` - no change
* `DelegationCallback` - incremented
* `RedeemStake` - no change
* `UndelegationCallback` - decremented

After the migration, the replacement field `RemainingDelegatedBalance` in staketia will update as follows:
* `LiquidStake` - no change (obviously)
* `ConfirmDelegation` - incremented (only occurs in one delegation cycle after the upgrade - see phase 2.5 in spec)
* `RedeemStake` - **decremented** so that we know when to switch over after the last token's been redeemed
* `ConfirmUndelegation` - no change
  * However **`TotalDelegations` in stakeibc will be decremented**

## Accounting Changelog
Based on the above, there are the following changes:
* `ConfirmDelegation` now increments `RemainingDelegatedBalance` instead of `DelegatedBalance`
* `RedeemStake` (handled in separate PR)
* `ConfirmUndelegation` now decrements `TotalDelegations` instead of `DelegatedBalance`
* `AdjustDelegatedBalance` now references `RemainingDelegatedBalance` instead of `DelegatedBalance`

## Remaining Changelog
* Renamed `DelegatedBalance` to `RemainingDelegatedBalance` on host zone
* Removed redemption rate from host zone
* Removed code that updates the redemption rate, as well as the redemption rate invariant check
* Read redemption rate from stakeibc in relevant places


riley-stride:
How do we handle rate limits when migrating the accounting / RR?

x/staketia/keeper/hooks.go
```diff
@@ -23,20 +23,6 @@ func (k Keeper) BeforeEpochStart(ctx sdk.Context, epochInfo epochstypes.EpochInf
 	// Every day, refresh the redemption rate and prepare delegations
 	// Every 4 days, prepare undelegations
 	if epochInfo.Identifier == epochstypes.DAY_EPOCH {
-		// Update the redemption rate
-		// If this fails, do not proceed to the delegation or undelegation step
-		// Note: This must be run first because it is used when refreshing the native token
-		// balance in prepare undelegation
-		if err := k.UpdateRedemptionRate(ctx); err != nil {
```
riley-stride, line 30: My understanding is that the ordering of refreshing the RR and sending delegations and undelegations matters. The comments here suggest the same.

If we move the update RR step to stakeibc, but keep the delegation and undelegation txs in staketia, we just need to make sure the update RR is run at the same epochly cadence. 

Just to triple check - it is, right? I.e. the cadence at which we ran it here matches the cadence at which we run it in stakeibc?

x/staketia/keeper/events.go
```diff
@@ -90,7 +90,6 @@ func EmitHaltZoneEvent(ctx sdk.Context, hostZone types.HostZone) {
 		sdk.NewEvent(
 			types.EventTypeHostZoneHalt,
 			sdk.NewAttribute(types.AttributeKeyHostZone, hostZone.ChainId),
-			sdk.NewAttribute(types.AttributeKeyRedemptionRate, hostZone.RedemptionRate.String()),
```
riley-stride, line 93: Why not emit this? Import stakeibc and continue emitting.

x/staketia/keeper/unbonding.go
```diff
@@ -56,10 +60,10 @@ func (k Keeper) RedeemStake(ctx sdk.Context, redeemer string, stTokenAmount sdkm
 
 	// Estimate a placeholder native amount with current RedemptionRate
```
riley-stride, line 61: Doesn't `RedeemStake` need to decrement `RemainingDelegatedBalance`?

riley-stride (commented):
Looking really clean. PR description and spec helped a ton in the review, ty for making those to clear. Left a few questions.

x/staketia/keeper/invariants.go
```diff
@@ -24,7 +24,7 @@ func (k Keeper) HaltZone(ctx sdk.Context) {
 	stDenom := utils.StAssetDenomFromHostZoneDenom(hostZone.NativeTokenDenom)
 	k.ratelimitKeeper.AddDenomToBlacklist(ctx, stDenom)
 
-	k.Logger(ctx).Error(fmt.Sprintf("[INVARIANT BROKEN!!!] %s's RR is %s.", hostZone.GetChainId(), hostZone.RedemptionRate.String()))
+	k.Logger(ctx).Error(fmt.Sprintf("[INVARIANT BROKEN!!!] %s's RR is %s.", hostZone.GetChainId()))
```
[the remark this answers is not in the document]
sampocs: this function is no longer called since the RR is now in stakeibc. I just had to edit this line to avoid a lint error

x/staketia/keeper/events.go
sampocs: this function is no longer called from staketia

x/staketia/keeper/unbonding.go
sampocs: great eye! yes it should have been decremented - the auditors caught it as well! https://github.com/Stride-Labs/stride/pull/1223/commits/def6f4a65f2b7345774adbfc23f3d6b46e23385f

x/staketia/keeper/hooks.go
sampocs: GREAT question! The cadence will be different in stakeibc (it runs every stride epoch instead of every day epoch)

i think how we have it right now, the RR will be ~6 hours stale when its used for unbondings here. But imo, since that's pretty negligible,  I'd prefer to keep it as is, rather than mess with the ordering. Wdyt?

riley-stride: Right, I see. 

So just to map out the worst case: some packets may be are stuck or reinvestment would be otherwise delayed for a long time, then we'd see a large RR jump that triggers within the "stale period". In that case, we'd use the stale RR for unbonding and the users who receive those unbondings would not receive their yield from the "stuck packet" period. Is that right?

Just to explore syncing up the epochs.... what if we had staketia epochs run every 6 hours? I imagine not much would change in the 6-hourly logic until the operator advances the state. But in that case, would this edge case or the "stale period" be solved?

x/staketia/keeper/unbonding.go
riley-stride: Nice! That fix looks good to me.

riley-stride: Good stress test of the auditors (:

riley-stride:
> How do we handle rate limits when migrating the accounting / RR - is removing the checks from staketia enough, since ratelimiting covers stakeibc?

@sampocs i think this question is still outstanding

x/staketia/keeper/hooks.go
sampocs: Yeah I think that worst case is righ!, But I think it's pretty unlikely that we have some multi-day bottleneck that just happens to clear in this 6 hour window. And even in that case, the impact is low - they just get the RR at the time of redemption (similar to how it used to be done).

Running the epochs every 6 hours could work, but I'd be a bit worried we have some implicit invariant about the length and not sure it's worth the effort to investigate based on the above

sampocs:
We don't have any tia RLs at the moment. Also RLs are module agnostic!

x/staketia/keeper/hooks.go
riley-stride: > even in that case, the impact is low - they just get the RR at the time of redemption (similar to how it used to be done).

Yeah good point that it'd revert to how it used to be done. We did get some complaints about that, but as long as we don't have a long packet backlog we're good. And ultimately this migration period is temporary so the time window within which we could see backlogs is bounded. 

Fwiw I'm mostly convinced to leave things as is, especially because of "we [may] have some implicit invariant about the length" could lead to a serious accounting issue if overlooked.

sampocs: Yeah good point about this being temporary!

also I assume the complaints were more in the case where the unbondings were delayed right? I’d imagine w/o a delay, getting the same RR that’s shown in the FE at time of redemption wouldnt upset anyone!

riley-stride:
I'm signed off here. Leaving [this](https://github.com/Stride-Labs/stride/pull/1214#discussion_r1687048822) open for other reviewers to opine on.

PR #1215, opened 2024-06-07
staketia migration - redeem stake

sampocs:
## Context
The redeem stake code contains an edge case for handling the last redemption that's processed through staketia. In this case, staketia handles whatever is left in the delegation account, and pushes the remainder over to stakeibc.

**Note to reviewer**: I'd recommend viewing commit by commit!!

## Brief Changelog on `RedeemStake`
* When checking if the protocol has enough delegated to handle the redemption, checks the `TotalDelegations` from stakeibc (instead of the `DelegatedBalance` in staketia)
* Decrement `RemainingDelegatedBalance` after the redemption (see #1214 for more context on accounting)
* On the last redemption, handles spillover case by calling stakeibc and enables stakeibc redemptions
* Disabled redemptions once the multisig account has been drained
* Moved stakeibc's redeem stake out of message server so that it can be called from staketia
* Added the `RedemptionsEnabled` bool to the stakeibc host zone and a function that sets it to true
* Added unit tests


riley-stride:
> The redeem stake code contains an edge case for handling the last redemption that's processed through staketia. In this case, staketia handles whatever is left in the delegation account, and pushes the remainder over to stakeibc.

I think if we wanted to, we could simplify this as follows: fail redemptions that redeems more than the exact remaining balance in staketia; assume someone will come along (prob will be a team member) to redeem the exact correct remaining amount.

x/stakeibc/keeper/redeem_stake.go
```diff
@@ -0,0 +1,156 @@
...
+	errorsmod "cosmossdk.io/errors"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
+
+	"github.com/Stride-Labs/stride/v22/utils"
+	recordstypes "github.com/Stride-Labs/stride/v22/x/records/types"
+	"github.com/Stride-Labs/stride/v22/x/stakeibc/types"
+)
+
+func (k Keeper) RedeemStake(ctx sdk.Context, msg *types.MsgRedeemStake) (*types.MsgRedeemStakeResponse, error) {
```
riley-stride, line 15: Note: I didn't diff this, but assuming it's a direct copy paste from the msg_server.

x/stakeibc/keeper/redeem_stake_test.go
```diff
@@ -0,0 +1,308 @@
...
+type RedeemStakeTestCase struct {
+	user                 Account
+	hostZone             stakeibctypes.HostZone
+	zoneAccount          Account
+	initialState         RedeemStakeState
+	validMsg             stakeibctypes.MsgRedeemStake
+	expectedNativeAmount sdkmath.Int
+}
+
+func (s *KeeperTestSuite) SetupRedeemStake() RedeemStakeTestCase {
```
riley-stride, line 28: Note: didn't diff this test but assuming it's a direct copy paste from the msg_server.

x/staketia/keeper/msg_server.go
```diff
@@ -29,7 +29,7 @@ func (k msgServer) LiquidStake(goCtx context.Context, msg *types.MsgLiquidStake)
 // User transaction to redeem stake stTokens into native tokens
 func (k msgServer) RedeemStake(goCtx context.Context, msg *types.MsgRedeemStake) (*types.MsgRedeemStakeResponse, error) {
 	ctx := sdk.UnwrapSDKContext(goCtx)
-	nativeToken, err := k.Keeper.RedeemStake(ctx, msg.Redeemer, msg.StTokenAmount)
+	nativeToken, err := k.Keeper.RedeemStake(ctx, msg.Redeemer, msg.Receiver, msg.StTokenAmount)
```
riley-stride, line 32: I don't understand the need for the receiver and how it connects to spillover. I also don't see us passing a Receiver into `NewMsgRedeemStake`.

x/staketia/keeper/unbonding.go
```diff
@@ -29,6 +35,12 @@ func (k Keeper) RedeemStake(ctx sdk.Context, redeemer string, stTokenAmount sdkm
 		return nativeToken, err
 	}
 
+	// If the remaining delegated balance for staketia is 0, that means we've undelegated
+	// all the stake in the MS account and the redemptions should be switched over to stakeibc
+	if hostZone.RemainingDelegatedBalance.LTE(sdkmath.ZeroInt()) {
+		return nativeToken, types.ErrRedemptionsDisabled
```
riley-stride, line 41: Where does `RedemptionsEnabled` get set to false?

x/staketia/keeper/unbonding.go
```diff
@@ -61,11 +73,37 @@ func (k Keeper) RedeemStake(ctx sdk.Context, redeemer string, stTokenAmount sdkm
...
+	// When checking if there's enough delegated TIA to handle the request,
+	// use the value from stakeibc instead of staketia
 	if nativeAmount.GT(stakeibcHostZone.TotalDelegations) {
 		return nativeToken, errorsmod.Wrapf(types.ErrUnbondAmountToLarge,
 			"cannot unstake an amount g.t. total staked balance: %v > %v", nativeAmount, stakeibcHostZone.TotalDelegations)
 	}
 
+	// Check if the requested unbonding is greater than what's in the multisig account
+	// If so, handle the spillover via stakeibc
+	if nativeAmount.GT(hostZone.RemainingDelegatedBalance) {
```
riley-stride, line 86: It would be nice to be able to clean this up and instead just assume we need a last redemption that matches the amount perfectly.

x/staketia/keeper/unbonding.go
```diff
@@ -102,6 +140,38 @@ func (k Keeper) RedeemStake(ctx sdk.Context, redeemer string, stTokenAmount sdkm
 	return nativeToken, nil
 }
 
+// Calls redeem stake for any requested redemption amount that exceeds what's in the staketia account
+// Returns the updated native and stTokens amounts that should be used in staketia
+func (k Keeper) HandleRedemptionSpillover(
```
riley-stride, line 145: https://github.com/Stride-Labs/stride/pull/1215/files#r1687073852

riley-stride (commented):
Looking good! Added a few comments. Main questions
1. Should we remove the spillover case?
2. Where do we set RedemptionsEnabled to false? If in the upgrade handler alone, what happens to redemptions submitted after we're at 0 remaining in staketia but before the upgrade

x/stakeibc/keeper/redeem_stake.go
sampocs: yeah [this](https://github.com/Stride-Labs/stride/pull/1215/commits/f8148e63a262d486f2b97c9a359e1dab9f626a52) commit was just a copy paste

x/stakeibc/keeper/redeem_stake_test.go
sampocs: yeah [this](https://github.com/Stride-Labs/stride/pull/1215/commits/f8148e63a262d486f2b97c9a359e1dab9f626a52) commit was just a copy paste

x/staketia/keeper/redemption_rate_test.go
```diff
@@ -71,7 +72,10 @@ func (s *KeeperTestSuite) TestCheckRedemptionRateExceedsBounds() {
 
 	for _, tc := range testCases {
 		s.Run(tc.name, func() {
-			s.App.StakeibcKeeper.SetHostZone(s.Ctx, tc.hostZone)
+			hostZone := tc.hostZone
+			hostZone.ChainId = types.CelestiaChainId
```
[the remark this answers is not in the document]
sampocs: cause `CheckRedemptionRateExceedsBounds` now looks up the celestia host zone from stakeibc

x/staketia/keeper/msg_server.go
sampocs: Great call! It was missing from `NewMsgRedeemStake`. Just added

And we need it cause stakeibc takes a receiver field but staketia did not originally

x/staketia/keeper/unbonding.go
sampocs: it's set to false in the upgrade handler

x/staketia/keeper/unbonding.go
sampocs: I don't have a strong opinion either way, but I think removing it leaves some messy edge cases and frontend complexity (e.g. what if a user tries to unbond a large amount). I think we discussed it with informal and they felt the same way

Here's the updated code [btw](https://github.com/Stride-Labs/stride/blob/07d8a838e664b4aa55859ab485f7a371ee6d7910/x/staketia/keeper/unbonding.go#L78)

I'd prefer to leave it as is cause I don't think removing it simplifies things _that_ much and we've already had it audited. But I could go either way

sampocs:
@riley-stride 

> what happens to redemptions submitted after we're at 0 remaining in staketia but before the upgrade

I'm not sure I follow the question. The upgrade is the first thing that happens so staketia wouldn't be at 0 until after the upgrade

x/staketia/keeper/msg_server.go
riley-stride: lgtm. resolved.

x/staketia/keeper/unbonding.go
riley-stride: yeah ig if it's already audited that's ok. don't think we'd hit the edge case because we could just submit the redemption ourselves when the account is low, but if it doesn't clean things up much maybe we just leave it.

sampocs: > don't think we'd hit the edge case because we could just submit the redemption ourselves when the account is low

I don't think we can just assume we'd be there to cover it. What if someone wants to unbond a large amount before the account gets low?

In that case, I think we'd have make changes to the frontend so that they unbond in two txs

riley-stride: Wouldn't the worst case there be that their first tx fails, then they have to wait until we unbond the right amount (which I think would trigger the switch to the new module), then they could unbond their full amount?

sampocs: Depends how we want to implement it. Imo that'd be a risky approach. What if they want to unbond $1M - we can't handle that ourselves, so they'd be unable to unbond for a while

I think the way we'd want to implement it is to have the frontend notice that they're trying to redeem more than the remaining balance, and submit the first tx for just that amount, and then prompt them to unbond the remainder

riley-stride: It's a good point about the large unbond. And the FE. I'm convinced. Resolving.

riley-stride:
> @riley-stride
> 
> > what happens to redemptions submitted after we're at 0 remaining in staketia but before the upgrade
> 
> I'm not sure I follow the question. The upgrade is the first thing that happens so staketia wouldn't be at 0 until after the upgrade

That clears it up. I'm signed off here.

PR #1216, opened 2024-06-07
staketia migration - upgrade handler function

sampocs:
## Context
This PR adds a function `InitiateMigration` that will be called in the upgrade handler when beginning the migration from staketia to stakeibc.

## Brief Changelog
* Registered the host zone in stakeibc
* Moved the redemption rate from the staketia host zone to stakeibc
* Set the total delegations on stakeibc to be the staketia delegated balance + sum of delegation records
* Transferred funds in the staketia deposit account to the stakeibc deposit account
* Transferred funds in the staketia fee account to the stakeibc fee account


x/staketia/keeper/migration.go
```diff
@@ -0,0 +1,159 @@
...
+	sdkmath "cosmossdk.io/math"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+
+	recordtypes "github.com/Stride-Labs/stride/v22/x/records/types"
+	stakeibctypes "github.com/Stride-Labs/stride/v22/x/stakeibc/types"
+	oldtypes "github.com/Stride-Labs/stride/v22/x/staketia/legacytypes"
+	"github.com/Stride-Labs/stride/v22/x/staketia/types"
+)
+
+// TODO [UPGRADE HANDLER]: Migrate stakeibc host zone (set redemptions enabled to true on each host zone)
```
riley-stride, line 16: Good flat - should we add this in here or is it tracking in a separate PR?

x/staketia/keeper/migration.go
```diff
@@ -0,0 +1,159 @@
...
+func (k Keeper) UpdateStakeibcHostZone(ctx sdk.Context, legacyHostZone oldtypes.HostZone) (stakeibctypes.HostZone, error) {
+	// Grab the newly created stakeibc host zone
+	stakeibcHostZone, found := k.stakeibcKeeper.GetHostZone(ctx, types.CelestiaChainId)
+	if !found {
+		return stakeibctypes.HostZone{}, errors.New("celestia host zone not found in stakeibc after registration")
+	}
+
+	// Disable redemptions and set the redemption rate to the one from stakeibc
+	stakeibcHostZone.RedemptionsEnabled = false
+	stakeibcHostZone.RedemptionRate = legacyHostZone.LastRedemptionRate
```
riley-stride, line 41: Why `LastRedemptionRate` instead of `RedemptionRate`?

x/staketia/keeper/migration.go
```diff
@@ -0,0 +1,159 @@
...
+		pendingDelegations = pendingDelegations.Add(delegationRecord.NativeAmount)
+	}
+	stakeibcHostZone.TotalDelegations = legacyHostZone.DelegatedBalance.Add(pendingDelegations)
+	k.stakeibcKeeper.SetHostZone(ctx, stakeibcHostZone)
+
+	return stakeibcHostZone, nil
+}
+
+// Migrates the protocol owned accounts (deposit and fee) to their stakeibc counterparts
+func (k Keeper) MigrateProtocolOwnedAccounts(
```
riley-stride, line 58: This feels particularly sensitive. Any way we can add some safeguards, or do you feel good about it?

e.g. record balance in sender addr before, compare to diff in receiver balance after and revert if not equal

riley-stride: Maybe compute RR using sender accounts, then send, then recompute using receiver balances and revert if RRs don't match.

x/staketia/keeper/migration.go
sampocs: I believe it was handled separately. Updated code is [here](https://github.com/Stride-Labs/stride/pull/1223/files#diff-8a0bb7f321cfb197fe6e15baf17d0db0ecd70e00f4d983d726f7764b1f620a19)

x/staketia/keeper/migration.go
sampocs: It should be `RedemptionRate`. It must have been [fixed](https://github.com/Stride-Labs/stride/pull/1223/files#diff-e29b0f57bc104470ceb763d46fdec16efae60e2708e7e98957c6eb060bd82f21R45-R49) at some point after this PR

x/staketia/keeper/migration.go
sampocs: Good question!

> record balance in sender addr before, compare to diff in receiver balance after and revert if not equal

I don't think doing checks like this add any value - cause they basically just add more code to sanity check that the bank module worked correctly

> Maybe compute RR using sender accounts, then send, then recompute using receiver balances and revert if RRs don't match.

This is an interesting idea. My only concern is we could accidentally footgun ourselves and stall our chain if there's actually a small, but acceptable discrepancy. I can take a pass at this and see what it would look like

Fwiw though, I feel pretty good about it and we'll end up doing some extensive testing around the RR before we ship

sampocs: Just took a pass here. Lmk what you think. It's a lot of new code which makes me nervous but it is an important invariant so I think I lean towards keeping it but no strong opinion.

https://github.com/Stride-Labs/stride/pull/1223/commits/13288dd3bbb330a2dc26c57a8995d83ed1b45120
https://github.com/Stride-Labs/stride/pull/1223/commits/ec24683e337be4651a0766e499662aa9dac34df2
https://github.com/Stride-Labs/stride/pull/1223/commits/046f992eb5c09e65a9b7da31f4c4adeef0d102f5

riley-stride: The first two PRs seem unrelated - are you sure they're adding sanity checks?

On the third, I echo your concern about footguns with small but acceptable discrepancies in code like [this](https://github.com/Stride-Labs/stride/commit/046f992eb5c09e65a9b7da31f4c4adeef0d102f5#diff-e29b0f57bc104470ceb763d46fdec16efae60e2708e7e98957c6eb060bd82f21R183-R184) but can't think of a reason why they should exist. 

Would be curious to hear opinions from @shellvish @asalzmann @assafmo as well.

sampocs: > The first two PRs seem unrelated - are you sure they're adding sanity checks?

Those are the individual commits. We had to refine the function to calculate each RR so I separated commits such that you could see the copy paste first and then see the relevant diffs

PR #1220, opened 2024-06-13
Unit Tests for stakeibc CLI

assafmo:


x/stakeibc/client/cli/tx.go
```diff
@@ -329,7 +329,16 @@ func CmdAddValidators() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "add-validators [host-zone] [validator-list-file]",
 		Short: "Broadcast message add-validators",
-		Args:  cobra.ExactArgs(2),
+		Long: strings.TrimSpace(
```
sampocs, line 332: thanks for adding this!

x/stakeibc/client/cli/query_test.go
```diff
@@ -0,0 +1,45 @@
+package cli_test
+
+import (
+	"testing"
+)
+
+func TestQueryParams(t *testing.T) {
```
sampocs, line 7: we can probably just remove this file if all these are empty

x/stakeibc/client/cli/tx_test.go
```diff
@@ -0,0 +1,308 @@
+package cli_test
```
sampocs, line 1: any reason to not include successful test cases?

x/stakeibc/client/cli/tx_test.go
```diff
@@ -0,0 +1,308 @@
...
+	// no input parsing to test
+}
+
+func TestCmdUpdateValidatorSharesExchRate(t *testing.T) {
+	// no input parsing to test
+}
+
+func TestCmdCalibrateDelegation(t *testing.T) {
+	// no input parsing to test
+}
```
sampocs, line 232: same thing here, I feel like we may as well just remove these

sampocs (commented):
looks great, and a the speedup is a game-changer!!

Was there any reason you only included error cases? I assume it was just cause the function would fail downstream anyways, but still feels like it'd make sense to include the success cases

x/stakeibc/client/cli/tx_test.go
assafmo: Done

x/stakeibc/client/cli/tx_test.go
assafmo: Done 74971a040

x/stakeibc/client/cli/query_test.go
assafmo: Done cfd6422

x/stakeibc/client/cli/tx_test.go
sampocs: hmm yeah, that's a good point. I was thinking the tx submission would at least not through an error (even if the code was non-zero). But I think that would only be if you had a network stood up without mocked state. I imagine with no network it will fail much earlier

One other option would be to, in the success case, just catch whatever exception is thrown during the broadcast step (to confirm it gets at least that far).

I'll defer to you!

assafmo: I'll play with it a bit

PR #1221, opened 2024-06-14
Replace `docker-compose` with `docker compose`

assafmo:
From July 2023 Compose V1 (`docker-compose`) stopped receiving updates. It’s also no longer available in new releases of Docker.

Compose V2 (`docker compose`), which was first released in 2020, is included with all currently supported versions of Docker. It offers an improved CLI experience, improved build performance with BuildKit, and continued new-feature development.

For more info see https://docs.docker.com/compose/migrate/


sampocs (approved):
long overdue, thank you!

PR #1223, opened 2024-06-20
batching delegations/undelegations + stTIA migration

sampocs:
## Context
We have a fixed number of undelegation messages that we can submit in a single tx for each chain due to gas limits. In order to be able to support a validator set of 100+ validators, we need to batch up the undelegation ICAs.

One motivating problem: We can only have 7 undelegations at one time on each validator. In particular, upon failure, we cannot re-unbond from the same validator again.

Another way to phrase this: for each unbonding record, we can only unbond one time per validator

## Brief Changelog
* Batched delegations #1201 
* Batched undelegations #1195 
* Misc audit fixes


x/stakeibc/keeper/redeem_stake.go
```diff
@@ -0,0 +1,153 @@
...
+		return nil, errorsmod.Wrapf(sdkerrors.ErrInvalidAddress, "invalid receiver address (%s)", err)
+	}
+
+	// construct desired unstaking amount from host zone
+	stDenom := types.StAssetDenomFromHostZoneDenom(hostZone.HostDenom)
+	nativeAmount := sdk.NewDecFromInt(msg.Amount).Mul(hostZone.RedemptionRate).TruncateInt()
+
+	if nativeAmount.GT(hostZone.TotalDelegations) {
+		return nil, errorsmod.Wrapf(types.ErrInvalidAmount, "cannot unstake an amount g.t. staked balance on host zone: %v", msg.Amount)
+	}
```
assafmo, line 57: nit, better to fail early

```suggestion
	if nativeAmount.GT(hostZone.TotalDelegations) {
		return nil, errorsmod.Wrapf(types.ErrInvalidAmount, "cannot unstake an amount g.t. staked balance on host zone: %v", msg.Amount)
	}

	// construct desired unstaking amount from host zone
	stDenom := types.StAssetDenomFromHostZoneDenom(hostZone.HostDenom)
	nativeAmount := sdk.NewDecFromInt(msg.Amount).Mul(hostZone.RedemptionRate).TruncateInt()
```

x/stakeibc/keeper/redeem_stake.go
```diff
@@ -0,0 +1,153 @@
...
+
+	// ensure the recipient address is a valid bech32 address on the hostZone
+	_, err = utils.AccAddressFromBech32(msg.Receiver, hostZone.Bech32Prefix)
+	if err != nil {
+		return nil, errorsmod.Wrapf(sdkerrors.ErrInvalidAddress, "invalid receiver address (%s)", err)
+	}
+
+	// construct desired unstaking amount from host zone
+	stDenom := types.StAssetDenomFromHostZoneDenom(hostZone.HostDenom)
+	nativeAmount := sdk.NewDecFromInt(msg.Amount).Mul(hostZone.RedemptionRate).TruncateInt()
```
assafmo, line 53: nit: I'd maybe rename `nativeAmount` to `baseDenomAmount`

x/stakeibc/keeper/redeem_stake.go
```diff
@@ -0,0 +1,153 @@
...
+	// safety checks on the coin
+	// 	- Redemption amount must be positive
+	if !nativeAmount.IsPositive() {
+		return nil, errorsmod.Wrapf(sdkerrors.ErrInvalidCoins, "amount must be greater than 0. found: %v", msg.Amount)
+	}
+	// 	- Creator owns at least "amount" stAssets
+	balance := k.bankKeeper.GetBalance(ctx, sender, stDenom)
+	if balance.Amount.LT(msg.Amount) {
+		return nil, errorsmod.Wrapf(sdkerrors.ErrInvalidCoins, "balance is lower than redemption amount. redemption amount: %v, balance %v: ", msg.Amount, balance.Amount)
+	}
```
assafmo, line 75: unnecessary check, if this fails then later on `SendCoins()` will fail with `"spendable balance %s is smaller than %s"` in https://github.com/cosmos/cosmos-sdk/blob/217e6419be/x/bank/keeper/send.go#L254

x/stakeibc/keeper/redeem_stake.go
```diff
@@ -0,0 +1,153 @@
...
+	// ----------------- UNBONDING RECORD KEEPING -----------------
+	// Fetch the record
+	redemptionId := recordstypes.UserRedemptionRecordKeyFormatter(hostZone.ChainId, epochTracker.EpochNumber, msg.Receiver)
+	userRedemptionRecord, userHasRedeemedThisEpoch := k.RecordsKeeper.GetUserRedemptionRecord(ctx, redemptionId)
+	if userHasRedeemedThisEpoch {
+		k.Logger(ctx).Info(fmt.Sprintf("UserRedemptionRecord found for %s", redemptionId))
+		// Add the unbonded amount to the UserRedemptionRecord
+		// The record is set below
+		userRedemptionRecord.StTokenAmount = userRedemptionRecord.StTokenAmount.Add(msg.Amount)
+		userRedemptionRecord.NativeTokenAmount = userRedemptionRecord.NativeTokenAmount.Add(nativeAmount)
```
assafmo, line 86: question: why add to an existing record instead of creating a record per action?

this seems to complicate the code

x/stakeibc/keeper/unbonding.go
```diff
@@ -408,40 +369,36 @@ func (k Keeper) UnbondFromHostZone(ctx sdk.Context, hostZone types.HostZone) err
...
-	//   (as if we were to unbond and then rebalance)
-	// This will serve as the starting point for determining how much to unbond each validator
-	// first, get the total delegations _excluding_ validators with slash_query_in_progress
+	// Determine the total eligible unbond amount - excluding delegations to validators with a slash query in progress
 	totalValidDelegationBeforeUnbonding := sdkmath.ZeroInt()
 	for _, validator := range hostZone.Validators {
 		if !validator.SlashQueryInProgress {
 			totalValidDelegationBeforeUnbonding = totalValidDelegationBeforeUnbonding.Add(validator.Delegation)
 		}
 	}
```
assafmo, line 397: IDK if related to the PR, but: What are the frequency and duration of a slash query? This might really mess up the rebalancing. Can we sequence it in such a way that once an epoch we do a slash query, then right after we unbond/rebalnce?

assafmo: Also true for https://github.com/Stride-Labs/stride/blob/c9b390dde/x/stakeibc/keeper/host_zone.go#L253-L259

x/stakeibc/keeper/redeem_stake.go
```diff
@@ -0,0 +1,153 @@
...
+	if nativeAmount.GT(hostZone.TotalDelegations) {
+		return nil, errorsmod.Wrapf(types.ErrInvalidAmount, "cannot unstake an amount g.t. staked balance on host zone: %v", msg.Amount)
+	}
+
+	// safety check: redemption rate must be within safety bounds
+	rateIsSafe, err := k.IsRedemptionRateWithinSafetyBounds(ctx, hostZone)
+	if !rateIsSafe || (err != nil) {
+		errMsg := fmt.Sprintf("IsRedemptionRateWithinSafetyBounds check failed. hostZone: %s, err: %s", hostZone.String(), err.Error())
+		return nil, errorsmod.Wrapf(types.ErrRedemptionRateOutsideSafetyBounds, errMsg)
+	}
```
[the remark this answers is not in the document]
sampocs: good call!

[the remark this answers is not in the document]
sampocs: err actually, I'm not sure the suggestion works cause `nativeAmount` is defined at the end but is used in the first if condition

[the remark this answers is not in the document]
sampocs: https://github.com/Stride-Labs/stride/pull/1223/commits/059843d5d25317d06acf8b2c6dd0b202853536fb

x/stakeibc/keeper/redeem_stake.go
sampocs: https://github.com/Stride-Labs/stride/pull/1223/commits/059843d5d25317d06acf8b2c6dd0b202853536fb

x/stakeibc/keeper/redeem_stake.go
sampocs: Good q! I think if we wanted to create a new record, we'd have to also add the block height to the record ID which is a bit messy imo.

I personally don't think combining records adds that much complexity (basically just these 3 lines) but open to suggestions if you had a better way of doing it!

x/stakeibc/keeper/unbonding.go
sampocs: So the slash query is a daisy chained query: first one determines if the validator was slashed, and, if so, the second one fires to get the magnitude of the slash. Only the second one is blocking. But the first one can be triggered permissionlessly, and can also be triggered by LSM liquid stakes.

i'm glad you called this out though - it's one of the most (unnecessarily) complex parts of our code base. But I think it's a tough problem cause slash queries can trigger at any time from LSM liquid stakes, and they're UX block when that does occur, so I think they need to take precedence. 

---

I think the best long term approach is actually to use ICQ from ibc-go which would simplify this a ton! We'd be able to just bundle the ICQ with the ICA and then we'd never have to worry about this overlap (here's a [doc](https://docs.google.com/document/d/1gpudfKI8mAfRM2Hix6Lkix8wJN2nzTa-BjHgyW12JMA/edit) for reference). 

Only blocker there though is having host chains [upgrade](https://docs.google.com/spreadsheets/d/18I813HXRCrz3IHq3F4dTAnIe7DaXCxYQWtPoJrCoSdI/edit?gid=0#gid=0)

x/stakeibc/keeper/redeem_stake.go
assafmo: Yeah, it's not worth it at this point. Any change to the ID format would require a state migration.

assafmo (approved):

sampocs:
closing in favor of #1263

PR #1224, opened 2024-06-20
added wasm light client

sampocs:
## Context
Added wasm light client to integrate with union bridge.

Referenced the following during the integration:
* ~[Wasm client integration guide](https://github.com/cosmos/ibc-go/blob/main/docs/docs/03-light-clients/04-wasm/03-integration.md)~ [Wasm client integration guide](https://ibc.cosmos.network/v7/ibc/light-clients/wasm/integration/)
* [Integrating light clients guide](https://github.com/cosmos/ibc-go/blob/main/docs/docs/01-ibc/02-integration.md#integrating-light-clients)
* [Osmosis integration](https://github.com/osmosis-labs/osmosis/pull/7637/files)

## Open Questions
* The [compatibility matrix ](https://github.com/cosmos/ibc-go/blob/main/docs/docs/03-light-clients/04-wasm/03-integration.md#importing-the-08-wasm-module) doesn't list ibc v7.4 - I used commit `7ee2a2452b79d0bc8316dc622a1243afa058e8cb` which is for v7.3
* The client router is not in ibc v7 so I excluded the following lines from app.go
```go
 clientRouter := app.IBCKeeper.ClientKeeper.GetRouter()
 wasmLightClientModule := wasm.NewLightClientModule(app.WasmClientKeeper)
 clientRouter.AddRoute(ibcwasmtypes.ModuleName, &wasmLightClientModule)
```

## Brief Changelog
* Added wasm client, sharing the same wasmVM as x/wasm
* Added the store key 
* Added the client name to `AllowedClients` in the upgrade handler

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -9,6 +9,7 @@ import (
 	"github.com/CosmWasm/wasmd/x/wasm"
 	wasmkeeper "github.com/CosmWasm/wasmd/x/wasm/keeper"
 	wasmtypes "github.com/CosmWasm/wasmd/x/wasm/types"
+	wasmvm "github.com/CosmWasm/wasmvm"
 	"github.com/Stride-Labs/ibc-rate-limiting/ratelimit"
 	ratelimitkeeper "github.com/Stride-Labs/ibc-rate-limiting/ratelimit/keeper"
 	ratelimittypes "github.com/Stride-Labs/ibc-rate-limiting/ratelimit/types"
@@ -88,6 +89,9 @@ import (
 	ibchooks "github.com/cosmos/ibc-apps/modules/ibc-hooks/v7"
 	ibchookskeeper "github.com/cosmos/ibc-apps/modules/ibc-hooks/v7/keeper"
 	ibchookstypes "github.com/cosmos/ibc-apps/modules/ibc-hooks/v7/types"
+	ibcwasm "github.com/cosmos/ibc-go/modules/light-clients/08-wasm"
+	ibcwasmkeeper "github.com/cosmos/ibc-go/modules/light-clients/08-wasm/keeper"
+	ibcwasmtypes "github.com/cosmos/ibc-go/modules/light-clients/08-wasm/types"
 	ica "github.com/cosmos/ibc-go/v7/modules/apps/27-interchain-accounts"
 	icacontroller "github.com/cosmos/ibc-go/v7/modules/apps/27-interchain-accounts/controller"
 	icacontrollerkeeper "github.com/cosmos/ibc-go/v7/modules/apps/27-interchain-accounts/controller/keeper"
@@ -227,6 +231,7 @@ var (
 		stakedym.AppModuleBasic{},
 		wasm.AppModuleBasic{},
 		ibchooks.AppModuleBasic{},
+		ibcwasm.AppModuleBasic{},
 	)
 
 	// module account permissions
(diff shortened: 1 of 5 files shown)


app/app.go
```diff
@@ -573,6 +593,15 @@ func NewStrideApp(
...
 	app.ContractKeeper = wasmkeeper.NewDefaultPermissionKeeper(app.WasmKeeper)
 
+	app.WasmClientKeeper = ibcwasmkeeper.NewKeeperWithVM(
+		appCodec,
+		keys[ibcwasmtypes.StoreKey],
+		app.IBCKeeper.ClientKeeper,
+		authtypes.NewModuleAddress(govtypes.ModuleName).String(),
+		wasmer,
+		app.GRPCQueryRouter(),
+	)
```
assafmo, line 603: What's the difference between [what Osmosis did](https://github.com/osmosis-labs/osmosis/pull/7503/files#diff-1ac782d2fab1d48ad7f7ef06cf2d58b37c99b6b4c98879a5ad86a985b73fad58R278-R286) and what you did here?

go.mod
```diff
@@ -16,6 +17,7 @@ require (
 	github.com/cosmos/gogoproto v1.4.10
 	github.com/cosmos/ibc-apps/middleware/packet-forward-middleware/v7 v7.1.3-0.20240228213828-cce7f56d000b
 	github.com/cosmos/ibc-apps/modules/ibc-hooks/v7 v7.0.0-20240403143657-8e64543c87e0
+	github.com/cosmos/ibc-go/modules/light-clients/08-wasm v0.1.2-0.20240412103620-7ee2a2452b79
```
assafmo, line 20: It compiles properly, so I'm okay with this. I guess the way to truly test this is with integration tests.

assafmo (approved):
Left a few comments, overall LGTM.

app/app.go
sampocs: > keys[ibcwasmtypes.StoreKey] / appKeepers.keys[ibcwasmtypes.StoreKey]

Osmosis just uses a different abstraction to manage keys/keepers/etc

> wasmer / ibcWasmConfig

I think the main difference here is they used a separate VM for ibc wasm (compared to x/wasm). [[docs](https://github.com/cosmos/ibc-go/blob/main/docs/docs/03-light-clients/04-wasm/03-integration.md#keeper-instantiation)] and [[osmosis note](https://github.com/osmosis-labs/osmosis/pull/7503/files#diff-1ac782d2fab1d48ad7f7ef06cf2d58b37c99b6b4c98879a5ad86a985b73fad58R278)]

go.mod
sampocs: are you referring to the question of whether the version is correct?

app/app.go
sampocs: also the IBC team just sent over this [better guide](https://ibc.cosmos.network/v7/ibc/light-clients/wasm/integration/) that's made for ibc v7

assafmo: > I think the main difference here is they used a separate VM

We do too

go.mod
assafmo: Whether hacking the commit hash like that is okay

sampocs: oh, it was done by `go get`

assafmo: Yeah I meant instead of getting an "official" tag

PR #1230, opened 2024-07-01
TESTNET: Fix wasmvm directory

assafmo:
In v22.1 we added the wasm IBC client. To not create two instances of wasmvm, we created it in `app.go` and passed it to `wasmkeeper.NewKeeper()` rather than have it create it inside. In doing so we accidentally passed the wrong directory to `wasmvm.NewVM()`.


sampocs (approved):
Great find!!

PR #1235, opened 2024-07-11
x/airdrop

assafmo:


proto/stride/airdrop/query.proto
```diff
@@ -0,0 +1,103 @@
...
+        "/Stride-Labs/stride/airdrop/user_allocations/{address}";
+  }
+
+  // Queries the state of an address for an airdrop (daily claim, claim & stake,
+  // upfront)
+  rpc UserClaimType(QueryUserClaimTypeRequest)
+      returns (QueryUserClaimTypeResponse) {
+    option (google.api.http).get =
+        "/Stride-Labs/stride/airdrop/user_claim_type/{id}/{address}";
+  }
```
sampocs, line 42: do you think we need a separate query for this? I was thinking we could maybe rework this into a `UserSummary` query that returns:
* Claim type 
* Amount claimed so far
* Amount remaining

Then that would be the main frontend query and we could use the `UserAllocations` more for auditing/introspection

proto/stride/airdrop/query.proto
```diff
@@ -0,0 +1,103 @@
...
+
+  // Queries all airdrop configurations
+  rpc Airdrops(QueryAirdropsRequest) returns (QueryAirdropsResponse) {
```
sampocs, line 45: nit
```suggestion
  rpc Airdrops(QueryAllAirdropsRequest) returns (QueryAllAirdropsResponse) {
```

proto/stride/airdrop/query.proto
```diff
@@ -0,0 +1,103 @@
...
+  uint64 id = 1;
+  string address = 2 [ (cosmos_proto.scalar) = "cosmos.AddressString" ];
+};
+message QueryUserClaimTypeResponse { ClaimType claim_type = 1; }
+
+// Airdrops
+message QueryAirdropsRequest {
+  cosmos.base.query.v1beta1.PageRequest pagination = 1;
+};
+message QueryAirdropsResponse {
```
sampocs, line 95:
```suggestion
message QueryAllAirdropsRequest {
  cosmos.base.query.v1beta1.PageRequest pagination = 1;
};
message QueryAllAirdropsResponse {
```

proto/stride/airdrop/query.proto
```diff
@@ -0,0 +1,103 @@
...
+  repeated AirdropRecord airdrop_config_records = 1
```
sampocs, line 96:
```suggestion
  repeated AirdropRecord airdrops = 1
```

proto/stride/airdrop/query.proto
```diff
@@ -0,0 +1,103 @@
...
+      [ (gogoproto.nullable) = false ];
+  cosmos.base.query.v1beta1.PageResponse pagination = 2;
+}
+
+// Airdrop
+message QueryAirdropRequest { uint64 id = 1; };
+message QueryAirdropResponse { AirdropRecord airdrop_config_record = 1; }
```
sampocs, line 103:
```suggestion
message QueryAirdropResponse { AirdropRecord airdrop = 1; }
```

proto/stride/airdrop/airdrop.proto
```diff
@@ -0,0 +1,96 @@
...
+    (gogoproto.customtype) = "github.com/cosmos/cosmos-sdk/types.Int"
+  ];
+
+  ClaimType claim_type = 5;
+}
+
+// AirdropRecord track the aggregate unbondings across an epoch
+message AirdropRecord {
+  // Airdrop ID
+  uint64 id = 1;
```
sampocs, line 70: should we use a string for this? e.g. airdrop name would be "sttia"

proto/stride/airdrop/tx.proto
```diff
@@ -0,0 +1,65 @@
...
+
+option go_package = "github.com/Stride-Labs/stride/v22/x/airdrop/types";
+
+enum OverwritableRecordType {
+  option (gogoproto.goproto_enum_prefix) = false;
+
+  RECORD_TYPE_DELEGATION = 0;
+  RECORD_TYPE_UNBONDING = 1;
+  RECORD_TYPE_REDEMPTION = 2;
+}
```
sampocs, line 18: what's this from?

proto/stride/airdrop/tx.proto
```diff
@@ -0,0 +1,65 @@
...
+
+  // User transaction to claim all the pending daily airdrop rewards
+  rpc Claim(MsgClaim) returns (MsgClaimResponse);
+
+  // User transaction to claim half of their total amount now, and forfeit the
+  // other half to be clawed back
+  rpc ClaimEarly(MsgClaimEarly) returns (MsgClaimEarlyResponse);
+
+  // User transaction to claim and automatically lock stake their whole airdrop
+  // amount now The funds can be unstaked by the user once the airdrop is over
```
sampocs, line 32:
```suggestion
  // amount now. The funds can be unstaked by the user once the airdrop is over
```

testutil/keeper/airdrop.go
```diff
@@ -0,0 +1,53 @@
+package keeper
```
sampocs, line 1: were you planning on using this testutil module? We don't really use it anymore so I think you can just scratch it unless you had other ideas!

tools/tools.go
```diff
@@ -0,0 +1,14 @@
+//go:build tools
```
sampocs, line 1: what's this for?

x/airdrop/keeper/keeper.go
```diff
@@ -0,0 +1,140 @@
...
+		bankKeeper:    bankKeeper,
+	}
+}
+
+// Logger returns a module-specific logger.
+func (k Keeper) Logger() log.Logger {
+	return k.logger.With("module", fmt.Sprintf("x/%s", types.ModuleName))
+}
+
+func (k Keeper) Claim(ctx sdk.Context, claimer string) error {
```
sampocs, line 51: imo it's a bit nicer to split these up into separate files.

`claim.go` for Claim, ClaimAndStake, ClaimEarly
`airdrop.go` for Airdrop keepers
`allocation.go` for user allocations

x/airdrop/keeper/keeper.go
```diff
@@ -0,0 +1,140 @@
...
+	}
+
+	// TODO implement logic
+
+	return nil
+}
+
+func (k Keeper) GetAirdropRecords(ctx sdk.Context) []types.AirdropRecord {
+	// TODO add pagination?
+	store := prefix.NewStore(ctx.KVStore(k.storeKey), types.AirdropRecordsKeyPrefix)
```
sampocs, line 86: nbd at all, but for future reference it's kinda nice to push these to a separate PR so comments are a little more manageable on each!

x/airdrop/keeper/msg_server.go
```diff
@@ -0,0 +1,59 @@
...
+	err := ms.Keeper.ClaimEarly(ctx, msg.Claimer)
+	if err != nil {
+		return nil, err
+	}
+
+	return &types.MsgClaimEarlyResponse{}, nil
+}
+
+// User transaction to claim and automatically lock stake their whole airdrop
+// amount now The funds can be unstaked by the user once the airdrop is over
```
sampocs, line 49:
```suggestion
// amount now. The funds can be unstaked by the user once the airdrop is over
```

x/airdrop/keeper/msg_server_test.go
```diff
@@ -0,0 +1,24 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+    keepertest "github.com/Stride-Labs/stride/v22/testutil/keeper"
+    "github.com/Stride-Labs/stride/v22/x/airdrop/types"
+    "github.com/Stride-Labs/stride/v22/x/airdrop/keeper"
+)
+
+func setupMsgServer(t testing.TB) (keeper.Keeper, types.MsgServer, context.Context) {
```
sampocs, line 14: would prefer to use the testing framework we use in the other modules but open to discuss it if you have a strong preference!

x/airdrop/keeper/query.go
```diff
@@ -0,0 +1,39 @@
...
+
+import (
+	"context"
+
+	"github.com/Stride-Labs/stride/v22/x/airdrop/types"
+)
+
+var _ types.QueryServer = Keeper{}
+
+func (k Keeper) AllAllocations(ctx context.Context, req *types.QueryAllAllocationsRequest) (*types.QueryAllAllocationsResponse, error) {
```
sampocs, line 11: nit: docstrings on these guis

x/airdrop/keeper/query.go
```diff
@@ -0,0 +1,39 @@
...
+	// TODO implement logic
+	return nil, nil
+}
+
+func (k Keeper) UserClaimType(ctx context.Context, req *types.QueryUserClaimTypeRequest) (*types.QueryUserClaimTypeResponse, error) {
+	// TODO implement logic
+	return nil, nil
+}
+
+func (k Keeper) Airdrops(ctx context.Context, req *types.QueryAirdropsRequest) (*types.QueryAirdropsResponse, error) {
```
sampocs, line 31:
```suggestion
func (k Keeper) AllAirdrops(ctx context.Context, req *types.QueryAirdropsRequest) (*types.QueryAirdropsResponse, error) {
```

x/airdrop/module.go
```diff
@@ -0,0 +1,235 @@
...
+	}
+}
+
+// ExportGenesis returns the module's exported genesis.
+func ExportGenesis(ctx sdk.Context, k keeper.Keeper) *types.GenesisState {
+	genesis := types.DefaultGenesis()
+	genesis.AirdropRecords = k.GetAirdropRecords(ctx)
+	genesis.AllocationRecords = k.GetAllocationRecords(ctx)
+
+	return genesis
```
sampocs, line 180: nit: imo it's a little nicer to move these to a genesis.go in the keeper

x/airdrop/module.go
```diff
@@ -0,0 +1,235 @@
...
+}
+
+// ----------------------------------------------------------------------------
+// App Wiring Setup
```
sampocs, line 184: is this SDK 50 stuff? haven't seen it before

x/airdrop/types/expected_keepers.go
```diff
@@ -0,0 +1,28 @@
+package types
+
+import (
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	authtypes "github.com/cosmos/cosmos-sdk/x/auth/types"
+	banktypes "github.com/cosmos/cosmos-sdk/x/bank/types"
+)
+
```
sampocs, line 8:
```suggestion
// TODO: trim these down to what's needed once module's complete
```

x/airdrop/types/keys.go
```diff
@@ -0,0 +1,38 @@
...
+// Generates a key byte prefix from a string
+func KeyPrefix(p string) []byte {
+	return []byte(p)
+}
+
+func AirdropRecordKeyPrefix(airdropId uint64) []byte {
+	return KeyPrefix(strconv.FormatUint(airdropId, 10))
+}
+
+func AllocationRecordKeyPrefix(airdropId uint64, userAddress string) []byte {
```
sampocs, line 36: imo we should use a string for the airdropId, but it's an open comment from above

x/airdrop/types/keys.go
```diff
@@ -0,0 +1,38 @@
...
+	ModuleName = "airdrop"
+
+	// StoreKey defines the primary module store key
+	StoreKey = ModuleName
+
+	// RouterKey defines the routing key
+	RouterKey = ModuleName
+
+	// Module Account for Fee Collection
+	FeeAddress = "staketia_fee_address"
```
sampocs, line 19: no need for this

x/airdrop/types/genesis_test.go
```diff
@@ -0,0 +1,42 @@
+package types_test
```
sampocs, line 1: you can scratch this for now until we implement validate genesis (since we'll probably use a different setup for that than what ignite used)

sampocs (commented):
Thanks for cranking this out so quickly!! Looks beautiful, all just nit comments!

proto/stride/airdrop/airdrop.proto
```diff
@@ -0,0 +1,96 @@
...
+  CLAIM_EARLY = 2;
+}
+
+// AllocationRecord tracks the status of an allocation for a user on a specific
+// airdrop
+message AllocationRecord {
+  // ID of the airdrop
+  uint64 airdrop_id = 1;
+
+  // Address of the account that is receiving the airdrop allocation
```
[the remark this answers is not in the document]
assafmo: Ohhh this is important! When we use AccAddress functions they might throw for non stride address, so we'll have to handle that with care and make sure this behavior is tested.

proto/stride/airdrop/airdrop.proto
```diff
@@ -0,0 +1,96 @@
...
+  google.protobuf.Timestamp claim_type_deadline_date = 5
+      [ (gogoproto.stdtime) = true ];
+
+  // Early claim penalty
+  string early_claim_penalty = 6 [
+    (gogoproto.customtype) = "github.com/cosmos/cosmos-sdk/types.Dec",
+    (gogoproto.nullable) = false
+  ];
+
+  // Distribution account
```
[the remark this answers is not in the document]
assafmo: Yeah this part is autogenerated, feel free to edit

x/airdrop/module.go
assafmo: Might be, I used ignite

x/airdrop/types/keys.go
assafmo: idc, go for it

x/airdrop/types/keys.go
assafmo: obv copied 😅😅

x/airdrop/keeper/keeper.go
sampocs: i realize now that this was probably auto generated from ignite haha

assafmo: no this was all me 😅

proto/stride/airdrop/query.proto
assafmo: This is just `UserState` from the spec doc. Feel free to change this.

x/airdrop/keeper/msg_server.go
```diff
@@ -0,0 +1,158 @@
...
+
+// Admin transaction to create a new airdrop
+func (ms msgServer) CreateAirdrop(goCtx context.Context, msg *types.MsgCreateAirdrop) (*types.MsgCreateAirdropResponse, error) {
+	ctx := sdk.UnwrapSDKContext(goCtx)
+
+	if _, found := ms.Keeper.GetAirdrop(ctx, msg.AirdropId); found {
+		return nil, types.ErrAirdropAlreadyExists.Wrapf("airdrop %s", msg.AirdropId)
+	}
+
+	airdrop := types.Airdrop{
```
assafmo, line 69: TODO: can anyone just create an airdrop?

x/airdrop/keeper/msg_server.go
```diff
@@ -0,0 +1,158 @@
...
+
+// Admin transaction to update an existing airdrop
+func (ms msgServer) UpdateAirdrop(goCtx context.Context, msg *types.MsgUpdateAirdrop) (*types.MsgUpdateAirdropResponse, error) {
+	ctx := sdk.UnwrapSDKContext(goCtx)
+
+	if _, found := ms.Keeper.GetAirdrop(ctx, msg.AirdropId); !found {
+		return nil, types.ErrAirdropNotFound.Wrapf("airdrop %s", msg.AirdropId)
+	}
+
+	airdrop := types.Airdrop{
```
assafmo, line 92: TODO: check airdrop admin

x/airdrop/keeper/msg_server.go
```diff
@@ -0,0 +1,158 @@
...
+
+// Admin transaction to add user allocations
+func (ms msgServer) AddAllocations(goCtx context.Context, msg *types.MsgAddAllocations) (*types.MsgAddAllocationsResponse, error) {
+	ctx := sdk.UnwrapSDKContext(goCtx)
+
+	if _, found := ms.Keeper.GetAirdrop(ctx, msg.AirdropId); !found {
+		return nil, types.ErrAirdropNotFound.Wrapf("airdrop %s", msg.AirdropId)
+	}
+
+	for _, rawAllocation := range msg.Allocations {
```
assafmo, line 115: TODO: check airdrop admin

x/airdrop/keeper/msg_server.go
```diff
@@ -0,0 +1,158 @@
...
+
+// Admin transaction to update a user's allocations
+func (ms msgServer) UpdateUserAllocation(goCtx context.Context, msg *types.MsgUpdateUserAllocation) (*types.MsgUpdateUserAllocationResponse, error) {
+	ctx := sdk.UnwrapSDKContext(goCtx)
+
+	if _, found := ms.Keeper.GetAirdrop(ctx, msg.AirdropId); !found {
+		return nil, types.ErrAirdropNotFound.Wrapf("airdrop %s", msg.AirdropId)
+	}
+
+	userAllocation, found := ms.Keeper.GetUserAllocation(ctx, msg.AirdropId, msg.UserAddress)
```
assafmo, line 142: TODO: check airdrop admin

x/airdrop/keeper/msg_server.go
```diff
@@ -0,0 +1,158 @@
...
+	}
+	userAllocation.Allocations = msg.Allocations
+	ms.Keeper.SetUserAllocation(ctx, userAllocation)
+
+	return &types.MsgUpdateUserAllocationResponse{}, nil
+}
+
+// Admin address to link a stride and non-stride address, merging their allocations
+func (ms msgServer) LinkAddresses(goCtx context.Context, msg *types.MsgLinkAddresses) (*types.MsgLinkAddressesResponse, error) {
+	ctx := sdk.UnwrapSDKContext(goCtx)
```
assafmo, line 154: TODO: check airdrop admin

x/airdrop/keeper/claim.go
```diff
@@ -0,0 +1,163 @@
...
+		}
+
+		// Update the amount claimed on the allocation record
+		userAllocation.Claimed = userAllocation.Claimed.Add(todaysRewards)
+
+		// Flag the user's claim type decision
+		userAllocation.ClaimType = types.CLAIM_EARLY
+
+		// Distribute rewards from the distributor, deducting the early penalty
+		distributedRewards := sdk.NewDecFromInt(todaysRewards).Mul(airdrop.EarlyClaimPenalty).TruncateInt()
```
assafmo, line 153: Qs:
- should we do `claimed += todaysRewards` or `claimed += distributedRewards`?
- should we rename `todaysRewards` to `pendingRewards` or something else? because later on `todaysRewards` is cut into `distributedRewards`
- should we add a `forfeitedRewards` field in `UserAllocation` (for accounting and ui purposes)?

sampocs: Great questions - this was bothering me as well.

Let's rename `todaysRewards ` to `pendingRewards `, increment `claimed` by `distributedRewards`, and add `forfeitedRewards `

assafmo: 714ffe8df

x/airdrop/keeper/claim.go
```diff
@@ -0,0 +1,198 @@
...
+	// If there are no rewards, alert the user with an error
+	if todaysRewards.IsZero() {
+		return types.ErrNoUnclaimedRewards
+	}
+
+	// Update the amount claimed on the allocation record
+	userAllocation.Claimed = userAllocation.Claimed.Add(todaysRewards)
+
+	// If this is their first time claiming, flag their decision
+	userAllocation.ClaimType = types.CLAIM_DAILY
```
riley-stride, line 53: how could it be anything else, based on the check on line 24?

sampocs: good flag! this was left over cause we used to have a `CLAIM_UNSPECIFIED` status. 

Just removed the line

x/airdrop/keeper/claim.go
```diff
@@ -0,0 +1,195 @@
...
+	airdrop, airdropFound := k.GetAirdrop(ctx, airdropId)
+	if !airdropFound {
+		return types.ErrAirdropNotFound.Wrapf("airdrop %s", airdropId)
+	}
+	userAllocation, userFound := k.GetUserAllocation(ctx, airdropId, claimer)
+	if !userFound {
+		return types.ErrUserAllocationNotFound.Wrapf("user %s for airdrop %s", claimer, airdropId)
+	}
+
+	// Confirm the user has not elected the non-daily claim types
```
riley-stride, line 80: nit
```suggestion
	// Confirm the user has not elected the daily claim types
```

x/airdrop/keeper/claim.go
```diff
@@ -0,0 +1,195 @@
...
+		userAllocation.Allocations[i] = sdkmath.ZeroInt()
+	}
+
+	// If there are no rewards, alert the user with an error
+	if totalAccruedRewards.IsZero() {
+		return types.ErrNoUnclaimedRewards
+	}
+
+	// Calculate rewards after claim early penalty
+	rewardsRemainingRate := sdk.OneDec().Sub(airdrop.EarlyClaimPenalty)
```
riley-stride, line 110: nit: should we sanity check this is /in [0,1]? ignore if we already have such checks on EarlyClaimPenalty

x/airdrop/keeper/claim.go
```diff
@@ -0,0 +1,195 @@
...
+	if !strideFound {
+		hostAllocation.Address = strideAddress
+		hostAllocation.ClaimType = types.CLAIM_DAILY
+		k.RemoveUserAllocation(ctx, airdropId, hostAddress)
+		k.SetUserAllocation(ctx, hostAllocation)
+		return nil
+	}
+
+	// Confirm the stride and host allocations are the same length
+	if len(strideAllocation.Allocations) != len(hostAllocation.Allocations) {
```
riley-stride, line 175: Aren't all allocations the same length?

x/airdrop/keeper/claim_test.go
```diff
@@ -0,0 +1,450 @@
...
+			initialAllocations:  []int64{10, 10, 10},
+			expectedAllocations: []int64{0, 0, 10},
+			initialClaimed:      100,
+			expectedClaimed:     100 + 20,
+			expectedNewRewards:  20,
+			initialClaimType:    types.CLAIM_DAILY,
+		},
+		{
+			// 10 rewards accrued on each of 3 days
+			// Claimed shortly into the second day, 20 total claimed
```
riley-stride, line 68: piconit
```suggestion
			// Claimed shortly into the third day, 30 total claimed
```

x/airdrop/keeper/claim_test.go
```diff
@@ -0,0 +1,450 @@
...
+		{
+			// Claimed 1 hour before the airdrop started
+			name:               "airdrop not started yet",
+			timeOffset:         (-1 * time.Hour), // before airdrop start
+			initialAllocations: []int64{},
+			initialClaimed:     100,
+			expectedError:      "airdrop distribution has not started",
+		},
+		{
+			// Claimed well after the airdrop ended
```
riley-stride, line 104: What about the case where they claim after the airdrop claim window, but have nonzero allocations?

x/airdrop/keeper/claim_test.go
```diff
@@ -0,0 +1,450 @@
...
+		{
+			// Claimed 1 hour before the airdrop started
+			name:               "airdrop not started yet",
+			timeOffset:         (-1 * time.Hour), // before airdrop start
+			initialAllocations: []int64{},
+			initialClaimed:     100,
+			expectedError:      "airdrop distribution has not started",
+		},
+		{
+			// Claimed well after the decision deadline
```
riley-stride, line 266: Are we allowing ClaimEarly throughout the airdrop? Thought the deadline to choose that option was earlier than the end of the airdrop. This test implies the deadlines are the same.

x/airdrop/keeper/claim_test.go
```diff
@@ -0,0 +1,450 @@
...
+			initialClaimed:      100,
+			expectedClaimed:     100 + 30,
+			expectedNewRewards:  30,
+			initialClaimType:    types.CLAIM_DAILY,
+		},
+		{
+			// Claimer already chose claim early
+			name:               "already chose to claim early",
+			timeOffset:         time.Hour,
+			initialAllocations: []int64{},
```
riley-stride, line 82: Is `[]int64{}` the same as `[]int64{0,0,0,...}`? That's what I expect an early claimer's allocation array to look like.

x/airdrop/keeper/claim_test.go
```diff
@@ -0,0 +1,450 @@
...
+		name                string
+		initialClaimType    types.ClaimType
+		initialClaimed      sdkmath.Int
+		strideAllocations   []int64
+		hostAllocations     []int64
+		expectedAllocations []int64
+		expectedError       string
+	}{
+		{
+			name:                "no stride allocations",
```
riley-stride, line 358: Is there a reverse case where StrideAllocations is non-nil and hostAllocations is nil or all zeros?

x/airdrop/keeper/claim_test.go
```diff
@@ -0,0 +1,450 @@
...
+			strideAllocations:   []int64{0, 0, 0},
+			hostAllocations:     []int64{0, 0, 0},
+			expectedAllocations: []int64{0, 0, 0},
+		},
+		{
+			name:              "different allocation lengths",
+			initialClaimType:  types.CLAIM_DAILY,
+			initialClaimed:    sdkmath.NewInt(100),
+			strideAllocations: []int64{10, 20, 30},
+			hostAllocations:   []int64{40, 50},
```
riley-stride, line 394: Similar to prev comment about lengths - how do we reach a state where the lengths aren't the same?

x/airdrop/types/msgs.go
```diff
@@ -0,0 +1,535 @@
...
+		return errors.New("clawback date must be specified")
+	}
+	if msg.ClaimTypeDeadlineDate == nil || *msg.ClaimTypeDeadlineDate == (time.Time{}) {
+		return errors.New("claim type deadline date must be specified")
+	}
+
+	if !msg.DistributionEndDate.After(*msg.DistributionStartDate) {
+		return errors.New("distribution end date must be after the start date")
+	}
+	if !msg.ClaimTypeDeadlineDate.After(*msg.DistributionStartDate) {
```
riley-stride, line 216:
```suggestion
	if !msg.ClaimTypeDeadlineDate.After(*msg.ClaimTypeDeadlineDate) {
```

riley-stride: shouldnt this have thrown errors in testing?

x/airdrop/types/msgs.go
```diff
@@ -0,0 +1,535 @@
...
+		return errors.New("claim type deadline date must be after the distribution start date")
+	}
+	if !msg.ClaimTypeDeadlineDate.Before(*msg.DistributionEndDate) {
+		return errors.New("claim type deadline date must be before the distribution end date")
+	}
+	if !msg.ClawbackDate.After(*msg.DistributionEndDate) {
```
riley-stride, line 222:
```suggestion
	if !msg.ClawbackDate.After(*msg.ClawbackDate) {
```

x/airdrop/keeper/claim.go
sampocs: i the original comment is correct

x/airdrop/keeper/claim.go
sampocs: it's handled in validate basic!

x/airdrop/keeper/claim.go
sampocs: it is in there when allos are initialized, this is more just a redundant check since we're about to iterate through each list

x/airdrop/keeper/claim_test.go
sampocs: I think those two tests are essentially the same since the check on the airdrop window doesn't consider the allocations

x/airdrop/keeper/claim.go
riley-stride: ah you're right

maybe for slightly more clarity we could make it "Confirm the user has elected the daily claim types" instead of the double negative? 

defer to you, and it's a nit point. feel free to resolve

x/airdrop/keeper/claim_test.go
sampocs: can you say more about why this test implies the deadlines are the same?

x/airdrop/keeper/claim_test.go
sampocs: it's not the same, but the allocations don't really matter since the check on the claim type is upstream. I'll flush it out though just to make the test easier to interpret

x/airdrop/keeper/claim_test.go
sampocs: just added a test case for nil host! (also covers the case where both are nil since it errors if host is nil)

x/airdrop/keeper/claim_test.go
sampocs: shouldn't be possible! Just a sanity check

x/airdrop/keeper/claim.go
sampocs: Yeah agreed it is confusing, it's also slightly stale since we used to have claim and stake.

The reason for the double negative is cause the user doesn't actually elect the daily option (it's assigned by default and stays as long as they don't elect ClaimEarly)

I'm just going to rephrase it to "Confirm the user has not elected to claim early"

x/airdrop/keeper/claim_test.go
riley-stride: Aren't there two separate deadlines? The first is the time at which you can no longer claim early, and the second is the time at which you can no longer claim. Or do we only have the latter

riley-stride: Ah understood. Then we're good here. Ty for explaining

x/airdrop/keeper/claim_test.go
```diff
@@ -0,0 +1,460 @@
+package keeper_test
```
shellvish, line 1: incredibly detailed test cases, thank you for being so methodical!

x/airdrop/keeper/msg_server.go
```diff
@@ -0,0 +1,148 @@
...
+
+	err := ms.Keeper.ClaimEarly(ctx, msg.AirdropId, msg.Claimer)
+	if err != nil {
+		return nil, err
+	}
+
+	return &types.MsgClaimEarlyResponse{}, nil
+}
+
+// Admin transaction to create a new airdrop
```
shellvish, line 48: Will we ever want to create an airdrop through the admin function? Or is this mainly "just-in-case"?

x/airdrop/types/msgs.go
```diff
@@ -0,0 +1,535 @@
...
+	}
+
+	if msg.AirdropId == "" {
+		return errors.New("airdrop-id must be specified")
+	}
+	if msg.RewardDenom == "" {
+		return errors.New("reward denom must be specified")
+	}
+
+	if msg.DistributionStartDate == nil || *msg.DistributionStartDate == (time.Time{}) {
```
shellvish, line 305: Very good coverage in the `ValidateBasic`s! 

nit, should we condense the validation in `MsgUpdateAirdrop` and `MsgCreateAirdrop`?

x/airdrop/client/cli/parser.go
```diff
@@ -0,0 +1,95 @@
...
+	"fmt"
+	"os"
+	"strings"
+
+	sdkmath "cosmossdk.io/math"
+
+	"github.com/Stride-Labs/stride/v22/x/airdrop/types"
+)
+
+// Parses an allocations CSV file consisting of allocations for various addresses
```
shellvish, line 16: Very cool! I had no idea you could read a CSV in a function call like this

x/airdrop/client/cli/parser.go
```diff
@@ -0,0 +1,95 @@
...
+		}
+		allAllocations = append(allAllocations, allocation)
+	}
+
+	return allAllocations, nil
+}
+
+// Parses a single user's allocations from a single line file with comma separate reward amounts
+// Ex: 10,10,20
+func ParseSingleUserAllocations(fileName string) (allocations []sdkmath.Int, err error) {
```
shellvish, line 64: Do we need this as a separate function? Could we not call `ParseMultipleUserAllocations` always? It feels _slightly_ safer to condense into one function (to avoid multiple code paths)

shellvish (commented):
Looks fantastic overall! Squeaky clean code, very well tested, and highly readable. A few misc questions throughout, but nothing major!

x/airdrop/keeper/claim.go
```diff
@@ -0,0 +1,195 @@
...
+	airdrop, airdropFound := k.GetAirdrop(ctx, airdropId)
+	if !airdropFound {
+		return types.ErrAirdropNotFound.Wrapf("airdrop %s", airdropId)
+	}
+	userAllocation, userFound := k.GetUserAllocation(ctx, airdropId, claimer)
+	if !userFound {
+		return types.ErrUserAllocationNotFound.Wrapf("user %s for airdrop %s", claimer, airdropId)
+	}
+
+	// Confirm the user has not elected the non-daily claim types
```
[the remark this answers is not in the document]
sampocs: updated in one of the earlier commits! 

to clarify though - the user is given type DAILY by default. So the check here is that they haven't elected claim early

x/airdrop/client/cli/tx.go
```diff
@@ -0,0 +1,535 @@
...
+		Use:   "update-user-allocation [airdrop-id] [user-address] [allocations-file]",
+		Short: "Update's a single user allocation's for a given airdrop",
+		Long: strings.TrimSpace(
+			fmt.Sprintf(`Update's a single user allocation's for a given airdrop
+
+Example file:
+ 0,10,10,20,30,40,...
+
+Example Command:
+  $ %[1]s tx %[2]s update-user-allocations airdrop-1 allocations.csv --from admin
```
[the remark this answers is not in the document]
sampocs: good catch! fixed

x/airdrop/keeper/claim.go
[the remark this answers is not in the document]
sampocs: ah actually nvm, the link above was for the other tx. Fixed the one your referencing [here](https://github.com/Stride-Labs/stride/pull/1235/commits/ed4d578c617cebd4020f9f3130cd2c259d7ca678)

x/airdrop/keeper/claim.go
```diff
@@ -0,0 +1,195 @@
...
+	if userAllocation.ClaimType != types.CLAIM_DAILY {
+		return types.ErrClaimTypeUnavailable.Wrapf("user has already elected claim option %s",
+			userAllocation.ClaimType.String())
+	}
+
+	// Get the index in the allocations array from the current date
+	// E.g. on the 10th day of distribution, this will map to the 9th index in the list
+	todaysIndex, err := airdrop.GetCurrentDateIndex(ctx)
```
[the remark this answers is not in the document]
sampocs: the stDYM airdrop would be a separate airdrop struct with a separate start date. lmk if that doesn't answer your question

x/airdrop/keeper/claim.go
```diff
@@ -0,0 +1,195 @@
...
+	if err != nil {
+		return err
+	}
+
+	// Sum the rewards up to that date and 0 them out in the process
+	todaysRewards := sdkmath.ZeroInt()
+	for i := 0; i <= todaysIndex; i++ {
+		rewardsOnDate := userAllocation.Allocations[i]
+		todaysRewards = todaysRewards.Add(rewardsOnDate)
```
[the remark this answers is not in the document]
sampocs: I think it'd be the same gas since gas is based on store read/writes! (low confidence though)

x/airdrop/keeper/claim.go
```diff
@@ -0,0 +1,195 @@
...
+	if !airdropFound {
+		return types.ErrAirdropNotFound.Wrapf("airdrop %s", airdropId)
+	}
+	userAllocation, userFound := k.GetUserAllocation(ctx, airdropId, claimer)
+	if !userFound {
+		return types.ErrUserAllocationNotFound.Wrapf("user %s for airdrop %s", claimer, airdropId)
+	}
+
+	// Confirm the user has not elected to claim early (meaning their type will be ClaimDaily)
+	if userAllocation.ClaimType != types.CLAIM_DAILY {
```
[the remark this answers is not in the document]
sampocs: that's what this check is doing actually! it's _erroring_ if the claim type is isn't DAILY

x/airdrop/client/cli/parser.go
```diff
@@ -0,0 +1,95 @@
...
+				return nil, fmt.Errorf("unable to parse allocation %s into sdk.Int", allocationString)
+			}
+			allocations = append(allocations, allocation)
+		}
+
+		allocation := types.RawAllocation{
+			UserAddress: userAddress,
+			Allocations: allocations,
+		}
+		allAllocations = append(allAllocations, allocation)
```
shellvish, line 56: should we verify all the `allocation`s have the same length?

x/airdrop/keeper/claim.go
```diff
@@ -0,0 +1,195 @@
...
+
+	return nil
+}
+
+// Admin transaction to merge allocations between a stride and non-stride address
+// If the stride address does not yet have an allocation, the host allocation will be overwritten
+// with the stride address
+// Otherwise, if the stride allocation already exists, the two allocations will be merged and set
+// in on the stride allocation
+// We can safely change the type back to DAILY because if a user claimed early, their allocations
```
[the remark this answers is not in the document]
sampocs: hmm that's a very good question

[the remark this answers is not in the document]
sampocs: I think I'd prefer to go with the simplest approach, since this situation shouldn't actually happen in practice (because our frontend requires users to link before they claim). So I think I'd lean towards leaving it as is since I think that would be the simpler implementation

x/airdrop/keeper/msg_server.go
sampocs: yeah this is actually how we're planning to do it! (not via the upgrade handler)

proto/stride/airdrop/airdrop.proto
```diff
@@ -0,0 +1,111 @@
...
+import "google/protobuf/timestamp.proto";
+
+option go_package = "github.com/Stride-Labs/stride/v22/x/airdrop/types";
+
+// ClaimType enum represents the possible claim types for a user getting an
+// airdrop
+enum ClaimType {
+  option (gogoproto.goproto_enum_prefix) = false;
+
+  // CLAIM_DAILY indicates that the airdrop rewards are accumulated daily
```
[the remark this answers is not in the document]
sampocs: definitely! just updated the comment

x/airdrop/types/msgs.go
sampocs: definitely! will update

x/airdrop/client/cli/parser.go
sampocs: yeah that's a good point. I don't have a strong opinion either way, down to consolidate them

x/airdrop/client/cli/parser.go
sampocs: it's handled in validate basic!

x/airdrop/types/msgs.go
sampocs: done

x/airdrop/client/cli/parser.go
sampocs: ok yeah just implemented it and this was def the right call - great suggestion!

proto/stride/airdrop/airdrop.proto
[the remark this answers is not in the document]
shellvish: Amazing, thank you! Resolving

x/airdrop/client/cli/parser.go
shellvish: Oh nice! Resolving

x/airdrop/keeper/claim.go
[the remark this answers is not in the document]
shellvish: Ah, if `userAllocation.Allocations[i]` is already 0, does setting it to `sdkmath.ZeroInt()` not cost any gas? 

Either way, I don't think this is a big deal, and down to leave as is. The gas considerations are fairly secondary

x/airdrop/keeper/claim.go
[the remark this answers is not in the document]
shellvish: Ah yeah sorry I misunderstood, I thought the claim should be in `types.CLAIM_EARLY` at the top of this function. Resolving!

x/airdrop/keeper/claim.go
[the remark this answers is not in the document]
shellvish: That makes sense, and agreed with the simpler implementation (particularly as this is an edge case). Resolving!

x/airdrop/keeper/msg_server.go
shellvish: Oh gotcha! Is that feasible? If there are 10k+ addresses, do you know if that's 100 txs, 10 txs, etc?

x/airdrop/keeper/claim.go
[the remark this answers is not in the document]
shellvish: Perfect, resolving!

x/airdrop/types/msgs.go
shellvish: Amazing, resolving!

x/airdrop/client/cli/parser.go
shellvish: Thank you for implementing so quickly! That was v speedy

x/airdrop/keeper/claim.go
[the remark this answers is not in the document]
sampocs: We don't actually set it in the store in this step! It's just updating the go variable

And then later on down below we write the full record to the store and the gas meter starts churning again. 

We should test this though!

x/airdrop/keeper/msg_server.go
sampocs: yeah we should test it - there's roughly 65k addresses in the airdrop. 

The way we were thinking about it was that it was a tradeoff between:
* being forced to run an upgrade to add allocations (e.g. we'd need to do another upgrade for stDYM) 
* vs doing it manually and having less confidence that all the txs went through

But I think we can mitigate the second point by just auditing the chain data after we add everything.

We should try and see how many we can get in a single message though. It's on my docket to test, but if you can get to to it before me that'd be great!

x/airdrop/keeper/claim.go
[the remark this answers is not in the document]
assafmo: > it'd be the same gas since gas is based on store read/writes

That's correct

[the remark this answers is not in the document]
assafmo:
```
store(x)
store(y)
while true:
  x+=1
```
and
```
store(x)
store(y)
```
consume the same amount of gas

x/airdrop/keeper/claim.go
```diff
@@ -0,0 +1,194 @@
...
+	distributorAccount := sdk.MustAccAddressFromBech32(airdrop.DistributionAddress)
+	claimerAccount := sdk.MustAccAddressFromBech32(userAllocation.Address)
+	rewardsCoin := sdk.NewCoin(airdrop.RewardDenom, todaysRewards)
+
+	if err := k.bankKeeper.SendCoins(ctx, distributorAccount, claimerAccount, sdk.NewCoins(rewardsCoin)); err != nil {
+		return errorsmod.Wrapf(err, "unable to distribute rewards")
+	}
+
+	// Update the reward record for to mark the progress
+	k.SetUserAllocation(ctx, userAllocation)
```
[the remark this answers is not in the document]
assafmo: Unless `SendCoins` can call `ClaimDaily` idt there's reentrency risk, but I've switched it anyway.

93dca5f07

x/airdrop/keeper/claim.go
```diff
@@ -0,0 +1,194 @@
...
+
+	// Update the amount claimed on the allocation record
+	userAllocation.Claimed = userAllocation.Claimed.Add(todaysRewards)
+
+	// Distribute rewards from the distributor
+	distributorAccount := sdk.MustAccAddressFromBech32(airdrop.DistributionAddress)
+	claimerAccount := sdk.MustAccAddressFromBech32(userAllocation.Address)
+	rewardsCoin := sdk.NewCoin(airdrop.RewardDenom, todaysRewards)
+
+	if err := k.bankKeeper.SendCoins(ctx, distributorAccount, claimerAccount, sdk.NewCoins(rewardsCoin)); err != nil {
```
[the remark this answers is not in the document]
assafmo: That's a good idea, but I'm not sure how to go about it.

proto/stride/airdrop/airdrop.proto
```diff
@@ -0,0 +1,120 @@
...
+  string id = 1;
+
+  // Denom used when distributing rewards
+  string reward_denom = 2;
+
+  // The first date that claiming begins and rewards are distributed
+  google.protobuf.Timestamp distribution_start_date = 3
+      [ (gogoproto.stdtime) = true ];
+
+  // The last date for rewards to be distributed. Immediately after this date
```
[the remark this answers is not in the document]
assafmo: Yes, and I think this behavior is tested [here](https://github.com/Stride-Labs/stride/blob/93dca5f07/x/airdrop/keeper/claim_test.go#L42-L77). Please let us know if these 3 tests are non-exhaustive in your opinion.

x/airdrop/keeper/claim.go
```diff
@@ -0,0 +1,194 @@
...
+	}
+
+	// Confirm we're not past the decision date
+	currentTime := ctx.BlockTime().Unix()
+	if currentTime > airdrop.ClaimTypeDeadlineDate.Unix() {
+		return types.ErrAfterDecisionDeadline
+	}
+
+	// Confirm the airdrop started
+	if currentTime < airdrop.DistributionStartDate.Unix() {
```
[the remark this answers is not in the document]
assafmo: You're absolutely correct a92dcca01

x/airdrop/keeper/claim.go
```diff
@@ -0,0 +1,194 @@
...
+	distributorAccount := sdk.MustAccAddressFromBech32(airdrop.DistributionAddress)
+	claimerAccount := sdk.MustAccAddressFromBech32(userAllocation.Address)
+
+	rewardsCoin := sdk.NewCoin(airdrop.RewardDenom, distributedRewards)
+	if err := k.bankKeeper.SendCoins(ctx, distributorAccount, claimerAccount, sdk.NewCoins(rewardsCoin)); err != nil {
+		return errorsmod.Wrapf(err, "unable to distribute rewards")
+	}
+
+	// Update the reward record for to mark the progress
+	k.SetUserAllocation(ctx, userAllocation)
```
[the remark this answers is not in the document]
assafmo: 78afa347c

x/airdrop/keeper/claim.go
```diff
@@ -0,0 +1,194 @@
...
+
+	return nil
+}
+
+// Admin transaction to merge allocations between a stride and non-stride address
+// If the stride address does not yet have an allocation, the host allocation will be overwritten
+// with the stride address
+// Otherwise, if the stride allocation already exists, the two allocations will be merged and set
+// in on the stride allocation
+// We can safely change the type back to DAILY because if a user claimed early, their allocations
```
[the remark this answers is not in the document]
assafmo: Good point. It was easier but not strictly necessary when we had claim&stake, but now there's an argument to be made that it's unnecessary. When a user claims, it'll fail anyway if they don't have future allocations. @sampocs wdyt?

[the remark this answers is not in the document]
assafmo: re: frontend, it can infer `ClaimType` from looking at `forfeited`. Not ideal, but removing `ClaimType` altogether can simplify the chain code.

x/airdrop/keeper/claim.go
```diff
@@ -0,0 +1,194 @@
...
+	}
+
+	// If the stride user does exist, merge the two allocations into the stride user
+	for i, strideRewards := range strideAllocation.Allocations {
+		hostReward := hostAllocation.Allocations[i]
+		strideAllocation.Allocations[i] = strideRewards.Add(hostReward)
+	}
+
+	// Reset the claim type to daily
+	// That's to support the use case of a stride user claiming early and then linking a host address
```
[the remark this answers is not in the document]
assafmo: Disallow what?

x/airdrop/keeper/claim.go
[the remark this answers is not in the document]
shellvish: Ah this makes a lot of sense, and is very helpful to update my mental model of gas! Thanks for the quick clarification, resolving

x/airdrop/keeper/claim.go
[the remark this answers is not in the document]
shellvish: Ah this makes sense, resolving

x/airdrop/keeper/claim.go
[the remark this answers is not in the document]
sampocs: the reason it wasn't in `ClaimDaily` is because it's checked in `GetCurrentDateIndex`

Probably doesn't hurt to add a redundant check though for readability

x/airdrop/keeper/claim.go
```diff
@@ -26,6 +26,17 @@ func (k Keeper) ClaimDaily(ctx sdk.Context, airdropId, claimer string) error {
 			userAllocation.ClaimType.String())
 	}
 
+	// Confirm we're not past the decision date
+	currentTime := ctx.BlockTime().Unix()
+	if currentTime > airdrop.ClaimTypeDeadlineDate.Unix() {
+		return types.ErrAfterDecisionDeadline
+	}
```
sampocs, line 33: i don't think we should do this for the daily check

x/airdrop/keeper/claim.go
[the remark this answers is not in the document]
sampocs: yeah it's a good point. We could remove the status and still add the "implied" type (based on `forfeited`) to the user summary query for the frontend

That said, the ClaimType does provide a nice short circuit in the event that a user claims twice (i.e. no need to loop through all the allocations again)

No strong opinions from me either way

proto/stride/airdrop/airdrop.proto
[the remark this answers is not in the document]
sampocs: I also do think we're missing some test cases on the end boundary too - I'll add some now

x/airdrop/keeper/claim.go
assafmo: oh :facepalm:

assafmo: fixed 1ef4a2f0b

x/airdrop/keeper/claim.go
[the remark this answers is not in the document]
assafmo: I haven't noticed it's already being checked in there, which makes me think "hiding" this in `GetCurrentDateIndex()` hurts readability and I personally prefer to double-check.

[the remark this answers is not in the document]
assafmo: lmk if you think otherwise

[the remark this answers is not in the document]
sampocs: yeah agreed!

proto/stride/airdrop/airdrop.proto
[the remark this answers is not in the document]
sampocs: added [here](https://github.com/Stride-Labs/stride/pull/1235/commits/05cc5607cfbdfcda7e10df79e9470c9d1f519e7f) and [here](https://github.com/Stride-Labs/stride/pull/1235/commits/3d4492c399f097828ecda6bd319155d67adcf7e0)

x/airdrop/keeper/claim.go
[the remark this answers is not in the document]
asalzmann: Discussed this morning, I don't have a strong view. @sampocs said he thinks it's cleaner to remove `ClaimType` and will open a diff

x/airdrop/keeper/claim.go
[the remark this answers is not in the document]
asalzmann: > the use case of a stride user claiming early and then linking a host address

Although we discussed this morning and I think we'll leave this functionality, so disregard

asalzmann:
@sampocs [05cc560](https://github.com/Stride-Labs/stride/pull/1235/commits/05cc5607cfbdfcda7e10df79e9470c9d1f519e7f) and [3d4492c](https://github.com/Stride-Labs/stride/pull/1235/commits/3d4492c399f097828ecda6bd319155d67adcf7e0) LGTM

x/airdrop/keeper/msg_server.go
sampocs: tested offline and confirmed it will probably be on the order of ~30 txs (assuming not submitted by a ledger admin)

marking resolved!

x/airdrop/keeper/claim.go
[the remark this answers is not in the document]
sampocs: removed [here](https://github.com/Stride-Labs/stride/pull/1235/commits/8ef66c0cf75aa19e0caca6426afdae2a52a992d4). I think i'm in favor

proto/stride/airdrop/airdrop.proto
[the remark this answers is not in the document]
sampocs: @asalzmann and I good to mark this resolved? (I see you reviewed the above two commits)

x/airdrop/keeper/claim.go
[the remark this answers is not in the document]
sampocs: I was wondering this as well. Only thing I could think of was to add some max claim amount to the airdrop struct

x/airdrop/keeper/claim.go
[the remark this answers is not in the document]
assafmo: agreed, it looks better

proto/stride/airdrop/airdrop.proto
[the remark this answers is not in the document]
asalzmann: Yes

x/airdrop/keeper/claim.go
```diff
@@ -159,11 +159,7 @@ func (k Keeper) ClaimEarly(ctx sdk.Context, airdropId, claimer string) error {
 // There's no need to merge the Claimed or Forfeited amounts because the host allocations cannot
 // be claimed through a non-stride address
 func (k Keeper) LinkAddresses(ctx sdk.Context, airdropId, strideAddress, hostAddress string) error {
-	// Fetch the airdrop and user's allocations
```
shellvish: is the thinking behind removing this check that the `hostFound` check below should cover it?

shellvish (approved):
Looks great to me! Only one small question

x/airdrop/keeper/claim.go
sampocs: it was moved into the main msg_server function!

x/airdrop/types/airdrop.go
```diff
@@ -28,16 +28,17 @@ func (a *Airdrop) GetCurrentDateIndex(ctx sdk.Context, windowLengthSeconds int64
...
+		elapsedDays = a.GetAirdropLength(windowLengthSeconds) - 1
 	}
 
 	return int(elapsedDays), nil
 }
 
 // Returns the length of the airdrop in days
-func (a *Airdrop) GetAirdropLength() int64 {
+// windowLengthSeconds is the time between each element in the allocations array
+func (a *Airdrop) GetAirdropLength(windowLengthSeconds int64) int64 {
```
shellvish, line 40: Nice, this makes a lot of sense, and I think this is an overall cleaner design! 

This is outside the scope of this PR I think, but would it make more sense to relabel "days" to "periods" with this new structure?

sampocs: agreed! 

Also I don't really love the naming of `windowLengthSeconds` or `allocation_window_seconds` - very open to better names if you can think of any!

sampocs: just renamed `GetAirdropLength` to `GetAirdropPeriods`

shellvish: Hm how about `periodLengthSeconds` and `period_length_seconds`? Open to other options too!

sampocs: I like it!

sampocs: done https://github.com/Stride-Labs/stride/pull/1235/commits/84abb9c5517b6f74b887c5d3fcbbf2ea89245330

x/airdrop/keeper/claim.go
sampocs: @shellvish can i mark this resolved?

shellvish: Yup, makes sense!

x/airdrop/types/airdrop.go
shellvish: Incredible, thank you for doing that!

x/airdrop/keeper/claim.go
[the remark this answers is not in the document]
sampocs: I also realized removing this helps simplify an edge case where a user claimed early, but then we needed to update their allocations (i.e. if we made an error originally). If we still had the claim type in there, they wouldn't be able to claim after the correction!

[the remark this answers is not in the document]
sampocs: marking this resolved

PR #1242, opened 2024-07-18
close delegation channel tx

sampocs:


x/stakeibc/types/message_close_delegation_channel.go
```diff
@@ -0,0 +1,79 @@
...
+		return errorsmod.Wrapf(sdkerrors.ErrInvalidAddress, "invalid creator address (%s)", err)
+	}
+	if err := utils.ValidateAdminAddress(msg.Creator); err != nil {
+		return err
+	}
+
+	if msg.ChannelId == "" {
+		return errorsmod.Wrapf(sdkerrors.ErrInvalidRequest, "channel ID must be specified")
+	}
+	if msg.PortId == "" {
```
riley-stride, line 64: nit: can we do more validation on the port?

x/stakeibc/keeper/msg_server.go
```diff
@@ -1001,6 +1001,21 @@ func (k msgServer) RestoreInterchainAccount(goCtx context.Context, msg *types.Ms
...
+// Admin transaction to close an ICA channel
+// This can be used if there are records stuck in state IN_PROGRESS after a channel has been re-opened after a timeout
+// After the closure, the a new channel can be permissionlessly re-opened with RestoreInterchainAccount
+func (k msgServer) CloseDelegationChannel(goCtx context.Context, msg *types.MsgCloseDelegationChannel) (*types.MsgCloseDelegationChannelResponse, error) {
+	ctx := sdk.UnwrapSDKContext(goCtx)
+	_, capability, err := k.IBCKeeper.ChannelKeeper.LookupModuleByChannel(ctx, msg.PortId, msg.ChannelId)
+	if err != nil {
+		return nil, errorsmod.Wrap(err, "could not retrieve capability from port ID and channel ID")
+	}
+	if err := k.IBCKeeper.ChannelKeeper.ChanCloseInit(ctx, msg.PortId, msg.ChannelId, capability); err != nil {
```
riley-stride, line 1013: is there a case where chancloseinit fails or gets stuck in a pending state? if so would we just re-send this `CloseDelegationChannel` tx?

riley-stride (approved):
Lgtm

Have you tested via CLI?

shellvish (approved):
LGTM

shellvish (approved):
LGTM

x/stakeibc/types/codec.go
```diff
@@ -22,6 +22,7 @@ func RegisterCodec(cdc *codec.LegacyAmino) {
 	cdc.RegisterConcrete(&AddValidatorsProposal{}, "stakeibc/AddValidatorsProposal", nil)
 	cdc.RegisterConcrete(&ToggleLSMProposal{}, "stakeibc/ToggleLSMProposal", nil)
 	cdc.RegisterConcrete(&MsgRestoreInterchainAccount{}, "stakeibc/RestoreInterchainAccount", nil)
+	cdc.RegisterConcrete(&MsgCloseDelegationChannel{}, "stakeibc/CloseDelegationChannel", nil)
```
asalzmann, line 25: Does this allow us to sign with Ledger / did we test with Ledger?

x/stakeibc/keeper/msg_server.go
sampocs: i don't think so cause it goes straight to state CLOSED on stride (there's not handshake process like with opening)

Either way, not relevant anymore now that we're doing the send packet instead

x/stakeibc/types/codec.go
sampocs: [this](https://github.com/Stride-Labs/stride/pull/1242/files#diff-588e281d6fae46aa27bd0a5a2a3cafcd412244b49fe96977892590da86095cb9R16) should be what allows us to sign with ledger. Haven't test it

asalzmann: I thought we need to register a legacy amino codec [like this](https://github.com/Stride-Labs/stride/blob/3e1e5db9b555e13cd6ab5fcf36220fb3e0e2c856/x/airdrop/types/codec.go#L15)

sampocs: yeah i guess it's both - but the issue we've had in the past is that we didn't implement the legacy interface

asalzmann: sorry disregard - I see this _is_ the LegacyAmino registration

riley-stride (approved):
lgtm

PR #1244, opened 2024-07-19
TS Integration Tests Framework 

assafmo:
Adds a framework for writing integration tests using typescript, stridejs & dockernet.


dockernet/config.sh
```diff
@@ -39,6 +39,9 @@ elif [[ "${#HOST_CHAINS[@]}" == "0" ]]; then
 fi
 REWARD_CONVERTER_HOST_ZONE=${HOST_CHAINS[0]}
 
+# TODO[airdrop]: revert before merging
```
assafmo, line 42: @sampocs the pr is full of these kind of things. brabched off airdrop-module and merged from main. 🤷‍♂️

go.mod
```diff
@@ -199,9 +199,11 @@ replace (
 	github.com/99designs/keyring => github.com/cosmos/keyring v1.2.0
 
 	github.com/btcsuite/btcd => github.com/btcsuite/btcd v0.22.2 //indirect
-	// fork SDK to fix SDKv0.47 Distribution Bug
+	// fork SDK to:
+	//   - fix SDKv0.47 Distribution Bug
+	//   - add better mempool error on sig verification failure
 	// TODO - Remove this patch and update Tokens in a subsequent upgrade handler
-	github.com/cosmos/cosmos-sdk => github.com/Stride-Labs/cosmos-sdk v0.47.10-stride-distribution-fix-0
+	github.com/cosmos/cosmos-sdk => github.com/Stride-Labs/cosmos-sdk v0.47.10-stride-distribution-fix-0-mempool-verbose-error-1
```
assafmo, line 206: better mempool errors

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,21 @@
+name: Integration Tests
```
sampocs, line 1: if this isn't working, I'd just remove it for now and we can revisit with the new infra

dockernet/config/relayer_config_ics.yaml
```diff
@@ -16,6 +16,7 @@ chains:
       gas-prices: 0.02ustrd
       debug: false
       timeout: 20s
+      min-loop-duration: 50ms
```
sampocs, line 19: these loops are pretty aggressive, i'd recommend removing it or setting it to 1s

dockernet/ts-tests/.gitignore
```diff
@@ -0,0 +1 @@
+node_modules
```
sampocs, line 1: nit, wanna just move this to the main `.gitignore`

dockernet/ts-tests/README.md
```diff
@@ -0,0 +1,34 @@
...
+# install deps
+pnpm i
+
+# run tests
+pnpm test
+```
+
+IMPORTANT: `@cosmjs/*` dependencies must match the versions used by stridejs. To get those versions, run e.g. `pnpm why @cosmjs/amino`.
+
+### debugging (vscode)
```
sampocs, line 16: 😍

dockernet/ts-tests/package.json
```diff
@@ -0,0 +1,33 @@
+{
+  "name": "stride-integration-tests",
+  "version": "1.0.0",
+  "description": "integration tests for stride chain",
+  "scripts": {
+    "test": "vitest run",
+    "test-watch": "vitest watch"
+  },
+  "keywords": [
+    "wow",
```
sampocs, line 10: what's this? lol

dockernet/ts-tests/test/utils.ts
```diff
@@ -0,0 +1,210 @@
...
+ * To get the IBC denom of STRD on mainnet Osmosis:
+ * ```
+ * ibcDenom([{incomingPortId: "transfer", incomingChannelId: "channel-326"}], "ustrd")
+ * ```
+ *
+ * @param {Object[]} paths An array of objects containing information about the IBC transfer paths.
+ * @param {string} coinMinimalDenom The minimal denom of the coin.
+ * @returns {string} The computed IBC denom of the token.
+ */
+export const ibcDenom = (
```
sampocs, line 70: nit: I think this will always be the transfer channel so you can change the param to take just the channel ID and denom!

dockernet/ts-tests/test/utils.ts
```diff
@@ -0,0 +1,210 @@
...
+ */
+export function pubkeyToAddress(
+  pubkey: Uint8Array,
+  prefix: string = "stride",
+): string {
+  return bech32.encode(prefix, bech32.toWords(ripemd160(sha256(pubkey))));
+}
+
+/**
+ * Convert a secp256k1 compressed public key to an address
```
sampocs, line 103:
```suggestion
 * Convert a base64 encoded secp256k1 compressed public key to an address
```

dockernet/ts-tests/test/utils.ts
```diff
@@ -0,0 +1,210 @@
...
+}
+
+/**
+ * Convert self delegator address to validator address
+ *
+ * @param {String} selfDelegator The self delegator bech32 encoded address
+ * @param {String} [prefix="stride"] The self delegator address' bech32 prefix. Defaults to `"stride"`.
+ * @returns the account's address
+ */
+export function selfDelegatorAddressToValidatorAddress(
```
sampocs, line 123: by "self delegator" does this just mean a regular non-validator address?

dockernet/ts-tests/test/utils.ts
```diff
@@ -0,0 +1,210 @@
...
+  selfDelegator: string,
+  prefix: string = "stride",
+): string {
+  return bech32.encode(`${prefix}valoper`, bech32.decode(selfDelegator).words);
+}
+
+/**
+ * Convert self delegator address to validator address
```
sampocs, line 131:
```suggestion
 * Convert self validator address to self delegator address
```

dockernet/ts-tests/test/utils.ts
```diff
@@ -0,0 +1,210 @@
...
+  prefix: string = "stride",
+): string {
+  return bech32.encode(
+    `${prefix}valcons`,
+    bech32.toWords(sha256(pubkey).slice(0, 20)),
+  );
+}
+
+/**
+ * Convert a secp256k1 compressed public key to an address
```
sampocs, line 162:
```suggestion
 * Convert a base64 encoded public key to a validator address
```

dockernet/ts-tests/test/main.test.ts
```diff
@@ -0,0 +1,180 @@
...
+        "turkey miss hurry unable embark hospital kangaroo nuclear outside term toy fall buffalo book opinion such moral meadow wing olive camp sad metal banner",
+    },
+    {
+      name: "val3",
+      mnemonic:
+        "tenant neck ask season exist hill churn rice convince shock modify evidence armor track army street stay light program harvest now settle feed wheat",
+    },
+  ];
+
+  // @ts-expect-error
```
sampocs, line 58: whats the `// @ts-expect-error` for?

dockernet/ts-tests/test/main.test.ts
```diff
@@ -0,0 +1,180 @@
...
+    // @ts-expect-error
+    // init accounts[name] as an empty object, then add the fields one by one
+    accounts[name] = {};
+
+    // setup signer
+    //
+    // IMPORTANT: we're using Secp256k1HdWallet from @cosmjs/amino because sending amino txs tests both amino and direct.
+    // that's because the tx contains the direct encoding anyway, and also attaches a signature on the amino encoding.
+    // the mempool then converts from direct to amino to verify the signature.
+    // therefore if the signature verification passes, we can be sure that both amino and direct are supported.
```
sampocs, line 71: this is genius! 🧠

dockernet/ts-tests/test/main.test.ts
```diff
@@ -0,0 +1,180 @@
...
+    },
+  ];
+
+  // @ts-expect-error
+  // init accounts as an empty object, then add the accounts in the loop
+  accounts = {};
+  for (const { name, mnemonic } of mnemonics) {
+    // @ts-expect-error
+    // init accounts[name] as an empty object, then add the fields one by one
+    accounts[name] = {};
```
sampocs, line 64: is this necessary since you just overwrite it at the end of this for loop?

dockernet/ts-tests/test/main.test.ts
```diff
@@ -0,0 +1,180 @@
+import { Secp256k1HdWallet } from "@cosmjs/amino";
+import { GasPrice, SigningStargateClient } from "@cosmjs/stargate";
+import { fromSeconds } from "@cosmjs/tendermint-rpc";
```
sampocs, line 3: I remember you were in talks with the telescope team - do these packages compete with it or are these just lower level?

dockernet/ts-tests/test/main.test.ts
```diff
@@ -0,0 +1,180 @@
...
+      broadcastPollIntervalMs: 50,
+      resolveIbcResponsesCheckIntervalMs: 50,
+    });
+  }
+
+  console.log("waiting for chain to start...");
+  while (true) {
+    try {
+      const block =
+        await accounts.user.query.cosmos.base.tendermint.v1beta1.getLatestBlock(
```
sampocs, line 90: do you have to execute all queries through a user like this?

dockernet/ts-tests/test/main.test.ts
```diff
@@ -0,0 +1,180 @@
...
+    expect(airdrop!.earlyClaimPenalty).toBe("0.5");
+  });
+});
+
+describe("ibc", () => {
+  test("MsgTransfer", async () => {
+    const stridejs = accounts.user;
+
+    const msg =
+      stridejs.types.ibc.applications.transfer.v1.MessageComposer.withTypeUrl.transfer(
```
sampocs, line 152: I wonder if we should add some utility functions to build messages to avoid having these long references

dockernet/ts-tests/test/main.test.ts
```diff
@@ -0,0 +1,180 @@
...
+    const tx = await stridejs.signAndBroadcast([msg]);
+    if (tx.code !== 0) {
+      console.error(tx.rawLog);
+    }
+    expect(tx.code).toBe(0);
+
+    const ibcAck = await tx.ibcResponses[0];
+    expect(ibcAck.type).toBe("ack");
+    expect(ibcAck.tx.code).toBe(0);
+  }, 30_000);
```
sampocs, line 179: is this the test timeout?

sampocs (commented):
These are BEAUTIFUL!! So excited to finally switch over to this, this will be a HUGE lift!

Mostly nit comments and then I would just gut the dockernet + CI changes and then lets merge this guy in!

dockernet/ts-tests/package.json
assafmo: swagger

dockernet/ts-tests/test/main.test.ts
assafmo: you might be right, this part has gone through a few iterations

dockernet/ts-tests/test/main.test.ts
assafmo: these are low level helper functions, we can re-export them from stridejs instead of doing this

dockernet/ts-tests/test/main.test.ts
assafmo: yes though it's the same query client. we can export it to a global query variable for these tests, but in general I've learned from secret that it's more convenient to have the query and tx clients under the same object

assafmo: open to more feedback though

dockernet/ts-tests/test/main.test.ts
assafmo: yes

assafmo: default is 5s, which is not enough for tests with ibc operations

dockernet/ts-tests/test/main.test.ts
sampocs: Gotchya - let's keep it for now and we can iterate on it!

Fwiw, the global query option sounds like a good option. I setup [firehose](https://github.com/Stride-Labs/firehose) with the query/tx client on the user and always felt like it was a bit of an antipattern

sampocs: just curious - what's the benefit you noticed from secret in having it be shared?

dockernet/ts-tests/test/main.test.ts
sampocs: No need to change anything - was just curious! I'm not really familiar with the ecosystem of JS packages

dockernet/ts-tests/test/main.test.ts
assafmo: you just have the one StrideClient object, you can traverse it just like the cli
stridejs.query.bank.balance...
stridejs.tx.bank.send...

we've got a lot of good feedback on this after multiple iterations, but mostly that it's nice that you can explore the object with the ide's intelisense like you would explore the cli with typing half a command and appendig --help

dockernet/ts-tests/test/main.test.ts
assafmo: `accounts` is defined as:
```ts
let accounts: {
    user: StrideClient;
    admin: StrideClient;
    val1: StrideClient;
    val2: StrideClient;
    val3: StrideClient;
}
```
and here we initialize it with all of its fields as `undefined`, which is a type error. to make this error go away with can either intialize all the fields with dummy values, mark this like as `// @ts-expect-error` or define `accounts` as:
```ts
let accounts: {
    user?: StrideClient;
    admin?: StrideClient;
    val1?: StrideClient;
    val2?: StrideClient;
    val3?: StrideClient;
}
```
the problem with defining it like that is that whenever we'd want to access one of the fields we'd have to do `account?.user` or check if it's not `undefined`, which is very annoying.

this is just an initialization code so I feel comfortable to tell typescript to shut up.

dockernet/ts-tests/test/main.test.ts
assafmo: 43bc9815ba8ea998536df5f2059eaf447132df85

dockernet/ts-tests/test/main.test.ts
sampocs: makes sense, good call!

dockernet/config/relayer_config_ics.yaml
assafmo: 468288391

dockernet/ts-tests/test/utils.ts
assafmo: with ibc contracts it's `wasm.{contract_address}`

dockernet/ts-tests/test/utils.ts
assafmo: 3a25d3700

dockernet/ts-tests/test/main.test.ts
sampocs: nice, make sense! I do think it breaks down a bit though when you have multiple accounts that you're sending txs from in one test. In which case, I could imagine it being cleaner to have a separate global query config

```
query.bank...
accounts.X.tx.bank.send(...)
accounts.Y.tx.bank.send(...)
```

But just ideating - obviously no need to change anything now!

dockernet/ts-tests/test/utils.ts
assafmo: that's the cosmos term for the account that controls a validator. it's basically the same bytes but with a different prefix: `stride` vs `stridevaloper`

dockernet/ts-tests/test/utils.ts
assafmo: "self delegator" is a cosmos term

dockernet/config/relayer_config_ics.yaml
sampocs: I just tagged this one instance, but applies to all the ones in the other config files too

dockernet/ts-tests/test/utils.ts
assafmo: ccfafc9fd

dockernet/ts-tests/.gitignore
assafmo: a17d5c502

dockernet/config/relayer_config_ics.yaml
assafmo: d4783a8db

.github/workflows/integration-tests.yml
assafmo: 099139e85516ae696ad68e53cae661f121addb5c

dockernet/ts-tests/test/utils.ts
sampocs: ah didn't realize that

sampocs (approved):
🚢 !

dockernet/ts-tests/test/utils.ts
assafmo: moved here https://github.com/Stride-Labs/stridejs/commit/2e3a23d

dockernet/ts-tests/test/utils.ts
assafmo: moved to stridejs https://github.com/Stride-Labs/stridejs/commit/2e3a23d

dockernet/ts-tests/test/main.test.ts
assafmo: I agree

PR #1251, opened 2024-07-25
updated airdrop query to include the date index and airdrop length

sampocs:
## Context
Modifies the existing airdrop query to return the airdrop length and date index for the frontend. This should be non-state breaking!

## Brief Changelog
* Updated the airdrop query to return the current date index and airdrop length
* Removed the current date index from the user summary query
* Added the prod scripts that Kier's been using
* Parameterized the dockernet startup so that we can change the period length
* Renamed `windowLengthSeconds` → `periodLengthSeconds`
* Updated unit tests


shellvish (approved):

PR #1255, opened 2024-08-06
Increase Memo, lower Receiver max char lengths

PoolPirate:

The character limits for IBC packets are necessary, however the current size limit of 2000 chars is too small.
There are valid and reasonable [transactions](https://www.mintscan.io/stride/tx/2A9CE9643F9386525CAB715E38A01CBD11FA3B24EC1B4EECCDE541C994786502)using ibc:hooks that can exceed that limit.

On the other hand the receiver field limit is set too high as there is no real reason for addresses to even reach the 2000 char limit.

## Brief Changelog

- Increase max memo length 2000 -> 4096
- Decrease max receiver length 2000 -> 1024

## Author's Checklist

I have...

(Don't have a local test setup for these, would be nice if they could be ran by somebody else)

## Reviewers Checklist

I have...

  - [x] Does this pull request introduce a new feature or user-facing behavior changes? 
      - [ ] not applicable
      - [ ] jira ticket `XXX` 
      - [x] README.md 
      - [ ] not documented

### DIFF
--- a/x/autopilot/README.md
+++ b/x/autopilot/README.md
@@ -49,6 +49,8 @@ Note: This will support more functions that can reduce number of users' operatio
 ### A Note on Parsing
 Since older versions of IBC do not have a `Memo` field, they must pass the routing information in the `Receiver` attribute of the IBC packet. To make autopilot backwards compatible with all older IBC versions, the receiver address must be specified in the JSON string. Before passing the packet down the stack to the transfer module, the address in the JSON string will replace the `Receiver` field in the packet data, regardless of the IBC version.
 
+The module also enforces a maximum length for both the `Memo` and `Receiver` fields of 4096 and 1024 characters respectively.
+
 ## Params
 
 ```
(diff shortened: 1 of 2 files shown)


sampocs (commented):
Thanks for flagging this! 

The memo change looks good, but we might have some host zones who are still on older versions of IBC without a memo field (in which case we need to use the receiver as the memo)

I’ll look up what IBC versions our host zones are on and then circle back!

sampocs:
@PoolPirate how did you arrive at the value of 4096?

Also, do you see any reason to support a receiver address over 100 characters?

PoolPirate:
The longest call chains I could reasonably find so far with skip go were in the 3000 character range. No particular reason for the 96 other than it being 2^12 tho. 

Don't see any reason not to decrease the receiver further to 100 chars if more is no longer needed by autopilot.

sampocs:
Closing in favor of #1273 

thanks again @PoolPirate!

PR #1256, opened 2024-08-06
integration test initial network setup

sampocs:
This is still very much a work in progress, but figured it'd be good to checkpoint and get something merged in before continuing! 

**Note to review: I'd recommend reviewing commit by commit!**

## Context
The main motivation for the new integration test framework is to have the ability to easily spin up multiple networks in parallel to accommodate different testing setups. 

The setup for each suite of tests are often mutually exclusive (e.g. some tests need long epochs, some tests need short epochs). And even for those that are not mutually exclusive, there's a compounding effect that comes with trying to use a single network across different tests. 

For instance, the dydx trade route has an involved startup cost (requires stride, dydx, osmo, noble, and relayers across each pair). There's no reason the airdrop tests should have to wait for this setup!

Our current integration test setup (dockernet), while great for iterative development, fell short of these requirements, and also made it difficult to write tests (since it was in bash). The new network framework introduced in this PR, coupled with the new testing client interface in #1244 (by @assafmo), will make it much easier to write tests and integrate them in our CI. 

## Brief Changelog
* Updated `add-comsumer-section` (testing cli command for ICS vals) to make it more flexible. It was previously hard coded to work only with dockernet
  * Now you can pass in either the validator home directories or the validator public keys directly, instead of having it look for directories named `stride1`, `stride2`, etc.
* Added stateful set validators using helm templating to make it easy to add different chains
* Added scripts to create the genesis.json file (`init-chain.sh`), start each validator (`init-node.sh`) and create validators/governors (`create-validators.sh`)
  * `init-chain.sh` and `init-node.sh` run in the pod's init container. `create-validators.sh` runs in a post start hook (since the network needs to be up first)
* Added API service to upload and download shared files (see `README` for explanation of design decision here)

## Next steps
* Add relayers
* Generalize/parameterize dockerfile setup for host zones
* Add an additional abstraction layer so that multiple networks can be started
* Setup the ingress with a wildcard so there's no need to add DNS entries for each chain


assafmo (approved):

assafmo (approved):
LGTM!

PR #1259, opened 2024-08-08
Batch undelegation integration tests

assafmo:
- Happy path
  - [x] step 1: liquid stake
  - [ ] step 2: trigger a batch undelegation
- TODO: failure paths


dockernet/ts-tests/test/main.test.ts
```diff
@@ -77,23 +82,27 @@ beforeAll(async () => {
...
+          resolveIbcResponsesCheckIntervalMs: 50,
+        }),
+      };
+    }
   }
+  console.log("waiting for stride to start...");
+  await waitForChain(accounts.user, "ustrd");
+
+  console.log("waiting for gaia to start...");
+  await waitForChain(gaiaAccounts.user, "uatom");
```
assafmo, line 105: just a refactor of the "waitForChain" logic

sampocs: were you running into any issues with the approach that checked for the latest block? Seems a bit more elegant that way rather than submitting a tx

PR #1272, opened 2024-08-27
removed submitTxsDayEpoch

sampocs:
## Context
The only ICA we submit that times-out at the day epoch is the undelegation ICA. However, in practice, it acts as if it times out at the stride epoch.

The reason for this is delegation ICAs are submitted immediately after the undelegation ICA (timing out at the end of the stride epoch), so if an undelegation ICA was not relayed by then, the delegation ICA would time out and close the channel.

Considering this, it makes more sense to keep our epochly time out's consistent.

This PR updates the undelegation ICA to timeout at the stride epoch, and removes the day epoch function.

**I'd recommend reviewing commit by commit**

## Brief Changelog
* Updated undelegation ICA to timeout at the stride epoch
* Removed `SubmitTxsDayEpoch`
* Combined `SubmitTxsEpoch` and `SubmitTxsStrideEpoch`
* Renamed `SubmitTxs` to `SubmitICATx`
* Renamed `SubmitTxsStrideEpoch` to `SubmitICATxStrideEpoch`
* Fixed error messages during submission


sampocs:
I actually don't think this works cause the day epoch triggers before the stride epoch. So i think when looking up the stride epoch for the timeout, it will grab the wrong one.

Closing this for now as this change isn't strictly necessary and it seems a bit risky to try to fix it at this point.

--

In the future, I think we should just remove the day epoch entirely, and just do everything on the same epoch off intervals (e.g. unbond at stride epoch % 4). Would make it a lot easier to test too since we could trigger both delegations and unbondings on the same epoch (instead of having to stagger them)

cc @riley-stride

PR #1275, opened 2024-08-27
changed osmosis swap type URL

sampocs:
## Context
The dydx trade route swapper current receives authz permissions to execute message `/osmosis.gamm.v1beta1.MsgSwapExactAmountIn`. However, this is a legacy message. This PR updates the permissions to use the pool manager message instead.

## Brief Changelog
* Removed osmosis protos (no longer needed since we don't have to actually build the message)
* Updated osmosis type url to the pool manager one

**NOTE**: We'll want to revoke the gamm permissions right before the upgrade and then grant this new message type right after the upgrade


shellvish (approved):

sampocs:
@shellvish Added a `--legacy` flag to allow us to still revoke the gamm type url if you can review the latest commit! https://github.com/Stride-Labs/stride/pull/1275/commits/392be8eb3969c6be45a60450a951c91627697637

shellvish (approved):
This looks great, thanks for doing this!

PR #1277, opened 2024-08-29
added log to undelegate icacallback

sampocs:


shellvish (approved):
Very elegant solution

PR #1280, opened 2024-08-29
loosened outer redemption rate bounds for all host zones

sampocs:


shellvish (approved):
Perfect, thanks for doing this!

PR #1286, opened 2024-09-05
Added celestia validators

sampocs:
## Context
Adding celestia validator set for ICA migration. Each validator should start with a 0 delegation (since they only track the delegations from the ICA account)


sampocs:
Open

sampocs:
Open

sampocs:
Open

shellvish (approved):
This looks great, thank you for doing this Sam! Verified on my end that the list of validators looks good too

PR #1308, opened 2024-11-26
Increment Rebalance Period by 1 Day

shellvish:
This PR increments the rebalance period to (`unbonding_period + 1`), to avoid any race conditions with the timing of rebalance ICAs


x/stakeibc/keeper/rebalance.go
```diff
@@ -38,7 +38,7 @@ func (k Keeper) RebalanceAllHostZones(ctx sdk.Context) {
 	}
 
 	for _, hostZone := range k.GetAllActiveHostZone(ctx) {
-		if dayEpoch.EpochNumber%hostZone.UnbondingPeriod != 0 {
+		if dayEpoch.EpochNumber%(hostZone.UnbondingPeriod+1) != 0 {
```
sampocs, line 41: mind just adding a comment as well so we don't forget (can either add it here, in the docstring, or in hooks)

sampocs (approved):
thank you!

x/stakeibc/keeper/rebalance.go
shellvish: Great idea, done!

PR #1318, opened 2024-12-17
Buyback and Burn

assafmo:
Unresolved discussions:
- https://github.com/Stride-Labs/stride/pull/1318#discussion_r1888669165


x/auction/keeper/msg_server.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+
+	return &types.MsgPlaceBidResponse{}, nil
+}
+
+// CreateAuction creates a new auction
+func (ms msgServer) CreateAuction(goCtx context.Context, msg *types.MsgCreateAuction) (*types.MsgCreateAuctionResponse, error) {
+	ctx := sdk.UnwrapSDKContext(goCtx)
+
+	// TODO check admin
```
assafmo, line 40: adding a comment so we don't forget this

x/icqoracle/keeper/msg_server.go
```diff
@@ -0,0 +1,64 @@
...
+	}
+
+	return &types.MsgRegisterTokenPriceQueryResponse{}, nil
+}
+
+// RemoveTokenPriceQuery removes a token from price tracking
+func (ms msgServer) RemoveTokenPriceQuery(goCtx context.Context, msg *types.MsgRemoveTokenPriceQuery) (*types.MsgRemoveTokenPriceQueryResponse, error) {
+	ctx := sdk.UnwrapSDKContext(goCtx)
+
+	// TODO check admin
```
assafmo, line 54: adding a comment so we don't forget this

proto/stride/auction/auction.proto
```diff
@@ -0,0 +1,59 @@
...
+  string payment_denom = 4;
+
+  // Whether auction is active
+  bool enabled = 5;
+
+  // Price multiplier (e.g. 0.95 for 5% discount)
+  string price_multiplier = 6 [
+    (gogoproto.customtype) = "cosmossdk.io/math.LegacyDec",
+    (gogoproto.nullable) = false
+  ];
```
assafmo, line 37: Sam:
i think we had a thread in the gdoc about this. "multiplier" is underspecified and kinda confusing imo - I think it should indicate that it's a lower bound

my preference would be either MinPriceMultiplier or FloorPriceMultiplier

fwiw, when we move onto different auction types, this field will still be relevant as the floor (and the term "discount" would make less sense)


Assaf:
I don't like "min" and "floor". They both sound like this multiplier number can be this or higher, which in my view is confusing.

I also agree that priceMultiplier is not that good, but at least it specifies what this number is used for.

claude suggests price_adjustment_factor, wdyt?

x/icqoracle/keeper/genesis.go
```diff
@@ -0,0 +1,32 @@
...
+	"github.com/Stride-Labs/stride/v24/x/icqoracle/types"
+)
+
+// Loads module state from genesis
+func (k Keeper) InitGenesis(ctx sdk.Context, genState types.GenesisState) {
+	k.SetParams(ctx, genState.Params)
+	for _, tokenPrice := range genState.TokenPrices {
+		tokenPrice.SpotPrice = math.LegacyZeroDec()
+		tokenPrice.UpdatedAt = time.Time{}
+		tokenPrice.QueryInProgress = false
```
assafmo, line 18: Sam:

i don't think you want to override these

The way I think about this function is, if we (hypothetically) ever had to fork the chain, we would want to export the full state and then initialize the new chain with that state. So in this world, I don't think we would want to mess with the fields.

Wdyt?

assafmo: Assaf:

This function is only called on genesis, where we don't even have IBC set up. In my mind these values are definitely stale at this point.

I guess there's no real reason to reset SpotPrice & UpdatedAt because there's also PriceExpirationTimeoutSec to make sure they are not used when stale. As for QueryInProgress, there's definitely no IBC at this point so I feel confident we can reset it and prevent potential issues down the road.

proto/stride/icqoracle/icqoracle.proto
```diff
@@ -0,0 +1,67 @@
...
+  // Token denom on Stride
+  string base_denom = 1;
+  // Quote denom on Stride
+  string quote_denom = 2;
+  // Token denom on Osmosis
+  string osmosis_base_denom = 3;
+  // Quote denom on Osmosis
+  string osmosis_quote_denom = 4;
+  // Pool ID on Osmosis
+  string osmosis_pool_id = 5;
```
assafmo, line 20: @sampocs sampocs Dec 17, 2024

i wonder if it's worth adding a human readable name to this too? or maybe we can just enrich it with the full denom trace during the RPC query

Thinking being that if we do this with all our host zone tokens, then we'll have a bunch of IBC denom's and it gets pretty clunky to try and manually inspect things

@assafmo assafmo Dec 17, 2024

yeah i don't like the fields here too, i'll try to think of a better way to represent this

proto/stride/icqoracle/icqoracle.proto
```diff
@@ -0,0 +1,67 @@
...
+  uint64 update_interval_sec = 3 [
+    (gogoproto.moretags) = "yaml:\"update_interval_sec\"",
+    (gogoproto.jsontag) = "update_interval_sec"
+  ];
+
+  // Max time before price is considered stale/expired
+  uint64 price_expiration_timeout_sec = 4 [
+    (gogoproto.moretags) = "yaml:\"price_expiration_timeout_sec\"",
+    (gogoproto.jsontag) = "price_expiration_timeout_sec"
+  ];
```
assafmo, line 60: @sampocs sampocs Dec 17, 2024

fwiw i kinda liked "stale" but no strong feelings haha

assafmo Dec 17, 2024

it was kinda confusing for me, i guess as a non native english speaker

x/auction/client/cli/tx.go
```diff
@@ -0,0 +1,180 @@
...
+`, version.AppName, types.ModuleName),
+		),
+		Args: cobra.ExactArgs(4),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			clientCtx, err := client.GetClientTxContext(cmd)
+			if err != nil {
+				return err
+			}
+
+			enabled := args[1] == "true"
```
assafmo, line 152: @sampocs sampocs Dec 17, 2024

nit: i think you can do ParseBool

assafmo Dec 17, 2024

parseBool is more lines and less easter eggs
e.g. "stfu" => false can sometimes brighten your day
otoh "tru" => false can f up your day

x/auction/keeper/auction_type.go
```diff
@@ -0,0 +1,103 @@
...
+
+// Map of auction types to their handlers
+var bidHandlers = map[types.AuctionType]AuctionBidHandler{
+	types.AuctionType_AUCTION_TYPE_FCFS: fcfsBidHandler,
+}
+
+// fcfsBidHandler handles bids for First Come First Serve auctions
+func fcfsBidHandler(ctx sdk.Context, k Keeper, auction *types.Auction, bid *types.MsgPlaceBid) error {
+	// Get token amount being auctioned off
+	moduleAddr := k.accountKeeper.GetModuleAddress(types.ModuleName)
```
assafmo, line 22: sampocs Dec 17, 2024

should we use a module account per auction? it would make it easier to support multiple auctions with the same denom

assafmo Dec 17, 2024

Yes, but not sure that we want to do it now. I've thought about it a bit, and for the MVP this will complicate the way we send fees to be auctioned off. For each fee token we'll have to map on-chain what auction gets what percentage.

Lmk what you think, I'll do what you decide because I don't have a strong feeling here

x/icqoracle/types/genesis.go
```diff
@@ -0,0 +1,33 @@
...
+	for i, tokenPrice := range gs.TokenPrices {
+
+		msg := NewMsgRegisterTokenPriceQuery(
+			"stride1palmssweatykneesweakarmsareheavy8ahm9u", // dummy address, not stored in token price
+			tokenPrice.BaseDenom,
+			tokenPrice.QuoteDenom,
+			tokenPrice.OsmosisPoolId,
+			tokenPrice.OsmosisBaseDenom,
+			tokenPrice.OsmosisQuoteDenom,
+		)
```
assafmo, line 26: @sampocs sampocs Dec 17, 2024

this is kinda hacky haha, you could just separate out the checks into its own function that's called by both this and validate basic

x/icqoracle/keeper/icq.go
```diff
@@ -0,0 +1,152 @@
...
+		Id:              queryId,
+		ChainId:         params.OsmosisChainId,
+		ConnectionId:    params.OsmosisConnectionId,
+		QueryType:       icqtypes.CONCENTRATEDLIQUIDITY_STORE_QUERY_WITH_PROOF,
+		RequestData:     icqtypes.FormatOsmosisKeyPool(osmosisPoolId),
+		CallbackModule:  types.ModuleName,
+		CallbackId:      ICQCallbackID_OsmosisClPool,
+		CallbackData:    tokenPriceBz,
+		TimeoutDuration: time.Duration(params.IcqTimeoutSec) * time.Second,
+		TimeoutPolicy:   icqtypes.TimeoutPolicy_RETRY_QUERY_REQUEST,
```
sampocs, line 89: not sure if we need to have this since I assume we'll just be retrying them right after the timeout

x/auction/keeper/msg_server.go
assafmo: @sampocs 3f51aa61dec4a72655c355d48cfafa0c0350b27d

x/auction/keeper/msg_server.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+
+	return &types.MsgCreateAuctionResponse{}, nil
+}
+
+// CreateAuction updates an existing auction
+func (ms msgServer) UpdateAuction(goCtx context.Context, msg *types.MsgUpdateAuction) (*types.MsgUpdateAuctionResponse, error) {
+	ctx := sdk.UnwrapSDKContext(goCtx)
+
+	// TODO check admin
```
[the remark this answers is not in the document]
assafmo: @sampocs 3f51aa61dec4a72655c355d48cfafa0c0350b27d

x/icqoracle/keeper/msg_server.go
```diff
@@ -0,0 +1,64 @@
...
+	return &msgServer{Keeper: keeper}
+}
+
+var _ types.MsgServer = msgServer{}
+
+// RegisterTokenPriceQuery registers a token to price tracking
+func (ms msgServer) RegisterTokenPriceQuery(goCtx context.Context, msg *types.MsgRegisterTokenPriceQuery) (*types.MsgRegisterTokenPriceQueryResponse, error) {
+	ctx := sdk.UnwrapSDKContext(goCtx)
+
+	// TODO check admin
```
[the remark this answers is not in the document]
assafmo: @sampocs 3f51aa61dec4a72655c355d48cfafa0c0350b27d

x/icqoracle/keeper/msg_server.go
assafmo: @sampocs 3f51aa61dec4a72655c355d48cfafa0c0350b27d

x/auction/client/cli/tx.go
assafmo: @sampocs4419b8ca808217c1776954e9f36c8f14914b56d6

x/icqoracle/keeper/icq.go
assafmo: be789efa31c628e6faf5eb2d528a6a03201f5a2b

x/icqoracle/types/genesis.go
assafmo: @sampocs 4188ddafa

proto/stride/icqoracle/icqoracle.proto
sampocs: i think either's fine! feel free to keep expiration

proto/stride/icqoracle/icqoracle.proto
sampocs: i think internally we should just keep it as is cause there's already so much on the struct. But I think we could add a query that figures out the denom traces

x/auction/keeper/auction_type.go
sampocs: can you say more on why we would have to figure out percentages? 

The way I was thinking about it was that we would make a module account using the auction name that gets created when you create an auction. And then in stakeibc or when we send fees, we would do something like:
```
for denom in reward collector balance:
    auctionName = f"stakeibc-fees-{denom}"
    auction, found = k.GetAuction(auctionName)
    if not found:
       continue

    k.SendCoins(from: rewardCollector, to: auction.Address)
```

That said, to your point, I don't think it has to be in the MVP, but if it's simple enough like that, I think we should do it

x/icqoracle/keeper/genesis.go
sampocs: > In my mind these values are definitely stale at this point.

Agreed with that

---

However, this function is also relevant when we do state exported testnets when we start up a forked chain with the current state. In this case, we would notice there are some differences (which we could visually verify are fine, but still kind of annoying to deal with).

I think it makes more sense to just keep values as what they are. The genesis comes from the JSON which is hard coded.In practice it will never be used, but in theory, if we had specific values set on the JSON, that probably means we want to use them (otherwise, we would just scrape them ourselves)

proto/stride/auction/auction.proto
sampocs: > They both sound like this multiplier number can be this or higher, which in my view is confusing.

I think the important point though is that the price from the bids _can_ be higher than this value.

I just asked claude too and he spit out the following (I never used the word floor in my prompt)

```
minimumBidRatio
minPriceMultiplier
floorPriceRatio
bidPriceThreshold
priceDiscountFactor
minimumPricePercentage
reservePriceRatio
bidFloorMultiplier
priceFloorFactor
minimumBidPercentage

Some related terms commonly used in auction terminology:

"Floor price" is a common auction term for minimum acceptable price
```

my vote is `minPriceMultiplier` so you can keep your `multiplier` and i keep my `min` lol 😄

x/auction/types/keys.go
```diff
@@ -0,0 +1,20 @@
...
+	// RouterKey defines the routing key
+	RouterKey = ModuleName
+
+	ParamsPrefix     = "params"
+	KeyAuctionPrefix = "auction"
+)
+
+func AuctionKey(auctionName string) []byte {
+	return []byte(fmt.Sprintf("%s|%s", KeyAuctionPrefix, auctionName))
+}
```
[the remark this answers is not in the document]
sampocs: the pattern i've usually seen is to do:

```
store := prefix.NewStore(ctx.KVStore(k.storeKey), types.KeyAuctionPrefix)
key := []byte(auctionName)
```

Slight preference to stick with that and use the prefix store, but either are technically valid

proto/stride/auction/auction.proto
assafmo: never thought about it this way and now I fully agree that this is actually providing the floor price for bids

assafmo: 59a9e7852

x/icqoracle/keeper/genesis.go
assafmo: okay, and what about `QueryInProgress`?

x/auction/keeper/auction_type.go
assafmo: My thinking is that it's possible there will be multiple auctions with the same selling token. E.g. maybe in the future we'll want to create 2 auctions for ATOM, one Dutch and the other English, and send 80% of ATOM to the Dutch auction and 20% to the English auction. To implement something like that we can add a `priority` field to auction and then have stakeibc calculate the percentages from the priorities. E.g. priorities of `(1,2,3) translate to percentages `(1/6,2/6,3/6)`.

I think with any of these approaches (all in one account, each auction has an account, each auction has an account + multiple auctions can use same selling denom) we could always migrate to a different approach later on, so I'm not too worried here.

Let's just decide what we want for the MVP and move on.

x/auction/types/keys.go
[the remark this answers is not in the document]
assafmo: much better 3b9150fe0

x/icqoracle/keeper/genesis.go
sampocs: same deal, my personal opinion is that if we had it set to true for some reason in genesis.json, we should keep it like that (otherwise we would have manually set it to false)

x/icqoracle/types/keys.go
```diff
@@ -0,0 +1,26 @@
...
+)
+
+var (
+	ParamsKey        = []byte("params")
+	PriceQueryPrefix = []byte("pricequery")
+)
+
+func TokenPriceQueryKey(baseDenom, quoteDenom, poolId string) []byte {
+	return []byte(fmt.Sprintf("%s|%s|%s", baseDenom, quoteDenom, poolId))
+}
```
sampocs, line 22: should this be `TokenPriceKey` instead to line up with the keepers and struct type?

x/icqoracle/types/keys.go
```diff
@@ -0,0 +1,26 @@
...
+
+func TokenPriceByDenomKey(baseDenom string) []byte {
+	return []byte(baseDenom)
+}
```
sampocs, line 26: I think you can remove this and just do the `[]byte(TokenPrice.BaseDenom)` in the keepers

x/auction/keeper/auction_type.go
sampocs: Imo, even in that world of split percentages, it should be the onus of the calling module (i.e. stakeibc) to determine those splits ahead of time - so not something the auction module has to take care of. I also think these two are the same: "each auction has an account" and "each auction has an account + multiple auctions can use same selling denom" cause if you have a different account, the denom doesn't matter.

I'll defer to you here since you're heads more in this / you have a better sense of how much work it would be to split accounts. My mental model with stuff like this is that I find small changes that you know for sure you'll want to make are often better done during the v0 since you're heads most into it and it's a little easier to add it now than, say, 6 months from now where you have to bring context back. 

So if (a) you're aligned this is the best solution, (b) it would be quick to implement and (c) it doesn't increase the testing complexity too much, then I would say probably makes sense to add it. Otherwise, I think punting it is fine. Defer to you!

assafmo:
unstale

x/icqoracle/keeper/keeper.go
```diff
@@ -0,0 +1,245 @@
...
+				// Calculate the price of 1 baseToken in quoteToken
+				price = baseTokenPrice.SpotPrice.Quo(quoteTokenPrice.SpotPrice)
+				break
+			}
+		}
+	}
+
+	if price.IsZero() {
+		return math.LegacyDec{}, fmt.Errorf(
+			"could not calculate price for baseToken='%s' quoteToken='%s' (foundCommonQuoteToken='%v', foundBaseTokenStalePrice='%v', foundQuoteTokenStalePrice='%v', foundQuoteTokenZeroPrice='%v', foundAlreadyHasStalePrice='%v')",
```
sampocs, line 213: police are here, break up the party

x/icqoracle/keeper/keeper.go
```diff
@@ -0,0 +1,245 @@
...
+
+	// Define flags to allow for better error messages
+	foundCommonQuoteToken := false
+	foundBaseTokenStalePrice := false
+	foundQuoteTokenStalePrice := false
+	foundQuoteTokenZeroPrice := false
+
+	// Find a common quote denom and calculate baseToken to quoteToken price
+	for quoteDenom1, baseTokenPrice := range baseTokenPrices {
+		for quoteDenom2, quoteTokenPrice := range quoteTokenPrices {
```
sampocs, line 184: might be a bit easier to follow if you rename `quoteDenom1/2` to `commonQuoteDenom1/2` or something of the like (otherwise it's easy to conflate with the actual quote Denom)

x/auction/keeper/auction_type.go
```diff
@@ -0,0 +1,118 @@
...
+	if err != nil {
+		return errorsmod.Wrapf(err, "error getting price for baseDenom='%s' quoteDenom='%s'", auction.SellingDenom, auction.PaymentDenom)
+	}
+
+	// Apply MinPriceMultiplier
+	bidsFloorPrice := price.Mul(auction.MinPriceMultiplier)
+
+	// if paymentAmount < sellingAmount * bidsFloorPrice
+	if bid.PaymentTokenAmount.ToLegacyDec().LT(bid.SellingTokenAmount.ToLegacyDec().
+		Mul(bidsFloorPrice)) {
```
sampocs, line 49:
```suggestion
	bidsFloorPrice := price.Mul(auction.MinPriceMultiplier)
	minPaymentRequired := bid.SellingTokenAmount.ToLegacyDec().Mul(bidsFloorPrice)

	// if paymentAmount < sellingAmount * bidsFloorPrice
	if bid.PaymentTokenAmount.ToLegacyDec().LT(minPaymentRequired) {
```

x/auction/keeper/auction_type.go
```diff
@@ -0,0 +1,118 @@
...
+		return fmt.Errorf("bid price too low: offered %s%s for %s%s, bids floor price is %s%s (price=%s %s/%s)",
+			bid.PaymentTokenAmount.String(),
+			auction.PaymentDenom,
+			bid.SellingTokenAmount.String(),
+			auction.SellingDenom,
+			bid.SellingTokenAmount.ToLegacyDec().Mul(bidsFloorPrice).String(),
```
sampocs, line 55: (after above change) 
```suggestion
			minPaymentRequired.string()
```

sampocs (commented):
Took another pass and it looks fantastic! Only substantive comment is around the ICQ'ing

I also see there's still some old comments from december, maybe on Monday we can hop on a call and cut through everything?

app/apptesting/test_helpers.go
```diff
@@ -81,7 +81,7 @@ type AppTestHelper struct {
 // AppTestHelper Constructor
 func (s *AppTestHelper) Setup() {
 	s.App = app.InitStrideTestApp(true)
-	s.Ctx = s.App.BaseApp.NewContext(false, tmtypesproto.Header{Height: 1, ChainID: StrideChainID})
+	s.Ctx = s.App.BaseApp.NewContext(false, tmtypesproto.Header{Height: 1, ChainID: StrideChainID, Time: time.Now().UTC()})
```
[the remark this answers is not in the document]
assafmo: to make sure price staleness checks work in tests

app/upgrades.go
```diff
@@ -399,6 +399,10 @@ func (app *StrideApp) setupUpgradeHandlers(appOpts servertypes.AppOptions) {
 		storeUpgrades = &storetypes.StoreUpgrades{
 			Added: []string{ibcwasmtypes.ModuleName, airdroptypes.ModuleName},
 		}
+	case "v25":
```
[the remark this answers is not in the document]
assafmo: c720da3d2 :grimacing:

app/app.go
```diff
@@ -759,6 +781,35 @@ func NewStrideApp(
...
+		appCodec,
+		keys[icqoracletypes.StoreKey],
+	)
+	icqOracleModule := icqoracle.NewAppModule(appCodec, app.ICQOracleKeeper)
+
+	// Register ICQ callbacks
+	err = app.InterchainqueryKeeper.SetCallbackHandler(icqoracletypes.ModuleName, app.ICQOracleKeeper.ICQCallbackHandler())
+	if err != nil {
+		return nil
+	}
```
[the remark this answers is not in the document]
assafmo: c8e0e1a81

app/upgrades/v26/upgrades.go
```diff
@@ -0,0 +1,39 @@
...
+) upgradetypes.UpgradeHandler {
+	return func(ctx sdk.Context, _ upgradetypes.Plan, vm module.VersionMap) (module.VersionMap, error) {
+		ctx.Logger().Info("Starting upgrade v26...")
+
+		err := icqoracleKeeper.SetParams(ctx, icqoracletypes.Params{
+			OsmosisChainId:            "osmosis-1",
+			OsmosisConnectionId:       "connection-2",
+			UpdateIntervalSec:         5 * 60,  // 5 min
+			PriceExpirationTimeoutSec: 10 * 60, // 10 min
+			IcqTimeoutSec:             2 * 60,  // 2 min
```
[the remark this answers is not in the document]
assafmo: Not sure I follow, IcqTimeoutSec refers to when to timeout the ibc query, UpdateIntervalSec is how often to send an ibc query.

dockernet/config.sh
```diff
@@ -35,7 +35,7 @@ ACCESSORY_CHAINS=()
 if [[ "${ALL_HOST_CHAINS:-false}" == "true" ]]; then 
   HOST_CHAINS=(GAIA EVMOS HOST)
 elif [[ "${#HOST_CHAINS[@]}" == "0" ]]; then 
-  HOST_CHAINS=(GAIA)
+  HOST_CHAINS=(GAIA OSMO)
```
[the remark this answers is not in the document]
assafmo: 5b956a815

dockernet/ts-tests/README.md
```diff
@@ -2,7 +2,7 @@
```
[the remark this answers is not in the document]
assafmo: 5b470c214

x/icqoracle/keeper/icq.go
```diff
@@ -0,0 +1,157 @@
...
+		return errorsmod.Wrapf(err, "Error converting osmosis pool id '%s' to uint64", tokenPrice.OsmosisPoolId)
+	}
+
+	tokenPriceBz, err := k.cdc.Marshal(&tokenPrice)
+	if err != nil {
+		k.Logger(ctx).Error(utils.LogWithTokenPriceQuery(tokenPrice.BaseDenom, tokenPrice.QuoteDenom, tokenPrice.OsmosisPoolId, "Error serializing tokenPrice '%+v' to bytes, error '%s'", tokenPrice, err.Error()))
+		return errorsmod.Wrapf(err, "Error serializing tokenPrice '%+v' to bytes", tokenPrice)
+	}
+
+	queryId := fmt.Sprintf("%s|%s|%s|%d", tokenPrice.BaseDenom, tokenPrice.QuoteDenom, tokenPrice.OsmosisPoolId, ctx.BlockHeight())
```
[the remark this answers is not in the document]
assafmo: oh wow this is actually getting ignored https://github.com/Stride-Labs/stride/blob/5b470c214/x/interchainquery/keeper/queries.go#L32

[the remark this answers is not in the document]
assafmo: 98470d7d5

x/icqoracle/keeper/keeper.go
assafmo: lol

x/icqoracle/keeper/abci.go
```diff
@@ -0,0 +1,49 @@
...
+		// Get last update time for this token
+		lastUpdate := tokenPrice.UpdatedAt
+
+		// Skip if there's already a query in progress
+		if tokenPrice.QueryInProgress {
+			continue
+		}
+
+		// If never updated or update interval has passed
+		if lastUpdate.IsZero() || currentTime.Sub(lastUpdate) >= time.Second*time.Duration(params.UpdateIntervalSec) {
```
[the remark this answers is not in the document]
assafmo: I don't like combining timeout time and update interval, but I do like timing out the query explicitly

x/icqoracle/keeper/keeper.go
assafmo: 823fdab09

x/auction/keeper/auction_type.go
assafmo: 210311ea3

x/auction/keeper/auction_type.go
assafmo: good eye!

app/apptesting/test_helpers.go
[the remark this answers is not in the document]
assafmo: nvm, reverted

a36b626e5

x/icqoracle/keeper/icq.go
```diff
@@ -0,0 +1,157 @@
...
+	return c
+}
+
+func (c ICQCallbacks) RegisterICQCallbacks() icqtypes.QueryCallbacks {
+	return c.
+		AddICQCallback(ICQCallbackID_OsmosisClPool, ICQCallback(OsmosisClPoolCallback))
+}
+
+// Submits an ICQ to get a concentrated liquidity pool from Osmosis' store
+func (k Keeper) SubmitOsmosisClPoolICQ(
```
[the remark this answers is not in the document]
sampocs: The error's are all logged in `abci` already though, so I don't think they'll get buried. 

In general, I think we should just always make sure the main calling function has a log and then instead of doing:
```
func inner():
   if err1: 
      k.Logger.Info("some inner error message", err1)
      return errorsmod("some inner error message", err1)
   if err2: 
      k.Logger.Info("some inner error message2", err2)
      return errorsmod("some inner error message2", err2)

func main():
   if err := inner():
      k.Logger.Info("some outer error message: %s", err) 
```

we should just do
```
func inner():
   if err1: 
      return errorsmod("some inner error message", err1)
   if err2: 
      return errorsmod("some inner error message2", err2)

func main():
   if err := inner():
      k.Logger.Info("some outer error message: %s", err) // has error message from inner
```

x/icqoracle/keeper/icq.go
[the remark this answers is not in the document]
sampocs: well to clarify - before it was getting ignored, but it was actually going down [this](https://github.com/Stride-Labs/stride/blob/5b470c214ecc2e6934f412ea84d504ca347d8562/x/interchainquery/keeper/queries.go#L30) route since forceUnique was true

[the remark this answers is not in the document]
sampocs: but to my other comment (and something we can discuss in a bit), I think we should make the ID _not_ unique

x/icqoracle/keeper/abci.go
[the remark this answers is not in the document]
sampocs: lets discuss live

x/icqoracle/keeper/genesis.go
assafmo: 058a06afc

proto/stride/icqoracle/icqoracle.proto
assafmo: 0452533dc

x/icqoracle/types/keys.go
assafmo: ded88301e

x/icqoracle/keeper/icq.go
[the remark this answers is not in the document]
assafmo: ea5aaefef

x/icqoracle/client/cli/tx.go
```diff
@@ -34,7 +34,7 @@ func GetTxCmd() *cobra.Command {
 
 func CmdAddTokenPrice() *cobra.Command {
 	cmd := &cobra.Command{
-		Use:   "add-token-price [base-denom] [quote-denom] [base-denom-decimals] [quote-denom-decimals] [osmosis-pool-id] [osmosis-base-denom] [osmosis-quote-denom]",
+		Use:   "add-token-price [base-denom] [quote-denom] [osmosis-pool-id] [osmosis-base-denom] [osmosis-quote-denom]",
```
assafmo, line 37: :crown:

PR #1319, opened 2024-12-17
Stride Burner Module

assafmo:
giphy-46


proto/stride/strdburner/query.proto
```diff
@@ -0,0 +1,26 @@
...
+
+option go_package = "github.com/Stride-Labs/stride/v24/x/strdburner/types";
+
+// Query defines the gRPC querier service.
+service Query {
+  // StrdBurnerAddress queries the address of the strdburner module
+  rpc StrdBurnerAddress(QueryStrdBurnerAddressRequest)
+      returns (QueryStrdBurnerAddressResponse) {
+    option (google.api.http).get = "/stride/strdburner/address";
+  }
```
sampocs, line 15: is there no query for total burned?

x/strdburner/keeper/abci.go
```diff
@@ -0,0 +1,20 @@
...
+	strdBurnerAddress := k.GetStrdBurnerAddress()
+
+	strdBalance := k.bankKeeper.GetBalance(ctx, strdBurnerAddress, "ustrd")
+
+	if strdBalance.IsPositive() {
+		err := k.bankKeeper.BurnCoins(ctx, types.ModuleName, sdk.NewCoins(strdBalance))
+		if err != nil {
+			k.Logger(ctx).Error("unable to burn %s", strdBalance.String())
+		}
+	}
```
sampocs, line 19: might make sense to also log when a burn was successful - shouldn't be too noisy since this will only happen once every 6 hours

sampocs (approved):
looks good! 

Does it make sense to track how much was burned and expose that with a query?

proto/stride/strdburner/query.proto
assafmo: good idea, will add

assafmo:
> Does it make sense to track how much was burned and expose that with a query?

good idea, will add!

PR #1321, opened 2024-12-19
Change reward distribution to auction off rewards

assafmo:
- Before: Liquid stake rewards and then send them to fee collector module
- After: Send rewards directly to auction module

### DIFF
--- a/x/stakeibc/keeper/reward_allocation.go
+++ b/x/stakeibc/keeper/reward_allocation.go
@@ -1,86 +1,32 @@
 package keeper
 
 import (
-	"fmt"
-
-	errorsmod "cosmossdk.io/errors"
 	sdk "github.com/cosmos/cosmos-sdk/types"
 
-	authtypes "github.com/cosmos/cosmos-sdk/x/auth/types"
-
+	auctiontypes "github.com/Stride-Labs/stride/v24/x/auction/types"
 	"github.com/Stride-Labs/stride/v24/x/stakeibc/types"
 )
 
-// Liquid Stake Reward Collector Balance
-func (k Keeper) LiquidStakeRewardCollectorBalance(ctx sdk.Context, msgSvr types.MsgServer) bool {
-	k.Logger(ctx).Info("Liquid Staking reward collector balance")
-	rewardCollectorAddress := k.AccountKeeper.GetModuleAccount(ctx, types.RewardCollectorName).GetAddress()
-	rewardedTokens := k.bankKeeper.GetAllBalances(ctx, rewardCollectorAddress)
-	if rewardedTokens.IsEqual(sdk.Coins{}) {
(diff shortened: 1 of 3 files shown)


assafmo:
destale

assafmo:
bump

x/stakeibc/keeper/reward_allocation.go
```diff
@@ -1,86 +1,32 @@
...
-		}
+	rewardCollectorBalances := k.bankKeeper.GetAllBalances(ctx, rewardCollectorAddress)
+	if rewardCollectorBalances.Empty() {
+		k.Logger(ctx).Info("No rewards to auction from RewardCollector")
+		return
 	}
-	k.Logger(ctx).Info(fmt.Sprintf("Sending %s stTokens from %s to %s", stTokens.String(), types.RewardCollectorName, authtypes.FeeCollectorName))
 
-	err := k.bankKeeper.SendCoinsFromModuleToModule(ctx, types.RewardCollectorName, authtypes.FeeCollectorName, stTokens)
+	err := k.bankKeeper.SendCoinsFromModuleToModule(ctx, types.RewardCollectorName, auctiontypes.ModuleName, rewardCollectorBalances)
```
sampocs, line 23: just flagging that we don't double check that there's an auction registered, but I think that's the right behavior!

Is the 2nd iteration of this when we plan to have separate module accounts for each auction?

sampocs (commented):
lgtm!

Did we ever discuss whether we should auction off the liquid staked tokens or the native tokens? Doesn't add a ton of complexity either way with the auction

x/stakeibc/keeper/reward_allocation.go
```diff
@@ -1,86 +1,32 @@
...
-	if rewardsFound := k.LiquidStakeRewardCollectorBalance(ctx, msgSvr); !rewardsFound {
-		k.Logger(ctx).Info("No accrued rewards in the reward collector account")
-		return
-	}
-	if err := k.SweepStTokensFromRewardCollToFeeColl(ctx); err != nil {
-		k.Logger(ctx).Error(fmt.Sprintf("Unable to allocate host zone reward, err: %s", err.Error()))
-	}
+// AllocateRewardsFromHostZones auctions off the reward collector balance
+func (k Keeper) AllocateRewardsFromHostZones(ctx sdk.Context) {
+	k.AuctionOffRewardCollectorBalance(ctx)
```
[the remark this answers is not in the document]
assafmo: dbdaf926d

x/stakeibc/keeper/reward_allocation.go
assafmo: IMO the next iteration will have a function in x/auction that gets a selling token and returns the address of the auction that handles this selling token.

PR #1335, opened 2025-01-28
fixed stuck lsm record

sampocs:
The LSM record cannot be detokenized due to an SDK rounding bug. To fix it, we can subtract 1 from the amount and retry it


shellvish (approved):

PR #1337, opened 2025-01-29
included delegation records in staketia migration check

sampocs:
Delegation records that have not been processed by the MS are not included in the RR calculation in the migration invariant check

### DIFF
--- a/x/staketia/keeper/migration.go
+++ b/x/staketia/keeper/migration.go
@@ -257,18 +257,18 @@ func GetStakeibcRedemptionRate(
 	// Staketia delegation records are not included in any of the above yet so we must
 	// gather than explicitly
 	// In practice, there will be no records here when the upgrade runs
-	staketiaUndelegatedBalance := sdk.ZeroDec()
+	staketiaPendingDelegations := sdk.ZeroDec()
 	for _, delegationRecord := range staketiaKeeper.GetAllActiveDelegationRecords(ctx) {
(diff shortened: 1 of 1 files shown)


x/staketia/keeper/migration.go
```diff
@@ -253,12 +254,21 @@ func GetStakeibcRedemptionRate(
 	undelegatedBalance := stakeibcKeeper.GetUndelegatedBalance(hostZone.ChainId, depositRecords)
 	nativeDelegation := sdk.NewDecFromInt(hostZone.TotalDelegations)
 
+	// Staketia delegation records are not included in any of the above yet so we must
+	// gather than explicitly
+	// In practice, there will be no records here when the upgrade runs
+	staketiaPendingDelegations := sdk.ZeroDec()
+	for _, delegationRecord := range staketiaKeeper.GetAllActiveDelegationRecords(ctx) {
```
shellvish, line 261: just checking: this counts records in `TRANSFER_IN_PROGRESS`, is that the right behavior?

shellvish (approved):
Looks great, thanks for doing this! Just one question

x/staketia/keeper/migration.go
shellvish: Makes sense! I was worried it'd double with the `GetDepositAccountBalance` check which also checks `TRANSFER_IN_PROGRESS` records, but I think that's just for `stakeibc`

PR #1339, opened 2025-01-29
fixed build docker image CI

sampocs:
The action was outdated and had a deprecated dependency causing this to fail. This PR updates the version and syntax


shellvish (approved):
Nice, good catch!

PR #1342, opened 2025-02-04
Updated dockernet hermes

sampocs:
## Context
For context, hermes is no longer used in dockernet, but is sometimes needed when adding new host zones. As a result, I left the startup script and dockerflie in there, but removed the local binary

## Brief Changelog
* Removed submodule dependency
* Removed local build step
* Added hermes scripts and configs from #1334 (when we last ran hermes)


assafmo (approved):
lgtm

PR #1344, opened 2025-02-04
updated buyback and burn ICQ logic

sampocs:
* Made query ID constant across queries
* Removed query timeout time (since we can just timeout at the update interval)
* Renamed `UpdateAt` to `LastQueryTime`


deps/osmosis
assafmo, line 1: what happened here?

x/icqoracle/keeper/keeper.go
```diff
@@ -67,6 +67,8 @@ func (k Keeper) SetTokenPriceQueryInProgress(ctx sdk.Context, baseDenom string,
 	}
 
 	tokenPrice.QueryInProgress = queryInProgress
+	tokenPrice.LastQueryTime = ctx.BlockTime()
```
assafmo, line 70: nice one

x/interchainquery/keeper/keeper.go
```diff
@@ -101,3 +101,14 @@ func (k *Keeper) RetryICQRequest(ctx sdk.Context, query types.Query) error {
...
+// Forces a query timeout of an in-progress query by removing it from the store
+func (k *Keeper) ForceQueryTimeout(ctx sdk.Context, queryId string) error {
+	query, found := k.GetQuery(ctx, queryId)
+	if !found {
+		return errorsmod.Wrapf(types.ErrQueryNotFound, "Query not found for %s", queryId)
+	}
+	EmitEventQueryResponse(ctx, query)
+	k.DeleteQuery(ctx, query.Id)
+	return nil
+}
```
assafmo, line 114: What about the timeout policy? I think we should call it, but not sure what's the best approach:

1. Add a 3rd param `timeoutPolicy TimeoutPolicy` and then do `query.TimeoutPolicy = timeoutPolicy` and then `k.HandleQueryTimeout()`. This way we could pass `TimeoutPolicy_REJECT_QUERY_RESPONSE` here.

```suggestion
// Forces a query timeout of an in-progress query by removing it from the store
func (k *Keeper) ForceQueryTimeout(ctx sdk.Context, queryId string, timeoutPolicy types.TimeoutPolicy) error {
	// Get query from store
	query, found := k.GetQuery(ctx, queryId)
	if !found {
		return errorsmod.Wrapf(types.ErrQueryNotFound, "Query not found for %s", queryId)
	}

	// Override TimeoutPolicy
	query.TimeoutPolicy = timeoutPolicy

	// Tell relayers this query has been handled
	EmitEventQueryResponse(ctx, query)

	// Update store
	k.DeleteQuery(ctx, query.Id)

	// Handle timeout acording to timeout policy
	// Note: dummyMsg.Result is used to call the callback in case of TimeoutPolicy_EXECUTE_QUERY_CALLBACK
	dummyMsg := &types.MsgSubmitQueryResponse{}
	err := k.HandleQueryTimeout(ctx, dummyMsg, query)
	if err != nil {
		return err
	}

	return nil
}
```

2. Add a 3rd param `enforceTimeoutPolicy bool` and do nothing if it's `false`, otherwise call `k.HandleQueryTimeout()`.

```suggestion
// Forces a query timeout of an in-progress query by removing it from the store
func (k *Keeper) ForceQueryTimeout(ctx sdk.Context, queryId string, enforceTimeoutPolicy bool) error {
	// Get query from store
	query, found := k.GetQuery(ctx, queryId)
	if !found {
		return errorsmod.Wrapf(types.ErrQueryNotFound, "Query not found for %s", queryId)
	}

	// Tell relayers this query has been handled
	EmitEventQueryResponse(ctx, query)

	// Update store
	k.DeleteQuery(ctx, query.Id)

	if enforceTimeoutPolicy {
		// Handle timeout acording to timeout policy
		// Note: dummyMsg.Result is used to call the callback in case of TimeoutPolicy_EXECUTE_QUERY_CALLBACK
		dummyMsg := &types.MsgSubmitQueryResponse{}
		err := k.HandleQueryTimeout(ctx, dummyMsg, query)
		if err != nil {
			return err
		}
	}

	return nil
}
```

deps/osmosis
sampocs: oh my b, must have accidentally committed it, will revert

x/interchainquery/keeper/keeper.go
sampocs: HandleQueryTimeout is just a no-op if the policy is REJECT_QUERY_RESPONSE (which is what we want)

But per my other message in slack, I realized right as I was going to bed that we don't actually have to time this out. Going to revert

x/icqoracle/types/genesis.go
```diff
@@ -11,7 +11,7 @@ func DefaultGenesis() *GenesisState {
 
 // Performs basic genesis state validation by iterating through all token prices and validating
 // using ValidateTokenPriceQueryParams().
-// We ignore the SpotPrice, UpdatedAt & QueryInProgress fields since they are reset in InitGenesis().
+// We ignore the SpotPrice, LastRequestTime & QueryInProgress fields since they are reset in InitGenesis().
```
assafmo, line 14: this line should actually by removed

```suggestion
```

x/icqoracle/keeper/abci.go
```diff
@@ -7,40 +7,31 @@ import (
...
-		}
-
-		// If never updated or update interval has passed
-		if lastUpdate.IsZero() || currentTime.Sub(lastUpdate) >= time.Second*time.Duration(params.UpdateIntervalSec) {
-			// Update price for this specific token
-			err := k.SubmitOsmosisClPoolICQ(ctx, tokenPrice)
-			if err != nil {
-				// Can't really do anything but log
+		// If never updated or update interval has passed, submit a new query for the price
+		// If a query was already in progress, it will be replaced with this new one that will
```
assafmo, line 25:
```suggestion
// If a query was already in progress, it will be replaced with this new one that
```

PR #1346, opened 2025-02-04
Osmo pool id type

sampocs:
Updated osmo pool ID to be a uint instead of a string since that's the type the osmo util expects


assafmo (approved):
I made it a string bc we never use it as a number, just pass the string along, but whatever

sampocs:
the way I was looking at it was that the only time we really use it is [here](https://github.com/Stride-Labs/stride/pull/1346/files#diff-d674df6400bade52fd35a386ead0b588b6cdf4b5219511d840f8478498bcc523L69-L72), and then in logging and the keepers. And the logging/keepers can work with either type

assafmo (approved):

PR #1347, opened 2025-02-04
add txBytes support to `strided tx decode` in addition to signBytes

assafmo:
- Attempts primary decoding signBytes with TxDecoder
- Falls back to direct protobuf unmarshaling (txBytes) if primary decode fails

This improvement provides better handling of different transaction encoding formats and increases decode reliability.

This also add support for decoding the output of mempool errors.


assafmo:
No downsides. I also [upstreamed it](https://github.com/cosmos/cosmos-sdk/pull/23611), hopefully will be merged soon.

PR #1354, opened 2025-02-06
bbb integration tests

assafmo:


integration-tests/network/scripts/init-chain.sh
```diff
@@ -148,6 +152,13 @@ update_stride_genesis() {
 # Genesis updates specific to non-stride chains
 update_host_genesis() {
     echo "Updating genesis.json with host configuration..."
+
+    if [[ "$CHAIN_NAME" == "osmosis" ]]; then
+        strd_on_osmo="ibc/FF6C2E86490C1C4FBBD24F55032831D2415B9D7882F85C3CC9C2401D79362BEA"
+        atom_on_osmo="ibc/6CDD4663F2F09CD62285E2D45891FC149A3568E316CE3EBBE201A71A78A69388" # through stride
+        jq_inplace '.app_state.concentratedliquidity.params.is_permissionless_pool_creation_enabled |= true' $genesis_json 
+        jq_inplace '.app_state.concentratedliquidity.params.authorized_quote_denoms += ["uosmo", "'$strd_on_osmo'", "'$atom_on_osmo'"]' $genesis_json
```
assafmo, line 160: This gets truncated somewhere, probably in osmo's genesis

assafmo: We can circumvent this by using osmo as token 1

x/icqoracle/keeper/icq.go
```diff
@@ -19,7 +19,7 @@ import (
 )
 
 const (
-	ICQCallbackID_OsmosisClPool = "osmosisclpool"
+	ICQCallbackID_OsmosisPool = "osmosispool"
```
sampocs, line 22: should we rename to OsmosisPrice?

x/icqoracle/keeper/query.go
```diff
@@ -120,6 +120,7 @@ func (k Keeper) TokenPriceToTokenPriceResponse(ctx sdk.Context, tokenPrices ...t
 			OsmosisPoolType:     tokenPrice.OsmosisPoolType,
 			SpotPrice:           tokenPrice.SpotPrice,
 			LastRequestTime:     tokenPrice.LastRequestTime,
+			LastResponseTime:    tokenPrice.LastResponseTime,
```
sampocs, line 123: ahh! 

I'm starting to think we should just change the response struct to be:
```
tokenPriceResult:
   tokenPrice: 
     ... struct in data
   unwrappedBaseDenom
   unwrappedQuoteDenom
```

Then we wouldn't have to keep making sure these were in sync

x/icqoracle/keeper/icq.go
```diff
@@ -69,11 +69,14 @@ func (k Keeper) SubmitOsmosisPoolICQ(
 		return errorsmod.Wrapf(err, "Error serializing tokenPrice '%+v' to bytes", tokenPrice)
 	}
 
+	var queryType string
 	var requestData []byte
 	switch tokenPrice.OsmosisPoolType {
 	case types.GAMM:
+		queryType = icqtypes.GAMM_STORE_QUERY_WITH_PROOF
```
sampocs, line 76: great catch!

x/icqoracle/keeper/query.go
assafmo: good idea

sampocs: its just annoying that it's not at the same indentation level

PR #1373, opened 2025-03-04
add utils.SafeSendCoins() and update various modules to use it

assafmo:
didn't change tests and past upgrade handlers

### DIFF
--- a/utils/safe_send_coins.go
+++ b/utils/safe_send_coins.go
@@ -1,0 +1,20 @@
+package utils
+
+import (
+	errorsmod "cosmossdk.io/errors"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
+)
+
+type BankKeeper interface {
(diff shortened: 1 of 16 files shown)


sampocs (commented):
looks good! 

I'm wondering if it would make more sense to just parameterize safe send coins with a `checkBlockedAddr` bool so that we could use it for every `SendCoins` instance instead of having to add the comment that says why we didn't use it

assafmo:
Good point. I'm not sure if it'll be more readable tho, from the outside it'll just be true/false. I'll do it and see how it looks.

Also, wdyt about adding a ci job to check that we're using SendCoins anywhere in the future?

sampocs:
Ooo CI job is a great idea! (assuming we stick with the switch to using SafeSend everywhere)

assafmo:
Yeah with the switch.

sampocs (commented):
Looks good! one last request - I would revert the test files and change the CI job to ignore test files. I think it's fine to use the bank one for those since we want them to be as non-verbose as possible

PR #1374, opened 2025-03-04
Fix auction query conflict

giansalex:
`/stride/auction/auctions` doesn't work because it matches with route `/stride/auction/{name}`.
i updated the route to be the same as [airdrops](https://github.com/Stride-Labs/stride/blob/main/proto/stride/airdrop/query.proto#L18)

image


assafmo (approved):

assafmo:
Thanks! :facepalm:

assafmo:
Included in https://github.com/Stride-Labs/stride/releases/tag/v26.0.1, thanks @giansalex!

PR #1375, opened 2025-03-05
fix CmdCreateAuction() to correctly parse minBidAmount

assafmo:


sampocs (commented):
it wont let me comment over it but can you also change the example command?

sampocs:
same with the example in update auction

sampocs:
and add-token-price

sampocs (approved):

PR #1381, opened 2025-03-11
sdk v0.50.x

assafmo:
## Upgraded Dependencies
* SDK: v0.47.16 → v0.50.13
* IBC: v7.9.2 → v8.7.0
* ICS: v4.0.0 → v6.4.1
* Comet: v0.37.15 → v0.38.17
* Wasmd: v0.46.0 → v0.54.0
* Wasmvm: v1.5.9 → v2.2.1


app/upgrades/v27/upgrades.go
```diff
@@ -0,0 +1,56 @@
...
+//
+// Steps to retrieve the Stride consumer chain ID from Cosmos Hub provider:
+//  1. First, obtain the client ID from Stride using the command:
+//     `strided q ccvconsumer provider-info` which returns "07-tendermint-1154"
+//  2. Then, use the Provider's QueryConsumerIdFromClientId endpoint to get the corresponding consumer ID:
+//     - API endpoint: https://rest.cosmos.directory/cosmoshub/interchain_security/ccv/provider/consumer_id/07-tendermint-1154
+//     - This endpoint implements the query defined in the Interchain Security repository at:
+//     https://github.com/cosmos/interchain-security/blob/307b1446/proto/interchain_security/ccv/provider/v1/query.proto#L132-L138
+func InitializeConsumerId(ctx sdk.Context, consumerKeeper consumerkeeper.Keeper) {
+	params := consumerKeeper.GetConsumerParams(ctx)
```
assafmo, line 53: Note for reviewer:

In [their example](https://github.com/cosmos/interchain-security/blob/v6.4.1/UPGRADING.md#consumer) they used `consumerKeeper.GetParams()`, however [this returns](https://github.com/cosmos/interchain-security/blob/307b1446/x/ccv/consumer/keeper/params.go#L31-L37) an empty `stakingtypes.Params` which `SetParams()` doesn't accept.

x/epochs/types/hooks.go
```diff
@@ -1,14 +1,12 @@
...
-)
+import context "context"
 
 type EpochHooks interface {
 	// the first block whose timestamp is after the duration is counted as the end of the epoch
-	AfterEpochEnd(ctx sdk.Context, epochInfo EpochInfo)
+	AfterEpochEnd(ctx context.Context, epochInfo EpochInfo)
 	// new epoch is next block of epoch end block
-	BeforeEpochStart(ctx sdk.Context, epochInfo EpochInfo)
+	BeforeEpochStart(ctx context.Context, epochInfo EpochInfo)
```
sampocs, line 9: why did this have to change if it was in our own module?

x/auction/types/codec.go
```diff
@@ -35,8 +34,4 @@ func init() {
 	RegisterLegacyAminoCodec(amino)
 	cryptocodec.RegisterCrypto(amino)
 	sdk.RegisterLegacyAminoCodec(amino)
-
-	// Register all Amino interfaces and concrete types on the authz and gov Amino codec so that this can later be
-	// used to properly serialize MsgSubmitProposal instances
-	RegisterLegacyAminoCodec(govcodec.Amino)
```
sampocs, line 41: why remove this? do we not have any gov txs for auction?

x/icaoracle/types/codec.go
```diff
@@ -39,8 +38,4 @@ func init() {
 	RegisterLegacyAminoCodec(amino)
 	cryptocodec.RegisterCrypto(amino)
 	sdk.RegisterLegacyAminoCodec(amino)
-
-	// Register all Amino interfaces and concrete types on the authz  and gov Amino codec so that this can later be
-	// used to properly serialize MsgSubmitProposal instances
-	RegisterLegacyAminoCodec(govcodec.Amino)
```
sampocs, line 45: same here - we definitely have gov txs for icaoracle

x/icqoracle/types/codec.go
```diff
@@ -35,8 +34,4 @@ func init() {
 	RegisterLegacyAminoCodec(amino)
 	cryptocodec.RegisterCrypto(amino)
 	sdk.RegisterLegacyAminoCodec(amino)
-
-	// Register all Amino interfaces and concrete types on the authz and gov Amino codec so that this can later be
-	// used to properly serialize MsgSubmitProposal instances
-	RegisterLegacyAminoCodec(govcodec.Amino)
```
sampocs, line 41: same here

app/app.go
```diff
@@ -389,15 +389,27 @@ func NewStrideApp(
...
 
 	keys := storetypes.NewKVStoreKeys(
-		authtypes.StoreKey, banktypes.StoreKey, stakingtypes.StoreKey,
-		minttypes.StoreKey, distrtypes.StoreKey, slashingtypes.StoreKey,
-		govtypes.StoreKey, paramstypes.StoreKey, ibchost.StoreKey, upgradetypes.StoreKey, feegrant.StoreKey,
-		evidencetypes.StoreKey, ibctransfertypes.StoreKey, capabilitytypes.StoreKey, // monitoringptypes.StoreKey,
+		authtypes.StoreKey,
+		banktypes.StoreKey,
+		stakingtypes.StoreKey,
+		minttypes.StoreKey,
```
sampocs, line 395: ty for finally lining these up lol

app/ante_handler.go
```diff
@@ -39,9 +40,6 @@ func NewAnteHandler(options HandlerOptions) (sdk.AnteHandler, error) {
 	if options.WasmConfig == nil {
 		return nil, errorsmod.Wrap(sdkerrors.ErrLogic, "wasm config is required for ante builder")
 	}
-	if options.TXCounterStoreKey == nil {
-		return nil, errorsmod.Wrap(sdkerrors.ErrLogic, "tx counter key is required for ante builder")
-	}
```
sampocs, line 44: where did this come from?

sampocs: ah is it no longer a pointer?

app/app.go
```diff
@@ -647,10 +647,21 @@ func NewStrideApp(
...
+		"iterator",
+		"staking",
+		"stargate",
+		"cosmwasm_1_1",
+		"cosmwasm_1_2",
+		"cosmwasm_1_3",
+		"cosmwasm_1_4",
+		"cosmwasm_2_0",
+		"cosmwasm_2_1",
+		"cosmwasm_2_2",
```
sampocs, line 660: do you understand what these are for? I've never known lol

app/upgrades/v17/upgrades_test.go
```diff
@@ -121,8 +123,10 @@ func (s *UpgradeTestSuite) checkCommunityPoolICAAccountsRegistered(chainId strin
 	expectedDepositPortId, _ := icatypes.NewControllerPortID(depositOwner)
 	expectedReturnPortId, _ := icatypes.NewControllerPortID(returnOwner)
 
-	depositPortIdRegistered := s.App.ICAControllerKeeper.IsBound(s.Ctx, expectedDepositPortId)
-	returnPortIdRegistered := s.App.ICAControllerKeeper.IsBound(s.Ctx, expectedReturnPortId)
+	scopedICAControllerKeeper := s.App.CapabilityKeeper.ScopeToModule(icacontrollertypes.SubModuleName)
```
sampocs, line 126: can you use `s.App.ScopedICAControllerKeeper`?

x/stakeibc/client/cli/gov.go
```diff
@@ -82,9 +82,6 @@ Where proposal.json contains:
 			if err != nil {
 				return err
 			}
-			if err := msg.ValidateBasic(); err != nil {
-				return err
-			}
```
sampocs, line 87: wdym these are now optional for gov messages?

sampocs (commented):
you're a beast!! Great stuff here

Planning to review a few files in more depth but dropping this here for now!

app/app.go
```diff
@@ -991,12 +987,17 @@ func NewStrideApp(
 		strdburnerModule,
 	)
 
+	app.mm.SetOrderPreBlockers(
+		upgradetypes.ModuleName,
+	)
+
+	app.SetPreBlocker(app.PreBlocker)
```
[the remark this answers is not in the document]
assafmo: yeah pretty sure

app/test_setup.go
```diff
@@ -115,15 +115,15 @@ func GenesisStateWithValSet(app *StrideApp) GenesisState {
 			Jailed:            false,
 			Status:            stakingtypes.Bonded,
 			Tokens:            bondAmt,
-			DelegatorShares:   sdk.OneDec(),
+			DelegatorShares:   sdkmath.LegacyOneDec(),
```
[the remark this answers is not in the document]
assafmo: in sdk 47, [`Dec` was `LegacyDec`](https://github.com/cosmos/cosmos-sdk/blob/0c4872928a/types/math.go#L50). in cosmossdk.io/math, `Dec` is the new Dec and `LegacyDec == sdk 47 LegacyDec == sdk 47 Dec`.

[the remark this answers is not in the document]
assafmo: https://www.perplexity.ai/search/explian-cosmos-sdk-math-legacy-dKtSr13yR1y6pXFNkIFOMQ

app/export.go
```diff
@@ -19,7 +19,7 @@ func (app *StrideApp) ExportAppStateAndValidators(
 	forZeroHeight bool, jailAllowedAddrs []string,
 ) (servertypes.ExportedApp, error) {
 	// as if they could withdraw from the start of the next block
-	ctx := app.NewContext(true, tmproto.Header{Height: app.LastBlockHeight()})
+	ctx := app.NewContextLegacy(true, tmproto.Header{Height: app.LastBlockHeight()})
```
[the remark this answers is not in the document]
assafmo: `NewContext()` doesn't accept `tmproto.Header` anymore which we actually want to use in all of our uses. `NewContext()` just uses an empty header which will change the behavior for us. I couldn't find an alternative way of doing this without `NewContextLegacy()` except for doing
```
ctx := app.NewContext(true)
ctx.WithBlockHeader(tmproto.Header{Height: app.LastBlockHeight()})
```
which seems silly.

x/epochs/types/hooks.go
assafmo: not mandatory, i did it for consistency with the sdk

x/auction/types/codec.go
assafmo: `govcodec.Amino` isn't available anymore, there's no way to do it in sdk 50 afaict. i have a personal todo to check this in integration tests (i think we already have a test for this there).

app/app.go
assafmo: haha i wasn't sure if you'll be happy or mad, because it screws up the diff

app/ante_handler.go
assafmo: yup, no longer nullable

app/app.go
assafmo: those are features that contracts can use. e.g. if a contract would try to iterate over its state without the "iterator" feature being enabled, cw will fail the tx.

x/icqoracle/types/expected_keepers.go
```diff
@@ -15,5 +15,5 @@ type IcqKeeper interface {
 
 // IbcTransferKeeper defines the expected interface needed to convert an ibc token hash to its denom on the source chain.
 type IbcTransferKeeper interface {
-	GetDenomTrace(ctx sdk.Context, denomTraceHash tmbytes.HexBytes) (ibctransfertypes.DenomTrace, bool)
+	GetDenomTrace(ctx sdk.Context, denomTraceHash []byte) (ibctransfertypes.DenomTrace, bool)
```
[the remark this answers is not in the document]
sampocs: ah gotchya

And yeah I was dropping comments as I scrolled commits so they should all say outdated, but I think this code is the same on the latest commit?

app/upgrades/v17/upgrades_test.go
assafmo: nice 2dcced3b9260d2b0197cc41c6626efd264b536f0

x/stakeibc/client/cli/gov.go
assafmo: ValidateBasic is now optional for msgs that doesn't need it, apparently

assafmo: there's a way of checking it if it's available but in this instance we know that message doesn't have ValidateBasic() anymore so there's no sense in checking if ValidateBasic() is available

app/test_setup.go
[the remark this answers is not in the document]
sampocs: i meant more that we probably should eventually migrate off the legacy ones, and that that would require a migration. But I don't think we need to do it in this upgrade

app/export.go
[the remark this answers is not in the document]
sampocs: i feel like that approach isn't that bad! seems slightly better than using the legacy function

x/stakeibc/client/cli/gov.go
assafmo: example: https://github.com/Stride-Labs/osmosis/blob/2ce269b07/osmoutils/osmocli/cli_tester.go#L64-L66

app/app.go
sampocs: always happy

x/stakeibc/client/cli/gov.go
sampocs: I think we do have validate basic for these though ([example](https://github.com/Stride-Labs/stride/blob/2dcced3b9260d2b0197cc41c6626efd264b536f0/x/stakeibc/types/gov.go#L46-L66))

Although, I think validateBasic is actually called in `GenerateOrBroadcastTxCLI` so no need to be redundant here

app/app.go
sampocs: what are the different cosmwasm_X_Y features?

x/icqoracle/types/expected_keepers.go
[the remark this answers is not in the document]
sampocs: oh weird 🤔

x/epochs/types/hooks.go
assafmo: `sdk.Context` now implements `context.Context` so we can pass `sdk.Context` to sdk functions and unwrap does nothing if it's already `sdk.Context`

assafmo: i'll follow your lead here, i don't have a strong preference but feel like it's the same [argument you made](https://github.com/Stride-Labs/stride/pull/1381#discussion_r2004087790) about `NewContextLegacy()` vs `NewContext()`

app/app.go
assafmo: https://github.com/CosmWasm/cosmwasm/blob/release/2.2/packages/std/Cargo.toml#L20-L55

x/stakeibc/client/cli/gov.go
assafmo: x/gov's MsgSubmitProposal doesn't have ValidateBasic() anymore. In this instance the internal msg can have ValidateBasic() but what was removed here was the call to the outer MsgSubmitProposal ValidateBasic(). We can maybe add a call to `proposal.ValidateBasic()`.

assafmo: wdyt about da4b5bdba631a6b588545cd0562262b989d7adce?

app/export.go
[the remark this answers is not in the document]
assafmo: a028b3f31753dab9c81691f3dbf00fc0768b7a39

x/stakeibc/client/cli/gov.go
sampocs: perfect

app/app.go
```diff
@@ -1602,7 +1589,7 @@ func initParamsKeeper(appCodec codec.BinaryCodec, legacyAmino *codec.LegacyAmino
 
 // SimulationManager implements the SimulationApp interface
 func (app *StrideApp) SimulationManager() *module.SimulationManager {
-	return app.sm
+	return nil
```
assafmo, line 1632: why?

app/app.go
```diff
@@ -535,7 +539,14 @@ func NewStrideApp(
 		runtime.NewKVStoreService(keys[feegrant.StoreKey]),
 		app.AccountKeeper,
 	)
-	app.UpgradeKeeper = upgradekeeper.NewKeeper(skipUpgradeHeights,
+
+	skipUpgradeHeights := map[int64]bool{}
+	for _, h := range cast.ToIntSlice(appOpts.Get(server.FlagUnsafeSkipUpgrades)) {
+		skipUpgradeHeights[int64(h)] = true
+	}
+	homePath := cast.ToString(appOpts.Get(flags.FlagHome))
```
assafmo, line 552: i don't really like it here. maybe better to do it outside and pass homePath into here like before?

assafmo: plus since wasmvm v2 that won't work because tests that use the same homedir will deadlock, that's why i added the MkdirTemp in the tests setup

app/app.go
sampocs: we didn't set the sim manager (it was commented out) so i just removed it from the App struct

sampocs (approved):

PR #1363, opened 2025-02-17
standardize amino names

assafmo:
- Standardize amino message names
- Add cosmos.AddressString scalar annotations
- Add cosmos.ValidatorAddressString scalars
- Fix typos in proto option names

Note: this may break amino clients (e.g. ledger, UIs), so we need to make sure to update stridejs after this goes live

PR #1386, opened 2025-03-20
standardize amino names

sampocs:
Replay of #1363 off of main (instead of BBB)


sampocs:
Last open question from the other PR was around whether to shorten the full message name everywhere or just the amino name for MsgUpdateInnerRedemptionRate and MsgUpdateValidatorExchangeRate and I think we landed on just shortening the amino name

assafmo (approved):

PR #1388, opened 2025-03-26
distribution fix

assafmo:
# Distribution Fix for Missing Slashing Event

## Problem
A missing slashing event for validator `stridevaloper1tlz6ksce084ndhwlq2usghamvh0dut9q4z2gxd` has prevented some users from withdrawing rewards. This issue affects delegations created before block height 4300034 (2023-06-28).

## Solution
This PR introduces a fix in the v27 upgrade handler that:

1. Inserts the missing 0.01% slashing event at:
   - Period: 3902
   - Block: 4673775

2. Sets historical rewards data for the slashing event period using data from period 3913 to ensure withdrawal calculations work correctly.

## Technical Details
- The fix uses hardcoded historical rewards data derived from mainnet state analysis
- The affected rewards are extremely minimal (~$0.0000000005910187 INJ and ~$0.0000000000284597 EVMOS)
- This approach simplifies the fix while having virtually no impact on users due to the microscopic values involved

## Testing
- Verified against mainnet state export data
- Tested withdrawal functionality for affected delegations


app/upgrades/v27/upgrades_test.go
```diff
@@ -30,3 +36,89 @@ func (s *UpgradeTestSuite) TestUpgrade() {
...
+	s.App.DistrKeeper.InitGenesis(s.Ctx, distGenesisState)
+
+	// Confirm that withdrawing rewards fails for delegations created before height 4300034
+	for _, delegation := range stakingGenesisState.Delegations {
+		delAddr := types.MustAccAddressFromBech32(delegation.DelegatorAddress)
+
+		period, err := s.App.DistrKeeper.GetDelegatorStartingInfo(s.Ctx, valAddr, delAddr)
+		s.Require().NoError(err)
+		s.Require().Positive(period.PreviousPeriod)
+		s.Require().Positive(period.Height)
```
assafmo, line 92: idk why but this fails sometimes

PR #1417, opened 2025-07-18
[tests] migrate core integration tests from bats to TypeScript framework

theghostmac:
## Context
The dockernet integration tests are getting more brittle by the day and don't work for SDK 50 chains. They're also just a pain to work with.

This starts the migration of the tests into the new typescript/k8s framework.

## Brief Changelog
* Added core integration tests in the new framework 
  * IBC transfer
  * Delegation
  * Unbonding
  * Reinvestment

## Future Work
Migration of remaining tests and fixing of the BBB tests to be done in a separate PR

### DIFF
--- a/.github/mergify.yml
+++ b/.github/mergify.yml
@@ -6,8 +6,10 @@ defaults:
 
 queue_rules:
   - name: default
-    conditions:
-      - "#approved-reviews-by>=1"
+    merge_method: squash
+    commit_message_template: |
+      {{ title }} (#{{ number }})
+      {{ body }}
 
 pull_request_rules:
   - name: automerge to the base branch with label automerge and branch protection passing
@@ -18,7 +20,3 @@ pull_request_rules:
     actions:
       queue:
         name: default
-        method: squash
-        commit_message_template: |
-           {{ title }} (#{{ number }})
-           {{ body }}


integration-tests/client/test/core.test.ts
```diff
@@ -0,0 +1,355 @@
...
+}, 45_000);
+
+describe("Core Tests", () => {
+    test("IBC Transfer", async () => {
+        const stridejs = strideAccounts.user;
+        const gaiajs = gaiaAccounts.user;
+        const transferAmount = 50000000;
+
+        // Get initial balances - only track what we can reliably query
+        const { balance: { amount: strideInitialStrideBalance } = { amount: "0" } } =
```
sampocs, line 156: Nit: i would rename these so it's clear when you mean the denom vs the chain
```suggestion
        const { balance: { amount: strideInitialStrdBalance } = { amount: "0" } } =
```

integration-tests/client/test/core.test.ts
```diff
@@ -0,0 +1,355 @@
...
+                denom: USTRD,
+            });
+
+        const { balance: { amount: strideInitialAtomBalance } = { amount: "0" } } =
+            await stridejs.query.cosmos.bank.v1beta1.balance({
+                address: stridejs.address,
+                denom: ATOM_DENOM_ON_STRIDE,
+            });
+
+        const gaiaInitialAtomBalance = await gaiajs.query.bank.balance(gaiajs.address, UATOM);
```
sampocs, line 168: should you also query for `gaiaInitialStrdBalance`

integration-tests/client/test/core.test.ts
```diff
@@ -0,0 +1,355 @@
...
+            coin: `${transferAmount}${UATOM}`,
+            sender: gaiajs.address,
+            receiver: stridejs.address,
+        });
+
+        // Wait a bit for transfers to complete
+        await sleep(5000);
+
+        // Get final balances
+        const { balance: { amount: strideFinalStrideBalance } = { amount: "0" } } =
```
sampocs, line 200:
```suggestion
        const { balance: { amount: strideFinalStrdBalance } = { amount: "0" } } =
```

integration-tests/client/test/core.test.ts
```diff
@@ -0,0 +1,355 @@
...
+                denom: USTRD,
+            });
+
+        const { balance: { amount: strideFinalAtomBalance } = { amount: "0" } } =
+            await stridejs.query.cosmos.bank.v1beta1.balance({
+                address: stridejs.address,
+                denom: ATOM_DENOM_ON_STRIDE,
+            });
+
+        const gaiaFinalAtomBalance = await gaiajs.query.bank.balance(gaiajs.address, UATOM);
```
sampocs, line 212: same here with gaia final strd balance

sampocs: also can you similarly unpack this so that it's an int instead of  a struct?

integration-tests/client/test/core.test.ts
```diff
@@ -0,0 +1,355 @@
...
+
+        console.log("Final balances:");
+        console.log(`Stride USTRD: ${strideFinalStrideBalance}`);
+        console.log(`Stride ATOM: ${strideFinalAtomBalance}`);
+        console.log(`Gaia ATOM: ${gaiaFinalAtomBalance.amount}`);
+
+        // Calculate and verify balance changes
+        const strideBalanceDiff = BigInt(strideInitialStrideBalance) - BigInt(strideFinalStrideBalance);
+        const strideAtomBalanceDiff = BigInt(strideFinalAtomBalance) - BigInt(strideInitialAtomBalance);
+        const gaiaAtomBalanceDiff = BigInt(gaiaInitialAtomBalance.amount) - BigInt(gaiaFinalAtomBalance.amount);
```
sampocs, line 222: nit: make them all (final - initial) since it's easier to track the sign that way, and make the denom explicit in the variable name
```suggestion
        const strideStrdBalanceDiff = BigInt(strideFinalStrdBalance) - BigInt(strideInitialStrdBalance);
        const strideAtomBalanceDiff = BigInt(strideFinalAtomBalance) - BigInt(strideInitialAtomBalance);
        const gaiaStrdBalanceDiff = BigInt(gaiaFinalStrdBalance) - BigInt(gaiaInitialStrdBalance);
        const gaiaAtomBalanceDiff = BigInt(gaiaFinalAtomBalance) - BigInt(gaiaInitialAtomBalance);
```

integration-tests/client/test/core.test.ts
```diff
@@ -0,0 +1,355 @@
...
+        // USTRD should decrease on Stride by at least the transfer amount (plus gas fees)
+        expect(strideBalanceDiff).toBeGreaterThanOrEqual(BigInt(transferAmount));
+        expect(strideBalanceDiff).toBeLessThan(BigInt(transferAmount + 1000000)); // increased gas fee limit
+
+        // ATOM should increase on Stride by exactly the transfer amount
+        expect(strideAtomBalanceDiff).toBe(BigInt(transferAmount));
+
+        // ATOM should decrease on Gaia by at least the transfer amount (plus gas fees)
+        expect(gaiaAtomBalanceDiff).toBeGreaterThanOrEqual(BigInt(transferAmount));
+        expect(gaiaAtomBalanceDiff).toBeLessThan(BigInt(transferAmount + 1000000)); // increased gas fee limit
```
sampocs, line 239: Ideally we could just do the following:
* Use strict equality
* Have the condition be +/- transfer amount based on the direction

```
expect(strideStrdBalanceDiff).toBe(BigInt(-transferAmount)); // strd sent out → decrease balance
expect(strideAomBalanceDiff).toBe(BigInt(transferAmount)); // atom received → increase balance

expect(gaiaStrdBalanceDiff).toBe(BigInt(transferAmount)); // strd received → increase balance
expect(gaiaAomBalanceDiff).toBe(BigInt(-transferAmount)); // atom sent out → increase decease
```

On the strict equality, I'm not sure the best way to get around fees. Some options:
1. Hard code the fee amount and then you can use that value exactly in the inequality (e.g. transferAmount + FEE)
2. Pay fees in different denom (might require a change at the network level)
3. Transfer a different denom (would require a change at the network level, but it would probably be simpler than (2) since you would just have to add a new denom to genesis)

Happy to talk through these offline!

integration-tests/client/test/core.test.ts
```diff
@@ -0,0 +1,355 @@
...
+        const stridejs = strideAccounts.user;
+        const gaiajs = gaiaAccounts.user;
+        const stakeAmount = 10000000;
+
+        // Get initial balances
+        const { balance: { amount: strideInitialAtomBalance } = { amount: "0" } } =
+            await stridejs.query.cosmos.bank.v1beta1.balance({
+                address: stridejs.address,
+                denom: ATOM_DENOM_ON_STRIDE,
+            });
```
sampocs, line 252: this structure is repeated in a lot of places, might make sense to move it to a utility function! where you could instead do:

```
const strideInitialAtomBalance = await getBalance({client: stridejs, address: stridejs.address, denom:  ATOM_DENOM_ON_STRIDE})
```

integration-tests/client/test/core.test.ts
```diff
@@ -0,0 +1,355 @@
...
+            await stridejs.query.cosmos.bank.v1beta1.balance({
+                address: stridejs.address,
+                denom: "stuatom", // stATOM denom
+            });
+
+        // Get delegation address
+        const { hostZone } = await stridejs.query.stride.stakeibc.hostZone({
+            chainId: GAIA_CHAIN_ID,
+        });
+        const delegationAddress = hostZone?.delegationIcaAddress || "";
```
sampocs, line 264: lets assert the delegationAddress exists here

integration-tests/client/test/core.test.ts
```diff
@@ -0,0 +1,355 @@
...
+        if (attempts >= maxAttempts) {
+            throw new Error("Timed out waiting for stATOM tokens to be minted");
+        }
+
+        // Get final ATOM balance on Stride
+        const { balance: { amount: strideFinalAtomBalance } = { amount: "0" } } =
+            await stridejs.query.cosmos.bank.v1beta1.balance({
+                address: stridejs.address,
+                denom: ATOM_DENOM_ON_STRIDE,
+            });
```
sampocs, line 313: similarly here, let's throw this into a helper so it  doesn't hurt the tests readability. Ideally we can just do something like:

```
const strideFinalStAtomBalance = await utils.waitForBalanceChange(...)
```

sampocs: if you want to re-use it for the while loop below too, maybe you can take it one step further and have like a `waitForStateChange` generic function that takes in a function as an argument that it retries. So in the above, you could pass the the get stride balance function and have it retry it until the response changes. And then below you can repeat the same thing but with a get delegation balance function

Possible this would be too much and muddy up the code, but worth a try

End goal is to keep the code in the actual test functions nice and concise so you can quickly see what's going on!

sampocs (commented):
Fantastic stuff here - excited to see these coming together!

Let's get these comments addressed and merge this guy, and then you can address the other tests in a different PR

integration-tests/client/test/core.test.ts
theghostmac: addressed.

integration-tests/client/test/core.test.ts
theghostmac: addressed

integration-tests/client/test/core.test.ts
theghostmac: addressed

integration-tests/client/test/core.test.ts
theghostmac: done. thanks.

integration-tests/client/test/utils.ts
```diff
@@ -229,3 +229,79 @@ export async function moduleAddress(
...
+                                             address,
+                                             denom,
+                                             initialBalance,
+                                             maxAttempts = 60,
+                                             intervalMs = 500,
+                                           }: {
+  client: StrideClient | CosmosClient;
+  address: string;
+  denom: string;
+  initialBalance: string;
```
sampocs, line 273: nit: you can probably remove initialBalance and just get the initial balance in the function itself

integration-tests/client/test/utils.ts
```diff
@@ -229,3 +229,79 @@ export async function moduleAddress(
     ).account as ModuleAccount
   ).baseAccount?.address!;
 }
+
+/**
+ * Generic function to wait for a state change by retrying a function until condition is met
+ */
+export async function waitForStateChange<T>(
```
sampocs, line 236: beautifully done!!

integration-tests/client/test/utils.ts
theghostmac: done. thank you for spotting that.

sampocs (approved):

PR #1422, opened 2025-07-30
v28 Upgrade Handler + Bounds Change

shellvish:
Adds an upgrade handler for the v28 upgrade.

Also adjusts the outer bounds for all host zones in the v28 upgrade.

### DIFF
--- a/app/upgrades/v28/upgrades.go
+++ b/app/upgrades/v28/upgrades.go
@@ -1,0 +1,70 @@
+package v28
+
+import (
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/cosmos/cosmos-sdk/types/module"
+	upgradetypes "github.com/cosmos/cosmos-sdk/x/upgrade/types"
+
+	stakeibckeeper "github.com/Stride-Labs/stride/v27/x/stakeibc/keeper"
+)
+
+var (
+	UpgradeName = "v28"
+
+	OsmosisChainId = "osmosis-1"
+
+	// Redemption rate bounds updated to give ~3 months of slack on outer bounds
+	RedemptionRateOuterMinAdjustment = sdk.MustNewDecFromStr("0.05")
+	RedemptionRateOuterMaxAdjustment = sdk.MustNewDecFromStr("0.25")
+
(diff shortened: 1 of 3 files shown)


app/upgrades/v28/upgrades.go
```diff
@@ -0,0 +1,70 @@
...
+
+	OsmosisChainId = "osmosis-1"
+
+	// Redemption rate bounds updated to give ~3 months of slack on outer bounds
+	RedemptionRateOuterMinAdjustment = sdk.MustNewDecFromStr("0.05")
+	RedemptionRateOuterMaxAdjustment = sdk.MustNewDecFromStr("0.25")
+
+	// Osmosis will have a slighly larger buffer with the redemption rate
+	// since their yield is less predictable
+	OsmosisRedemptionRateBuffer = sdk.MustNewDecFromStr("0.02")
```
asalzmann, line 22: Do we need this? Outer bounds are already wide

app/upgrades/v28/upgrades.go
```diff
@@ -0,0 +1,70 @@
...
+		// Give osmosis a bit more slack since OSMO stakers collect real yield
+		outerAdjustment := RedemptionRateOuterMaxAdjustment
+		if hostZone.ChainId == OsmosisChainId {
+			outerAdjustment = outerAdjustment.Add(OsmosisRedemptionRateBuffer)
+		}
+
+		outerMinDelta := hostZone.RedemptionRate.Mul(RedemptionRateOuterMinAdjustment)
+		outerMaxDelta := hostZone.RedemptionRate.Mul(outerAdjustment)
+
+		outerMin := hostZone.RedemptionRate.Sub(outerMinDelta)
```
asalzmann, line 62: Do we need to lower the min? We definitely can but I think we mainly care about the max

app/upgrades/v28/upgrades.go
shellvish: I don't think so, we've just separated out Osmosis in the past. I'll remove it from this one for simplicity

asalzmann (approved):

PR #1423, opened 2025-07-31
disable community pool liquid stake feature

sampocs:
## Context
This feature spams the relayer with a bunch a queries to the community pool ICA accounts which can brick the relayer on SDK 50 chains during the non-membership proof.

Since this feature was only used by DYDX and is no longer used, we're just disabling it entirely but can re-enable it easily with an upgrade if needed.


shellvish (approved):
Looks good, thanks for making this change!

PR #1424, opened 2025-07-31
add DelayedVestingAccount code and test

asalzmann:
**Summary**
* Adds code to vest 4 million STRD over 1 year to account 
* Sends tokens to account and converts it from BaseAccount to DelayedVestingAccount

**Test plan**
Adds a test to simulate mainnet
- Fund account
- Send a tx
- Receive tokens and convert to DelayedVestingAccount type
- Verify balances


app/upgrades/v28/upgrades_test.go
```diff
@@ -40,14 +43,46 @@ func TestKeeperTestSuite(t *testing.T) {
...
 	checkRedemptionRates()
+	checkLockedTokens()
+}
+
+func (s *UpgradeTestSuite) SetupTestDeliverLockedTokens() func() {
+
+	// Init BaseAccount (which is the type of the account pre-upgrade)
+	deliveryAccountAddress, err := sdk.AccAddressFromBech32(v28.DeliveryAccount)
+	s.Require().NoError(err)
+	deliveryAccount := s.App.AccountKeeper.GetAccount(s.Ctx, sdk.MustAccAddressFromBech32(v28.DeliveryAccount))
```
shellvish, line 63:
```suggestion
	deliveryAccount := authtypes.NewBaseAccount(deliveryAccountAddress, nil, 476582, 0)
```
I think we need to create this account, as it doesn't exist yet in the chain state

Note: we'll also have to import ```authtypes "github.com/cosmos/cosmos-sdk/x/auth/types"``` at the top of the file

shellvish:
Do we have any `DelayedVestingAccounts` live on mainnet? If not, we might want to use `PeriodicVestingAccounts` instead, just because they're more tested in prod. If we do have `DelayedVestingAccounts` already live, then I think it's fine to leave as-is

app/upgrades/v28/upgrades_test.go
asalzmann: >I think we need to create this account, as it doesn't exist yet in the chain state

Are you sure? The test does run

app/upgrades/v28/upgrades.go
```diff
@@ -56,3 +78,29 @@ func UpdateRedemptionRateBounds(ctx sdk.Context, k stakeibckeeper.Keeper) {
...
+func DeliverLockedTokens(ctx sdk.Context, ak authkeeper.AccountKeeper, bk bankkeeper.Keeper) error {
+	// Get account
+	from := sdk.MustAccAddressFromBech32(FromAccount)
+	to := sdk.MustAccAddressFromBech32(DeliveryAccount)
+	amt := sdk.NewCoins(sdk.NewInt64Coin("ustrd", LockedTokenAmount))
+
+	// Destination must exist and be a plain BaseAccount.
+	base, isBaseAccount := ak.GetAccount(ctx, to).(*authtypes.BaseAccount)
+	if !isBaseAccount {
+		// Maybe return an error here?
```
shellvish, line 91: I think the tradeoff is that in the chance the upgrade fails:
- Return error = chain's halted until validators can apply a patch
- Return nil = no downtime but need to circulate a second upgrade
I think I slightly lean towards returning no error, but could go either way

shellvish (approved):
This looks great to me, and really great testing on `DelayedVestingAccount` on mainnet too

PR #1456, opened 2025-09-17
placeholder

sampocs:


asalzmann (approved):

utils/poa.go
```diff
@@ -0,0 +1,25 @@
+package utils
+
+import sdkmath "cosmossdk.io/math"
+
+// Validators are paid 15% of revenue
+var PoaValPaymentRate = sdkmath.LegacyMustNewDecFromStr("0.15")
```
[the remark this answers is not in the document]
sampocs: added warning comment and CI job

Changing the filename like that is a bit unorthodox imo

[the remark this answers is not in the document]
asalzmann: Thanks for adding

app/upgrades/v29/upgrades.go
```diff
@@ -0,0 +1,45 @@
...
+		ctx.Logger().Info("Updating CCV params...")
+		DisableCcvRewards(ctx, consumerKeeper)
+
+		return versionMap, nil
+	}
+}
+
+func DisableCcvRewards(ctx sdk.Context, ck consumerkeeper.Keeper) {
+	params := ck.GetConsumerParams(ctx)
+	params.ConsumerRedistributionFraction = "1.0"
```
riley-stride, line 43: Why do we need both changes? If we're hardcoding 15%, why also change the ConsumerRedistributionFraction? This creates two sources of truth

Would it be cleaner to just delete this param, or set to an invalid number (if possible e.g. -1)

x/stakeibc/keeper/reward_allocation_test.go
```diff
@@ -88,13 +104,26 @@ func (s *KeeperTestSuite) TestLiquidStakeRewardCollectorBalance_Success() {
...
-	// Distribute rewards using new logic: 85% liquid staked to ConsumerToSendToProvider, 85% to auction
+	// Distribute rewards using new logic: 15% liquid staked to PoA validators, 85% to auction
 	s.App.StakeibcKeeper.AuctionOffRewardCollectorBalance(s.Ctx)
 
-	// Check ConsumerToSendToProvider module balance (should have liquid staked stTokens - 15% of original)
-	providerPortion := sdkmath.NewInt(150) // 15% of 1000
-	s.checkModuleAccountBalance(ccvtypes.ConsumerToSendToProviderName, StAtom, providerPortion)
-	s.checkModuleAccountBalance(ccvtypes.ConsumerToSendToProviderName, StOsmo, providerPortion)
+	// Check PoA validators received stTokens (15% of original amount gets liquid staked)
+	// Since the redemption rate is 1:1, 15% of 1000 = 150 stTokens total
```
riley-stride, line 111: Do we need tests for edge cases: small amounts that don't divide evenly, zero balances, and validator payment failures. What happens with 1 token distributed to 7 validators?

utils/poa.go
```diff
@@ -0,0 +1,27 @@
...
+// - cosmosvalcons1m7fg8k39k2tyym5hgwrpf5wx9hqsr8vywuyrtm
+// - cosmosvalcons1c5e86exd7jsyhcfqdejltdsagjfrvv8xv22368
+// - cosmosvalcons1pdpwglc4fcjdzqvyhvfwxg684trpc6uqck5sxk
+// - cosmosvalcons1px0zkz2cxvc6lh34uhafveea9jnaagckmrlsye
+// - cosmosvalcons1vz42ewp04wwepjed7z4qenj925gpakgvap4q2u
+// - cosmosvalcons1upc05nc9pwhhagnkr3f2dft327qxsxfeyvajsu
+// - cosmosvalcons1f6cjsfn47ujttypx7gtncglsmjvndugc2zelqx
+
+var PoaValidatorSet = []string{
+	"stride1gp957czryfgyvxwn3tfnyy2f0t9g2p4pxxdj7c",
```
riley-stride, line 20: Should we validate these are real addresses in code? I assume you checked them manually on mintscan

riley-stride (commented):
small comments; overall looks solid

app/upgrades/v29/upgrades.go
sampocs: I would think of them as two different params. This one is for CCV specifically (in charge of fees that are not from LS'ing revenue). So we want to first pause these rewards to the hub

And then the other one is specifically for our agreement with the validators

utils/poa.go
sampocs: they're technically validated here: https://github.com/Stride-Labs/stride/pull/1456/files#diff-47c484455a6d234bf341de50b1221206d9e4a1acc603055c0096612bee25e8b0R67

and tested that that doesn't panic with the k8s net

x/stakeibc/keeper/reward_allocation_test.go
sampocs: Good call, I'll add a 0 test case! (will cover zero balance and case where it truncates to 0)

The main test case has truncation already for the not dividing evenly case

It's not really possible to force the payment failure in tests

sampocs: actually the 0 case is already in there: `TestLiquidStakeRewardCollectorBalance_NoRewardsAccrued`

PR #1496, opened 2026-04-30
v33: migrate from ICS consumer to POA

sampocs:
## TODO
Throwaway, but re-do before shipping:
* `poa-migration-rehearsal`: Tests the rehearsal
* `poa-migration-ig-tests`: Tests upgrade + test-core
Merge updated scripts: #1497
---

## Summary

Stride v33 swaps the chain's block-producing validator set from ICS-consumer-driven (validators pushed by the Cosmos Hub via PSS) to **POA-driven** (validator set controlled by an admin multisig via `x/poa`). Govenator staking, delegations, governance, and the liquid-staking product economics are unchanged. After this upgrade, Stride no longer leaks 15% of inflation to the Hub.

## Architecture at a glance

- **Adds** `x/poa` (the SDK enterprise POA module) as the source of validator-set updates. Its admin is an off-chain multisig (`AdminMultisigAddress` in `app/upgrades/v33/constants.go`).
- **Removes from the module manager** `ccvconsumer`, `ccvdistr`, `slashing`, and `evidence`. Their keepers and store keys stay mounted — full deletion is a v34 follow-up.
- **Keeps `ccvstaking`** (the wrapper around `x/staking`) so govenator delegations, voting, and proposals continue to work exactly as before.
- **Reconfigures `x/distribution`** to read fees from the standard `fee_collector` (was `cons_redistribute`).
- **Routes tx fees** to POA's module account in the ante handler via `DeductFeeDecorator.WithFeeRecipientModule(poatypes.ModuleName)`.
- **Leaves the CCV IBC route bound** so late VSC packets can still land safely on the open channel until it times out naturally. The provider channel cleanup is a coordinated v34 step.

## Upgrade handler

`app/upgrades/v33/upgrades.go::CreateUpgradeHandler` orchestrates four steps, all of which are individually tested:

1. `mm.RunMigrations(...)` — silently skips the four removed modules.
2. `SnapshotValidatorsFromICS` — reads the current CCV validator set from `ConsumerKeeper` and converts each entry into a POA `Validator` (pubkey + power + metadata).
3. `InitializePOA` — calls `poaKeeper.InitGenesis(...)` with the snapshotted set and the admin multisig address.
4. `SweepICSModuleAccounts` — moves any residual balance in `cons_redistribute` and `cons_to_send_to_provider` to the community pool (post-v33, nothing else deposits to those accounts).

The intent is **no validator-set discontinuity** at the upgrade block: POA's first `EndBlock` emits a diff that resolves to the same set CometBFT was already signing with.

## Scope

**In v33:**
- Module-manager surgery (POA in, four ICS-era modules out)
- Distribution + ante-handler rewiring
- Upgrade handler that snapshots ICS → POA on the upgrade block
- Removal of the now-dead stakeibc → ConsumerKeeper whitelist plumbing

**Deferred to v34 (intentionally):**
- Deleting `ConsumerKeeper`, `SlashingKeeper`, `EvidenceKeeper`, their store keys, and the IBC route binding
- Dropping `interchain-security` from `go.mod`
- Closing the CCV channel on the provider side

## Notable design choices

- **OperatorAddress sourced from the existing PoA validator set.** POA enforces uniqueness on `OperatorAddress` and validates that it's a real `stride1...` account. We reuse the eight validator-controlled addresses already maintained in `utils.PoaValidatorSet` — the same set the existing reward-allocation pipeline pays out to. The snapshot helper does a two-step join: `hex_cons_addr → moniker` (from `validators.json`) → `moniker → operator` (from `utils.PoaValidatorSet`). If either join misses, the upgrade halts loudly so any drift between the two sources of truth surfaces at upgrade time, not silently. `utils.PoaValidatorSet` was reshaped into `[]PoaValidator{Moniker, Operator, HubAddress}` to carry monikers alongside the addresses; reward allocation just reads `.Operator`.
- **Validator monikers via `//go:embed`.** ICS doesn't carry monikers on the consumer side — they live on the Hub. We pre-bake a hex-cons-addr → moniker map, sourced from `app/upgrades/v33/validators.json`. Regeneration script: `scripts/fetch_validator_monikers.sh` (joins the validator-supplied CSV, Stride RPC, and Hub gRPC).
- **CCV IBC route stays bound through v33.** Removing it while the channel is open would leave incoming VSC packets, acks, and timeouts with no module to dispatch to. Late packets land in `ConsumerKeeper.OnRecvPacket` and queue but never apply (because `ccvconsumer.EndBlock` no longer runs to drain the queue). Cleaner to wait for the channel to time out naturally and clean up everything together in v34.

## Testing

- **Per-helper unit tests** in `app/upgrades/v33/helpers_test.go` — happy path (full hex_cons_addr → moniker → operator join), wrong-count guard, missing moniker (halts), unknown moniker not in PoaValidatorSet (halts), invalid admin bech32, empty-account sweep.
- **Integration test** in `app/upgrades/v33/upgrades_test.go::TestUpgradeTestSuite` — seeds 8 ICS validators, 3 govenators with delegations, both ICS reward accounts, and an active gov proposal; runs the upgrade handler; verifies POA validators match the ICS snapshot element-wise (pubkey + power), admin is set, govenator state is untouched, ICS reward accounts are drained to community pool, the active gov proposal is unaffected, and POA's `EndBlocker` produces the expected validator updates.
- **Mainnet-export integration test** is documented in `app/upgrades/v33/testdata/README.md` but deferred until ops produces a real state export. Synthetic-state coverage is the primary protection in the meantime.

Full `./app/...` test sweep passes (all 30 suites, including the v3–v33 upgrade-test history).

## Pre-release checklist

- [x] Replace `AdminMultisigAddress` in `app/upgrades/v33/constants.go` with the final ops-provided multisig
- [x] Re-run `app/upgrades/v33/scripts/fetch_validator_monikers.sh` close to release in case the validator set changes; commit the regenerated `validators.json`
- [x] Produce a mainnet state export and run the mainnet-export upgrade test (procedure: `app/upgrades/v33/testdata/README.md`)

## Outstanding (non-blocking, can be follow-ups)

- A few CLI integration test suites (`x/{claim,icacallbacks,icaoracle,records}/client/cli`) regress because `cosmos-sdk/testutil/network`'s default app constructor pulls POA in via depinject without fully wiring it. This is test-infrastructure fallout, not production logic — worth a separate fix-up PR.
- `utils.TestCreateModuleAccount` was already failing on `66e491db7` (pre-v33), unrelated.
- Module-path bump (`stride/v32` → `stride/v33`) is intentionally deferred to a separate end-of-cycle PR per the multi-dep-upgrade convention.

## References

- Spec: `docs/superpowers/specs/2026-04-27-ics-to-poa-migration-design.md`
- Plan: `docs/superpowers/plans/2026-04-27-ics-to-poa-migration.md`


app/upgrades/v14/upgrades.go
```diff
@@ -388,19 +388,19 @@ func SetConsumerParams(ctx sdk.Context, ck *ccvconsumerkeeper.Keeper, sibc stake
 	ck.SetParams(ctx, ccvconsumerparams)
 
 	// Then, add the stTokens to the reward list
-	ctx.Logger().Info("Registering stTokens to consumer reward denom whitelist...")
-	hostZones := sibc.GetAllHostZone(ctx)
-	allDenoms := []string{}
+	// ctx.Logger().Info("Registering stTokens to consumer reward denom whitelist...")
+	// hostZones := sibc.GetAllHostZone(ctx)
+	// allDenoms := []string{}
```
aljo242, line 393: should just remove these?

aljo242 (approved):

app/upgrades/v14/upgrades.go
sampocs: they're never run again, so we just leave em commented out as documentation

sampocs: we just had to change it since ccv was removed, otherwise would've just left it alone

CHANGELOG.md
```diff
@@ -42,6 +42,31 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+3. Reconfigure `x/distribution` to use the standard `fee_collector` (was
+   `cons_redistribute`). Wrap the standard `distr.AppModule` with a new in-tree
+   `app/distrwrapper` that overrides `BeginBlock` to: (a) iterate bonded
+   staking validators by stake (no `ValidatorByConsAddr` lookup, which would
+   otherwise halt the chain on POA validators that lack a staking shadow due
+   to ICS key-assignment), and (b) preserve the pre-v33 85/15 split with the
+   15% slice now routed to the POA module account instead of the previous
+   non-functional IBC-to-Hub target.
+4. Route tx fees to the POA module account via the ante handler.
+5. Inflation flow economic profile is unchanged — same 85/15 split, just with
```
vladjdk: I don't think this is true on tx fees - everything gets sent to the POA module by design.

app/upgrades/v33/upgrades.go
```diff
@@ -0,0 +1,68 @@
...
+
+		// 2. Snapshot current ICS validator set into POA-shaped Validators.
+		ctx.Logger().Info("v33: snapshotting ICS validator set...")
+		poaValidators, err := SnapshotValidatorsFromICS(ctx, consumerKeeper)
+		if err != nil {
+			return nil, err
+		}
+
+		// 3. Initialize POA state with that set + admin.
+		ctx.Logger().Info("v33: initializing POA state...")
```
vladjdk, line 54: Can optionally remove state for slashing and evidence stores since we remove them modules.

app/upgrades/v33/helpers.go
```diff
@@ -0,0 +1,156 @@
...
+// joined to a Stride-side operator address via utils.PoaValidatorSet — the
+// same address the existing reward-allocation pipeline pays out to. Both joins
+// must succeed; a missing entry on either side is a configuration drift
+// between the two sources of truth and halts the upgrade.
+func SnapshotValidatorsFromICS(
+	ctx sdk.Context,
+	consumerKeeper ccvconsumerkeeper.Keeper,
+) ([]poatypes.Validator, error) {
+	ccVals := consumerKeeper.GetAllCCValidator(ctx)
+	if len(ccVals) != ExpectedValidatorCount {
```
vladjdk: There's a risk that a validator can get jailed and this check fails mid-upgrade.

vladjdk (approved):
Overall, LGTM.

I would strongly recommend adding a (multinode) systemtest and running this on testnet to ensure proper behaviour.

For the systemtest in particular, you should test for the features you care about:
1. Distribution is properly working post-upgrade for both transactions and inflation.
2. Validator set transitions cleanly at the upgrade height and full nodes catch up.
3. POA admin path actually works end-to-end.
4. Govenator state survives intact (and functions as you intend it to)
5. Liquid staking is unbroken (run your common transactions)
6. Late packets from the Hub don't cause issues

For extra safety, you can run the test on an exported genesis from a real stride mainnet node.

app/upgrades/v33/upgrades.go
sampocs: you mean by removing the store key? And if so, can we defer it to next upgrade? Feels a bit safer/eaiser to rollback if needed

app/upgrades/v33/helpers.go
sampocs: good flag - would you recommend instead we just ignore them and add them to the poa set later?

fwiw, we only have 8 vals, so the risk feels pretty low

vladjdk: That would work - better to have that as a fallback than have the upgrade fail

CHANGELOG.md
sampocs: good call - will update!

app/upgrades/v33/upgrades.go
vladjdk: Yeah, and that sounds fine! As long as you're aware

sampocs:
open

riley-stride (approved):

sampocs:
open

PR #1498, opened 2026-05-03
[DO NOT MERGE] PoA migration rehearsal

sampocs:
## ⚠️ DO NOT MERGE

Throwaway rehearsal branch. Test-only constants are hardcoded into v33 source (`validators.json`, `constants.go`, `utils/poa.go`), each carrying a `// REHEARSAL ONLY — DO NOT MERGE` marker. This branch exists to run the migration on a live network, not to merge.

## Purpose

Run the v32 → v33 (ICS → POA) migration on an **8-validator, Stride-only** network — 5 govenators (with x/staking records) + 3 PSS-only validators (without) — to validate the migration handler under multi-node consensus and exercise POA-specific behavior.

This complements #1505, which runs the same migration on the standard 3-validator network with host zones and then runs `make test-core`. Where #1505 asks *"does the existing product still work after the upgrade,"* this branch asks *"is the upgrade itself correct at mainnet shape."* The 5/3 split is the piece #1505 structurally cannot cover.

## Rehearsal result — all tests pass

Run on the `integration` cluster, 8 Stride validators, upgrade at height 109.

**The upgrade itself:** all 8 validators swapped binaries via cosmovisor with **0 pod restarts**, `applying ABCI validator updates total=8`, `Upgrade v33 complete.` POA came up with 8 validators at total power 800 and the expected admin.

| Task | Test | Result |
|---|---|---|
| 15 | Block production continues; validator hash unchanged | ✅ `DHwama1jFBvzdBVQFdYhkKAvO7N72reAiWW+MCBZx6Q=` identical at heights 58, 109 and 130 |
| 16 | Tx fees route to POA, not `fee_collector` | ✅ POA `390,462,303 → 485,124,831` (+94,662,528); `fee_collector` drained to **0** |
| 17 | Govenator rewards accrue and withdraw | ✅ rewards `1,653,757,035 → 1,689,239,412` (+35,482,377); withdrew **1,724,716,788** net of fee |
| 18 | `MsgWithdrawFees` payout, incl. PSS-only validator | ✅ val1 **+115,360,737**; val6 **+322,424,704** |

### Task 18 is the load-bearing one

The 5/3 split was confirmed directly on-chain:

- `strided q staking validators` returns exactly `val1 val2 val3 val4 val5`
- **val6 has no x/staking record at all** (0 matches)
- POA nonetheless credited val6 `320,950,055ustrd` withdrawable, and `MsgWithdrawFees` paid out `322,424,704`

That confirms POA distributes to validators the staking module doesn't know about — the case the 5/3 split exists to prove.

### One evidence caveat

val1's `MsgWithdrawFees` was confirmed by **balance delta**, not a tx receipt — the `q tx` lookup never resolved (tx-indexing quirk on that node). The delta is unambiguous (+115,360,737 against a 112,407,064 withdrawable, with no other tx in flight, and POA's lazy checkpoint pays what is owed at execution rather than at query time), but it is inference from balances. val6's withdrawal returned `code: 0` cleanly.

## Two fixes were required to run this at all

Both cherry-picked from #1505; neither is specific to this branch's fixtures.

1. **`integration-tests/dockerfiles/Dockerfile.stride-upgrade` — idempotent `start.sh`.** The old one-liner moved `COSMOVISOR_HOME` into the daemon home unconditionally. The state volume is an `emptyDir` that survives container restarts while image layers do not, so the move succeeded only on the first start; afterwards `mv` nested the source inside the destination and the next start died on the collision. Since `DAEMON_RESTART_AFTER_UPGRADE=true` restarts the container the moment cosmovisor swaps binaries, **every upgrade test killed its own validators at the upgrade height**, then masked the original error behind `mv` failures on all subsequent restarts. This fix is generally useful and has already landed on `poa-migration-scripts` as `b4b85308a`.

2. **`app/upgrades/v33/upgrades.go` — skip absent host zones in `UpdateValidatorWeights`.** It hard-failed on the first missing host zone, and its constants reference ten mainnet host zones that no test network has. The upgrade halted at `failed to add validator celestiavaloper1nxdm2... to celestia: Host Zone (celestia) not found`. Now skips with an `Error` log, matching `ReconcileOsmosisDelegations` directly above it, which already documents the same reasoning. Mainnet is unaffected — it has all ten zones, verified separately against a mainnet state export where this step applied 331 weight changes successfully.

## Runbook corrections found while executing

The plan doc in `docs/superpowers/plans/2026-04-30-poa-migration-rehearsal.md` has stale commands:

- `strided q ccvconsumer all-cc-validators` does not exist in this ICS version (only `next-fee-distribution`, `params`, `provider-info`, `throttle-state`). Verify the consensus set via each pod's `strided status .validator_info.address` instead.
- All `poa` subcommands need **`strided2`** — `strided` on `PATH` is the v32 binary and has no `poa` command. Only the node itself is swapped to v33 by cosmovisor.
- Rewards and withdrawable-fees are now coin **strings** (`"1582792281.504000000000000000ustrd"`), not `{denom, amount}` objects, so `jq '.total[] | select(.denom=="ustrd")'` errors. `withdrawable-fees` also nests under `.fees.fees[0]`.
- `staking validators[0]` is not val1; ordering is not guaranteed. Take val1's operator address from its own `distribution rewards` response.

## Merge posture

Nothing here merges. The only reusable fix is already on `poa-migration-scripts`. Pre-merge defense:

```bash
git diff main..poa-migration-rehearsal -- \
    app/upgrades/v33/validators.json \
    app/upgrades/v33/validators.json.REHEARSAL_ONLY \
    app/upgrades/v33/constants.go \
    utils/poa.go
```

Non-empty means the branch still carries throwaway state.

### DIFF
--- a/app/upgrades/v33/constants.go
+++ b/app/upgrades/v33/constants.go
@@ -8,9 +8,11 @@ import (
 // UpgradeName is the SDK upgrade plan name. Match the binary release tag.
 const UpgradeName = "v33"
 
-// AdminMultisigAddress is the bech32 address that POA recognizes as its admin
-// post-upgrade
-const AdminMultisigAddress = "stride1fduug6m38gyuqt3wcgc2kcgr9nnte0n7ssn27e"
+// REHEARSAL ONLY — DO NOT MERGE
+// AdminMultisigAddress points at the integration-tests `admin` account so the
+// rehearsal can drive POA admin behavior with a known signer. The mainnet
+// multisig address lives on the `main` branch.
+const AdminMultisigAddress = "stride1u20df3trc2c2zdhm8qvh2hdjx9ewh00sv6eyy8"
 
 // ExpectedValidatorCount is enforced by the upgrade handler. Panics if
 // consumerKeeper.GetAllCCValidator returns a different count.
--- a/app/upgrades/v33/upgrades.go
+++ b/app/upgrades/v33/upgrades.go
@@ -199,6 +199,14 @@ func ReconcileOsmosisDelegations(ctx sdk.Context, sk stakeibckeeper.Keeper, rk r
 // Phase 2: Set all validator weights to their target values
 func UpdateValidatorWeights(ctx sdk.Context, sk stakeibckeeper.Keeper) error {
 	for _, chainId := range utils.StringMapKeys(NewValidators) {
+		// Skip rather than error, for the same reason ReconcileOsmosisDelegations does: an error
+		// here fails the upgrade and halts the chain, and non-mainnet environments (dockernet,
+		// testnets, integration-tests) carry none of the mainnet host zones these weights target.
+		if _, found := sk.GetHostZone(ctx, chainId); !found {
+			ctx.Logger().Error(fmt.Sprintf("v33: host zone %s not found, skipping new validators", chainId))
+			continue
+		}
+
 		validators := NewValidators[chainId]
 		ctx.Logger().Info(fmt.Sprintf("Adding %d new validators to %s...", len(validators), chainId))
 		for _, val := range validators {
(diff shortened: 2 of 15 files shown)


sampocs:
open

sampocs:
open

sampocs:
Closing as testing's been completed

PR #1508, opened 2026-07-07
fix(stakeibc): v32 unbonding remediation — slashed-validator rounding fix, v33 Osmosis phantom reconciliation, retry observability

moonyandfriends:
## Context

After the v32 validator-weight rebalance, redemptions on **injective-1** and **osmosis-1** became permanently stuck in `UNBONDING_RETRY_QUEUE`: every retry rebuilt the same doomed undelegation, the host chain rejected it, the atomic batch reverted, and no event or error was emitted — users waited 25+ days past their unbonding dates. The two chains failed for different reasons.

Full incident report with on-chain evidence: https://gist.github.com/moonyandfriends/2a324c33542c37e4068f4222245f51df

## Bug 1 — Injective: invalid-shares rejection on slashed-validator full drains

Stride stores each validator's shares-to-tokens rate as an 18-decimal `Dec` rounded slightly above the exact tokens/shares ratio. A full-balance `MsgUndelegate` on a slashed validator (rate < 1) can therefore imply marginally more shares than the ICA holds; the host rejects it with `invalid shares amount` (ABCI code 18), reverting the whole batch. The overshoot is bounded by `shares × 5e-19`, so it only manifests on 18-decimal denoms — INJ but never ATOM.

**Fix** (543b22dd): when fully draining a slashed validator, leave a rounding buffer of `amount / 1e17` (floor 1 base unit) so implied shares can never exceed the ICA's actual shares. The dust stays delegated and reconciles at the next calibration. Deterministic regression test included; 5bf9b6a4 guards a nil rate. The stuck epochs self-heal via the existing per-day-epoch retry, no manual action.

## Bug 2 — Osmosis: phantom delegations from a lost acknowledgement

During the v32 rebalance an undelegation ICA executed on Osmosis but its ack was never delivered (ordered channel closed first), leaving 614,423 OSMO of phantom delegations on three weight-0 validators. Every subsequent batch tried to fully drain delegations that don't exist and reverted. The undelegated tokens are not lost — they sit liquid on the delegation ICA.

**Fix** (bb8faeca): the v33 upgrade handler sets the three validators' records to their true on-chain values, reduces `TotalDelegations` by the phantom, and credits the identical amount back as a `DELEGATION_QUEUE` deposit record. The credit exactly offsets the reduction, so **the redemption rate is unchanged** (a naive decrement would cut it ~13%), and the stranded liquid re-delegates through the normal flow. 46d016c2 makes the handler skip (not halt) if the zone is absent. ⚠️ Constants must be re-verified against live state immediately before the upgrade height.

## Observability

- 2abb8312: emit an event when an undelegation fails into the retry queue (v32 loops silently — this incident went unnoticed for ~3 weeks)
- e7332c8: log-only delegation-records consistency check in BeginBlocker

## Testing

Validated end-to-end on dockernet running the real v32.0.0 binary, upgrading via cosmovisor + live gov proposal:

1. Recreated the Osmosis failure organically (deny-filtered relayer, recv-only manual delivery, surgical single-sequence timeout closing the ordered channel) — phantom + stranded liquid + redemption wedged in `UNBONDING_RETRY_QUEUE`, silent on v32, exactly like mainnet.
2. Upgraded at a gov-scheduled height: handler reconciled the phantom, redemption rate preserved (1.000124 → 1.000144, rewards drift only), stuck epoch retried to `CLAIMABLE` with zero manual record surgery.
3. Recreated the Injective shape (validator slashed 3×5% while holding Stride delegations): weight-0 full drain succeeded first-try with the rounding buffer leaving its designed dust.
4. Test account claimed every redemption, including both previously stuck epochs.

Unit tests: deterministic regression test for the rounding buffer, upgrade-handler tests, consistency-check tests.

## Ride-along: dockernet SDK-53 repairs

Five commits fixing dockernet bitrot from the SDK 53 bump (cosmovisor builder image go 1.23, explicit keyring-backend/chain-id flags, `create-validator` JSON format, modern `tx upgrade software-upgrade` command) — needed to run the upgrade test above.

### DIFF
--- a/app/upgrades.go
+++ b/app/upgrades.go
@@ -42,6 +42,7 @@ import (
 	v30 "github.com/Stride-Labs/stride/v32/app/upgrades/v30"
 	v31 "github.com/Stride-Labs/stride/v32/app/upgrades/v31"
 	v32 "github.com/Stride-Labs/stride/v32/app/upgrades/v32"
+	v33 "github.com/Stride-Labs/stride/v32/app/upgrades/v33"
 	v4 "github.com/Stride-Labs/stride/v32/app/upgrades/v4"
 	v5 "github.com/Stride-Labs/stride/v32/app/upgrades/v5"
 	v6 "github.com/Stride-Labs/stride/v32/app/upgrades/v6"
@@ -417,6 +418,17 @@ func (app *StrideApp) setupUpgradeHandlers(appOpts servertypes.AppOptions) {
 			app.ModuleManager,
 			app.configurator,
 			app.GovKeeper,
+			app.RecordsKeeper,
+			app.StakeibcKeeper,
+		),
+	)
+
+	// v33 upgrade handler
+	app.UpgradeKeeper.SetUpgradeHandler(
+		v33.UpgradeName,
+		v33.CreateUpgradeHandler(
+			app.ModuleManager,
+			app.configurator,
 			app.RecordsKeeper,
 			app.StakeibcKeeper,
 		),
--- a/app/upgrades/v33/upgrades.go
+++ b/app/upgrades/v33/upgrades.go
@@ -1,0 +1,140 @@
+package v33
+
+import (
+	"context"
+	"fmt"
+
+	errorsmod "cosmossdk.io/errors"
+	sdkmath "cosmossdk.io/math"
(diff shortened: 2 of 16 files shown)


sampocs:
closing in favor of #1510

PR #1510, opened 2026-07-17
fix(stakeibc): remediate v32 stuck unbondings

sampocs:
## Context

This supersedes #1508 and rebases its v32 unbonding remediation onto current `main` after the v33 ICS-to-POA migration. It intentionally excludes the dockernet SDK repair commits and the delegation-record consistency invariant; all other source changes are retained. The Osmosis reconciliation is composed into the existing v33 handler immediately after module migrations.

After the v32 validator-weight rebalance, redemptions on `injective-1` and `osmosis-1` became stuck in `UNBONDING_RETRY_QUEUE`. The host chains rejected each retry, the atomic batches reverted, and the failures had no on-chain event.

Original incident report with on-chain evidence: https://gist.github.com/moonyandfriends/2a324c33542c37e4068f4222245f51df

## Injective

A full-balance undelegation from a slashed validator can imply marginally more shares than the ICA holds because the stored 18-decimal shares-to-tokens rate rounds above the exact ratio. The host rejects the message with `invalid shares amount`.

The fix leaves a bounded rounding buffer when fully draining a slashed validator. The remainder cascades through the existing undelegation split, and a nil-rate guard preserves existing validator behavior.

## Osmosis

A lost acknowledgement during the v32 rebalance left phantom tracked delegations on three weight-zero validators even though the undelegation executed on Osmosis. The v33 handler now reconciles those records to their on-chain values, reduces `TotalDelegations` by the phantom amount, and credits the same amount as a `DELEGATION_QUEUE` deposit record so the redemption rate remains unchanged and stranded liquid is redelegated normally.

The handler logs and skips reconciliation if `osmosis-1` is absent. The hard-coded reconciliation values must be re-verified against live state immediately before the upgrade height.

## Observability

Failed undelegations now emit an `undelegation_failed` event with the host zone and affected epoch numbers when records move to the retry queue.

## Verification

- `go test ./app/upgrades/v33 ./x/stakeibc/keeper -count=1`
- `go test ./app ./x/stakeibc/types -count=1`
- `git diff --check origin/main...HEAD`
- Confirmed no diff under `dockernet/` or in the removed invariant files

A repository-wide `go test ./...` reaches the same unrelated `utils/TestCreateModuleAccount` uniqueness panic reproduced before the rebase; this branch has no `utils` changes.

Carries forward the work and commit authorship from #1508 by @moonyandfriends.

### DIFF
--- a/app/upgrades.go
+++ b/app/upgrades.go
@@ -439,6 +439,8 @@ func (app *StrideApp) setupUpgradeHandlers(appOpts servertypes.AppOptions) {
 			app.BankKeeper,
 			app.AccountKeeper,
 			app.DistrKeeper,
+			app.RecordsKeeper,
+			app.StakeibcKeeper,
 		),
 	)
 
--- a/app/upgrades/v33/upgrades.go
+++ b/app/upgrades/v33/upgrades.go
@@ -6,14 +6,24 @@ import (
 
 	ccvconsumerkeeper "github.com/cosmos/interchain-security/v7/x/ccv/consumer/keeper"
 
+	errorsmod "cosmossdk.io/errors"
+	sdkmath "cosmossdk.io/math"
+
 	"github.com/cosmos/cosmos-sdk/codec"
 	poakeeper "github.com/cosmos/cosmos-sdk/enterprise/poa/x/poa/keeper"
 	sdk "github.com/cosmos/cosmos-sdk/types"
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
 	"github.com/cosmos/cosmos-sdk/types/module"
 	authkeeper "github.com/cosmos/cosmos-sdk/x/auth/keeper"
 	bankkeeper "github.com/cosmos/cosmos-sdk/x/bank/keeper"
 	distrkeeper "github.com/cosmos/cosmos-sdk/x/distribution/keeper"
 	upgradetypes "github.com/cosmos/cosmos-sdk/x/upgrade/types"
+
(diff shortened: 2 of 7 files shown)


sampocs:
Re-verified the Osmosis reconciliation constants against live chain state.

Evidence:
- Chain: `osmosis-1`
- Height: `66,558,211`
- Block time: `2026-07-17T23:58:55.243444971Z`
- Delegation ICA: `osmo1npfl4vmmmf4yqhcemz95mvqujgdnlhrlxfzhlhz2gru8g2t749xqr9zm5e`
- Queried independently through `rpc.osmosis.zone` and `osmosis-rpc.polkachu.com`

Delegations:
- `osmovaloper1clpqr4nrk4khgkxj78fcwwh6dl3uw4ep88n0y4`: no delegation record (`0 uosmo`)
- `osmovaloper15urq2dtp9qce4fyc85m6upwm9xul3049wh9czc`: no delegation record (`0 uosmo`)
- `osmovaloper14n8pf9uxhuyxqnqryvjdr8g68na98wn5amq3e5`: `60,662,797,124 uosmo`

These exactly match `OsmosisPhantomDelegations`; no code update is required.

The delegation ICA liquid balance at the same height was `624,326,793,891 uosmo`. The reconciliation credit is `614,423,102,876 uosmo`, leaving `9,903,691,015 uosmo` of headroom. Re-verify again immediately before the actual upgrade height if it is scheduled later.

asalzmann (approved):

PR #1524, opened 2026-09-14
feat(upgrade): v34 - reconcile Injective delegations and unblock stuck redemption sweep

moonyandfriends:
## Summary

Reconciles the `injective-1` host zone and unblocks its stuck redemption sweep, as part of the **v34** upgrade handler (moved from a standalone v35 handler per review — v34 is still unreleased).

Between mid-July and early September 2026 the `injective-1` DELEGATION ICA channel closed seven times — no relayer covered the `icacontroller` path, so each epoch's ICA timeouts closed the ordered channel. Every close stranded the acks of `MsgDelegate` packets that had **already executed on Injective**, so their callbacks never ran and the host zone never booked those delegations. Actual delegations on-chain therefore exceed tracked delegations by ~200.48 INJ.

That INJ is liquid balance the redemption sweep expects to find in the delegation ICA. Because the sweep is all-or-nothing, it has failed with insufficient funds every epoch, blocking 12 user redemptions dating back to epoch 1369.

## What the v34 handler does (after the POA validator swap)

- **Reconciles delegations**: applies per-validator deltas (measured on-chain — see `app/upgrades/v34/injective.go`) to each validator's tracked delegation and to `TotalDelegations`, so the books match Injective.
- **Re-queues the over-swept unbondings**: moves the already-unbonded records that can't fit the liquid balance to `UNBONDING_RETRY_QUEUE`, so the normal unbonding flow undelegates them a second time and unwinds the accidentally-staked redemption funds. The remaining records then fit inside the liquid balance and sweep at the next epoch; the re-queued records sweep after they unbond.
- **Hardens the redemption callback** (`x/stakeibc/keeper/icacallbacks_redemption.go`): on a failure ack it now only resets records still in `EXIT_TRANSFER_IN_PROGRESS`, so a record re-queued while a sweep is in flight is not pulled back into the sweep. This removes any constraint on the upgrade height relative to epoch boundaries.

## Testing

- `app/upgrades/v34/injective_test.go` — reconciliation and re-queue unit tests (deltas applied, `TotalDelegations` kept consistent, missing host zone / missing validator / stale-negative-delta and wrong-status records skipped rather than halting).
- `x/stakeibc/keeper/icacallbacks_redemption_test.go` — `TestRedemptionCallback_FailureOnlyResetsInProgressRecords` covers the callback hardening.
- Exercised end-to-end on a local dockernet Stride+Gaia network: reproduced the lost-ack shortfall, ran the upgrade through governance/cosmovisor, and confirmed the handler reconciles the books to match the host chain and re-queues the over-swept record.

## Note for reviewers

The delta table in `injective.go` is a point-in-time on-chain measurement. **Re-measure it against current `injective-1` state immediately before submitting the on-chain upgrade proposal**, in case delegations move in the interim.

(The PR branch is still named `v35-injective-reconciliation` for continuity of this PR; the change itself now lives entirely in `app/upgrades/v34/`.)


sampocs:
closing in favor of 1526

PR #1525, opened 2026-09-14
v34: reset stuck slash queries, delete timed-out ICQs, update gov params

sampocs:
## Summary

Adds stuck-ICQ cleanup and a gov params change to the (unreleased) v34 upgrade handler:

1. **Resets `SlashQueryInProgress` to false** on 12 cosmoshub-4 validators that have the flag set but no delegation ICQ in the store.
2. **Deletes 30 stuck ICQs** that are past their timeout and never received a response.
3. **Updates gov params:** quorum goes from 33.4% to **25%**, and the voting period from 3 days to **5 days** (259200s → 432000s). Every other gov param is unchanged. The new params are validated with `ValidateBasic` before they're written. The expedited voting period (1 day) is still shorter than the new voting period.

The validator and query lists are hardcoded constants in `app/upgrades/v34/constants.go`, measured against mainnet at height **40276875** (2026-09-14). Anything that gets stuck after that measurement is intentionally left alone.

## How the lists were built

- **Validators:** `strided q stakeibc list-host-zone` → every validator with `slash_query_in_progress: true`. All 12 are on cosmoshub-4: whispernode, kira, onblocnode, pupmos, kraken, gatadao, cosvalidator, klubstaking, huobi, validatornetwork, zoomerlabs, ethicalnode.
- **Cross-reference:** `strided q interchainquery list-pending-queries` returned 33 queries and **none** had the `delegation` (slash query) callback, so no flagged validator had a real query in flight. The ICQ EndBlocker marks every new query `RequestSent=true` within its block, so the pending list covers the whole query store.
- **Stuck queries:** 30 of the 33 are past their timeout: 18 evmos validator queries (Jan–Feb), 6 comdex calibrate queries (Jun), 1 celestia validator query (Jan), umee calibrate/validator/withdrawalbalance, and stargaze/evmos withdrawalbalance (Jul). evmos, stargaze and umee are halted. The 3 excluded queries (comdex withdrawalbalance, injective feebalance, laozi withdrawalbalance) were still within their timeout and are live.

## Likely root cause of the stuck flags

All 12 validators have zero delegation and zero weight. When a sharesToTokens-rate drop triggers `SubmitDelegationICQ`, the host has no delegation record, so the ICQ response is contentless. `SubmitQueryResponse` deletes contentless responses and returns *before* invoking the callback, and the callback is the only place the flag gets cleared. Because `SubmitDelegationICQ` skips validators with the flag set, those validators never get another slash check. This PR only clears the current state; it doesn't change the contentless-response path.

## Safety

- No code path depends on a query's existence other than the ICQ module itself. Deleting timed-out queries is equivalent to the eventual reject/retry, minus the response. The calibrate queries can be re-triggered by the admin if needed.
- Missing state is logged and skipped rather than returned as an error, so the cleanup can never halt the upgrade. For example, a query answered before the upgrade is a no-op to delete.

## Notes for reviewers

- **LSM liquid stakes from these 12 validators are allowed again after the upgrade.**
  - Today the stuck flag is the only thing blocking them. `GetValidatorFromLSMTokenDenom` doesn't check weight.
  - The reset puts them back to the same behavior as every other zero-weight validator without the flag. Their stored sharesToTokens rates were refreshed when the slash was detected.
- **The flag can get stuck again**, since the contentless-response path isn't changed here. A manual `UpdateValidatorSharesExchRate`, or an LSM stake that crosses the checkpoint on a zero-delegation validator, would reproduce it. Candidate follow-ups: skip `SubmitDelegationICQ` when `Delegation` is zero, or treat an empty delegation response as zero shares.
- **Re-verify the constants just before tagging.** The tests build state from the constants, so they can't catch a typo. As of height ~40277082 they still match mainnet exactly.
- **Resubmitted query IDs are kept.**
  - Calibrate and withdrawal-balance queries use `forceUnique=false`, so resubmitting reuses the same ID with a fresh timeout. That happens if a halted zone is un-halted or calibrate is re-run before the upgrade.
  - `DeleteStuckQueries` re-checks `HasTimedOut` at upgrade time, so it skips any listed query that is live again.
- **Gov params and proposals already in flight:**
  - Each proposal stores its voting end time when voting starts, so a proposal already in voting keeps its 3-day window.
  - Quorum is read at tally time, so any proposal that ends after the upgrade is tallied against 25%.
  - A failed expedited proposal that converts to a regular one gets the 5-day period, and so do proposals still in deposit.
  - Mainnet currently has no proposals in deposit or voting.
- **Test networks:** after upgrading to v34, the integration and in-place test networks inherit the 5-day voting period, overriding their 30s config. Any gov-driven step after the upgrade (e.g. `client/test/auction.test.ts`, or a second upgrade rehearsal) needs an expedited proposal or a period reset.
- **Pre-existing, not changed here:** mainnet's `expedited_min_deposit` is `50000000stake`, a denom that doesn't exist on Stride. Validation passes only because the coin sets share no denom. This upgrade could fix it if we want.
- #1524 also changes the `v34.CreateUpgradeHandler` signature and `app/upgrades.go`. Whichever PR merges second will get a small, mechanical conflict when adding its keeper params.

## Testing

- `TestUpgradeClearsStuckSlashQueries` runs the full v34 handler and checks that:
  - listed validators are reset
  - an unlisted flagged validator and a same-named validator on another host zone stay flagged
  - listed timed-out queries are deleted, while both an unlisted query and a listed query with a fresh timeout (a resubmitted deterministic ID) survive; the latter case fails if the timeout guard is removed
  - a listed validator or query missing from state is skipped
- It fails when the two new handler calls are removed (`validator whispernode slash query in progress`).
- `TestResetStuckSlashQueriesMissingHostZone` checks that a missing host zone doesn't panic.
- `TestUpgradeUpdatesGovParams` runs the full handler from mainnet's pre-upgrade values (0.334, 3 days).
  - It checks the literal results (`0.250000000000000000`, 5 days) rather than the constants, and that every other gov param is unchanged.
  - It fails when the `UpdateGovParams` call is removed (quorum stays at `0.334`).
- `go test ./app/upgrades/v34/...` passes. `golangci-lint run ./app/...` reports 0 issues.
- Full `go test ./...` passes in 70 packages. The only failure is `utils/TestCreateModuleAccount`, which already fails on `main`.

https://claude.ai/code/session_01XQjLn5EodPCcSpwxa1fquE


riley-stride (approved):

sampocs:
## Summary

This PR extends the v34 upgrade handler to clear 12 stuck Cosmos Hub slash-query flags, remove 30 snapshotted timed-out ICQs, lower governance quorum from 33.4% to 25%, and extend the regular voting period from three to five days while preserving and validating the remaining governance parameters.

## Findings

### Blocking

None.

### Should fix

1. **Re-check each query's timeout at upgrade execution before deleting it.** [`app/upgrades/v34/slash_queries.go:45-50`](https://github.com/Stride-Labs/stride/blob/1ef82975ec79d115ece2d59f67c36ca242a4aedc/app/upgrades/v34/slash_queries.go#L45-L50) deletes every listed ID solely because it was timed out at snapshot height 40276875. That ID is not always a historical identity: `SubmitICQRequest` assigns a new timeout and then overwrites an existing query with the same ID ([`x/interchainquery/keeper/keeper.go:62-85`](https://github.com/Stride-Labs/stride/blob/1ef82975ec79d115ece2d59f67c36ca242a4aedc/x/interchainquery/keeper/keeper.go#L62-L85)), while both calibration and withdrawal-balance submissions use `forceUnique=false` ([`x/stakeibc/keeper/interchainqueries.go:171-183`](https://github.com/Stride-Labs/stride/blob/1ef82975ec79d115ece2d59f67c36ca242a4aedc/x/stakeibc/keeper/interchainqueries.go#L171-L183), [`x/stakeibc/keeper/interchainqueries.go:215-226`](https://github.com/Stride-Labs/stride/blob/1ef82975ec79d115ece2d59f67c36ca242a4aedc/x/stakeibc/keeper/interchainqueries.go#L215-L226)). Therefore, if (for example) one of the Comdex calibrations is re-submitted after the snapshot but before v34 executes, the fresh in-flight request replaces the stale record under the same listed ID and this loop deletes it. Fetch each existing query and call `query.HasTimedOut(ctx.BlockTime())` before deletion; log and preserve fresh queries. Add a regression case with a listed ID and a future `TimeoutTimestamp`.

### Nit

None.

### Question

None.

## Test coverage notes

- `go test ./app/upgrades/v34/...` passes at PR head `1ef82975ec79d115ece2d59f67c36ca242a4aedc`.
- The full-handler cleanup test covers listed versus unlisted validators/queries, missing entries, and same-named validators on another host zone, but every listed query fixture contains only an ID ([`app/upgrades/v34/slash_queries_test.go:42-66`](https://github.com/Stride-Labs/stride/blob/1ef82975ec79d115ece2d59f67c36ca242a4aedc/app/upgrades/v34/slash_queries_test.go#L42-L66)); it does not cover a listed deterministic ID that has been refreshed and is no longer timed out.
- The governance path reads the stored params, changes only quorum and voting period, runs `ValidateBasic`, and then writes them ([`app/upgrades/v34/gov_params.go:13-30`](https://github.com/Stride-Labs/stride/blob/1ef82975ec79d115ece2d59f67c36ca242a4aedc/app/upgrades/v34/gov_params.go#L13-L30)). The handler test verifies the literal 25%/five-day results and equality of every other field ([`app/upgrades/v34/gov_params_test.go:14-35`](https://github.com/Stride-Labs/stride/blob/1ef82975ec79d115ece2d59f67c36ca242a4aedc/app/upgrades/v34/gov_params_test.go#L14-L35)); I found no governance-parameter validity issue.

**Overall verdict:** The upgrade and governance wiring look sound, but the deletion loop should guard on the query's timeout at execution time before this is merged.

sampocs:
Addressed the should-fix in e1074281f: `DeleteStuckQueries` now fetches each listed ICQ and only deletes it if `HasTimedOut(ctx.BlockTime())` at upgrade time, so a deterministic ID resubmitted with a fresh timeout is kept. `TestUpgradeClearsStuckSlashQueries` adds a listed query with a future timeout that must survive; it fails if the guard is removed.

PR #1526, opened 2026-09-14
feat(upgrade): v34 - reconcile Injective delegations and unblock the stuck redemption sweep

sampocs:
## Summary

Supersedes #1524 (thanks @moonyandfriends — the reconciliation, delta table, and redemption-callback hardening are cherry-picked verbatim from that branch, commit `ef22bcd7e`; the record re-queue is replaced by the mechanism below).

Between mid-July and early September 2026 the `injective-1` DELEGATION ICA channel closed seven times, stranding the acks of `MsgDelegate` packets that had already executed on Injective. Two consequences:

1. **Ledger drift** — the delegate callbacks never ran, so actual on-chain delegation exceeds tracked delegation by **200.476671 INJ** (net).
2. **Redemption funds staked** — `RestoreInterchainAccount` reset the in-flight deposit records and they were re-delegated, drawing on the liquid balance where completed unbondings wait for the sweep. ~200 INJ of redeemers' money is staked instead of liquid.

The all-or-nothing redemption sweep (one `MsgSend` for the sum of every `EXIT_TRANSFER_QUEUE` record) has failed every epoch since: records total 1,298.91 INJ vs 1,103.20 INJ liquid. 11 host-zone unbonding records are stuck back to epoch 1369.

## What v34 does

| Mechanism | Fixes | Where |
|---|---|---|
| Reconcile per-validator deltas (cherry-picked) | ledger drift | `app/upgrades/v34/injective.go` |
| One-shot undelegation of exactly the applied excess, via the normal undelegate pipeline | funds staked | handler stores a pending amount; day-epoch hook submits it |
| Per-record redemption sweep, scoped to `injective-1` | all-or-nothing sweep | `x/stakeibc/keeper/redemption_sweep.go` |

- **Reconciliation** applies the delta table to each validator and `TotalDelegations` **all-or-nothing, never as an upgrade error**: a missing host zone (non-mainnet) is a no-op; if the host zone exists but any entry's validator is missing or the result would go negative, nothing is written, zero is returned with an Error log, and the upgrade proceeds with the reconciliation deferred to a later upgrade. A partial application would return a partial sum (dropping the one large negative entry alone turns the ~200 INJ excess into ~867 INJ) and that sum is what gets undelegated; erroring instead would halt the chain for an accounting fix. With the validator list and weights unchanged, the check can only trip if acks are lost again. `ReconcileInjectiveDelegations` returns the applied delta, and the handler writes it to a new store key `PendingUndelegation-value-injective-1`.
- **Pending undelegation** (`x/stakeibc/keeper/pending_undelegation.go`): `SubmitPendingUndelegations` runs in `BeforeEpochStart` on the day epoch. It builds messages with the extracted `GetUndelegateMessagesForAmount` (the normal capacity/priority logic, refactored out of `UnbondFromHostZone` with no behavior change) and submits through the unchanged `BatchSubmitUndelegateICAMessages` with no record ids, so the existing `UndelegateCallback` books the ack (validator + `TotalDelegations` decremented, zero burn, in-progress counters cleared). The amount **stays stored until each batch acks successfully** (that batch's split total is then subtracted; key removed at zero); a per-chain in-flight batch counter blocks a duplicate submission; FAILURE/TIMEOUT only release the batch so the next eligible day epoch resubmits; `RestoreInterchainAccount` clears the counter so a batch stranded on a dead channel is resubmitted on the new one (one store delete — no scan of the 21k-entry icacallbacks store, restore gas unchanged); a record-less ack arriving on a stale (restored) channel is ignored. The host zone's own unbonding epoch is skipped to avoid a same-snapshot capacity collision. Store-driven and a no-op when nothing is pending, so nothing to remove later.
- **Per-record sweep**: for `types.PerRecordSweepChainId` (`"injective-1"`, `TODO [cleanup]` for v35) records are submitted oldest-first as one ICA each, so the records that fit the liquid balance are paid at the first epoch after the upgrade; the rest fail-and-retry individually. Bundled path is byte-for-byte unchanged for every other host zone.
- **Redemption callback hardening** (from #1524): a failure ack only resets records still `EXIT_TRANSFER_IN_PROGRESS`.

Expected timeline on mainnet: first stride epoch after the upgrade, 9 of 11 records (1,044.65 INJ) sweep; first non-unbonding day epoch, 200.476671 INJ is undelegated; ~21 days later records 1426 and 1429 sweep. Ledger ends at exactly the pre-incident `TotalDelegations` (21,514.925033 INJ), RR net change zero.

Design doc: `docs/superpowers/specs/2026-09-14-v34-injective-reconciliation-design.md` (includes why the record re-queue and an upgrade-handler ICA were rejected).

## k8s rehearsal (v33.0.0 → v34, 3 Stride + 3 Gaia validators, real relayers)

Run on the integration cluster from a throwaway branch (`v34-injective-rehearsal`; results in `integration-tests/rehearsal/injective-results.md` there). The incident was **reproduced**, not seeded: a delegate ICA executed on Gaia with its ack stranded, the channel closed, the account was restored, and the retrying re-delegate consumed 300 of the ~910 ATOM that two redemptions had just unbonded, leaving the bundled sweep failing every epoch with on-chain − tracked = 309.64 ATOM. Then, through the real upgrade: reconciliation booked the measured deltas, the per-record sweep paid the record that fit and kept retrying the one that didn't, the pending undelegation executed and brought `TotalDelegations` back to exactly the pre-incident value with the RR back to its exact pre-upgrade value, the last record swept after the host unbonding, and both claims paid the user to the uatom.

The first run also found the gap fixed in `9708df412`/`543ef9955`: the delegation channel died with the undelegate ICA in flight behind a packet whose timeout is never processable after restore, so the delete-on-submit + re-queue-on-timeout design lost the amount.

**Rerun on the final code (rehearsal branch rebased onto `543ef9955`), same choreography, all steps green:** drift 300,760,825 uatom reproduced and stable; upgrade at day epoch 18 booked the three deltas (`TotalDelegations adjusted by 300760825 to 2711300823`, RR 1.004412 → 1.129824); first stride epoch the per-record sweep paid record 10 (501,010,934) and error-acked record 11 alone; day epoch 19 (first non-unbonding epoch) `Submitting pending undelegation of 300760825uatom` once, record-less `UNDELEGATE ICACALLBACK` success with zero burn, `TotalDelegations` back to 2,413,112,179 and the RR at 1.004793 = the pre-upgrade accrual line; 240s later record 11 swept, redemption ICA = 901,819,681 = R1+R2; both claims paid the user exactly that; every validator's tracked delegation == Gaia to the uatom and no resubmission through four more day epochs (two of them unbonding epochs). The in-flight release / restore / stale-channel paths were not hit in this run (the channel stayed healthy) and rely on the unit tests.

## Verification of the numbers

The delta table was taken by `git cherry-pick` from `moonyandfriends/stride@bcfa27712` and never retyped: `git diff bcfa27712 -- app/upgrades/v34/injective.go` touches only the removed re-queue code and the return signature, and the sha256 over the 32 `mustInt` constants is identical on both branches (`5bb4b5eb…c43a2e`).

Verified live on 2026-09-14 (Stride 40280619 / Injective 183003315): all 32 deltas match on-chain delegation minus tracked delegation to 18 decimals; Σ = 200.476671 INJ = on-chain total − `TotalDelegations`; no drift since 2026-09-08. Liquid balance 1,104.02 INJ minus 0.82 INJ pending deposit = 1,103.20 available; the 9 records that fit total 1,044.65. RR after reconciliation ≈ 1.5557 vs `max_inner_redemption_rate` 1.5687.

## Release checklist

- **Re-measure the table against `injective-1` immediately before submitting the upgrade proposal.** With the all-or-nothing check, a stale table is skipped (logged, nothing written, nothing queued) rather than moving money on wrong numbers — but then the reconciliation has to be redone in v35.
- **Upgrade timing.** Between the upgrade block and the undelegate ack the RR reads ~0.93% high (+0.0144). The RR updates every stride epoch (01/07/13/19 UTC); the undelegation goes out at the first non-unbonding day boundary (19:00 UTC, day epochs with `epoch % 4 != 0`; Sep 18 / 22 / 26 / 30 are unbonding days). Scheduling 1–3h before such a boundary limits the inflated RR to one epoch. **Target: 2026-09-23 15:00–16:00 UTC** (day 1481 boundary at 19:00). The RR is deliberately not adjusted for the pending amount.
- **Inner bounds.** A bounds breach halts the host zone, which would also stop the sweep and unbondings. Projected RR on Sep 23 ≈ 1.5432 → ≈ 1.5576 after reconciliation vs `max_inner_redemption_rate` 1.56871 today. The inner bounds are set manually — confirm `max_inner_redemption_rate ≥ 1.575` on upgrade day (or raise to 1.60 the day before and restore after the ack).
- Confirm the `icacontroller` relayer path is healthy — a lost ack on the one-shot undelegation would re-create the drift it fixes.

## Testing

Unit tests only (no dockernet rehearsal, by decision).

- `go test ./x/... ./app/...` — 70 packages ok.
- `app/upgrades/v34`: reconciliation (deltas applied, `TotalDelegations` consistent, missing host zone / validator / negative-result skips everything with nothing written, applied delta returned), upgrade handler queues the pending amount / queues nothing without the host zone or with an inconsistent table (upgrade still succeeds).
- **Mainnet-export release gate** (`mainnet_export_test.go`): the fixture now includes the real `injective-1` host zone (height 40302363) alongside the POA set, and the suite asserts the delta table applies in full against it — every address is a host zone validator, no delegation goes negative, and the handler queues exactly the whole-table sum. Since the handler skips rather than errors on an inconsistent table, this is what turns a stale table into a red build. Regenerate per `testdata/README.md` if the Injective validator set changes before release.
- `x/stakeibc/keeper`: pending key round-trip; hook submits an ICA and removes the key, increments `DelegationChangesInProgress`; closed channel keeps the key; unbonding epoch skipped; zero amount / zero weights / genuine insufficient capacity keep the key; `UndelegateCallback` with no record ids — success books delegations with zero burn, failure/timeout re-queue additively; `UnbondFromHostZone` tests unchanged after the refactor; per-record sweep submits N ICAs ascending by epoch with one id and the matching amount each; bundled path unchanged; redemption-callback failure only resets in-progress records.
- Whole-branch review (two Important findings — failure/timeout re-queue and unbonding-epoch skip — fixed) and scoped re-review: ready to merge.
- Codex review (comment below): RR lifecycle traced (stage 1 overstatement bounded by scheduling, stages 2–3 correct); its blocking finding (partial-table reconciliation) is fixed in `4a41aeef8` with two abort tests that fail against the old function.

https://claude.ai/code/session_01SF2rXDgGxUwdmq53pHYR3h


sampocs:
_Review by Codex (OpenAI), requested via Claude Code — posted on its behalf because the Codex sandbox has no GitHub access. Focus: redemption-rate behavior across the reconciliation → undelegation → sweep lifecycle._

---

## Summary

Request changes. The pending-undelegation, callback retry, and per-record sweep mechanics are generally sound, but the current implementation temporarily credits ~200.48 INJ of redemption liabilities to the outstanding stINJ supply, producing an economically incorrect redemption-rate increase. Separately, the skip-and-continue reconciliation can turn stale validator state into a much larger undelegation than intended.

## Redemption rate analysis

The formula is explicit: `nativeDelegation := ...hostZone.TotalDelegations`, then `nativeTokensLocked := depositAccountBalance.Add(undelegatedBalance).Add(tokenizedDelegation).Add(nativeDelegation)`, divided by stToken supply (`x/stakeibc/keeper/redemption_rate.go:42-62`). Host-zone unbonding records and the delegation/redemption ICA bank balances are not inputs; only qualifying deposit/LSM records are (`redemption_rate.go:49-52,87-143`). It is recomputed before the sweep on each eligible stride epoch (`x/stakeibc/keeper/hooks.go:43-68,90-92`; the default RR interval is 1 at `x/stakeibc/types/params.go:13-18`).

Assumption/magnitude: using the design snapshot—not a fresh chain query, because the public endpoint was unavailable—pre-incident `TotalDelegations` is 21,514.925033 INJ (`docs/superpowers/specs/2026-09-14-v34-injective-reconciliation-design.md:44-49`), and the document gives RR 1.541302 -> ~1.555664 with max inner 1.568710 (`:29-32`). Holding unrelated deposits, rewards, and supply changes constant, the rounded rates imply ~13,958.83 stINJ supply: `200.476671 / (1.555664 - 1.541302)`. Thus the correction makes `TotalDelegations` 21,715.401704 INJ (+0.9318%) and RR +0.014362 (+0.9318%).

| Stage | `TotalDelegations` / other RR inputs | Resulting RR and bounds risk |
|---|---|---|
| 1. Upgrade applied; pending or ICA in flight, before success ack | `ReconcileInjectiveDelegations` adds the full applied delta to validator delegations and `TotalDelegations` (`app/upgrades/v34/injective.go:103-130`), and the handler queues the same amount (`app/upgrades/v34/upgrades.go:59-69`). Supply and every other RR input are unchanged. The stored RR does not change in the upgrade handler itself, but the next RR tick before the ack sees the +200.476671 numerator. Submission also removes the pending key before the ack (`x/stakeibc/keeper/pending_undelegation.go:109-124`). | RR jumps unnecessarily from ~1.541302 to ~1.555664. On the measured snapshot this is below max-inner 1.568710, leaving 0.013046 RR (~182.11 INJ) of headroom. The design does not report the live outer maximum, and `GetInnerSafetyBounds` clamps to the stricter outer/inner values (`x/stakeibc/keeper/keeper.go:149-160`), so outer safety is an explicit assumption and both maxima must be checked pre-upgrade. It is also not guaranteed safe if supply/rate/bounds move before the upgrade: the new RR is stored before checking bounds (`redemption_rate.go:67-75`), and `BeginBlocker` halts/blacklists the zone when the inner/outer check fails (`x/stakeibc/keeper/invariants.go:32-53`; `x/stakeibc/keeper/abci.go:17-28`). |
| 2. Success ack; ~21-day unbonding period | The nil-record callback subtracts each split from validator delegation and `TotalDelegations` (`x/stakeibc/keeper/icacallbacks_undelegate.go:95-115,214-224`). `UpdateHostZoneUnbondingsAfterUndelegation` iterates zero record IDs and returns a zero burn (`:262-329`), so supply and deposit/LSM inputs remain unchanged; the in-flight unbonding is also not an RR input. | Next RR tick falls by the same ~0.014362 to the pre-incident ~1.541302. That is the correct value: these funds back already-created redemption liabilities, not live stINJ. Assuming the pre-incident rate remains inside the lower bounds, this stage adds no new bounds risk. |
| 3. Unbonding completes and funds sweep | No accounting field used by RR changes. The already-existing `EXIT_TRANSFER_QUEUE` records become sweep-eligible by time/status (`x/stakeibc/keeper/redemption_sweep.go:20-55`); the sweep moves funds on Injective from the delegation ICA to the redemption ICA (`:58-75`) and the success callback marks those records `CLAIMABLE` (`x/stakeibc/keeper/icacallbacks_redemption.go:96-110`). Neither ICA balance nor those records enter the RR formula. | RR remains ~1.541302; there is no accounting jump or new bounds risk. The old records' stINJ was already burned when their original undelegate callback transitioned them to `EXIT_TRANSFER_QUEUE` (`icacallbacks_undelegate.go:285-313,331-349`), so adding the recovered/unbonded/swept amount to RR would double-count assets owed to redeemers. |

Conclusion: stage 1 is an unnecessary and user-visible overstatement; stages 2 and 3 are correct without adding the unbonding or swept funds to RR. Track the recovery as a durable outstanding redemption-liability offset across both queued **and in-flight** states, subtract it from the RR numerator, and reduce/clear it only as successful nil-record undelegate callbacks decrement `TotalDelegations`. The current pending key alone is insufficient because it is deleted at submission. Add a lifecycle test proving RR stays at the baseline before submit, while in flight, after ack/unbonding, and after sweep.

## Findings on the rest of the diff

### Blocking

- `app/upgrades/v34/injective.go:103-130` skips a missing validator or a delta that would make its delegation negative, but still commits all other deltas and returns their partial sum; `app/upgrades/v34/upgrades.go:67-69` then undelegates that partial sum. Skipping a negative entry can greatly over-unbond—for example, omitting blackpanther's -666.635876... entry (`injective.go:43`) changes the intended ~200.48 INJ net to ~867.11 INJ. Validate the full table before mutating anything; on `stride-1`, any missing/invalid entry should abort the reconciliation (and queue nothing), while the existing missing-host-zone no-op can cover non-mainnet environments. Add tests for skipped positive and skipped negative entries asserting all-or-nothing behavior.

### Should fix

- `x/stakeibc/keeper/pending_undelegation_test.go:137-179` verifies submission leaves `TotalDelegations` unchanged, and `x/stakeibc/keeper/icacallbacks_undelegate_test.go:368-399` verifies the later no-record ack decreases it without changing supply, but no test runs the RR calculation between those states. Add the lifecycle regression described above so this accounting invariant cannot regress.

### Nit

- None.

## Things verified OK

- The handler queues exactly the delta actually returned; no key is written for a missing host zone/zero delta.
- Pending submission uses cached state, preserves the key on local errors/zero messages, skips the host-zone unbonding epoch, increments in-progress counters, and re-queues failed/timed-out nil-record batches additively (`x/stakeibc/keeper/pending_undelegation.go:79-131`; `x/stakeibc/keeper/icacallbacks_undelegate.go:55-81,195-212`).
- The normal undelegation path is a mechanical extraction around `GetUndelegateMessagesForAmount`; the zero-message guard remains (`x/stakeibc/keeper/unbonding.go:406-536`).
- Injective's per-record sweep is scoped and oldest-first, while other zones retain the bundled path (`x/stakeibc/keeper/redemption_sweep.go:141-165`); failed redemption callbacks now reset only records still in progress (`x/stakeibc/keeper/icacallbacks_redemption.go:61-84`). Changelog coverage is present.
- Fresh verification: `go test ./app/upgrades/v34 ./x/stakeibc/keeper` passed, and `git diff --check main...HEAD` was clean.

riley-stride (approved):

